PR #11, opened 2016-01-25
Initial api

jakemac53:
This is my first cut at the initial api, any and all suggestions welcome.

There is also a very rough implementation included (lib/generate/build.dart) but I wouldn't focus on that, it is just enough to get the example up and running.

Also you will probably note there are no tests yet, I will add those once I have general buy-in on the api. There is an example though (e2e_example folder) so you can see how the api comes together from a user perspective.

I am also open to other package names, there seems to be a general consensus that `build` is a bit generic.


jakemac53:
cc @kegluneq @munificent @nex3 @sigmundch @kevmoo

lib/src/asset/id.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: You can probably ignore this file, no changes currently from the barback version

e2e_example/lib/copy_builder.dart
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+library build.example.copy_builder;
```
nex3: Down with the tyranny of library tags!

For real though, you can omit these without consequence now.

jakemac53: Are they only needed if you use parts then?

lib/src/asset/asset.dart
```diff
@@ -0,0 +1,19 @@
...
+
+import 'id.dart';
+
+/// A fully realized asset whose content is available synchronously.
+class Asset {
+  /// The id for this asset.
+  final AssetId id;
+
+  /// The content for this asset.
+  final String contents;
```
nex3: How confident are you that no one will need to read in a non-text file?

e2e_example/lib/copy_builder.dart
nex3: Yes.

lib/src/asset/asset.dart
jakemac53: Definitely an open question, I don't know of any transformers today which read in by bytes. Primarily I made the decision to only support String apis to start, but I have been using `*AsString` in other api so I should probably at least change this to `stringContents` or `contentsAsString`.

Basically, I don't want to exclude the possibility of supporting `Stream<int>` later on but I also don't want to actually add support for it until somebody actually needs it.

lib/src/builder/build_step.dart
```diff
@@ -0,0 +1,80 @@
...
+class BuildStep {
+  /// The primary input for this build step.
+  final Asset input;
+
+  /// The list of all outputs which are expected/allowed to be output from this
+  /// step.
+  final List<AssetId> expectedOutputs;
+
+  /// The actual outputs of this build step.
+  UnmodifiableListView<Asset> get outputs => new UnmodifiableListView(_outputs);
```
kegluneq: It feels wrong to me that `outputs` and `dependencies` are "Unmodifiable" but may observably change. Can we hide access to those lists until we are sure they won't change?

lib/src/builder/build_step.dart
```diff
@@ -0,0 +1,80 @@
...
+  final List<Asset> _outputs = [];
+
+  /// A future that completes once all outputs current are done writing.
+  ///
+  /// TODO(jakemac): This is not typically needed inside of a Builder, we could
```
kegluneq: nit: Move this TODO out of docs.

lib/src/generate/build.dart
```diff
@@ -0,0 +1,153 @@
...
+void _validatePhases(List<List<Phase>> phaseGroups) {
+  if (phaseGroups.length > 1) {
+    // Don't support using generated files as inputs yet, so we only support
+    // one phase.
+    throw new UnimplementedError(
+        'Only one phase group is currently supported.');
+  }
+}
+
+/// Validates the phases, for now that just means making sure they are all only
```
kegluneq: Is this comment accurate?

lib/src/generate/input_set.dart
```diff
@@ -0,0 +1,22 @@
...
+
+  /// The file patterns from [package] to use as inputs, glob syntax is
+  /// supported.
+  ///
+  /// Note: If the [package] is a package dependency, then only files under
+  /// `lib` will be available, but for the application package any files can be
+  /// listed.
+  final List<String> filePatterns;
+
+  InputSet(this.package, {Iterable<String> filePatterns})
```
kegluneq: Create `Glob`s from these once in the ctor so they don't have to be recompiled.

Maybe accept `Glob`s in addition to, or instead of, `String`s?

lib/src/generate/build.dart
```diff
@@ -0,0 +1,153 @@
...
+/// running from the root of the package.
+class _SimpleAssetReader implements AssetReader {
+  @override
+  Future<String> readAsString(AssetId id, {Encoding encoding: UTF8}) async {
+    assert(id.package == _localPackageName);
+    return new File(id.path).readAsString(encoding: encoding);
+  }
+}
+
+AssetReader _reader = new _SimpleAssetReader();
```
kegluneq: Could be `const`?

lib/src/generate/build.dart
```diff
@@ -0,0 +1,153 @@
...
+  @override
+  Future writeAsString(Asset asset, {Encoding encoding: UTF8}) async {
+    assert(asset.id.package == _localPackageName);
+    var file = new File(path.join(_outputDir, asset.id.path));
+    await file.create(recursive: true);
+    await file.writeAsString(asset.contents, encoding: encoding);
+  }
+}
+
+AssetWriter _writer = new _SimpleAssetWriter('generated');
```
kegluneq: Could be `const`?

lib/src/builder/build_step.dart
nex3: It seems strange to me that there's so many introspection properties on this class at all. Do you expect the builder not to know its own expected outputs, or to want to read its concrete outputs? If not, exposing these just makes the API more cluttered and commits you to more public API surface.

lib/src/generate/build.dart
```diff
@@ -0,0 +1,153 @@
...
+Future _runBuilder(
+    Builder builder, List<AssetId> inputs, List<Asset> outputs) async {
+  for (var input in inputs) {
+    var expectedOutputs = builder.declareOutputs(input);
+    var inputAsset = new Asset(input, await _reader.readAsString(input));
+    var buildStep =
+        new BuildStep(inputAsset, expectedOutputs, _reader, _writer);
+    await builder.build(buildStep);
+    await buildStep.outputsCompleted;
+    outputs.addAll(buildStep.outputs);
```
nex3: Every time you modify input parameters instead of setting a return value I shed a tear.

lib/src/builder/build_step.dart
jakemac53: Definitely agree with both pieces of input here, I was already kind of thinking the same thing with the TODO on the next field and I think the same logic applies for this one. If only we had friend libraries I could easily leave these as private :).

Being that we don't have friend libraries, and would like to avoid parts, what do you think the best solution is? Create an InternalBuildStep class which wraps a BuildStep and exposes these private fields as public ones (but then hide it when exporting this library in `lib/build.dart`)? I would really rather not go the `part` route but that would solve this issue as well....

lib/src/generate/build.dart
jakemac53: Nope!

lib/src/generate/build.dart
jakemac53: Agreed that using a return value is a simpler pattern in this scenario (and in general). Sometimes my c++ days come back to haunt me.

lib/src/generate/build.dart
jakemac53: done

lib/src/generate/input_set.dart
jakemac53: Done, didn't add ability to supply a `Glob` directly for now, we could add this later if needed.

lib/src/asset/asset.dart
jakemac53: Changed to `stringContents`

jakemac53:
Ok, I have  updated everything except for the `BuildStep` class, haven't decided how I want to solve that one yet.

lib/src/builder/build_step.dart
jakemac53: I am going to go with a `BuildStepImpl` class which implements the `BuildStep` abstract class. You can technically use mirrors (or import BuildStepImpl from the `src` dir) to get at this if you really wanted, but I think that's fine and also consistent with other libraries.

jakemac53:
Ok, all comments addressed

lib/src/builder/build_step.dart
nex3: I like the controller pattern for this. [`barback` uses it](https://github.com/dart-lang/barback/blob/master/lib/src/transformer/aggregate_transform.dart) and [so does `test`](https://github.com/dart-lang/test/blob/master/lib/src/backend/live_test_controller.dart). Basically you'd define both a `BuildStep` and a `BuildStepController` class in the same library, with the controller exposing a `BuildStep` field and having all the extra fields and methods you need. You'd only export `BuildStep` externally, while internally using the controller to create and control it.

jakemac53: For now I think I am fine with the `BuildStepImpl` solution, its a bit simpler to have only the one class.

kegluneq:
LGTM

PR #12, opened 2016-01-28
add transformer tests

jakemac53:
closes https://github.com/dart-lang/build/issues/2

### DIFF
--- a/lib/src/transformer/transformer.dart
+++ b/lib/src/transformer/transformer.dart
@@ -14,7 +14,17 @@ import '../asset/writer.dart';
 import '../builder/builder.dart';
 import '../builder/build_step_impl.dart';
 
-abstract class BuilderTransformer implements Transformer, DeclaringTransformer {
+/// A [Transformer] which runs multiple [Builder]s.
+/// Extend this class and define the [builders] getter to create a [Transformer]
+/// out of your custom [Builder]s.
+abstract class BuilderTransformer implements Transformer {
+  /// The only thing you need to implement when extending this class. This
+  /// declares which builders should be ran.
+  ///
+  /// **Note**: All [builders] are ran in the same phase, and there are no
+  /// ordering guarantees. Thus, none of the [builders] can use the outputs of
+  /// other [builders]. In order to do this you must create a [TransformerGroup]
+  /// with multiple [BuilderTransformer]s.
   List<Builder> get builders;
 
   @override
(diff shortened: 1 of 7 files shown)


jakemac53:
cc @munificent @nex3 @kevmoo

lib/src/transformer/transformer.dart
```diff
@@ -14,7 +14,11 @@ import '../asset/writer.dart';
 import '../builder/builder.dart';
 import '../builder/build_step_impl.dart';
 
-abstract class BuilderTransformer implements Transformer, DeclaringTransformer {
+/// A [Transformer] which runs multiple [Builder]s. Extend this class and define
```
kegluneq: uber nit: line break after the first sentence here, per [Effective Dart](https://www.dartlang.org/effective-dart/documentation/#do-make-the-first-sentence-a-standalone-paragraph)

lib/src/transformer/transformer.dart
```diff
@@ -30,23 +34,32 @@ abstract class BuilderTransformer implements Transformer, DeclaringTransformer {
     var reader = new _TransformAssetReader(transform);
     var writer = new _TransformAssetWriter(transform);
 
-    var futures = <Future>[];
-    for (var builder in builders) {
+    var allExpected = new Set<build.AssetId>();
+    await Future.forEach(builders, (builder) async {
```
kegluneq: This will run each builder serially - is that the intention?

We should probably document this, one way or another.

lib/src/transformer/transformer.dart
```diff
@@ -68,8 +81,9 @@ class _TransformAssetWriter implements AssetWriter {
   _TransformAssetWriter(this.transform);
 
   @override
-  Future writeAsString(build.Asset asset, {Encoding encoding: UTF8}) async =>
-      transform.addOutput(_toBarbackAsset(asset));
+  Future writeAsString(build.Asset asset, {Encoding encoding: UTF8}) async {
```
kegluneq: disclaimer: totally my opinion here. Since there is no inherently `async` operation here, I think it would be clearer to remove `async` and `return new Future.value(null);` here.

WDYT?

test/transformer/transformer_test.dart
```diff
@@ -0,0 +1,117 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+@TestOn('vm')
+import 'package:test/test.dart';
+import 'package:transformer_test/utils.dart';
+
+import '../common/common.dart';
+
+main() {
+  var singleCopyTransformer =
```
kegluneq: Consider moving the creation of these to a `setUp` method.

lib/src/transformer/transformer.dart
jakemac53: Good catch, changed to `Future.wait(builders.map(....))`

lib/src/transformer/transformer.dart
jakemac53: done

lib/src/transformer/transformer.dart
jakemac53: Personally I like just using `async`, because it is less to type/scan. Unless you think there is a significant performance impact?

jakemac53: It does have the interesting side-effect of running `transform.addOutput` asynchronously as well, which is a bit interesting.

test/transformer/transformer_test.dart
jakemac53: Since they are stateless, I prefer them to be set up just once.

lib/src/transformer/transformer.dart
jakemac53: Ya actually I agree with you just because of the weirdness of running `transform.addOutput` async, if nothing else ;).

kegluneq: I'm not particularly worried about performance here, I think returning a Future from a synchronous method makes it easier for a reader to understand what's happening here.

lib/src/transformer/transformer.dart
```diff
@@ -14,7 +14,12 @@ import '../asset/writer.dart';
 import '../builder/builder.dart';
 import '../builder/build_step_impl.dart';
 
-abstract class BuilderTransformer implements Transformer, DeclaringTransformer {
+/// A [Transformer] which runs multiple [Builder]s.
+/// Extend this class and define the [builders] getter to create a [Transformer]
```
kegluneq: I still think we should be explicit about our (lack of) ordering guarantees here, maybe just as simple as "there are no guarantees about the order in which [Builder]s will be run."

kegluneq:
LGTM!

lib/src/transformer/transformer.dart
jakemac53: done

PR #17, opened 2016-01-29
added hasInput funtion to Reader/BuildStep

jakemac53:
cc @nex3 @munificent @kevmoo

### DIFF
--- a/lib/src/builder/build_step.dart
+++ b/lib/src/builder/build_step.dart
@@ -13,6 +13,12 @@ abstract class BuildStep {
   /// The primary input for this build step.
   Asset get input;
 
+  /// Checks if an [Asset] by [id] exists as an input for this [BuildStep].
+  ///
+  /// If [trackAsDependency] is true, then [id] will be marked as a dependency
+  /// of all [expectedOutputs].
+  Future<bool> hasInput(AssetId id, {bool trackAsDependency: true});
+
   /// Reads an [Asset] by [id] as a [String] using [encoding].
   ///
   /// If [trackAsDependency] is true, then [id] will be marked as a dependency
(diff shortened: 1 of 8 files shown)


lib/src/builder/build_step.dart
```diff
@@ -13,6 +13,12 @@ abstract class BuildStep {
   /// The primary input for this build step.
   Asset get input;
 
+  /// Checks if an [Asset] by [id] exists as an input for this [BuildStep].
+  ///
+  /// If [trackAsDependency] is true, then [id] will be marked as a dependency
+  /// of all [expectedOutputs].
+  Future<bool> hasInput(AssetId id, {bool trackAsDependency: true});
```
kegluneq: Is `trackAsDependency` something we want to expose to users, or should it be an implementation detail?

I feel like you had a use case for it, but I can't remember what it is right now.

kegluneq:
LGTM

lib/src/builder/build_step.dart
munificent: What does it mean if the user passes *false* for this? Presumable, the caller uses the boolean result of this method to do *something* that affects their build output. If that boolean were to change without the build step being re-run (because it passed `trackAsDependency: false`), won't you end up with stale or inconsistent output?

jakemac53: After some offline discussions I am going to remove this as part of the public api, at least for now, as well as removing the `addDependency` function. These combined would allow you to write a builder that cheats a bit in terms of marking its dependencies, which specifically when doing resolution would be useful. We may consider adding it back in the future, but for now it doesn't need to be there.

jakemac53: I will do that in a followup cl though

jakemac53:
This is merged into master now

PR #19, opened 2016-01-29
remove custom build apis from public api for now

jakemac53:
These apis are untested and unstable, so removing them from the public api surface for now.


kegluneq:
LGTM

PR #22, opened 2016-02-02
File based asset system

jakemac53:


lib/src/asset/exceptions.dart
```diff
@@ -11,3 +12,32 @@ class AssetNotFoundException implements Exception {
   @override
   String toString() => 'AssetNotFoundException: $assetId';
```
kegluneq: Consider reusing [`Error#safeToString`](https://api.dartlang.org/1.12.0/dart-core/Error/safeToString.html) in exception `toString` methods.

lib/src/asset/file_based.dart
```diff
@@ -0,0 +1,77 @@
...
+
+  FileBasedAssetWriter(this.packageGraph);
+
+  @override
+  Future writeAsString(Asset asset, {Encoding encoding: UTF8}) async {
+    if (asset.id.package != packageGraph.root.name) {
+      throw new InvalidOutputException(asset);
+    }
+
+    var file = _fileFor(asset.id, packageGraph);
```
kegluneq: Should this method `throw` if the file exists?

lib/src/asset/file_based.dart
```diff
@@ -0,0 +1,77 @@
...
+    await file.create(recursive: true);
+    await file.writeAsString(asset.stringContents, encoding: encoding);
```
kegluneq: I *think* the answer is no, but should we be `flush`ing here?

test/asset/file_based_test.dart
```diff
@@ -0,0 +1,94 @@
...
+    test('can output files in the application package', () async {
+      var asset = makeAsset('basic_pkg|test_file.txt', 'test');
+      await writer.writeAsString(asset);
+      var id = asset.id;
+      var file = new File('test/fixtures/${id.package}/${id.path}');
+      expect(await file.exists(), isTrue);
+      expect(await file.readAsString(), 'test');
+      await file.delete();
+    });
+
```
kegluneq: If `FileBasedAssetWriter` can't overwrite files, add a test here for that case.

kegluneq:
LG with a few minor questions

lib/src/asset/file_based.dart
jakemac53: The writer doesn't have enough knowledge to know whether that is ok or not (it is fine to overwrite a generated file from a previous build, just not a previous _phase_).

lib/src/asset/file_based.dart
jakemac53: I don't know a ton about when that would be necessary, but I don't think it is? This will eventually be caching these files as well, at which point it really won't matter.

test/asset/file_based_test.dart
jakemac53: n/a

lib/src/asset/exceptions.dart
jakemac53: Tried that out but the message doesn't seem very good? (all you get is the type)

kegluneq:
All responses SGTM!

Issue #24, opened 2016-02-02
Make it easy and fast to use the analyzer


jakemac53:

jakemac53:
We may be able to just piggyback off of `Resolver` for now, but eventually we probably want to port it.

PR #25, opened 2016-02-02
add `resolve(id)` to BuildStep

jakemac53:
This adds essentially the `code_transformers` package logic directly to this package. It is up in the air still if this is really a good idea or not though ;).

cc @munificent @nex3 

closes https://github.com/dart-lang/build/issues/24

### DIFF
--- a/lib/src/analyzer/resolver.dart
+++ b/lib/src/analyzer/resolver.dart
@@ -1,0 +1,118 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+import 'package:analyzer/analyzer.dart' show Expression;
+import 'package:analyzer/src/generated/constant.dart' show EvaluationResult;
+import 'package:analyzer/src/generated/element.dart';
+import 'package:code_transformers/resolver.dart' as code_transformers
+    show Resolver;
+import 'package:source_maps/refactor.dart';
+import 'package:source_span/source_span.dart';
+
+import '../asset/id.dart';
+import '../builder/build_step.dart';
+import '../util/barback.dart';
+
+class Resolver {
+  final code_transformers.Resolver resolver;
+
+  Resolver(this.resolver);
+
+  /// Update the status of all the sources referenced by the entry points and
+  /// update the resolved library. If [entryPoints] is omitted, the input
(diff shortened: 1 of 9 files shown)


lib/src/analyzer/resolver.dart
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+import 'package:analyzer/src/generated/ast.dart' show Expression;
```
kegluneq: Use the "visible" import `package:analyzer/analyzer.dart` instead

lib/src/builder/build_step_impl.dart
```diff
@@ -5,16 +5,24 @@ import 'dart:async';
...
 import '../asset/writer.dart';
+import '../util/barback.dart';
 import 'build_step.dart';
 import 'exceptions.dart';
 
 /// A single step in the build processes. This represents a single input and
 /// its expected and real outputs. It also handles tracking of dependencies.
 class BuildStepImpl implements BuildStep {
+  static code_transformers.Resolvers _resolvers =
+      new code_transformers.Resolvers.fromMock(
```
kegluneq: I don't really understand `fromMock` - let's discuss offline.

lib/src/util/barback.dart
```diff
@@ -0,0 +1,77 @@
...
+        var message = new StringBuffer();
+        message.write('$level: ');
+        if (asset != null) {
+          message.write('From $asset');
+          if (span != null) {
+            message.write('()$span)');
+          }
+          message.write(': ');
+        }
+        message.write(message);
```
kegluneq: We must go deeper (two `message`s in scope here.)

jakemac53: hahaha good catch :)

lib/src/builder/build_step_impl.dart
jakemac53: Will switch this to use the actual dart sdk until we figure out how we want to make it configurable.

lib/src/analyzer/resolver.dart
jakemac53: done

jakemac53:
updated

kegluneq:
For future reviews, please push a commit with just review changes overtop the original one. You can always merge them before committing.

LGTM

Issue #23, opened 2016-02-02
Add generalized caching layer for Assets


jakemac53:
I think this should just be implemented as a general CachedAssetReader which takes an instance of AssetReader, and delegates to that whenever there isn't a cached asset available.  There would also be a CachedAssetWriter which takes an instance of AssetWriter and a CachedAssetReader (so it can populate the readers cache with new values).

Another option is to just implement an AssetCache class which gets passed into individual implementations of AssetReader/AssetWriter, but then each of those implementations needs to know about the cache.

jakemac53:
Or maybe a combo, so CachedAssetReader and CachedAssetWriter each take an AssetCache.

PR #29, opened 2016-02-04
add caching of assets

jakemac53:
closes https://github.com/dart-lang/build/issues/23


jakemac53:
reassigned, for some reason I thought Tim was not around

lib/src/asset/asset.dart
```diff
@@ -13,5 +13,10 @@ class Asset {
 
   Asset(this.id, this.stringContents);
 
+  bool operator ==(other) =>
```
kegluneq: Please override `hashCode`

lib/src/asset/cache.dart
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/// An [AssetReader] which takes both an [AssetCache] and an [AssetReader]. It
+/// will first look up values in the [AssetCache], and if not present it will
+/// use the [AssetReader] and add the [Asset] to the [AssetCache].
+class CachedAssetReader extends AssetReader {
+  final AssetCache _cache;
+  final AssetReader _reader;
+  /// Cache of ongoing reads by [AssetId].
+  final Map<AssetId, Future<String>> _pendingReads = {};
```
kegluneq: I think these should be
```dart
final _pendingReads = <AssetId, Future<String>>{};
```

lib/src/asset/cache.dart
```diff
@@ -0,0 +1,95 @@
...
+/// Writes are done synchronously to the cache, and then asynchronously to the
+/// [AssetWriter].
+class CachedAssetWriter extends AssetWriter {
+  final AssetCache _cache;
+  final AssetWriter _writer;
+
+  CachedAssetWriter(this._cache, this._writer);
+
+  @override
+  Future writeAsString(Asset asset, {Encoding encoding: UTF8}) {
```
kegluneq: I can't think specifically what, but should `AssetWriter` complete to some specific type?

test/common/matchers.dart
```diff
@@ -5,6 +5,7 @@ import 'package:test/test.dart';
 
 import 'package:build/build.dart';
 
+final argumentError = new isInstanceOf<ArgumentError>();
```
kegluneq: Exists in `package:matcher`

https://github.com/dart-lang/matcher/blob/master/lib/src/error_matchers.dart#L9

kegluneq: Actually, you can replace `throwsA(argumentError)` with `throwsArgumentError` from package:test

https://github.com/dart-lang/test/blob/master/lib/src/frontend/throws_matchers.dart#L10

lib/src/asset/cache.dart
jakemac53: good point, this one always gets me :(

lib/src/asset/cache.dart
jakemac53: I don't think it needs to, it just completes when the asset is done being written. If anything it could return the asset that was passed in but that doesn't really provide any value imo.

lib/src/asset/asset.dart
jakemac53: I am actually just going to make this a test only utility, thats all I was using it for.

test/common/matchers.dart
jakemac53: done

jakemac53:
updated

kegluneq:
LGTM

PR #37, opened 2016-02-08
Add ability to recover from a deleted cache

### LINKED ISSUE #34: Add method of recognizing generated files
After a build, the next build needs to know which files were generated previously, so that they aren't treated as inputs.

Proposed solution:

Output a `build_outputs.json` file under a new hidden folder `.build`, eventually this will be replaced by the dependency graph file, but for now it will just be a list of all output file ids. Just using JSON for now as its the easiest, eventually more formats need to be evaluated.

On startup, check if the file exists and:

- If it does exists, read the file and ignore all listed files as inputs. Probably also delete those files from disk (otherwise we can easily end up with stranded files, we can be smarter about this later on).
- If it doesn't exist, call `declareOutputs` on all builders in all phases, recording all declared outputs which would conflict with an existing file on disk. If there are conflicts then summarize the results to the user to confirm deletion of these files, and then start over.

jakemac53:
closes #34

### DIFF
--- a/lib/src/generate/build.dart
+++ b/lib/src/generate/build.dart
@@ -158,7 +158,8 @@ Future _deletePreviousOutputs(List<List<Phase>> phaseGroups) async {
         done = true;
         break;
       case 'n':
-        done = true;
+        stdout.writeln('Exiting...');
+        exit(1);
         break;
       case 'l':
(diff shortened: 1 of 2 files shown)


kegluneq:
LGTM

Issue #5, opened 2016-01-22
Keep track of a build dependency graph


jakemac53:
Don't need to do anything with it yet

PR #42, opened 2016-02-10
add AssetGraph/AssetNode

jakemac53:
closes https://github.com/dart-lang/build/issues/5

### DIFF
--- a/lib/src/asset_graph/exceptions.dart
+++ b/lib/src/asset_graph/exceptions.dart
@@ -1,0 +1,13 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'node.dart';
+
+class DuplicateAssetNodeException implements Exception {
+  final AssetNode assetNode;
+
+  DuplicateAssetNodeException(this.assetNode);
+
+  @override
+  String toString() => 'DuplicateAssetNodeError: $assetNode';
+}
--- a/lib/src/asset_graph/graph.dart
+++ b/lib/src/asset_graph/graph.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../asset/id.dart';
+import 'exceptions.dart';
+import 'node.dart';
+
+/// Represents all the [Asset]s in the system, and all of their dependencies.
+class AssetGraph {
+  /// All the [AssetNode]s in the system, indexed by [AssetId].
+  final _nodesById = <AssetId, AssetNode>{};
+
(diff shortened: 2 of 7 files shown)


lib/src/asset_graph/exceptions.dart
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'node.dart';
+
+class DuplicateAssetNodeException implements Exception {
+  final AssetNode assetNode;
+
+  DuplicateAssetNodeException(this.assetNode);
```
kegluneq: up to you whether to change, but this could be `const`

lib/src/asset_graph/graph.dart
```diff
@@ -0,0 +1,40 @@
...
+
+  /// Adds [node] to the graph.
+  void add(AssetNode node) {
+    if (_nodesById.containsKey(node.id)) {
+      throw new DuplicateAssetNodeException(node);
+    }
+    _nodesById[node.id] = node;
+  }
+
+  /// Adds the node  returned by [ifAbsent] to the graph, if [id] doesn't
```
kegluneq: nit: extra space

lib/src/asset_graph/node.dart
```diff
@@ -0,0 +1,42 @@
...
+
+/// A node in the asset graph.
+///
+/// This class specifically represents normal (ie: non-generated) assets.
+class AssetNode {
+  /// The asset this node represents.
+  final AssetId id;
+
+  /// The [AssetId]s of all generated assets which depend on this node.
+  final outputs = new Set<AssetId>();
```
kegluneq: This seems fine for now, but I want to be careful about conflating information about the graph structure with details of the build system.

It feels like we should be able to swap out the graph implementation relatively easily without changing much about the rest of the implementation.

Not sure we need to do anything about this now but I wanted to bring it up.

lib/src/asset_graph/node.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  AssetNode(this.id);
+
+  String toString() => 'AssetNode: $id';
```
kegluneq: `@override`, here & in `GeneratedAssetNode`

lib/src/asset_graph/exceptions.dart
jakemac53: can't since AssetNode isn't const :(

kegluneq:
LGTM

lib/src/asset_graph/graph.dart
jakemac53: done

lib/src/asset_graph/node.dart
jakemac53: I think this will shake out in the coming cls as this thing actually gets used.

lib/src/asset_graph/node.dart
jakemac53: done

Issue #43, opened 2016-02-10
Refactor `build` function


jakemac53:
Need to make a `Build` class to wrap up internal state, instead of zone-local values. Its getting a bit out of control and probably hurting performance ;).

PR #44, opened 2016-02-10
refactor build fn to use a class for managing state

jakemac53:
closes https://github.com/dart-lang/build/issues/43

### DIFF
--- a/lib/src/generate/build_impl.dart
+++ b/lib/src/generate/build_impl.dart
@@ -1,0 +1,296 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:path/path.dart' as path;
+
+import '../asset/asset.dart';
+import '../asset/exceptions.dart';
+import '../asset/id.dart';
+import '../asset/reader.dart';
+import '../asset/writer.dart';
+import '../asset_graph/graph.dart';
+import '../asset_graph/node.dart';
+import '../builder/builder.dart';
+import '../builder/build_step_impl.dart';
+import '../package_graph/package_graph.dart';
+import 'build_result.dart';
+import 'exceptions.dart';
+import 'input_set.dart';
+import 'phase.dart';
+
+/// Class which manages running builds.
+class BuildImpl {
+  final AssetGraph _assetGraph;
+  final AssetReader _reader;
+  final AssetWriter _writer;
+  final PackageGraph _packageGraph;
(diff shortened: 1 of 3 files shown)


jakemac53:
Essentially this just moves all top level functions to instance methods, and zone-local values to properties. I also simplified some argument signatures by adding additional properties.

lib/src/generate/build_impl.dart
```diff
@@ -0,0 +1,290 @@
...
+import 'phase.dart';
+
+/// Class which manages running builds.
+class BuildImpl {
+  final AssetGraph _assetGraph;
+  final AssetReader _reader;
+  final AssetWriter _writer;
+  final PackageGraph _packageGraph;
+  final List<List<Phase>> _phaseGroups;
+  Map<String, Set<AssetId>> _inputsByPackage;
```
kegluneq: nit: Make this `final` & assign right away.

Change the assignment in `_initializeInputsByPackage` to a call to `clear()`

lib/src/generate/build_impl.dart
```diff
@@ -0,0 +1,290 @@
...
+  bool _buildRunning = false;
+
+  BuildImpl(this._assetGraph, this._reader, this._writer, this._packageGraph,
+      this._phaseGroups);
+
+  /// Runs a build.
```
kegluneq: Consider adding an explanation of what is returned, something like:

```dart
/// Runs a build
/// 
/// The returned [Future] is guaranteed to complete with a [BuildResult]. If 
/// an exception is thrown by any phase, [BuildResult#status] will be set to 
/// [BuildStatus.Failure]. The exception and stack trace that caused the failure
/// will be available as [BuildResult#exception] and [BuildResult#stackTrace]
/// respectively.
```

lib/src/generate/build_impl.dart
```diff
@@ -0,0 +1,290 @@
...
+      final groupOutputIds = <AssetId>[];
+      for (var phase in group) {
+        var inputs = _matchingInputs(phase.inputSets);
+        for (var input in inputs) {
+          for (var builder in phase.builders) {
+            var outputs = builder.declareOutputs(input);
+
+            groupOutputIds.addAll(outputs);
+            for (var output in outputs) {
+              if (tempInputsByPackage[output.package]?.contains(output) == true) {
```
kegluneq: Aw, I see why this needs to happen, but it makes me sad.

kegluneq:
LGTM

lib/src/generate/build_impl.dart
jakemac53: done, I copied your description cause I liked it :)

lib/src/generate/build_impl.dart
jakemac53: Ya :(, at least its not hitting the file system twice now lol :)

PR #45, opened 2016-02-12
added top level watch function

jakemac53:
First part of https://github.com/dart-lang/build/issues/6. For now this just does full rebuilds, not incremental ones.

### DIFF
--- a/e2e_example/tool/build.dart
+++ b/e2e_example/tool/build.dart
@@ -1,6 +1,8 @@
 // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+import 'dart:io';
+
 import 'package:build/build.dart';
 
 import 'package:e2e_example/copy_builder.dart';
@@ -16,5 +18,9 @@ main() async {
 
   var result = await build([phases]);
 
-  print(result);
+  if (result.status == BuildStatus.Success) {
+    stdout.writeln(result);
+  } else {
+    stderr.writeln(result);
+  }
 }
--- a/lib/src/generate/build.dart
+++ b/lib/src/generate/build.dart
@@ -14,8 +14,9 @@ import '../asset_graph/graph.dart';
 import '../package_graph/package_graph.dart';
 import 'build_impl.dart';
 import 'build_result.dart';
-import 'phase.dart';
 import 'directory_watcher_factory.dart';
+import 'phase.dart';
+import 'watch_impl.dart';
 
 /// Runs all of the [Phases] in [phaseGroups].
 ///
@@ -52,28 +53,21 @@ Future<BuildResult> build(List<List<Phase>> phaseGroups,
       new CachedAssetWriter(cache, new FileBasedAssetWriter(packageGraph));
 
   var buildImpl = new BuildImpl(
-      cache, new AssetGraph(), reader, writer, packageGraph, phaseGroups);
+      new AssetGraph(), reader, writer, packageGraph, phaseGroups);
 
   /// Run the build!
   var futureResult = buildImpl.runBuild();
 
-  // Handle the first SIGINT, and shut down gracefully.
-  terminateEventStream ??= ProcessSignal.SIGINT.watch();
-  int numEventsSeen = 0;
-  var terminateListener = terminateEventStream.listen((_) {
-    numEventsSeen++;
-    if (numEventsSeen == 1) {
-      new Logger('Build').info('Waiting for build to finish...');
-    } else {
-      exit(2);
-    }
+  // Stop doing new builds when told to terminate.
+  var listener = _setupTerminateLogic(terminateEventStream, () {
+    new Logger('Build').info('Waiting for build to finish...');
+    return futureResult;
   });
 
-  return futureResult.then((result) async {
-    await logListener.cancel();
-    await terminateListener.cancel();
-    return result;
-  });
+  var result = await futureResult;
+  listener.cancel();
+  logListener.cancel();
+  return result;
 }
 
 /// Same as [build], except it watches the file system and re-runs builds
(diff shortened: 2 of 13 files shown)


jakemac53:
Test is flaky on the bots right now, I will work on fixing that.

e2e_example/tool/build.dart
```diff
@@ -18,21 +16,5 @@ main() async {
...
-Build Failed :(
-
-Type: ${result.buildType}
-Outputs: ${result.outputs}
-
-Exception: ${result.exception}
-Stack Trace:
-${result.stackTrace}''');
-  }
+  print(result);
```
kegluneq: Consider `stdout` for success & `stderr` for failures

lib/src/generate/build.dart
```diff
@@ -45,12 +51,92 @@ Future<BuildResult> build(List<List<Phase>> phaseGroups,
...
+    if (numEventsSeen == 1) {
+      new Logger('Build').info('Waiting for build to finish...');
+    } else {
+      exit(2);
+    }
+  });
 
-  return result;
+  return futureResult.then((result) async {
+    await logListener.cancel();
```
kegluneq: nit: do we need to wait for `logListener.cancel()` before canceling `terminateListener`? If not, consider:

```dart
await Future.wait([logListener.cancel(), terminateListener.cancel()]);
return result;
```

lib/src/generate/build.dart
```diff
@@ -45,12 +51,92 @@ Future<BuildResult> build(List<List<Phase>> phaseGroups,
...
+      new CachedAssetWriter(cache, new FileBasedAssetWriter(packageGraph));
+  directoryWatcherFactory ??= defaultDirectoryWatcherFactory;
+  var watchImpl = new WatchImpl(directoryWatcherFactory, debounceDelay, cache,
+      new AssetGraph(), reader, writer, packageGraph, phaseGroups);
+
+  var resultStream = watchImpl.runWatch();
+
+  // Handle the first SIGINT, and shut down gracefully. Second one causes a
+  // full shutdown.
+  terminateEventStream ??= ProcessSignal.SIGINT.watch();
```
kegluneq: Move `terminateEventStream` logic to a helper

lib/src/generate/build_impl.dart
```diff
@@ -143,6 +153,7 @@ class BuildImpl {
         case 'y':
```
kegluneq: Add handling for upper case (or just `switch (input.toLowerCase())`

lib/src/generate/build_impl.dart
```diff
@@ -294,3 +305,155 @@ class BuildImpl {
     }
   }
 }
+
+/// Extends [BuildImpl] to add the [runWatch] method, which watches all
```
kegluneq: Avoid implementation details in class comments.

This looks like a good place for composition over inheritance.

lib/src/generate/build_impl.dart
```diff
@@ -294,3 +305,155 @@ class BuildImpl {
...
+  final Duration _debounceDelay;
+
+  /// Whether or not we are currently watching and running builds.
+  bool _runningWatch = false;
+
+  /// Controller for the results stream.
+  StreamController<BuildResult> _resultStreamController;
+
+  /// All file listeners currently set up.
+  final _allListeners = <StreamSubscription>[];
```
kegluneq: nit: This is personal preference, but I like putting a class's `final` properties above its reassignable ones.

lib/src/generate/build_impl.dart
```diff
@@ -294,3 +305,155 @@ class BuildImpl {
...
+    _runningWatch = false;
+    for (var listener in _allListeners) {
+      await listener.cancel();
+    }
+    _allListeners.clear();
+    if (_currentBuild != null) {
+      _logger.info('Waiting for ongoing build to finish.');
+      await _currentBuild;
+    }
+    await _resultStreamController.close();
```
kegluneq: This could be `null` if a user calls `terminate` prior to `runWatch` -- check above & throw a `StateError` if this happens.

lib/src/generate/build_impl.dart
```diff
@@ -294,3 +305,155 @@ class BuildImpl {
...
+    _logger.info('Build watching terminated.');
+  }
+
+  /// Runs a build any time relevant files change.
+  ///
+  /// Only one build will run at a time, and changes are batched.
+  Stream<BuildResult> runWatch() {
+    assert(_runningWatch == false);
```
kegluneq: `assert(!_runningWatch)`?

lib/src/generate/build_result.dart
```diff
@@ -24,6 +24,25 @@ class BuildResult {
   BuildResult(this.status, this.buildType, List<Asset> outputs,
       {this.exception, this.stackTrace})
       : outputs = new List.unmodifiable(outputs);
+
+  String toString() {
```
kegluneq: `@override`

lib/src/generate/build.dart
jakemac53: Weird, these are both returning `null` instead of a `Future`. Apparently this is by design, but definitely annoying :(.

jakemac53: Actually, from reading the docs further it looks like there is no need to wait for the `Future` if one is returned.

lib/src/generate/build.dart
jakemac53: good call

lib/src/generate/build_impl.dart
jakemac53: good point

lib/src/generate/build_impl.dart
jakemac53: reorganized these a bit

lib/src/generate/build_impl.dart
jakemac53: changed and also now throwing a StateError

lib/src/generate/build_impl.dart
jakemac53: done

lib/src/generate/build_impl.dart
jakemac53: Ya I only used inheritance in order to get at the private variables, because I forgot momentarily how privacy works in Dart lol.

jakemac53: Actually, I just broke this out to its own library anyways, turns out I didn't need much private stuff and the stuff I did need is already passed in the constructor.

jakemac53:
Ok updated, still working on fixing the test though

kegluneq:
LGTM when LGT Travis

PR #46, opened 2016-02-12
incremental rebuilds with watch

jakemac53:
closes https://github.com/dart-lang/build/issues/6

### DIFF
--- a/lib/src/generate/build_impl.dart
+++ b/lib/src/generate/build_impl.dart
@@ -9,7 +9,6 @@ import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
 
 import '../asset/asset.dart';
-import '../asset/cache.dart';
 import '../asset/exceptions.dart';
 import '../asset/id.dart';
 import '../asset/reader.dart';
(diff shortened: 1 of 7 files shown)


lib/src/generate/build_impl.dart
```diff
@@ -85,15 +84,21 @@ class BuildImpl {
...
           JSON.decode(await _reader.readAsString(_buildOutputsId));
       await _writer.delete(_buildOutputsId);
       _inputsByPackage[_buildOutputsId.package]?.remove(_buildOutputsId);
-      await Future.wait(previousOutputs.map((output) {
+      await Future.wait(previousOutputs.map((output) async {
         var outputId = new AssetId.deserialize(output);
         _inputsByPackage[outputId.package]?.remove(outputId);
-        return _writer.delete(outputId);
+        var node = _assetGraph.get(outputId);
+        if (node == null || (node as GeneratedAssetNode).needsUpdate) {
```
kegluneq: Will this always be a safe cast to `GeneratedAssetNode`? If so, why not type `_assetGraph`?

kegluneq:
Just the one comment, otherwise LGTM

lib/src/generate/build_impl.dart
jakemac53: This should always be a `GeneratedAssetNode`, otherwise it would be an error (all outputs are `GeneratedAssetNode`s).

Issue #41, opened 2016-02-09
Evaluate other storage formats for dependency graph. 


jakemac53:
Will start with JSON just to get things up and running, but proto etc should be evaluated.

natebosch:
[msgpack](https://pub.dartlang.org/packages/msgpack) is super easy to drop in to replace JSON.

On e2e_example I didn't see a performance difference and file size dropped from 1.5M to 1.1M.
On angular_components_example I saw (sample size 2) read time drop from 1.7s to 1.5s, and write time drop from 350-430ms to 250-275ms. File size dropped from 8M to 4.9M

The package looks like it might be unmaintained though.

natebosch:
The JSON file gzips very well - <10% of it's original size. There is maybe a slight regression on serialize/deserialize time though.

PR #49, opened 2016-02-16
make AssetGraph serializable

jakemac53:


lib/src/asset_graph/exceptions.dart
```diff
@@ -11,3 +11,14 @@ class DuplicateAssetNodeException implements Exception {
...
 }
+
+class AssetGraphVersionError implements Exception {
+  final int versionSeen;
+  final int currentVersion;
+
+  AssetGraphVersionError(this.versionSeen, this.currentVersion);
+
+  @override
+  String toString() => 'AssetGraphVersionError: saw version $versionSeen but '
```
kegluneq: Is it possible to be more concrete? What component of the system saw (read? deserialized?) the incorrect version?

jakemac53: Actually I need to rename this to `Exception` not `Error`. In practice I plan on catching this Exception and just starting the build from scratch in this case, so the user should never see the message.

lib/src/asset_graph/node.dart
```diff
@@ -16,32 +15,56 @@ class AssetNode {
 
   AssetNode(this.id);
 
+  factory AssetNode.deserialize(List serializedNode) {
```
kegluneq: I'll put in a quick plug for protos here. They're not perfect but you get backwards-compatible serialization, composition, etc. for free and their implementations will improve without you doing anything. I think it's worth the up front cost.

jakemac53: Ya I plan on evaluating protos soon, see https://github.com/dart-lang/build/issues/41.

kegluneq: Cool - I have experience with getting them set up, so feel free to reach out to chat about them when the time comes.

kegluneq:
LGTM

PR #50, opened 2016-02-17
Cache the asset graph after each build, and read it in at startup

jakemac53:
closes https://github.com/dart-lang/build/issues/7

### DIFF
--- a/lib/src/generate/build_impl.dart
+++ b/lib/src/generate/build_impl.dart
@@ -13,6 +13,7 @@ import '../asset/exceptions.dart';
 import '../asset/id.dart';
 import '../asset/reader.dart';
 import '../asset/writer.dart';
+import '../asset_graph/exceptions.dart';
 import '../asset_graph/graph.dart';
 import '../asset_graph/node.dart';
 import '../builder/builder.dart';
@@ -25,7 +26,9 @@ import 'phase.dart';
 
 /// Class which manages running builds.
 class BuildImpl {
-  final AssetGraph _assetGraph;
+  AssetGraph _assetGraph;
+  AssetGraph get assetGraph => _assetGraph;
+
   final AssetReader _reader;
   final AssetWriter _writer;
   final PackageGraph _packageGraph;
@@ -34,8 +37,7 @@ class BuildImpl {
   bool _buildRunning = false;
   final _logger = new Logger('Build');
 
-  BuildImpl(this._assetGraph, this._reader, this._writer, this._packageGraph,
-      this._phaseGroups);
+  BuildImpl(this._reader, this._writer, this._packageGraph, this._phaseGroups);
 
   /// Runs a build
   ///
@@ -49,6 +51,12 @@ class BuildImpl {
       if (_buildRunning) throw const ConcurrentBuildException();
       _buildRunning = true;
 
+      /// Initialize the [assetGraph] if its not yet set up.
+      if (_assetGraph == null) {
+        _logger.info('Reading cached dependency graph');
+        _assetGraph = await _readAssetGraph();
+      }
+
       /// Wait while all inputs are collected.
       _logger.info('Initializing inputs');
       await _initializeInputsByPackage();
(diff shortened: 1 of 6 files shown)


lib/src/generate/build_impl.dart
```diff
@@ -79,27 +83,45 @@ class BuildImpl {
...
 
-  /// Asset containing previous build outputs.
-  AssetId get _buildOutputsId =>
-      new AssetId(_packageGraph.root.name, '.build/build_outputs.json');
+  /// Asset containing previous asset dependency graph.
+  AssetId get _assetGraphId =>
+      new AssetId(_packageGraph.root.name, '.build/asset_graph.json');
+
+  /// Reads in the [assetGraph] from disk.
+  Future _readAssetGraph() async {
```
kegluneq: `Future<AssetGraph>`

lib/src/generate/build_impl.dart
```diff
@@ -79,27 +83,45 @@ class BuildImpl {
...
+      /// TODO(jakemac): Only invalidate nodes which need invalidating, which
+      /// will give us incremental builds on startup.
+      var generatedNodes =
+          graph.allNodes.where((node) => node is GeneratedAssetNode);
+      for (var node in generatedNodes) {
+        node.needsUpdate = true;
+      }
+      return graph;
+    } on AssetGraphVersionException catch (_) {
+      return new AssetGraph();
```
kegluneq: In context, this is a reasonable return, but taken alone it has a bad code smell. Add a comment explaining why ignoring the error & returning an empty `AssetGraph` is the right way to handle this.

Is it worthwhile to log the error?

lib/src/generate/build_impl.dart
```diff
@@ -79,27 +83,45 @@ class BuildImpl {
...
+  /// Reads in the [assetGraph] from disk.
+  Future _readAssetGraph() async {
+    if (!await _reader.hasInput(_assetGraphId)) return new AssetGraph();
+    try {
+      _logger.info('Reading cached asset graph.');
+      var graph = new AssetGraph.deserialize(
+          JSON.decode(await _reader.readAsString(_assetGraphId)));
+      /// TODO(jakemac): Only invalidate nodes which need invalidating, which
+      /// will give us incremental builds on startup.
+      var generatedNodes =
```
kegluneq: Feels a bit strange to mix `where` & `for (var ... in ...)`

Maybe:
```dart
graph.allNodes
    .where((node) => node is GeneratedAssetNode)
    .forEach((node) {
      node.needsUpdate = true;
    });
```

lib/src/generate/watch_impl.dart
```diff
@@ -126,40 +127,53 @@ class WatchImpl {
...
+          AssetId id, AssetId primaryInput, ChangeType rootChangeType) async {
         var node = _assetGraph.get(id);
         if (node == null) return;
-        if (node is GeneratedAssetNode) {
-          node.needsUpdate = true;
-        }
         _assetCache.remove(id);
         for (var output in node.outputs) {
-          clearNodeAndDeps(output, rootChangeType);
+          await clearNodeAndDeps(output, primaryInput, rootChangeType);
```
kegluneq: At the risk of premature optimization, all of these can be processed concurrently.

```dart
await Future.wait(node.outputs.map((output) => clearNodeAndDeps(output, primaryInput, rootChangeType));
```

lib/src/generate/watch_impl.dart
```diff
@@ -126,40 +127,53 @@ class WatchImpl {
...
       }
-      for (var input in updatedInputs.keys) {
-        clearNodeAndDeps(input, updatedInputs[input]);
-      }
-      updatedInputs.clear();
 
-      _logger.info('Starting build');
-      _currentBuild = _buildImpl.runBuild();
-      _currentBuild.then((result) {
+      () async {
```
kegluneq: This feels a bit strange in Dart, I think the preferred pattern here is `new Future(() { ... });` but I'm not 100% sure.

jakemac53: Basically, I just wanted to be able to use async/await for this logic, but I need the first part of this function to run synchronously so I didn't want to add async to the entire thing. ie: my end goal was not to schedule all this work for later (although this does have that effect), but instead just be able to use async/await.

lib/src/generate/watch_impl.dart
jakemac53: I thought there might be some race conditions here, but now that I think harder on it I don't think there will be, so I can update this.

lib/src/generate/build_impl.dart
jakemac53: done

lib/src/generate/build_impl.dart
jakemac53: done

lib/src/generate/build_impl.dart
jakemac53: done, logged a message as well

jakemac53:
updated

lib/src/generate/watch_impl.dart
kegluneq: This still looks odd to me. If you prefer to leave the IIFE, please add a comment explaining why.

kegluneq:
LGTM pending added comment re: IIFE

lib/src/generate/watch_impl.dart
jakemac53: I just changed it, also updated to use `Future.wait` which makes the use of async/await less necessary.

PR #51, opened 2016-02-17
add lastModified to AssetReader

jakemac53:
Need this in order to invalidate asset nodes on startup


jakemac53:
following up with some tests for this as well, I just realized I forgot to add those for AssetReader/CachedAssetReader

kegluneq:
LGTM

PR #52, opened 2016-02-18
Add incremental builds on startup using cached dependency graph

jakemac53:


lib/src/asset/cache.dart
```diff
@@ -40,18 +40,18 @@ class AssetCache {
 /// will first look up values in the [AssetCache], and if not present it will
 /// use the [AssetReader] and add the [Asset] to the [AssetCache].
 class CachedAssetReader extends AssetReader {
-  final AssetCache _cache;
+  final AssetCache cache;
```
kegluneq: Consider adding an `evict` or `remove` method to the `CachedAssetReader` rather than exposing the `cache` itself.

jakemac53: good call, adding `evictFromCache` method

kegluneq:
LGTM

Issue #53, opened 2016-02-18
Invalidate the asset graph if the build script changes


jakemac53:
This needs to see if any dependencies change as well.

Allan-Nava:
How can I fix this problem?

jakemac53:
This should be fixed, what problem are you seeing?

Allan-Nava:
> This should be fixed, what problem are you seeing?

> 
> Error compiling dartdevc module:dart_noipy|web/main.ddc.js
> 
> [error] The getter 'AppComponentFactory' isn't defined for the class 'ng'. (/web/main.dart, line 38, col 13)
>

jakemac53:
This looks like a static error in the code - to me it looks like `ng` is an import prefix for a generated template file, and that file doesn't contain something you expect it to?

Allan-Nava:
> This looks like a static error in the code - to me it looks like `ng` is an import prefix for a generated template file, and that file doesn't contain something you expect it to?

This project: https://github.com/NoipyDevOPS/Dart-Noipy/blob/master/web/web/main.dart#L38

Allan-Nava:
> > This looks like a static error in the code - to me it looks like `ng` is an import prefix for a generated template file, and that file doesn't contain something you expect it to?
> 
> This project: https://github.com/NoipyDevOPS/Dart-Noipy/blob/master/web/web/main.dart#L38

I fix! thanks for support

PR #54, opened 2016-02-19
trigger a full rebuild when the build script changes

jakemac53:
closes https://github.com/dart-lang/build/issues/53

### DIFF
--- a/lib/src/generate/build_impl.dart
+++ b/lib/src/generate/build_impl.dart
@@ -4,6 +4,7 @@
 import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
+import 'dart:mirrors';
 
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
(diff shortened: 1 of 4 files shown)


lib/src/generate/build_impl.dart
```diff
@@ -120,6 +133,51 @@ class BuildImpl {
     }
   }
 
+  /// Checks if the [_assetGraph] needs to be completely invalidated.
+  ///
+  /// For now this just means looking at the current running program, and seeing
+  /// if any of its sources are newer than the asset graph itself.
+  Future<bool> _shouldInvalidateAssetGraph() async {
+    var shouldRebuild = false;
+    await Future.wait(currentMirrorSystem().libraries.keys.map((Uri uri) async {
```
kegluneq: Once `shouldRebuild == true`, we should be able to return `true` without waiting for the results of other futures.

Let's discuss offline.

test/common/in_memory_reader.dart
```diff
@@ -35,9 +36,29 @@ class InMemoryAssetReader implements AssetReader {
     }
   }
 
+  Map<Uri, LibraryMirror> __allLibraries;
```
kegluneq: Why `__`?

jakemac53: changed to just a final var

lib/src/generate/build_impl.dart
jakemac53: Will switch to use a completer

jakemac53:
updated

kegluneq:
LGTM

Issue #55, opened 2016-02-19
Handle edits to the build file while the script is running


jakemac53:
Probably the best thing to do here is just quit with some sort of error message?

PR #56, opened 2016-02-19
Error if the build script gets updated after it starts running

jakemac53:
closes https://github.com/dart-lang/build/issues/55


kegluneq:
LGTM

PR #58, opened 2016-02-22
Add colors based on severity

jakemac53:
This only colors the `[${record.level}]` portion, so its not overly obnoxious, lmk what you think.


kegluneq:
Seems fine to me - what does this do on Windows?

jakemac53:
Heh, good question. I will try to figure that out.

jakemac53:
OK, I just put in a thing to disable colors on windows for now, as windows support as a whole needs to be evaluated.

kegluneq:
I'd vote for a helper function to avoid the logic duplication, but it doesn't seem terribly urgent. Up to you.

LGTM.

PR #76, opened 2016-03-04
run all actions in a single phase together

jakemac53:


kevmoo:
LGTM – although I wonder if we pay a cost for this kind of parallelism – if there are many...hrm

jakemac53:
There shouldn't be any issues afaik, we do this type of thing quite a lot

PR #77, opened 2016-03-07
make sure the CachedAssetReader/CachedAssetWriter dont throw synchronously

jakemac53:
The easier solution here would be to just mark all the methods as `async`, but we need to deal with the caches in a synchronous fashion. Those same caches can throw however, so wrapping everything in a try/catch seems to be the simplest solution.

### DIFF
--- a/lib/src/asset/cache.dart
+++ b/lib/src/asset/cache.dart
@@ -57,28 +57,46 @@ class CachedAssetReader extends AssetReader {
   }
 
   @override
+  // Can't use async keyword here because we need to update
+  // `_pendingHasInputChecks` synchronously.
   Future<bool> hasInput(AssetId id) {
-    if (_cache.contains(id)) return new Future.value(true);
+    try {
+      if (_cache.contains(id)) return new Future.value(true);
 
-    return _pendingHasInputChecks.putIfAbsent(id, () async {
-      var exists = await _reader.hasInput(id);
-      _pendingHasInputChecks.remove(id);
-      return exists;
-    });
+      return _pendingHasInputChecks.putIfAbsent(id, () async {
+        try {
+          return await _reader.hasInput(id);
+        } finally {
+          _pendingHasInputChecks.remove(id);
+        }
+      });
+    } catch (e, s) {
+      return new Future.error(e, s);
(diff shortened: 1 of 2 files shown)


lib/src/asset/cache.dart
```diff
@@ -58,27 +58,41 @@ class CachedAssetReader extends AssetReader {
...
+
+      return _pendingHasInputChecks.putIfAbsent(id, () async {
+        try {
+          return await _reader.hasInput(id);
+        } finally {
+          _pendingHasInputChecks.remove(id);
+        }
+      });
+    } catch (e, s) {
+      return new Future.error(e, s);
```
kevmoo: Or just make it `async`?

lib/src/asset/cache.dart
```diff
@@ -58,27 +58,41 @@ class CachedAssetReader extends AssetReader {
...
+          var content = await _reader.readAsString(id, encoding: encoding);
+          _cache.put(new Asset(id, content));
+          return content;
+        } finally {
+          // Make sure we always remove the pending read
+          _pendingReads.remove(id);
+        }
+      });
+    } catch(e, s) {
+      return new Future.error(e, s);
```
kevmoo: Or just make it `async`?

kevmoo:
Yeah – why not make these methods `async`? Then you can get rid of all of the explicit `Future` creation

jakemac53:
Because of the caches, which need to be updated synchronously. For instance given the code:

```dart
Future.wait(assetIds.map((assetId) => reader.readAsString(assetId)));
``

All of those calls are dispatched synchronously, so the `_pendingReads` map needs to be updated synchronously.

File: lib/src/asset/cache.dart [hunk 3]
```diff
@@ -58,27 +58,41 @@ class CachedAssetReader extends AssetReader {
 
   @override
   Future<bool> hasInput(AssetId id) {
```
kevmoo: Add comment as to why this is not `async` – it makes sense now, but docs would help.

kevmoo:
LGTM – just add comments as to why we don't use `async` for future maintainers.

PR #79, opened 2016-03-07
make strong-mode happy

jakemac53:


kevmoo:
Woo hoo!

You'll have to disable travis on Stable – maybe w/ a note that we should re-enable once 1.15 is stable

PR #80, opened 2016-03-07
get basic windows support working

jakemac53:


kevmoo:
> Hello, friend. Have you heard of our lord and savior, `pkg/path`?

Sooo much better than randomly handing paths n such

jakemac53:
I am using `path` where relevant, but in most places this package only deals in unix style paths, which is why you still see a fair bit of hardcoded `/` in there.  I tend not to use the `path` package when dealing with those types of paths because it just adds a lot of boilerplate in that case, and you have to remember to use `path.url.*` everywhere which is easy to forget and actually more error prone.

kevmoo:
LGTM

PR #83, opened 2016-03-09
Better logging

jakemac53:


lib/src/generate/options.dart
```diff
@@ -57,9 +57,11 @@ class BuildOptions {
       } else {
         color = _red;
       }
-      var message = '$color[${record.level}]$_endColor ${record.loggerName}: '
+      var message = '${Platform.isWindows ? '' : '\x1b[2K\r'}'
```
kegluneq: Just curiosity, what's this craziness do?

kegluneq:
LGTM. What does it mean that "info logs now overwrite each other by default"?

lib/src/generate/options.dart
jakemac53: Oh ha, it deletes the entire previous line and sets the cursor back to the beginning. Seems to be the easiest way.

jakemac53:
They don't cause a new line in your console, they just overwrite the current line. This just helps to get rid of the spammy logs and only highlight warnings/errors.

This only happens with the default logger implementation (which is a console logger).

Issue #68, opened 2016-03-02
Discuss location/naming of .build


kevmoo:

kevmoo:
Current thought: go with `.cache/build` – give an opportunity for other packages to play along if they like.

jakemac53:
I think `.cache` is probably to generic, but I like the general idea

jakemac53:
Ok, going with `.dart_tool/build`

PR #84, opened 2016-03-10
Change cache location

jakemac53:
closes https://github.com/dart-lang/build/issues/68


lib/src/generate/build_impl.dart
```diff
@@ -168,7 +168,7 @@ class BuildImpl {
 
   /// Asset containing previous asset dependency graph.
   AssetId get _assetGraphId =>
-      new AssetId(_packageGraph.root.name, '.build/asset_graph.json');
+      new AssetId(_packageGraph.root.name, '.dart_tool/build/asset_graph.json');
```
kevmoo: DRY this up. Make a const in `lib/src/some_util.dart` or something.

https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

kevmoo: ...I'd say three times, if you count the help text :smile:

Issue #87, opened 2016-03-11
Multiple build scripts?


jakemac53:
Today only one build script per package is supported because all scripts share the same asset graph.

We should at least provide some errors or something around this.

PR #88, opened 2016-03-11
asset graph path is now based on script uri

jakemac53:
closes https://github.com/dart-lang/build/issues/87


kevmoo:
LGTM – although does the content of the script also factor in here? Should it?

jakemac53:
There is separate logic elsewhere to invalidate the caches when the script changes, it just uses last modified timestamps to avoid reading the files.

PR #92, opened 2016-03-14
dont prompt to delete files if not in a terminal

jakemac53:
This helps things make sense when running in intellij/atom, or whenever stdin isn't a normal terminal.

### DIFF
--- a/lib/src/generate/build_impl.dart
+++ b/lib/src/generate/build_impl.dart
@@ -167,10 +167,17 @@ class BuildImpl {
       _logger.info('Succeeded after ${watch.elapsedMilliseconds}ms with '
           '${result.outputs.length} outputs\n\n');
     } else {
+      if (result.exception is FatalBuildException) {
+        exitCode = 1;
+      }
       var exceptionString =
           result.exception != null ? '\n${result.exception}' : '';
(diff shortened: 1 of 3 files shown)


lib/src/generate/build_impl.dart
```diff
@@ -375,15 +375,29 @@ class BuildImpl {
...
         'which already exist on disk. This is likely because the'
         '`$cacheDir` folder was deleted, or you are submitting generated '
         'files to your source repository.');
+
+    // If not in a standard terminal then we just exit, since there is no way
+    // for the user to provide a yes/no answer.
+    if (stdioType(stdin) != StdioType.TERMINAL) {
+      _logger.severe('Exiting since not in interactive mode and'
+          '`deleteFilesByDefault` was not set to true.');
+      exit(1);
```
kevmoo: Could we throw a special exception here and turn it into an exit code at the `main` level?

See https://github.com/dart-lang/test/blob/master/lib/src/executable.dart#L152

jakemac53:
Ok, added a `FatalBuildException` class, and when those are seen then the `exitCode` is set. The watcher also now listens for these and terminates itself.

kevmoo:
LGTM - squash & rebase on master

PR #97, opened 2016-03-16
Kevmoo working

kevmoo:


lib/src/asset/file_based.dart
```diff
@@ -20,7 +20,7 @@ import 'exceptions.dart';
 /// files from disk.
 class FileBasedAssetReader implements AssetReader {
   final PackageGraph packageGraph;
-  final ignoredDirs;
+  final List ignoredDirs;
```
jakemac53: might as well do `List<String>`

lib/src/asset/id.dart
```diff
@@ -86,11 +89,12 @@ class AssetId implements Comparable<AssetId> {
   AssetId changeExtension(String newExtension) =>
       new AssetId(package, pathos.withoutExtension(path) + newExtension);
 
+  @override
   String toString() => "$package|$path";
 
   /// Serializes this [AssetId] to an object that can be sent across isolates
   /// and passed to [deserialize].
-  serialize() => [package, path];
+  List serialize() => [package, path];
```
jakemac53: This was actually intentionally dynamic, but you could mark that explicitly (maybe as `Object` actually?). The format is likely to change in the future and is meant to be treated as opaque.

lib/src/asset_graph/node.dart
```diff
@@ -11,7 +11,7 @@ class AssetNode {
   final AssetId id;
 
   /// The [AssetId]s of all generated assets which depend on this node.
-  final outputs = new Set<AssetId>();
+  final Set<AssetId> outputs = new Set<AssetId>();
```
jakemac53: type shouldn't be necessary here?

lib/src/util/constants.dart
```diff
@@ -6,10 +6,10 @@ import 'dart:io';
 import 'package:crypto/crypto.dart';
 
 /// Relative path to the asset graph from the root package dir.
-final assetGraphPath = '$cacheDir/${scriptHash}/asset_graph.json';
+final String assetGraphPath = '$cacheDir/$scriptHash/asset_graph.json';
```
jakemac53: these types shouldn't be necessary either

test/asset/cache_test.dart
```diff
@@ -4,14 +4,13 @@
 @TestOn('vm')
 import 'dart:async';
 
-import 'package:test/test.dart';
-
 import 'package:build/build.dart';
 import 'package:build/src/asset/cache.dart';
+import 'package:test/test.dart';
```
jakemac53: This was intentional, https://www.dartlang.org/effective-dart/style/#prefer-placing-third-party-package-imports-before-other-imports

test/asset/file_based_test.dart
```diff
@@ -4,17 +4,17 @@
 @TestOn('vm')
 import 'dart:io';
 
+import 'package:build/build.dart';
```
jakemac53: ditto

test/asset/cache_test.dart
jakemac53: Same applies elsewhere as well

test/common/fake_watcher.dart
```diff
@@ -17,18 +20,22 @@ class FakeWatcher implements DirectoryWatcher {
...
 
+  @override
   Future get ready => new Future(() {});
 
+  @override
   bool get isReady => true;
 
   /// All watchers.
-  static final watchers = <FakeWatcher>[];
+  static final List<FakeWatcher> watchers = <FakeWatcher>[];
```
jakemac53: same comment about the type on the left

test/common/matchers.dart
```diff
@@ -25,7 +27,7 @@ class _AssetMatcher extends Matcher {
   const _AssetMatcher(this._expected);
 
   @override
-  bool matches(item, _) =>
+  bool matches(Object item, _) =>
```
jakemac53: What value is they type here providing?

test/common/matchers.dart
```diff
@@ -45,7 +47,7 @@ class _AssetGraphMatcher extends Matcher {
   const _AssetGraphMatcher(this._expected, this._checkValidAsOf);
 
   @override
-  bool matches(item, _) {
+  bool matches(dynamic item, _) {
```
jakemac53: same comment here

jakemac53:
Generally looks good, we should talk about the types on the left hand side for final variables, afaik there is no reason for those? Also the import ordering, I have been following the guideline to put imports for the current package below imports for other packages.

lib/src/asset/file_based.dart
kevmoo: ack

lib/src/asset/id.dart
kevmoo: ack

lib/src/util/constants.dart
kevmoo: In theory, no. But the lint is yelling about it.

lib/src/asset_graph/node.dart
kevmoo: In theory, no. But the lint is yelling about it.

test/common/matchers.dart
kevmoo: Hrm – trying to make the linter happy. I really don't like this guideline at all.

test/common/matchers.dart
kevmoo: Ditto

jakemac53:
So is the `type_annotate_public_apis` lint rule the one that caused types on the LHS even for final fields as well as the explicit dynamic type on arguments? In my opinion, we should remove that lint then because its providing negative value overall.

PR #102, opened 2016-04-29
Fix GeneratedAssetNode bug

jakemac53:


kevmoo:
LGTM – weird that the test crashes unless I pub upgrade.

Silly nit – I notice you use triple slashes within method bodies.

jakemac53:
It might have been a different test which failed, one of the tests relies on a certain version of analyzer (they changed their error messages).

In terms of the triple slashes, I know that's not the "correct" way but I prefer to keep it consistent with the rest of the file for now, and I don't think there is any harm in doing it.

PR #107, opened 2016-05-23
improve invalid output error messages, fixes #106

### LINKED ISSUE #106: Error message when output is also an input is misleading
Triggered here

https://github.com/dart-lang/build/blob/master/lib/src/generate/build_impl.dart#L523 

causes the error message "Files may only be output in the root (application) package." which is misleading. It should be something like, "Cannot overwrite input."

jakemac53:


kevmoo:
LGTM – slap dartfmt over this before you commit

PR #116, opened 2016-06-21
Fix #109, convert packages paths to absolute paths in file watcher

### LINKED ISSUE #109: Repeated build catches changes, watcher doesn't
I have a case where

 - I'm using a watch.dart that's identical to the build.dart, it just calls 'watch' instead of 'build'
 - When I make changes and re-run build.dart, it notices the changes and correctly updates everything
 - But if I run watch.dart, and make the same changes
  --> it notices that I saved a file and triggers a build
  --> but, the build does nothing -- says '0 output files'

Am I correct that running the watcher should be equivalent to repeatedly running the build script -- or did I miss something there?

Thanks!

jakemac53:


lib/src/generate/watch_impl.dart
```diff
@@ -195,17 +195,45 @@ class WatchImpl {
...
+                path.startsWith(absolutePackagePath))
+            .toList();
 
         var watcher = _directoryWatcherFactory(absolutePackagePath);
         watchers.add(watcher);
         _allListeners.add(watcher.events.listen((WatchEvent e) {
+          var changePath = e.path;
+
+          // Convert `packages` paths to absolute paths.
+          () {
```
kevmoo: Why this lambda thingy here? Feels weird...

kevmoo:
One question inline.

LGTM

...you ran dartfmt, right?

lib/src/generate/watch_impl.dart
jakemac53: Oh I just used that to encapsulate all the local variables so I didn't have to worry about using somewhat generic names. Really though I should have just broken this out into a helper method or something, I will do that instead.

kevmoo: Sweet

davidmorgan:
Nice, thanks!

PR #122, opened 2016-08-22
allow customizing logging in testPhases

yjbanov:
/cc @jakemac53

### DIFF
--- a/build_test/lib/src/utils.dart
+++ b/build_test/lib/src/utils.dart
@@ -63,13 +63,58 @@ Future<BuildResult> nextResult(List results) {
   return done.future;
 }
 
+/// Runs build [phases] in a test environment.
+///
+/// The test environment supplies in-memory build [inputs] to the builders under
+/// test. [outputs] may be optionally provided to verify that the builders
+/// produce the expected output.
+///
+/// The keys in [inputs] and [outputs] are paths to file assets and the values
+/// are file contents. The paths must use the following format:
+///
+///     PACKAGE_NAME:PATH_WITHIN_PACKAGE
+///
+/// Where `PACKAGE_NAME` is the name of the package, and `PATH_WITHIN_PACKAGE`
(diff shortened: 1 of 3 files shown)


build_test/lib/src/utils.dart
```diff
@@ -63,13 +63,58 @@ Future<BuildResult> nextResult(List results) {
...
+/// The keys in [inputs] and [outputs] are paths to file assets and the values
+/// are file contents. The paths must use the following format:
+///
+///     PACKAGE_NAME:PATH_WITHIN_PACKAGE
+///
+/// Where `PACKAGE_NAME` is the name of the package, and `PATH_WITHIN_PACKAGE`
+/// is the path to a file relative to the package. `PATH_WITHIN_PACKAGE` must
+/// include `lib`, `web`, `bin` or `test`. Example: "myapp|lib/utils.dart".
+///
+/// [packageGraph] supplies dependent packages necessary to process [inputs] in
```
jakemac53: All this actually does in the in memory asset reader/writer today is define the root package (the one you are allowed to output files in). Anything else is not used.

Ideally it would be able to do more than that though, such as load other packages from disk.

test/build_test/utils_test.dart
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: I wonder if this should live under build_test/test? We would then need to add something to the travis file as well to run the tests under there.

jakemac53:
Generally lgtm, just some minor comments

build_test/lib/src/utils.dart
yjbanov: Done

test/build_test/utils_test.dart
yjbanov: Done

PR #124, opened 2016-09-06
Refactor BuilderTransformer and Resolver

natebosch:
Makes it feasible to create new Resolver implementations and plumb them through to the build step.


matanlurey:
I think you need to update `build_test` as well.

e2e_example/lib/transformer.dart
```diff
@@ -1,13 +0,0 @@
...
-// BSD-style license that can be found in the LICENSE file.
-import 'package:build/build.dart';
-
-import 'copy_builder.dart';
-
-class CopyTransformer extends BuilderTransformer {
-  @override
-  final List<Builder> builders = [new CopyBuilder()];
-
-  CopyTransformer.asPlugin(_);
```
sigmundch: do we want to keep an example like this one to illustrate how to plug a build transformer as a regular pub-transformer (which requires you to have a `asPlugin` constructor that pub can reflectively use to allocate the transformer)?

It can be simpler with your change though, for example:
```dart
class CopyTransformer extends BuildTransformer {
  CopyTransformer.asPlugin(_) : super([new CopyBuilder()]);
}
```

natebosch: I didn't realize that asPlugin had extra meaning which made this useful. Restored it.

natebosch:
Since this repo has 2 pub packages it's a bit tough to get travis to pass during version bumps. Options are:
- Submit changes to `build`, update pub, submit changes to `build_test`. This leaves the git repo in an inconsistent state for some time
- Add a dependency_overrides to `build_test` pubspec.yaml to use the local copy. This leaves us open to failing to bump the version number for build upgrades since we'll never see the travis failure
- Submit with a failing travis, update pub, then travis will go back green

Discussed offline and we'll go with the last option

sigmundch:
lgtm!

Issue #135, opened 2016-11-03
Removing a package seems to cause issues


zoechi:
I removed the package `ng_bootstrap` from my project and re-run the build.
(I did also some other work in this project since I last run build)
Now I get

`(master) $ dart tool/build.dart 
[SEVERE] Build: Failed after 234ms
PackageNotFoundException: ng_bootstrap
package:build/src/asset/file_based.dart 115:5      _fileFor
package:build/src/asset/file_based.dart 30:12      FileBasedAssetReader.hasInput.<async>
dart:async                                         _Completer.completeError
package:build/src/asset/file_based.dart 33:3       FileBasedAssetReader.hasInput.<async>
dart:async                                         Future.Future.microtask
package:build/src/asset/file_based.dart            FileBasedAssetReader.hasInput
package:build/src/asset/cache.dart 67:26           CachedAssetReader.hasInput.<fn>.<async>
dart:async                                         Future.Future.microtask
package:build/src/asset/cache.dart                 CachedAssetReader.hasInput.<fn>
dart:collection                                    _HashVMBase&MapMixin&&_LinkedHashMapMixin.putIfAbsent
package:build/src/asset/cache.dart 65:37           CachedAssetReader.hasInput
package:build/src/generate/build_impl.dart 259:34  BuildImpl._getUpdates.<async>.<fn>.<async>
dart:async                                         Future.wait
package:build/src/generate/build_impl.dart 254:18  BuildImpl._getUpdates.<async>
dart:async                                         Future.Future.microtask
package:build/src/generate/build_impl.dart         BuildImpl._getUpdates
package:build/src/generate/build_impl.dart 99:34   BuildImpl.runBuild.<async>.<fn>.<async>.<fn>.<async>
dart:async                                         _SyncCompleter.complete
package:build/src/generate/build_impl.dart 187:7   BuildImpl._readAssetGraph.<async>
dart:async                                         _SyncCompleter.complete
package:build/src/asset/cache.dart 91:11           CachedAssetReader.readAsString.<fn>.<async>
package:build                                      BuildImpl.runBuild.<async>.<fn>.<async>.<fn>
package:build/src/logging/logging.dart 19:28       logWithTime.<async>
dart:async                                         Future.Future.microtask
package:build/src/logging/logging.dart             logWithTime
package:build/src/generate/build_impl.dart 89:15   BuildImpl.runBuild.<async>.<fn>.<async>
package:stack_trace                                Chain.capture
[...]

PR #137, opened 2016-11-04
better support rebuilds after package graph updates

jakemac53:
fixes https://github.com/dart-lang/build/issues/135


e2e_example/tool/watch.dart
```diff
@@ -17,5 +18,8 @@ Future main() async {
   /// the transitive deps issue.
   CopyBuilder.addPhases(phases, graph);
 
+  /// Adds all the phases necessary to copy all files into a fake packages dir!
```
natebosch: might want to fix existing cases in this file - these aren't doc comments and shouldn't use triple slash

PR #139, opened 2016-11-12
support latest version of build_test

kevmoo:
Need to make sure we do this after every release


kevmoo:
@jakemac53 would you just publish this? Seems we forgot to...

kevmoo:
...the code is already there...

PR #142, opened 2016-11-21
Split into more focused packages

### LINKED ISSUE #140: Split into more focused packages
Packages like source_gen want to be able to use the interfaces, and a few utilities, from build but don't necessarily need things like the runner or barback.

natebosch:
Fixes #140 

Allows development and use of the Builder/BuildStep api without dependencies through to, for example, barback. The `build` package will be the most stable and we can have multiple code generation approaches that become interchangeable by using these apis.

- Create two new top level directories build_runner and build_barback with separate packages for each.
- Pull out code specific to running builds, or to code_transformers/barback, into these separate packages
- Update build_test to depend on both build and build_runner and to expose some of the testing utilities that were previously only available inside tests for build
- Add a run_builder with a easier API on top of create BuildSteps and running builds
- Add a ManagedBuildStep interface with a factory through to the BuildStepImpl. Builders continue to only care about the smaller BuildStep interface, while code running builds uses the expanded ManagedBuildStep
- Reduce the interface of BuildStepImpl. Take the code for tracking inputs/outputs and move to the build_runner package
- Reduce the interface of AssetReader by moving methods only used from build_runner to a subclass in that package


README.md
```diff
@@ -4,15 +4,13 @@ These packages provide utilities for generating Dart code.
 
 ## Build
```
kevmoo: Use `build` on the header – so it's clear you're talking about the subdirectory/package

README.md
```diff
@@ -4,15 +4,13 @@ These packages provide utilities for generating Dart code.
...
 
-### Interfaces
-
 Defines the basic pieces of how a build happens and how they interact.
 
-### Runner
+## Build Runner
```
kevmoo: Ditto...

README.md
```diff
@@ -4,15 +4,13 @@ These packages provide utilities for generating Dart code.
...
 
 Provides `build` and `watch` utilities to enact builds.
 
-### Barback
+## Build Barback
```
kevmoo: ditto

build/README.md
```diff
@@ -1,8 +1,23 @@
 # [![Build Status](https://travis-ci.org/dart-lang/build.svg?branch=master)](https://travis-ci.org/dart-lang/build) [![Coverage Status](https://coveralls.io/repos/dart-lang/build/badge.svg?branch=master)](https://coveralls.io/r/dart-lang/build)
 
-This package provides a way of generating files using Dart code, outside of
-`pub`. These files are generated directly on disk in the current package folder,
-and rebuilds are _incremental_.
+# Build
```
kevmoo: Talking about the package, right? `build`

build/README.md
```diff
@@ -1,8 +1,23 @@
...
+
+Defines the basic pieces of how a build happens and how they interact.
+
+## Builder
```
kevmoo: Talking about the class? Maybe `Builder` class? etc...

README.md
```diff
@@ -2,22 +2,20 @@
 
 These packages provide utilities for generating Dart code.
 
-## Build
-
-### Interfaces
+## `build`
 
 Defines the basic pieces of how a build happens and how they interact.
```
jakemac53: Lets add some more details in each of these sections, specifically around how dependencies should be set up:

## build

Defines the interfaces for creating a `Builder` which is a way of doing codegen that is compatible across build systems (pub, bazel, standalone runner, etc).

In general core packages that want to do code generation should only have a public dependency on this package (although they may have a dev dependency on one or more of the other packages).

It is recommended that transformers are exposed via a separate package, which has a dependency on `build_barback`. This allows core packages to not leak the dependency on `barback` and all of its transitive deps to all consumers.

## build_runner

Provides `build` and `watch` utilities to enact builds.

This package generally only be a `dev_dependency` as it is used to run actual standalone builds. The only exception to that would be if somebody was wrapping the `build` and `watch` methods with some other package.


## build_test

Stub implementations for classes in `Build` and some utilities for running instances of builds and checking their outputs.

This package generally only be a `dev_dependency` as it introduces a dependency on `package:test`. The exception to that would be if you were creating another testing-only package that wraps this one.

build_runner/pubspec.yaml
```diff
@@ -0,0 +1,34 @@
...
+author: Dart Team <misc@dartlang.org>
+homepage: https://github.com/dart-lang/build
+
+environment:
+  sdk: '>=1.9.1 <2.0.0'
+
+dependencies:
+  analyzer: ">=0.27.1 <0.30.0"
+  build: ^0.6.0
+  build_barback: 0.0.1
```
jakemac53: This should be removed from regular deps right?

build_runner/pubspec.yaml
```diff
@@ -0,0 +1,34 @@
...
+  shelf: ^0.6.5
+  shelf_static: ^0.2.3
+  stack_trace: ^1.6.0
+  watcher: ^0.9.7
+  yaml: ^2.1.0
+
+dependency_overrides:
+  build:
+    path: ../build
+  build_barback:
```
jakemac53: Remove this or move to dev_dependencies if its used for testing

build_runner/pubspec.yaml
```diff
@@ -0,0 +1,34 @@
...
+    path: ../build_barback
+
+dev_dependencies:
+  build_test:
+    path: ../build_test
+  test: ^0.12.0
+  transformer_test: ^0.2.0
```
jakemac53: This shouldn't be necessary for this package?

build_runner/tool/create_test_all.dart
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: This file probably exists for all packages.... should we stick it in some top level tool dir?

build_test/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-## 0.2.0+1
+## 0.3.0
```
jakemac53: Is this actually a breaking change for this package, or is it all internal to the package?

build_test/lib/src/test_builder.dart
```diff
@@ -0,0 +1,85 @@
...
+import 'package:logging/logging.dart';
+import 'package:test/test.dart';
+
+import 'package:build/build.dart';
+
+import 'in_memory_writer.dart';
+import 'in_memory_reader.dart';
+import 'assets.dart';
+
+void _checkOutputs(Map<String, /*String|Matcher*/ dynamic> outputs,
```
jakemac53: Can we re-use `checkOutputs` from `test_phases.dart`?

build/lib/src/builder/build_step_impl.dart
```diff
@@ -18,38 +17,23 @@ import 'exceptions.dart';
...
-    _logger ??= new Logger(input.id.toString());
-    return _logger;
-  }
-
-  Logger _logger;
+  final Logger logger;
 
-  /// The actual outputs of this build step.
-  UnmodifiableListView<Asset> get outputs => new UnmodifiableListView(_outputs);
-  final List<Asset> _outputs = [];
```
natebosch: Why do we need this in the bazel case? We haven't needed it there before

build/lib/src/builder/build_step_impl.dart
```diff
@@ -59,26 +43,25 @@ class BuildStepImpl implements BuildStep {
...
-      this._writer, this._rootPackage, this._resolvers)
-      : expectedOutputs = new List.unmodifiable(expectedOutputs) {
-    /// The [input] is always a dependency.
-    _dependencies.add(input.id);
+  BuildStepImpl(Asset input, Iterable<AssetId> expectedOutputs, this._reader,
+      this._writer, this._rootPackage, this._resolvers,
+      {Logger logger})
+      : this.input = input,
+        this.logger = logger ?? new Logger('${input.id}') {
+    _expectedOutputs.addAll(expectedOutputs);
```
natebosch: It's not exposed outside this library anymore and I generally like UnmodifiableList as a signal that we're being defensive because it's visible to other code

build/pubspec.yaml
```diff
@@ -21,6 +21,12 @@ dependencies:
   watcher: ^0.9.7
   yaml: ^2.1.0
 
+dependency_overrides:
```
natebosch: These are needed for travis tests to pass locally. I didn't realize you can't publish with them... Removing

build/lib/src/builder/build_step_impl.dart
jakemac53: ah kk, sgtm

build_runner/lib/src/asset/cache.dart
```diff
@@ -4,11 +4,10 @@
 import 'dart:async';
```
natebosch: My intention was to keep package:build as small as possible until there is evidence it needs to grow. Since only build_runner stuff is using this so far I moved it there. If we come across something using that that isn't using other stuff from build_runner we can move it back and re-export it

build_runner/pubspec.yaml
natebosch: Unfortunately this one still needs build_barback for now since it uses `BarbackResolvers` as a default implementation. Once we can find another way to provide a reasonable default without code_transformers we can break this dependency

build_barback/pubspec.yaml
```diff
@@ -0,0 +1,35 @@
...
+  shelf: ^0.6.5
+  shelf_static: ^0.2.3
+  stack_trace: ^1.6.0
+  watcher: ^0.9.7
+  yaml: ^2.1.0
+
+dependency_overrides:
+  build:
+    path: ../build
+  build_runner:
```
jakemac53: ohhhh, because they are transitive deps sad :(

build_runner/lib/src/asset/cache.dart
jakemac53: I am fine w/that, the bazel stuff also does something similar though.

build_test/lib/src/test_builder.dart
natebosch: They are a little different. We can look at refactoring in a followup

build/pubspec.yaml
jakemac53: ack(nowledged)

build_test/CHANGELOG.md
natebosch: Ah good catch. At one point I had made a breaking change but ended up reverting it.

build_runner/lib/src/asset/cache.dart
natebosch: Right now this one depends on details like `InputSet` from `build_runner`. Depending on whether we actually need that full interface in the places we're using this cache today we might need to have 2 copies.

I'll look at whether we should re-add a cache in `build` or add a different implementation only the package dealing with bazel in a followup

build_runner/tool/create_test_all.dart
natebosch: Good call

natebosch: Turns out it's easier to duplicate it. It needs to be under one of the packages with a pubspec.yaml...

PR #148, opened 2016-11-29
Drop build_runner dep from build_test

natebosch:
Most consumers of build_test will need to test Builder instances. The
testPhases utility is most useful from within the build_runner package.

- Drop dep on build_runner from build_test
- Add dev_dependency on build_barback from build. Previously this was
  picked up transitively build_test -> build_runner -> build_barback
- Move utilities specific to classes in build_runner from build_test
  into build_runner/test/common. This includes duplicating the
  InMemoryReader/Writer so that the version in build_runner can
  implement the extended interfaces
- Clean up unnecessary method implementations for asset reader/writer
  implementations
- Remove duplicated test for BarbackResolvers
- Remove tests from build_test package. Previous test was only useful
  for testPhases
- Clean up signature of testBuilder. It had been copied from testPhases
  but some arguments were unused


kevmoo:
Going to let @jakemac53 run with this...

tool/travis.sh
```diff
@@ -22,7 +22,6 @@ function run_coveralls {
 pushd build_test
 pub upgrade
 dartanalyzer --fatal-warnings lib/build_test.dart
-pub run test
```
jakemac53: Probably doesn't hurt to leave this in? Even if there are no tests I think it still passes, and that way if we do add tests they will run.

natebosch: It does cause it to fail. It prints
"No test files were passed and the default "test/" directory doesn't exist." and usage and returns a non-zero exit code.

PR #153, opened 2016-12-02
Fix log forwarding in BuilderTransformer

natebosch:
Fixes a bug where there would be runtime failures when double wrapping
(a Builder as a Transformer as a Builder).

- Set hierarchialLoggingEnabled to true so that a `detached` logger can
  actually stand outside the logging hierarchy rather than forward all
  logs to root
- Use a detached logger instead of a logger participating in chain since
  we don't want these logs to go anywhere other than to the
  TransformLogger

### DIFF
--- a/build_barback/lib/src/transformer/transformer.dart
+++ b/build_barback/lib/src/transformer/transformer.dart
@@ -60,12 +60,11 @@ class BuilderTransformer implements Transformer, DeclaringTransformer {
       return;
     }
 
-    // Run the build step.
-    var logger = new Logger('${input.id}');
-    Logger.root.level = Level.ALL;
-    var logSubscription = logger.onRecord.listen((LogRecord log) {
-      if (log.loggerName != logger.fullName) return;
+    hierarchicalLoggingEnabled = true;
 
+    var logger = new Logger.detached('BuilderTransformer-${input.id}');
+    logger.level = Level.ALL;
+    var logSubscription = logger.onRecord.listen((LogRecord log) {
       if (log.level <= Level.CONFIG) {
         transform.logger.fine(_logRecordToMessage(log));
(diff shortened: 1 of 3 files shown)


build_barback/lib/src/transformer/transformer.dart
```diff
@@ -60,12 +60,11 @@ class BuilderTransformer implements Transformer, DeclaringTransformer {
       return;
     }
 
-    // Run the build step.
-    var logger = new Logger('${input.id}');
-    Logger.root.level = Level.ALL;
-    var logSubscription = logger.onRecord.listen((LogRecord log) {
-      if (log.loggerName != logger.fullName) return;
+    hierarchicalLoggingEnabled = true;
```
jakemac53: Note that build_runner also messes with this setting.... in general I really hate how this setting can't be scoped somehow to a specific set of loggers :(

build_barback/lib/src/transformer/transformer.dart
```diff
@@ -60,12 +60,11 @@ class BuilderTransformer implements Transformer, DeclaringTransformer {
...
 
+    var logger = new Logger.detached('BuilderTransformer-${input.id}');
```
jakemac53: Ah nice, never heard of this. Do we really need to enable hierarchical logging then?

build_barback/lib/src/transformer/transformer.dart
natebosch: Yeah, I *think* this should be safe since
a) we're turning it on just in time (and turning it back on if someone went behind our back since the last time). Someone would have to sneak in during an async event loop and flip it to false to mess with this
b) If you're expecting the `false` case I'm pretty sure you're also fine with the behavior of the `true` case, other than possibly some worse performance

PR #157, opened 2016-12-09
Add writeFromFuture to BuildStepImpl

natebosch:
Transform has a synchronous addOutput method, and barback Assets cannot
be read synchronously. This combination means we can't even
synchronously call the writeAsString method which would also record that
an asset needs to be written before the build step is complete. So far
this has been OK since we haven't come across cases where a Transformer
does other asynchronous work and to make the call happen after the
Builder's Future has finished. To catch the case where the call does
happen after this future add a hook to record the pending write
synchronously.

- Change some exceptions to hold an AssetId rather than an Asset since
  in some cases that is all we have, and the AssetId is the criticial
  piece of information.
- Add writeFromFuture to BuildStepImpl. Since we don't expect many use
  cases to need this don't expose it further
- Add a src import and cast to BuildStepImpl from build_barback to use
  the new method

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.6.3
+
+- Add hook for `build_barback` to write assets from a Future
+
 ## 0.6.2
 
 - Remove unused dependencies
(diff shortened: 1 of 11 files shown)


natebosch:
Travis won't pass unless I add a commit that switches to path dependencies - but I'd have to remove this again for publishing.

kevmoo:
No tests?

PR #164, opened 2017-01-13
DO NOT SUBMIT: update package:build interfaces to support reading/writing as bytes

jakemac53:
Preview of a number of changes to the apis, primarily to support reading/writing as bytes,
as this is going to inevitably be a required feature.

### New Features
- The `AssetWriter` class now has a
  `Future writeAsBytes(AssetId id, List<int> bytes)` method.
- The `AssetReader` class now has a `Future<List<int>> readAsBytes(AssetId id)`
  method.

### Breaking Changes
- The `Asset` class has been removed entirely.
- The `AssetWriter#writeAsString` signature has changed to
  `Future writeAsString(AssetId id, String contents, {Encoding encoding})`.
- The type of the `AssetWriterSpy#assetsWritten` getter has changed from an
  `Iterable<Asset>` to an `Iterable<AssetId>`.
- `BuildStep#input` has been changed to `BuildStep#inputId`, and its type has
  changed from `Asset` to `AssetId`. This means you must now use
  `BuildStep#readAsString` or `BuildStep#readAsBytes` to read the primary input,
  instead of it already being read in for you.

**Note**: The changes to `AssetReader` and `AssetWriter` also affect `BuildStep`
and other classes that implement those interfaces.

### New Features
- The `testBuilder` method now accepts `List<int>` values for both
  `sourceAssets` and `outputs`.

### Breaking Changes
- The `testBuilder` method now requires a `RecordingAssetWriter` instead of
  just an `AssetWriter` for the `writer` parameter.
- Deleted the `makeAsset` and `makeAssets` methods. There is no more `Asset`
  class so these don't really have any value any more.
- The signature of `addAssets` has changed to
  `void addAssets(Map<AssetId, dynamic> assets, InMemoryAssetWriter writer)`.
  Values of the map may be either `String` or `List<int>`.
- `InMemoryAssetReader#assets` and `InMemoryAssetWriter#assets` have changed to
  a type of `Map<AssetId, DatedValue>` from a type of
  `Map<AssetId, DatedString>`. `DatedValue` has both a `stringValue` and
  `bytesValue` getter.
- `InMemoryAssetReader` and `InMemoryAssetWriter` have been updated to implement
  the new `AssetReader` and `AssetWriter` interfaces (see the `build` package
  CHANGELOG for more details).
- `InMemoryAssetReader#cacheAsset` has been changed to two separate methods,
  `void cacheStringAsset(AssetId id, String contents)` and
  `void cacheBytesAsset(AssetId id, List<int> bytes)`.
- The `equalsAsset` matcher has been removed, since there is no more `Asset`
  class.


jakemac53:
Resolves https://github.com/dart-lang/build/issues/165

jakemac53:
Fwiw, I also considered changing `writeAsString` and `writeAsBytes` to not take an `Asset`, but just and `AssetId` and a value (`String` or `List<int>`), and then possibly adding a `writeAsset` method which just takes an `Asset` (of any type). This would be a larger breaking change though.

jakemac53:
Updated based on offline talks, removed the `Asset` class entirely to simplify/unify things.

build/CHANGELOG.md
```diff
@@ -1,3 +1,29 @@
+## 0.7.0-dev
+
+A number of changes to the apis, primarily to support reading/writing as bytes,
+as this is going to inevitably be a required feature. This will hopefully be the
+last breaking change before the `1.0` release, but it is a fairly large one.
```
natebosch: [nit] remove this comment since we have another breaking change planned

build/lib/src/asset/writer.dart
```diff
@@ -4,24 +4,33 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'asset.dart';
+import 'id.dart';
 
+/// Abstract class that can write [Asset]s.
```
natebosch: [nit] This comment isn't adding much information over the signature. We get a link to [Asset] but I don't think it's worth the comment

build/lib/src/asset/writer.dart
```diff
@@ -4,24 +4,33 @@
...
+
+  Future writeAsString(AssetId id, String contents, {Encoding encoding: UTF8});
 }
 
-/// An [AssetWriter] which tracks all assets written during it's lifetime.
+/// An [AssetWriter] which tracks all [AssetId]s written during it's lifetime.
 class AssetWriterSpy implements AssetWriter {
   final AssetWriter _delegate;
-  final List<Asset> _assetsWritten = [];
+  final List<AssetId> _assetsWritten = [];
```
natebosch: while we're in this file... this can be a Set

build/lib/src/builder/build_step.dart
```diff
@@ -26,19 +25,29 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   @override
   Future<bool> hasInput(AssetId id);
 
+  /// Reads an [Asset] by [id] as a [List<int>].
+  @override
+  Future<List<int>> readAsBytes(AssetId id);
```
natebosch: I should have noticed this earlier... but repeating these signatures here isn't adding any value. The `implements` statement is all we need here. You can move these doc comments to the other classes if they aren't already there.

build/lib/src/builder/build_step.dart
```diff
@@ -26,19 +25,29 @@ abstract class BuildStep implements AssetReader, AssetWriter {
...
+
   /// Reads an [Asset] by [id] as a [String] using [encoding].
   @override
   Future<String> readAsString(AssetId id, {Encoding encoding: UTF8});
 
-  /// Outputs an [Asset] using the current [AssetWriter], and adds [asset] to
-  /// [outputs].
+  /// Outputs an [BytesAsset] using the current [AssetWriter].
+  ///
+  /// Most `Builder` implementations should not need to `await` this Future
```
natebosch: I suppose this bit of information could be worth adding. Feel free to keep the method signature and doc comment here

build/lib/src/builder/build_step_impl.dart
```diff
@@ -43,11 +42,11 @@ class BuildStepImpl implements ManagedBuildStep {
...
 
-  BuildStepImpl(Asset input, Iterable<AssetId> expectedOutputs, this._reader,
-      this._writer, this._rootPackage, this._resolvers,
+  BuildStepImpl(AssetId inputId, Iterable<AssetId> expectedOutputs,
+      this._reader, this._writer, this._rootPackage, this._resolvers,
       {Logger logger})
-      : this.input = input,
-        this.logger = logger ?? new Logger('${input.id}') {
+      : this.inputId = inputId,
+        this.logger = logger ?? new Logger('${inputId}') {
```
natebosch: [nit] remove the {} from the string interpolation

build/pubspec.yaml
```diff
@@ -18,3 +18,9 @@ dev_dependencies:
   glob: ^1.1.0
   test: ^0.12.0
   transformer_test: ^0.2.0
+
+dependency_overrides:
```
natebosch: drop before merging

build/CHANGELOG.md
jakemac53: I am going to leave it in for now, as I think we should wait to release until we have all breaking changes done. I set up a milestone for this (0.7.0) to track other changes, so we can continue to add to the changelog for this version.

build/lib/src/asset/writer.dart
jakemac53: and `Asset` isn't even a thing any more anyways haha, done

build/lib/src/asset/writer.dart
jakemac53: done

build/lib/src/builder/build_step.dart
jakemac53: done, I left the `buildStep` specific comments here

build/lib/src/builder/build_step.dart
jakemac53: yep added this paragraph but removed the general comment

build/lib/src/builder/build_step_impl.dart
jakemac53: done

build/pubspec.yaml
jakemac53: Actually, I was planning on leaving these in until the next release. Otherwise no tests will pass etc.

jakemac53: I added an additional issue to the milestone to remove these

jakemac53:
Ok, I pushed a commit that updates `build_test` now as well, PTAL.

This is necessary to get tests and things passing, I will update the other packages separately.

build_test/lib/src/assets.dart
```diff
@@ -14,22 +14,15 @@ AssetId makeAssetId([String assetIdString]) {
...
-    var asset = makeAsset(idString, content);
-    assets[asset.id] = asset;
+void addAssets(Map<AssetId, dynamic> assets, InMemoryAssetWriter writer) {
+  assets.forEach((id, value) {
+    if (value is String) {
+      writer.assets[id] = new DatedString(value);
+    } else if (value is List<int>) {
+      writer.assets[id] = new DatedBytes(value);
+    } else {
+      throw 'asset values must be of type `String` or `List<int>`, got '
```
natebosch: throw new ArgumentError?

Issue #177, opened 2017-01-27
Resolver should maybe throw rather than return null for non-library assets


natebosch:
Today if we try to resolve the LibraryElement for an asset which isn't a Dart library we return null.
This is fine when it is handled by Builder authors - but if it isn't handled explicitly the error message isn't helpful - usually a null reference exception.

We should consider whether we want to instead throw an exception for this case.

IMO we shouldn't force Builder authors to use a try/catch so we should also add a method to check whether an asset is a Dart library.

PR #178, opened 2017-01-27
Add `isLibrary` method to `Resolver`

natebosch:
Fixes https://github.com/dart-lang/build/issues/177

- Add isLibrary to check whether a source file is a Dart library
- Throw instead of return null when the asset is not a Dart library
- Implement the new interface in build_barback
- Clean up some repeated Doc comments and add @override instead

### DIFF
--- a/build_barback/CHANGELOG.md
+++ b/build_barback/CHANGELOG.md
@@ -6,6 +6,7 @@ package, and the removal of the `Asset` class.
 ### New Features
 - `BuilderTransformer` now supports wrapping transformers that read or write
   their inputs as bytes.
+- The Resolver implementation can check whether an Asset is a Library.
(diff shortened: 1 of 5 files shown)


build_barback/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ package, and the removal of the `Asset` class.
 ### New Features
 - `BuilderTransformer` now supports wrapping transformers that read or write
   their inputs as bytes.
+- The Resolver implementation can check whether an Asset is a Library.
```
jakemac53: Maybe mention the `isLibrary` name directly?

natebosch: Done

Issue #193, opened 2017-02-03
PackageGraph chokes on 'hosted' dependencies


lpodmolik:
PackageGraph only understands pub, git and path dependencies.  In our projects we often use hosted dependencies like this:

```
some_package:
    hosted:
        name: some_package
        url: http://bld-01.acme.com:8085
    version: ^0.1.3
```

In checked mode this results in:

```
Unhandled exception:
'package:build_runner/src/package_graph/package_graph.dart': Failed assertion: line 162 pos 10: 'source.keys.length == 1': is not true.
```

Unchecked, it hits the default case in the subsequent switch statement on type, and throws a different exception.

PR #195, opened 2017-02-06
add support for hosted dependencies

jakemac53:
Fixes https://github.com/dart-lang/build/issues/193

### DIFF
--- a/build_runner/lib/src/package_graph/package_graph.dart
+++ b/build_runner/lib/src/package_graph/package_graph.dart
@@ -153,23 +153,24 @@ class PackageNode {
 
 /// The type of dependency being used. This dictates how the package should be
 /// watched for changes.
-enum PackageDependencyType { pub, github, path }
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -153,23 +153,24 @@ class PackageNode {
...
+    switch (key) {
+      case 'git':
+        return PackageDependencyType.github;
+      case 'hosted':
+        return PackageDependencyType.hosted;
+      case 'path':
+        return PackageDependencyType.path;
+    }
   }
+  throw 'Unable to determine dependency type:\n$source';
```
natebosch: Would this be valid?

```
dependencies:
  foo:
    version: ^1.0.0
```

I don't think it's handled here.

jakemac53: Doesn't look like it, I get an error that it doesn't have a source specified.

PR #198, opened 2017-02-10
fix #197

### LINKED ISSUE #197: Why build does not accept input at root folder ?
I have a package that need a config file (ex: config.yaml) directly on my root folder.

```
-> ./config.yaml
-> ./lib/..
-> ./bin/..
...
```

I am generating code from that file, but `build` does not accept that input, why ?

I know it is a choice by looking at the code
https://github.com/dart-lang/build/blob/0b4f98e568399af8df3d4d7d32100b87d26e517a/build_runner/lib/src/generate/build_impl.dart#L501

lejard-h:


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -498,7 +500,6 @@ class BuildImpl {
   bool _isValidInput(AssetId input) {
     var parts = path.split(input.path);
     // Files must be in a top level directory.
```
natebosch: remove this comment as well

lejard-h:
it should be good now :)

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -478,6 +478,8 @@ class BuildImpl {
     for (var input in allInputs) {
       if (_isValidInput(input)) {
         _inputsByPackage[input.package].add(input);
+      } else {
```
jakemac53: I don't think we want this else, we just want to skip the invalid inputs at this stage. Users aren't trying to read inputs yet, we are just collecting the list of possible inputs (and skipping the invalid ones).

Also, given the globs above it doesn't seem like this should ever be hit? Maybe this whole section can be removed.

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -465,7 +465,7 @@ class BuildImpl {
     }
 
     var inputSets = packages.map((package) => new InputSet(
-        package, [package == _packageGraph.root.name ? '**/*' : 'lib/**']));
+        package, [package == _packageGraph.root.name ? '**' : 'lib/**']));
```
jakemac53: Seems like a pre-existing issue, but this looks like its collecting a lot of things we don't actually want (such as hidden dirs like `.git` etc). Technically you could consume these I guess, but it seems like a bad idea. Not sure the best way to get around this though

I filed https://github.com/dart-lang/build/issues/200 to investigate this, as I don't think this change makes it any worse :).

natebosch:
Thanks for the fix!

Issue #191, opened 2017-02-02
Add some re-usable utilities somewhere or new package


matanlurey:
Almost every code generator/builder wants to do the following things, easily. I've added really simple proof of concept APIs that could be added to avoid code duplication, but open to other suggestions. We could call this `build_common`, or put in one of the existing packages? No idea.

- [x] Setup `package:logging` to emit to stdout/err 

```dart
import 'package:build_common/build_common.dart' as build_common;

void main() {
  build_common.initLogging();
}
```

- [x] Add a top-level logging API that can be imported/shared without passing `BuildStep`

```dart
import 'package:build_common/build_common.dart' as build_common;

String generateCode() {
  if (somethingBad) {
    build_common.logger.warning('Something bad occurred!');
  }
}
```

- [ ] Use `package:stack_trace`, at least in dev-mode, to get good stack traces.

```dart
import 'package:build_common/build_common.dart' as build_common;

void main() {
  build_common.runGuarded(() {
    doActualGeneration();
  }, onUnhandledException: (e, s) {
    ...
  });
}
```

Could be used in combination with the above logging issue to write unhandled exceptions.

```dart
import 'package:build_common/build_common.dart' as build_common;

void main() {
  build_common.initiLogging();
  build_common.runGuarded(() {
    doActualGeneration();
  });
  // Now all unhandled exceptions get written as "severe" to logger?
}
```

/cc @yjbanov

natebosch:
Not sure on "Add a top-level logging API that can be imported/shared without passing BuildStep"

Logs need to be tied to a particular instance of a BuildStep. Currently we don't run each BuildStep in it's own zone in any implementation - but that would be required if we want to have this globally accessible log available.

matanlurey:
> Not sure on "Add a top-level logging API that can be imported/shared without passing BuildStep"
> 
> Logs need to be tied to a particular instance of a BuildStep. Currently we don't run each BuildStep in it's own zone in any implementation - but that would be required if we want to have this globally accessible log available.

This could be required then that you use this top-level zone, i.e. the `runGuarded` or perhaps `runBuildZone` or something that makes sense. @yjbanov does something like this in `inject` internally if you want to see an example.

> 
> Edit: As a consequence of this it wouldn't quite as optional. I'd think we'd need to define the behavior in `package:build` and force all implementations that run builders to guarantee behavior.
> 

Hmm. I see what you mean. I do think this would be very useful, though.

natebosch:
[...]

PR #202, opened 2017-02-10
Add a top-level `log` getter

natebosch:
Discussed in #191

- Run builds inside a zone to hold the logger
- Add a `log` getter which pulls from the zone-scoped logger
- Deprecate ManagedBuildStep - prefer for all use cases to go through
  runBuilder so they can get the log
- Point `BuildStep.logger` at the new top-level getter and mark it
  deprecated

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 0.7.1
+
+- Add a top-level `log` getter which is scoped to running builds and can be used
+  anywhere within a build rather than pass around a logger. This replaces the
+  `BuildStep.logger` field.
+- Deprecate `BuildStep.logger` - it is replaced by `log`
+- Deprecate `ManagedBuildStep`, all build runs should go through `runBuilders`.
+
 ## 0.7.0
 
 A number of changes to the apis, primarily to support reading/writing as bytes,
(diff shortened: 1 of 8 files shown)


natebosch:
I'm not catching `print`, for now we'll still do that in `_bazel_codegen` only. I think the importance is lower since it should be easier for more use-cases to use the log instead of 'print'

build/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.7.1
+
+- Add a top-level `log` getter which is scoped to running builds and can be used
+  anywhere within a build rather than pass around a logger. This replaces the
```
nshahan: (nit) pass -> passing

natebosch: done

nshahan: In practice what would be the scope of the top level getter log? It is within the build steps right? Otherwise the name seems too generic.

nshahan:
In practice what would be the scope of the top level getter log? It is within the build steps right? Otherwise the name seems too generic.

natebosch:
It's within the scope of a Builder and across steps (different inputs run together).

See the discussion on naming here: https://github.com/dart-lang/build/issues/191#issuecomment-277110402

Issue #192, opened 2017-02-02
AssetId.parse but for relative imports


matanlurey:
For example, someone has the directive:

```dart
import 'some/relative/path.dart';
```

I'd like to create an `AssetId` representing that file in my builder so I can read it.

jakemac53:
Ya this would be nice, what about a `AssetId relative(AssetId to)` instance method on `AssetId`?

matanlurey:
Here is my hack today:

```dart
  AssetId assetId;
    try {
      assetId = new AssetId.parse(directiveUri);
    } on FormatException catch (_) {
      assetId = new AssetId(
        p.normalize(source.package),
        p.join(source.path, directiveUri),
      );
    }
```

(Yes it could be better and not use try/catch, but just playing around)

matanlurey:
+1 to @jakemac53's suggestion.

jakemac53:
We could also add a more general `AssetId.resolve(String uri, {AssetId from})` constructor, which would support absolute `package:` uris as well as relative ones if `from` is provided.

natebosch:
At one point I had a change that added the concept of a 'canonical' Uri. With that solution this use case would have looked like:

```dart
var canonicalUri = normalize(directiveUri, from: buildStep.inputId.asUri);
var assetId = new AssetId.fromUri(canonicalUri);
```

matanlurey:
On a related note, I forgot to add `lib/` when doing this manually.

Hopefully `new AssetId.from` or whatever we come up with handles this for me. Easy to get wrong.

PR #207, opened 2017-02-15
Add AssetId.resolve

matanlurey:
Closes https://github.com/dart-lang/build/issues/192.

### DIFF
--- a/build/lib/src/asset/id.dart
+++ b/build/lib/src/asset/id.dart
@@ -54,7 +54,7 @@ class AssetId implements Comparable<AssetId> {
             pathos.joinAll(parsedUri.pathSegments.skip(1)));
       }
       throw new UnsupportedError(
-        'Cannot resolve $uri; only "package" supported',
+        'Cannot resolve $uri; only "package" supported'
       );
     }
     if (from == null) {
--- a/build/lib/src/asset/id.dart
+++ b/build/lib/src/asset/id.dart
@@ -22,12 +22,56 @@ class AssetId implements Comparable<AssetId> {
   /// Gets the file extension of the asset, if it has one, including the ".".
   String get extension => pathos.extension(path);
 
-  /// Creates a new AssetId at [path] within [package].
+  /// Creates a new [AssetId] at [path] within [package].
   ///
   /// The [path] will be normalized: any backslashes will be replaced with
   /// forward slashes (regardless of host OS) and "." and ".." will be removed
   /// where possible.
   AssetId(this.package, String path) : path = _normalizePath(path);
+
+  /// Creates a new [AssetId] from an [uri].
+  ///
+  /// This gracefully handles `package:` URIs, which is useful when creating an
+  /// [AssetId] from an `import` or `export` directive pointing to a package:
+  /// ```dart
+  /// AssetId assetOfDirective(UriReferencedElement element) {
+  ///   return new AssetId.resolve(element.uri);
+  /// }
+  /// ```
+  ///
+  /// Optionally, specifying the origin asset ([from]) will also allow resolving
+  /// relative to another asset - otherwise an [ArgumentError] will be thrown.
+  /// ```dart
+  /// AssetId assetOfDirective(AssetId origin, UriReferencedElement element) {
+  ///   return new AssetId.resolve(element.uri, from: origin);
+  /// }
+  /// ```
+  factory AssetId.resolve(String uri, {AssetId from}) {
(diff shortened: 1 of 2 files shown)


build/lib/src/asset/id.dart
```diff
@@ -22,13 +22,57 @@ class AssetId implements Comparable<AssetId> {
...
   /// where possible.
   AssetId(this.package, String path) : path = _normalizePath(path);
 
+  /// Creates a new [AssetId] from an [uri].
+  ///
+  /// This gracefully handles `package:` URIs, which is useful when creating an
+  /// [AssetId] from an `import` or `export` directive pointing to a package:
+  /// ```dart
+  /// AssetId assetOfDirective(UriReferencedElement element) {
+  ///   return new AssetId.resolve(element.uri);
```
jakemac53: Could we pass a sensible from here using the containing library?

build/lib/src/asset/id.dart
```diff
@@ -22,13 +22,57 @@ class AssetId implements Comparable<AssetId> {
...
+  /// }
+  /// ```
+  factory AssetId.resolve(String uri, {AssetId from}) {
+    final parsedUri = Uri.parse(uri);
+    // If a dart:, package:, or file: URI was used, assume an absolute path.
+    if (parsedUri.hasScheme) {
+      if (parsedUri.scheme == 'package') {
+        return new AssetId(
+          parsedUri.pathSegments.first,
+          pathos.joinAll(parsedUri.pathSegments.skip(1)),
```
jakemac53: Nit: we don't use trailing comments in this package for constructors. It's really unfortunate that this "feature" introduced effectively multiple styles 😢.

build/lib/src/asset/id.dart
```diff
@@ -22,13 +22,57 @@ class AssetId implements Comparable<AssetId> {
...
+  /// Creates a new [AssetId] from an [uri].
+  ///
+  /// This gracefully handles `package:` URIs, which is useful when creating an
+  /// [AssetId] from an `import` or `export` directive pointing to a package:
+  /// ```dart
+  /// AssetId assetOfDirective(UriReferencedElement element) {
+  ///   return new AssetId.resolve(element.uri);
+  /// }
+  /// ```
+  ///
```
kevmoo: Extra new-line?

build/lib/src/asset/id.dart
matanlurey: No worries, sorry, bad habit from Angular where we have a crazy amount of args.

build/lib/src/asset/id.dart
matanlurey: `element.librarySource.source.uri.toString()`?

build/lib/src/asset/id.dart
matanlurey: Done

build/lib/src/asset/id.dart
jakemac53: So `element.librarySource.source` used to be an `AssetBasedSource`, which had an `id` field. However I think with the summaries resolver that may not be the case which makes things a bit more complicated. Would you just file an issue to support more than `package:` URIs and we can follow up.

matanlurey: Sure. Here, right?

matanlurey: https://github.com/dart-lang/build/issues/208

matanlurey:
@natebosch Anything else? Ok to squash and merge?

build/lib/src/asset/id.dart
```diff
@@ -22,13 +22,49 @@ class AssetId implements Comparable<AssetId> {
...
+  /// Optionally, specifying the origin asset ([from]) will also allow resolving
+  /// relative to another asset - otherwise an [ArgumentError] will be thrown.
+  /// ```dart
+  /// AssetId assetOfDirective(AssetId origin, UriReferencedElement element) {
+  ///   return new AssetId.resolve(element.uri, from: origin);
+  /// }
+  /// ```
+  factory AssetId.resolve(String uri, {AssetId from}) {
+    final parsedUri = Uri.parse(uri);
+    // If a dart:, package:, or file: URI was used, assume an absolute path.
```
natebosch: [nit] this comment doesn't look like it matches behavior. Maybe remove it?

matanlurey: Done.

matanlurey:
Thanks! This will help :)

Issue #211, opened 2017-02-17
Update PackageGraph to allow Flutter as a dependency


matanlurey:

natebosch:
Can you clarify this?

matanlurey:
We choke right now when we see the `sdk: ...` property.

I am working with Eric to get an example of `.packages` for a Flutter app so we can support `build_runner`.

matanlurey:
We need to publish a new build_runner to fix this.

@jakemac53 @natebosch @kevmoo

natebosch:
published

PR #212, opened 2017-02-17
Add support for Flutter packages.

matanlurey:
For now, treat `sdk:` as ~ the same thing as `path:`.

This will break down once Flutter is published another way, but allows `build_runner` to start working without crashing when running within a Flutter package. We will (eventually) need to figure out additional things, like how to use the right analyzer to recognize `dart:ui`, but this is a start.

Also ran `dartfmt`, and added a better error message when a dependency is not found (not NPE).

Closes https://github.com/dart-lang/build/issues/211.


natebosch:
LGTM but Jake is more familiar

eseidelGoogle:
I don't really have any context here.  Looks reasonable.  If you need stronger technical consideration consider asking cbracken, devoncarew or abarth.

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -92,7 +92,11 @@ class PackageGraph {
       deps.forEach((name, source) {
         var dep = nodes[name];
         if (dep == null) {
-          var pubspec = _pubspecForUri(packageLocations[name]);
+          var uri = packageLocations[name];
```
jakemac53: nit: move this error detection to the `_pubspecForUri` method?

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -167,6 +171,7 @@ PackageDependencyType _dependencyType(source) {
       case 'hosted':
         return PackageDependencyType.hosted;
       case 'path':
+      case 'sdk': // Until Flutter supports another type, assume same as path.
```
jakemac53: I have no context here, is it actually equivalent to a path dependency? Or is it coming from pub?

matanlurey: That is my understanding. Maybe @cbracken or @yjbanov or @devoncarew know better.

matanlurey:
@jakemac53 and @natebosch - OK with merging?

kevmoo:
@natebosch ?

build_runner/lib/src/package_graph/package_graph.dart
abarth: It means that it comes from the Flutter SDK, which the developer downloads.

PR #225, opened 2017-03-16
Add lib/ when resolving package:

matanlurey:
See occurrences of this bug in `angular2`: https://github.com/dart-lang/angular2/blob/2506ea4977ec4c4e6fefd142dac0f69b409ca8a0/lib/src/source_gen/template_compiler/template_processor.dart#L45 (as an example).

When `package:app/app.dart` is resolved, it should point to `app|lib/app.dart`.


matanlurey:
@natebosch Will add before merging.

natebosch:
SG

Issue #223, opened 2017-03-15
Add glob or directory listing API in AssetReader


natebosch:
This first came up in https://github.com/dart-lang/build/issues/213

A glob api would be the most flexible option since directory listing can easily be done with a glob.

Proposed API:

`List<AssetId> findAssets(String glob)`

I have it take in a String or glob from `package:glob`: I'm not sure it's worth the extra set of dependencies in `build` even if all implementations choose to use package:glob anyway.

jakemac53:
Should we also allow a package name?

natebosch:
Allowing a package name should be feasible, and we can default to the root package...

`List<AssetId> findAssets(String glob, {String package})`

matanlurey:
+1 that API, only nit is maybe `Iterable<AssetId>` in the case someone is going to try and find the first file that matches something?

jakemac53:
It is worth noting that we can't support globbing other packages in pub, so we might want to hold off on actually adding that support until it is needed?

If we do decide to add it, then I think we should also consider adding sync apis (which also won't work in pub).

natebosch:
I retract my earlier statement... `build` already transitively depends on `glob` so for the sake of stricter API (and potentially better performance) I'm going to use

```
Iterable<AssetId> findAssets(Glob glob);
```

natebosch:
Some offline discussion:

This API is going to mean that all implementations of `AssetReader` will need to have an understanding of a 'default' or 'root' package. This shouldn't be too much of a problem in general but will cause a few bits of weirdness:

- In tests the `InMemoryAssetReader` will either always need a default package, or will need to `throw` if there is no default package. Should be doable.
- It'll be harder to pass in a `RunnerAssetReader` for building with build_runner. We'll likely have to change this to a factory `RunnerAssetReader forPackageGraph(PackageGraph graph)`

natebosch:
With https://github.com/dart-lang/build/pull/253 this is now rolled out everywhere

PR #228, opened 2017-03-21
Preemptively implement findAssets API

natebosch:
We will be implementing a findAssets API in AssetReader soon. Since
build_barback can't implement the api including a `packageName`
argument, but build_runner needs this we'll still have a
`RunnerAssetReader` interface for now.

First step for #223 

- Refactor listAssetIds to iterate over input sets and globs in
  build_impl.dart rather than the asset reader to facility reuse of the
  more general findAssets API.
- Update InMemoryAssetReader to the new API
- Update tests against listAssetIds to instead test findAssetIds

### DIFF
--- a/build_runner/test/common/in_memory_reader.dart
+++ b/build_runner/test/common/in_memory_reader.dart
@@ -7,6 +7,7 @@ import 'dart:mirrors';
 import 'package:build/build.dart';
 import 'package:build_runner/build_runner.dart';
 import 'package:build_test/build_test.dart';
+import 'package:glob/glob.dart';
 
 class InMemoryRunnerAssetReader extends InMemoryAssetReader
     implements RunnerAssetReader {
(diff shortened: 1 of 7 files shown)


natebosch:
I'll bump to a non `-dev` release once the build package is updated and I can change the pubspec constraint.

build_runner/test/common/in_memory_reader.dart
```diff
@@ -7,22 +7,16 @@ import 'dart:mirrors';
...
-  Stream<AssetId> listAssetIds(Iterable<InputSet> inputSets) async* {
-    for (var id in assets.keys) {
-      var matches = inputSets.any((inputSet) {
-        if (inputSet.package != id.package) return false;
-        return inputSet.globs.any((glob) => glob.matches(id.path));
-      });
-      if (matches) yield id;
-    }
-  }
+  Iterable<AssetId> findAssets(Glob glob, {String packageName}) => assets.keys
```
jakemac53: `assert(packageName != null)`? In general this exposes a bit of a weirdness in the api I think. A `BuildStep` definitely has a default package, but that isn't necessarily the case for an arbitrary `AssetReader`. Maybe it doesn't matter though, and we do want them to share the same api.

jakemac53: Ah looking at the other PR I see this is only available for `RunnerAssetReader` anyways, so nvm.

natebosch: I believe all implementations of AssetReader today *could* have a default package.

With the upcoming change in `build` the `InMemoryAssetReader` is going to end up needing a default package. I think that should work out...

PR #232, opened 2017-03-24
Add findAssets to test reader implementations

natebosch:
Change the signature of InMemoryAssetReader to take an optional root
package and use named parameters instead of positional optional
parameters.

### DIFF
--- a/build_test/lib/src/stub_reader.dart
+++ b/build_test/lib/src/stub_reader.dart
@@ -5,6 +5,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:build/build.dart';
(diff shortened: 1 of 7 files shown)


build_test/lib/src/stub_reader.dart
```diff
@@ -15,4 +16,7 @@ class StubAssetReader implements AssetReader {
 
   @override
   Future<String> readAsString(_, {Encoding encoding}) => new Future.value(null);
+
+  @override
+  Iterable<AssetId> findAssets(Glob glob) => const [];
```
jakemac53: I think this could even just return null

Issue #168, opened 2017-01-17
Need a way to test against real dependencies


natebosch:
build_test makes it easy to test with *all* fake sources, but there is no way to test with a fake primary input but real dependencies. Currently we need clunky end-to-end tests.

natebosch:
Ideally this should work with the same test with and without bazel.

matanlurey:
Current API:

```dart
Future testBuilder(
  Builder builder,
  Map<String, dynamic> sourceAssets, {
  Set<String> generateFor,
  bool isInput(String assetId),
  String rootPackage,
  RecordingAssetWriter writer,
  Map<String, dynamic> outputs,
  void onLog(LogRecord log)
})
```

Potential future API (just a scratchboard):

```dart
Future testBuilderOn(
  Builder builder,
  String inputName,
  dynamic inputContents, {
  Iterable<String> dependencies,
  ...
})
```

Idea for use:

```dart
testBuilderOn(myBuilder, 'input.dart', '''
  import 'package:angular2/angular2.dart';

  class Foo {}
''', dependencies: [
  'package:angular2/**/*.dart',  
]);
```

(I realize this isn't sound, just want to start a conversation)

--- 

There is also the possibility of just exposing a `TestBuildStep`, i.e:

```dart
main() {
  TestBuildStep buildStep;

  setUp(() {
    buildStep = new TestBuildStep(...)
  });

  test('CopyBuilder should copy all primary inputs', () async {
    buildStep.addPrimaryInput('package:a/a.dart', '// Hello World');
    var buildResult = await testBuilderWith(const CopyBuilder(), buildStep);
    expect(buildResult, wroteFile('package:a/a.dart.copy', '// Hello World'));
  });
}
```

This actually might be more immediately useful for AngularDart, but will require more work - i.e. a `TestBuildStep` API.

natebosch:
We have some ideas around this but any solution we can come up with is either:

- very complex/hacky
- requires a different setup on bazel vs pub

matanlurey:
Even something like a `TestBuildStep` @natebosch?

jakemac53:
What if we required you to have a physical dart import to anything you wanted to load, and we used the dart vm itself to resolve the imports? I don't know 100% that we have all the features needed available to us, but it might be worth trying...

matanlurey:
FWIW, I'm able to get bazel-only testing by doing the following. Just thought I'd share so we can think about how to make a build-system independent API:

```dart
main() {
[...]

PR #239, opened 2017-03-29
Add PackageResolver support.

matanlurey:
Partial work towards https://github.com/dart-lang/build/issues/168.

It looks like in my testing that `PackageResolver.current` receives the appropriate package configuration (in the Dart VM) for both pub and bazel executables. If this holds I should be able to use it to handle reading files in upcoming PRs.

(Happy to make this package-private if you are worried about use)

### DIFF
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -32,13 +32,7 @@ class PackageAssetReader implements AssetReader {
   }
 
   File _resolve(AssetId id) {
-    String path = id.path;
-    // The PackageResolver API can only resolve files in lib/.
-    if (path.startsWith('lib/')) {
-      path = path.substring('lib/'.length);
-      return new File.fromUri(_packageResolver.urlFor(id.package, path));
-    }
-    var nonLibPath = '${_packageResolver.packagePath(id.package)}/$path';
+    var nonLibPath = '${_packageResolver.packagePath(id.package)}/${id.path}';
     return new File(nonLibPath);
   }
 
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -9,7 +9,7 @@ import 'dart:io';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 import 'package:package_resolver/package_resolver.dart';
-import 'package:path/path.dart' as path;
+import 'package:path/path.dart' as p;
 
 /// Resolves using a [SyncPackageResolver] before reading from the file system.
 class PackageAssetReader implements AssetReader {
@@ -33,7 +33,7 @@ class PackageAssetReader implements AssetReader {
   }
 
   File _resolve(AssetId id) =>
-      new File(path.join(_packageResolver.packagePath(id.package), id.path));
+      new File(p.join(_packageResolver.packagePath(id.package), id.path));
 
   @override
   Iterable<AssetId> findAssets(Glob glob) {
@@ -48,7 +48,7 @@ class PackageAssetReader implements AssetReader {
   }
 
   AssetId _toAsset(String assetPath, String rootPackagePath) {
-    var relative = path.relative(assetPath, from: rootPackagePath);
+    var relative = p.relative(assetPath, from: rootPackagePath);
     return new AssetId(_rootPackage, relative);
   }
 
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -9,6 +9,7 @@ import 'dart:io';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 import 'package:package_resolver/package_resolver.dart';
+import 'package:path/path.dart' as path;
 
 /// Resolves using a [SyncPackageResolver] before reading from the file system.
 class PackageAssetReader implements AssetReader {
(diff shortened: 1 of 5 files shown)


build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  File _resolve(AssetId id) {
+    String path = id.path;
+    // The PackageResolver API can only resolve files in lib/.
+    if (path.startsWith('lib/')) {
+      path = path.substring('lib/'.length);
+      return new File.fromUri(_packageResolver.urlFor(id.package, path));
+    }
+    var nonLibPath = '${_packageResolver.packagePath(id.package)}/$path';
```
matanlurey: Yep. Teaches me to code late at night. Done.

build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,72 @@
...
+  const PackageAssetReader(this._packageResolver, [this._rootPackage]);
+
+  /// A reader that can resolve files known to the current isolate.
+  static Future<PackageAssetReader> currentIsolate({String rootPackage}) async {
+    var resolver = PackageResolver.current;
+    return new PackageAssetReader(await resolver.asSync, rootPackage);
+  }
+
+  File _resolve(AssetId id) {
+    String path = id.path;
```
jakemac53: Lets make this function platform independent, something like:

```dart
File _resolve(AssetId id) {
  var parts = path.split(id.path);
  if (parts.first == 'lib') {
    return new File.fromUri(_packageResolver.urlFor(id.package, path.joinAll(parts.skip(1))));
  }
  var nonLibPath = path.joinAll([_packageResolver.packagePath(id.package)]..addAll(parts));
  return new File(nonLibPath);
}
```

build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,72 @@
...
+    }
+    var rootPackagePath = _packageResolver.packagePath(_rootPackage);
+    return glob
+        .listSync(root: rootPackagePath)
+        .where((entity) => entity is File)
+        .map((file) => _toAsset(file.path, rootPackagePath, glob));
+  }
+
+  AssetId _toAsset(String path, String rootPackagePath, Glob glob) {
+    var relative = glob.context.relative(path, from: rootPackagePath);
```
jakemac53: I would just import `package:path/path.dart` and use that instead of the glob.context (then you don't need the glob parameter).

build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  @override
+  Future<bool> hasInput(AssetId id) async => _resolve(id).exists();
+
+  @override
+  Future<List<int>> readAsBytes(AssetId id) => _resolve(id).readAsBytes();
+
+  @override
+  Future<String> readAsString(AssetId id, {Encoding encoding: UTF8}) async {
```
jakemac53: Imo it's fine to use `=>` syntax here to be consistent with the other lines above even though you will have line wrap.

build_test/test/package_reader_test.dart
```diff
@@ -0,0 +1,45 @@
...
+  group('$PackageAssetReader', () {
+    AssetReader reader;
+
+    final buildTest = new AssetId('build_test', 'lib/build_test.dart');
+    final buildMissing = new AssetId('build_test', 'lib/build_missing.dart');
+    final thisFile = new AssetId('build_test', 'test/package_reader_test.dart');
+
+    setUp(() async {
+      reader = await PackageAssetReader.currentIsolate(
+        rootPackage: 'build_test',
```
jakemac53: I wonder if this could be inferred? Might not be worth the effort/hacks that would be involved though.

build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,66 @@
...
+  const PackageAssetReader(this._packageResolver, [this._rootPackage]);
+
+  /// A reader that can resolve files known to the current isolate.
+  static Future<PackageAssetReader> currentIsolate({String rootPackage}) async {
+    var resolver = PackageResolver.current;
+    return new PackageAssetReader(await resolver.asSync, rootPackage);
+  }
+
+  File _resolve(AssetId id) {
+    var nonLibPath = '${_packageResolver.packagePath(id.package)}/${id.path}';
```
jakemac53: Now that this is simplified you could just use [p.canonincalize](https://www.dartdocs.org/documentation/path/1.4.1/path/canonicalize.html) to get a platform independent path

build_test/test/package_reader_test.dart
matanlurey: Yeah I assumed we could figure it out, but seemed easy enough to define it for now and we can always fill in if null in the future.

build_test/lib/src/package_reader.dart
matanlurey: Done.

build_test/lib/src/package_reader.dart
matanlurey: Done.

build_test/lib/src/package_reader.dart
matanlurey: Done.

build_test/lib/src/package_reader.dart
matanlurey: Versus p.join?

build_test/lib/src/package_reader.dart
```diff
@@ -0,0 +1,64 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+import 'package:package_resolver/package_resolver.dart';
+import 'package:path/path.dart' as path;
```
jakemac53: I would import this `as p`, `path` is a very common local variable or parameter name and it tends to get highly confusing :)

build_test/lib/src/package_reader.dart
jakemac53: I don't think that p.join will fix up the slashes `id.path` portion.

PR #242, opened 2017-03-30
Add MultiAssetReader (to pkg/build_test)

matanlurey:
Partial work towards https://github.com/dart-lang/build/issues/168:

For running test code that is a combination of "fake" (stub) code and "real" code (such as dependencies you'd expect to be resolved, such as annotations, other classes, types, etc), this will allow something like:

```dart
runBuilder(..., new MutliAssetReader([
  new InMemoryAssetReader(...), // Fakes/Stubs
  new PackageAssetReader(...), // Real code on disk
])
```

I could see this having value in `pkg/build` (i.e. `new AssetReader.fromMulti(...)`), but I figure it's probably better to not add to the public API until it's stabilized and supports the #168 use-case, so I've kept it package private to `build_test`.

Test cases look good, so next step will be using it with `testBuilder` in some more e2e-y way.

### DIFF
--- a/build_test/lib/src/multi_asset_reader.dart
+++ b/build_test/lib/src/multi_asset_reader.dart
@@ -1,0 +1,61 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:collection/collection.dart';
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+
+/// An [AssetReader] that delegates to multiple other asset readers.
+///
+/// [MultiAssetReader] attempts to check every provided [AssetReader] to see if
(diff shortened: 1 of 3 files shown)


build_test/lib/src/multi_asset_reader.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  @override
+  Future<String> readAsString(AssetId id, {Encoding encoding: UTF8}) async =>
+      (await _readerWith(id)).readAsString(id, encoding: encoding);
+
+  /// Returns all readable assets matching [glob] under the root package.
+  ///
+  /// **NOTE**: This is a combined view of all provided readers.
+  @override
+  Iterable<AssetId> findAssets(Glob glob) => new CombinedIterableView(
```
natebosch: Will it ever be possible for the same `AssetId` to show up twice? Do we care?

I don't think it should matter but might want to call it out in the comment

matanlurey: Ostensibly it could if you tried to override an existing file in a stub. 

I could do a `.toSet(...)` with a custom equality, but maybe it's easier to call out for now.

PR #244, opened 2017-03-30
Add a top-level `resolveSource` to pkg/build_test

matanlurey:
Branched from https://github.com/dart-lang/build/pull/242.

Added a simple `resolve_source_test.dart`. This should invalidate hacks I added for:
https://github.com/dart-lang/angular2/blob/master/test/meta/types_test.dart

Partial work towards https://github.com/dart-lang/build/issues/168. We could consider it closed since you could always spin up a `MultiAssetReader` with a `PackageAssetReader` and get the same functionality with `testBuilder`. Up to you folks.

I'm planning on adding a `_bazel_codegen` e2e example as well.


build_test/lib/src/package_reader.dart
```diff
@@ -62,3 +54,25 @@ class PackageAssetReader implements AssetReader {
   Future<String> readAsString(AssetId id, {Encoding encoding: UTF8}) =>
       _resolve(id).readAsString(encoding: encoding);
 }
+
+/// Returns all assets that match [glob] in [package] with a [path].
+///
+/// **INTERNAL ONLY**.
```
natebosch: should this be private?

build_test/lib/src/resolve_source.dart
```diff
@@ -0,0 +1,85 @@
...
+
+import 'in_memory_reader.dart';
+import 'in_memory_writer.dart';
+import 'multi_asset_reader.dart';
+import 'package:package_resolver/package_resolver.dart';
+import 'package_reader.dart';
+
+/// Returns a future that completes with a source [Resolver] for [inputSource].
+///
+/// Example use:
```
natebosch: should the example show code that does something with the result?

build_test/lib/src/resolve_source.dart
```diff
@@ -0,0 +1,85 @@
...
+/// ```dart
+/// await resolveSource(r'''
+///   import 'dart:collection';
+///
+///   external Map createMap();
+/// ''', rootPackage: 'some_pkg')
+/// ```
+///
+/// **NOTE**: All `package` dependencies are resolved using [PackageAssetReader]
+/// - by default, [PackageAssetReader.currentIsolate]. A custom [reader] may be
```
natebosch: how would one provide a custom reader? I don't see it as an argument

build_test/lib/src/resolve_source.dart
```diff
@@ -0,0 +1,85 @@
...
+    rootPackage: inputId.package,
+  );
+  // We don't care about the results of this build.
+  // ignore: unawaited_futures
+  runBuilder(
+    builder,
+    inputs,
+    new MultiAssetReader([
+      inMemory,
+      reader,
```
natebosch: [nit] would prefer if this didn't use trailing comma if it can fit on 1 line

build_test/lib/src/package_reader.dart
matanlurey: It is used in two different files (libraries).

build_test/lib/src/resolve_source.dart
matanlurey: Added a bit more.

build_test/lib/src/resolve_source.dart
matanlurey: Sorry meant resolver. Decided not to expose a custom reader quite yet.

build_test/lib/src/resolve_source.dart
matanlurey: Done.

build_test/lib/src/package_reader.dart
natebosch: not in this change...

build_test/lib/src/package_reader.dart
```diff
@@ -62,3 +54,25 @@ class PackageAssetReader implements AssetReader {
...
+Iterable<AssetId> globAssets(String package, String path, Glob glob) {
+  if (package == null) {
```
jakemac53: These are overkill, they aren't marked as optional args so if somebody wants to pass in nulls it is fine for it to blow up.

matanlurey:
Sg. I'll make them assert for sanity - the issue is if the package doesn't
resolve it sort of half works and the error you get is hard to understand.

build_test/lib/src/package_reader.dart
matanlurey: Ah, I didn't end up needing that code. Made it private :)

matanlurey:
@jakemac53 doe this look fine for you? No rush, just didn't want to merge until LGTM.

build_test/lib/src/resolve_source.dart
```diff
@@ -0,0 +1,86 @@
...
+  // ignore: unawaited_futures
+  runBuilder(
+    builder,
+    inputs,
+    new MultiAssetReader([inMemory, reader]),
+    new InMemoryAssetWriter(),
+    const BarbackResolvers(),
+    rootPackage: inputId.package,
+  );
+  return completer.future;
```
jakemac53: Note that `runBuilder` also calls `buildStep.complete` which does some cleanup and nothing (including the Resolver) is guaranteed to work after that point. It seems like since you aren't awaiting the `runBuilder` call there is also a bit of a race condition at play here. The future for the completer might run before `buildStep.complete` actually gets invoked, or it might be later. Any other future you await would probably not finish until after `buildStep.complete` though, which could lead to subtle bugs.

You could potentially fix this by passing a future into the builder, which it awaits on inside its `build` method after completing `completer` with the resolver. You wouldn't complete that future until `tearDown` most likely.

matanlurey: Done. Added an optional `tearDown` `Future<Null>`.

build_test/lib/src/resolve_source.dart
```diff
@@ -0,0 +1,86 @@
...
+/// expect(element.getType('Foo'), isNotNull);
+/// ```
+///
+/// **NOTE**: All `package` dependencies are resolved using [PackageAssetReader]
+/// - by default, [PackageAssetReader.currentIsolate]. A custom [resolver] may
+/// be provided to map files not visible to the current package's runtime.
+Future<Resolver> resolveSource(
+  String inputSource, {
+  AssetId inputId,
+  PackageResolver resolver,
```
matanlurey: Done.

build_test/lib/src/resolve_source.dart
```diff
@@ -69,16 +86,18 @@ Future<Resolver> resolveSource(
...
+  final Future<Null> _tearDown;
 
-  const _ResolveSourceBuilder(this._resolver);
+  const _ResolveSourceBuilder(this._resolver, this._tearDown);
 
   @override
-  Future<Null> build(BuildStep buildStep) {
-    return new Future.sync(() {
+  Future<Null> build(BuildStep buildStep) async {
+    await new Future.sync(() {
```
jakemac53: Now that this is an `async` method you should be able to remove this explicit Future?

matanlurey: Done.

PR #247, opened 2017-04-13
Upgrade build_barback to build 0.8.0

natebosch:
We won't implement the findAssets API for now, if a use case comes up
needing this we will re-evaluate.

### DIFF
--- a/build_barback/lib/src/transformer/transformer.dart
+++ b/build_barback/lib/src/transformer/transformer.dart
@@ -4,10 +4,11 @@
 import 'dart:async';
 import 'dart:convert';
 
-import 'package:build/build.dart' as build;
-import 'package:build/build.dart' hide AssetId;
 import 'package:barback/barback.dart' as barback show AssetId;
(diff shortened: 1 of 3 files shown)


build_barback/lib/src/transformer/transformer.dart
```diff
@@ -121,6 +122,10 @@ class _TransformAssetReader implements AssetReader {
   @override
   Future<String> readAsString(build.AssetId id, {Encoding encoding: UTF8}) =>
       transform.readInputAsString(toBarbackAssetId(id), encoding: encoding);
+
+  @override
+  Iterable<build.AssetId> findAssets(Glob glob) => throw new UnsupportedError(
+      'findAssets cannot be used within a Transformer');
```
jakemac53: Lets add a link to the github repo where they can file a bug about this if they need it.

natebosch: Done

Issue #234, opened 2017-03-24
build_runner needs to block reading generated files from current/later phases


natebosch:
When a set of inputs changes only the impacted outputs are invalidated and deleted. This means that the non-invalidated outputs are still present on disk and can show up as existing with `hasInput` when they shouldn't.

This problem becomes a lot more major once we have builders that are using a `findAssets` glob api.

PR #248, opened 2017-04-13
Prepare to publish build, build_test, build_barback

natebosch:
build_runner will remain unpublished until #224 is resolved

- Drop dependency overrides
- Drop `-dev` from version numbers


jakemac53:
wrong issue link in description?

natebosch:
Yeah wrong issue link, should have been #234.
Will correct when merging.

PR #253, opened 2017-04-14
Disallow reading assets from current/future phases

### LINKED ISSUE #234: build_runner needs to block reading generated files from current/later phases

natebosch:
Fixes #234

- Expand AssetReaderSpy into SinglePhaseReader. It is now has the
additional responsibility of preventing reads to generated assets
- Add a phaseNumber concept to GenereatedAssetNode. The number is
assigned while looping over phases in build_impl
- Upgrade to package:build 0.8.0 and mark that the AssetReader
overrides findAssets

### DIFF
--- a/build_runner/lib/src/asset/reader.dart
+++ b/build_runner/lib/src/asset/reader.dart
@@ -14,7 +14,7 @@ abstract class RunnerAssetReader extends AssetReader {
   Future<DateTime> lastModified(AssetId id);
 }
 
-final _false = new Future.value(false);
+final _asyncFalse = new Future.value(false);
 
 /// An [AssetReader] with a lifetime equivalent to that of a single Phase in a
 /// build.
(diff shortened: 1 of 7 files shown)


build_runner/lib/src/asset/reader.dart
```diff
@@ -3,37 +3,63 @@ import 'dart:convert';
...
 abstract class RunnerAssetReader extends AssetReader {
+  @override
   Iterable<AssetId> findAssets(Glob glob, {String packageName});
 
   /// Asynchronously gets the last modified [DateTime] of [id].
   Future<DateTime> lastModified(AssetId id);
 }
 
-class AssetReaderSpy implements AssetReader {
+final _false = new Future.value(false);
```
natebosch: Renamed to _asyncFalse. We can change to `FutureOr<bool>` in build 0.9.0 and clean this up.
Unfortunately it is a breaking change in case anyone does `reader.hasInput(input).then()`

PR #259, opened 2017-04-18
Fix backwards incompatibilities

natebosch:
- Upgrade AssetGraph serialization version number so we don't try and
  fail to read existing graphs from disk
- Make the check for unreadable assets asychronous so we don't
  synchronously throw and break the BarbackResolvers as it tries to read
  reachable sources


natebosch:
File https://github.com/dart-lang/build/issues/260 to add a regression test. Would like to get this out quickly

Issue #256, opened 2017-04-14
Rename hasInput


natebosch:
The distinction between 'input' and 'primaryInput' is not drastic enough - some authors seem to get confused about what they are allowed to read etc.

We should name it `hasAsset` or `canRead` or similar.

Will want to deprecate the current version and leave it in for a while.

jakemac53:
I like `exists`, as it maps to regular `File` apis.

natebosch:
I would agree if it was something like AssetId.exists... but `buildStep.exists(someAssetId)` looks strange to me.

natebosch:
We still have a hasInput for now on BuildStep, marked as deprecated. Will remove in next breaking release.

PR #262, opened 2017-04-20
Rename hasInput as canRead

### LINKED ISSUE #254: Return FutureOr from hasInput
Technically this is a breaking change even though it won't impact most users since you won't be able to call `.then` on the result. We should bundle this up with the 0.9 release

natebosch:
First step on #256
Fixes #254

- Rename hasInput on AssetReader
- Add canRead on BuildStep but keep hasInput marked deprecated. Limits
  the impact on most clients.

This will break implementations, so build_test, build_runner, and
_bazel_codegen will need to be updated.

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.5.3-dev
 
 - Support build 0.9.0
-  - Rename hasInput to canReade in AssetReader implementations
+  - Rename hasInput to canRead in AssetReader implementations
 
 ## 0.5.2
 
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.5.3-dev
+
+- Support build 0.9.0
+  - Rename hasInput to canReade in AssetReader implementations
+
 ## 0.5.2
 
 - Add `MultiAssetReader` to the public API.
--- a/build_test/lib/src/in_memory_reader.dart
+++ b/build_test/lib/src/in_memory_reader.dart
@@ -17,19 +17,17 @@ class InMemoryAssetReader implements AssetReader {
       : assets = sourceAssets ?? <AssetId, DatedValue>{};
 
   @override
-  Future<bool> hasInput(AssetId id) {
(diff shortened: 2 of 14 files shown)


build_test/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.5.3-dev
+
+- Support build 0.9.0
+  - Rename hasInput to canReade in AssetReader implementations
```
kevmoo: spelling oops

build_test/lib/src/in_memory_reader.dart
```diff
@@ -17,19 +17,17 @@ class InMemoryAssetReader implements AssetReader {
       : assets = sourceAssets ?? <AssetId, DatedValue>{};
 
   @override
-  Future<bool> hasInput(AssetId id) {
-    return new Future.value(assets.containsKey(id));
-  }
+  FutureOr<bool> canRead(AssetId id) => assets.containsKey(id);
```
jakemac53: Since `hasInput` is just deprecated don't we still need it here (and other impls below)

build_test/test/package_reader_test.dart
```diff
@@ -22,16 +22,16 @@ void main() {
     });
 
     test('should be able to read `build_test.dart`', () async {
-      expect(await reader.hasInput(buildTest), isTrue);
+      expect(await reader.canRead(buildTest), isTrue);
```
jakemac53: It might be worth testing both `hasInput` and `canRead` for now...

build_test/CHANGELOG.md
natebosch: Fixed

build_test/lib/src/in_memory_reader.dart
natebosch: Discussed offline. hasInput is removed from AssetReader (very few people impacted but is technically breaking for client code, not just build system implementations) Only BuildStep keeps the old method.

build_test/test/package_reader_test.dart
natebosch: tested in build_step_impl_test.dart

PR #264, opened 2017-04-20
Stricter analyzer

natebosch:
- Make some hints that we never want to allow behave as errors
- Turn on a bunch more lints
- Fix the newly discovered lints
- Temporarily disable await_only_futures until the linter understands
  FutureOr


build_runner/lib/src/generate/directory_watcher_factory.dart
```diff
@@ -3,6 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 import 'package:watcher/watcher.dart';
 
+export 'package:watcher/watcher.dart' show DirectoryWatcher;
```
jakemac53: I don't think we should export things from other packages generally - if this api changed then that would now be a breaking change in `build`.

natebosch: Fair enough. Was hoping to be able to keep some references in a Doc comment elsewhere, but I'll change them to backticks instead.

PR #267, opened 2017-04-20
Replace declareOutputs with buildExtensions

### LINKED ISSUE #222: Replace declareOutputs with inputExtension/outputExtensions
Today in a `Builder` there is a lot of flexibility in deciding what outputs to generate based on an input name. This flexibility is almost completely unused and makes future work more difficult. It's also a detriment to cross-compatibility since we can't support executing Dart code to determine outputs in bazel.

# Proposal

Replace `declareOutputs` with `inputExtension` and `outputExtensions`. Every files ending with `inputExtension` would be a primary input and produce a new files with the extension replaced with every value in `outputExtensions`. The extensions would need to exactly map to the configuration in `build.yaml` when bazel is used.

# Limitations/Drawbacks

One known use case that we'd be making harder is sass - today it is easy to declare no outputs for inputs that start with a leading underscore. We are not aware of other use cases that this would break.

The workaround means slightly more work on the consuming side. Rather than the `Builder` author being able to run only on files that don't start with underscore, the consumer would use a `generate_for` glob that matches only files getting an output.

# Benefits

- Parity between implementations. Arbitrary Dart code to determine outputs is not something we can support in bazel.
- Simpler mental model. The outputs can always be easily predicted.
- Will allow us to extend the configuration expressivity. Today we cannot express that a Builder will output only with the file doesn't already exist - and we cannot check for existence of other files during `declareOutputs`.  If `build_runner` can predict outputs by extension without running code in the Builder we can allow configuration to express that an output is allowed to be superseded by an existing file of the same name.

natebosch:
Fixes #222

- Add a top level method `findOutputs` to create the Set<AssetId>.
  Export this in the public API
- Use findOutputs from runBuilder
- Update MultiplexingBuilder to merge the buildExtensions maps
- Update some Builder implementations to the new style

### DIFF
--- a/build/lib/build.dart
+++ b/build/lib/build.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 export 'src/analyzer/resolver.dart';
@@ -12,3 +12,4 @@ export 'src/builder/exceptions.dart';
 export 'src/builder/logging.dart' show log;
 export 'src/builder/multiplexing_builder.dart';
 export 'src/generate/run_builder.dart';
+export 'src/generate/find_outputs.dart';
--- a/build/lib/build.dart
+++ b/build/lib/build.dart
@@ -11,5 +11,5 @@ export 'src/builder/builder.dart';
 export 'src/builder/exceptions.dart';
 export 'src/builder/logging.dart' show log;
 export 'src/builder/multiplexing_builder.dart';
+export 'src/generate/expected_outputs.dart';
 export 'src/generate/run_builder.dart';
-export 'src/generate/find_outputs.dart';
--- a/build/lib/src/builder/builder.dart
+++ b/build/lib/src/builder/builder.dart
@@ -1,9 +1,8 @@
-// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:async';
 
-import '../asset/id.dart';
 import 'build_step.dart';
 
 /// The basic builder class, used to build new files from existing ones.
@@ -11,11 +10,15 @@ abstract class Builder {
   /// Generates the outputs for a given [BuildStep].
   Future build(BuildStep buildStep);
 
-  /// Declares all potential output assets given [inputId]. If an empty [List]
-  /// is returned then the asset will never be provided to [build].
+  /// Configuration for which files will be created based on the file extension
+  /// of inputs.
   ///
-  /// Note: Reading the contents of [inputId] is not supported during this
-  /// phase. You can however choose later on not to actually output assets which
-  /// you return here, but no other [Builder] will be able to output them.
-  List<AssetId> declareOutputs(AssetId inputId);
+  /// All input sources matching any key in this map will be passed as build
+  /// step to this builder. Only files with the same basename and an extension
+  /// from the values in this map are expected as outputs. If an empty key
+  /// exists, all inputs are considered matching. A builder must always return
+  /// the same configuration. Typically this will be `const` but may vary based
+  /// on build arguments. Most builders will use a single input extension and
+  /// one or more output extensions.
+  Map<String, List<String>> get buildExtensions;
 }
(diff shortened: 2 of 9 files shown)


build/lib/build.dart
```diff
@@ -12,3 +12,4 @@ export 'src/builder/exceptions.dart';
 export 'src/builder/logging.dart' show log;
 export 'src/builder/multiplexing_builder.dart';
 export 'src/generate/run_builder.dart';
+export 'src/generate/find_outputs.dart';
```
jakemac53: nit: alphabetize

build/lib/src/generate/find_outputs.dart
```diff
@@ -0,0 +1,28 @@
...
+}
+
+Iterable<AssetId> _replaceExtensions(
+        AssetId assetId, String oldExtension, List<String> newExtensions) =>
+    newExtensions.map((n) => _replaceExtension(assetId, oldExtension, n));
+
+AssetId _replaceExtension(
+    AssetId assetId, String oldExtension, String newExtension) {
+  var path = assetId.path;
+  return new AssetId(
```
jakemac53: Might be worth asserting `path.endsWith(oldExtension)`? And/or add a comment that `assetId.path` must end with `oldExtension`.

build/lib/src/builder/builder.dart
```diff
@@ -1,21 +1,24 @@
...
 import 'build_step.dart';
 
 /// The basic builder class, used to build new files from existing ones.
 abstract class Builder {
   /// Generates the outputs for a given [BuildStep].
   Future build(BuildStep buildStep);
 
-  /// Declares all potential output assets given [inputId]. If an empty [List]
-  /// is returned then the asset will never be provided to [build].
+  /// Configuration for which files will be created based on the file extension
```
jakemac53: what about `Mapping of input file extension to output file extensions.`

build/lib/src/builder/builder.dart
```diff
@@ -1,21 +1,24 @@
...
-  /// you return here, but no other [Builder] will be able to output them.
-  List<AssetId> declareOutputs(AssetId inputId);
+  /// All input sources matching any key in this map will be passed as build
+  /// step to this builder. Only files with the same basename and an extension
+  /// from the values in this map are expected as outputs. If an empty key
+  /// exists, all inputs are considered matching. A builder must always return
+  /// the same configuration. Typically this will be `const` but may vary based
+  /// on build arguments. Most builders will use a single input extension and
+  /// one or more output extensions.
+  Map<String, List<String>> get buildExtensions;
```
jakemac53: I am not sure how I feel about this name, but I can't think of a great alternative either. Possibly `outputExtensions` or something... this is probably ok though.

build/lib/src/generate/find_outputs.dart
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import '../asset/id.dart';
+import '../builder/builder.dart';
+
+/// Collects the expected AssetIds created by [builder] when given [input] based
+/// on the extension configuration.
+Set<AssetId> findOutputs(Builder builder, AssetId input) {
```
jakemac53: What about `expectedOutputs` instead? `find` to me implies file io or something.

build/lib/src/builder/builder.dart
```diff
@@ -1,21 +1,24 @@
...
-  /// Declares all potential output assets given [inputId]. If an empty [List]
-  /// is returned then the asset will never be provided to [build].
+  /// Configuration for which files will be created based on the file extension
+  /// of inputs.
   ///
-  /// Note: Reading the contents of [inputId] is not supported during this
-  /// phase. You can however choose later on not to actually output assets which
-  /// you return here, but no other [Builder] will be able to output them.
-  List<AssetId> declareOutputs(AssetId inputId);
+  /// All input sources matching any key in this map will be passed as build
```
jakemac53: Lets break this up into a few bullet points, specifically the bit about empty String keys.

build/lib/build.dart
natebosch: Done

build/lib/src/builder/builder.dart
natebosch: Done

build/lib/src/builder/builder.dart
natebosch: Done

build/lib/src/builder/builder.dart
natebosch: Yeah I don't know. `outputExtensions` feels less accurate to me, `extensionMapping` just repeats the type in the name... Will leave it as is for now and see if we come up with anything else before publishing

build/lib/src/generate/find_outputs.dart
natebosch: Good call, done

build/lib/src/generate/find_outputs.dart
natebosch: Done

PR #271, opened 2017-04-21
Upgrade build_barback to build 0.9.0

natebosch:
- Change expectedOutputs to an Iterable. Some use cases need to check
  for duplicates. This means it will be illegal for the
  `buildExtensions` to express two ways to output the same file.
- Take `buildExtensions` as an argument to `TransformerBuilder`. We
  cannot automatically translate from declareOutputs to the new
  approach. This does mean that non DeclaringTransfomer instances can be
  wrapped.
- Update Builder, AssetReader, and AssetWriter implementations for the
  breaking changes in build

### DIFF
--- a/build_barback/lib/src/transformer/transformer.dart
+++ b/build_barback/lib/src/transformer/transformer.dart
@@ -1,4 +1,4 @@
-// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 import 'dart:async';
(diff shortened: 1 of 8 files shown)


build_barback/lib/src/transformer/transformer.dart
```diff
@@ -137,15 +138,13 @@ class _TransformAssetWriter implements AssetWriter {
   _TransformAssetWriter(this.transform);
 
   @override
-  Future writeAsBytes(build.AssetId id, List<int> bytes) {
-    transform.addOutput(barbackAssetFromBytes(id, bytes));
-    return new Future.value(null);
+  Future writeAsBytes(build.AssetId id, FutureOr<List<int>> bytes) async {
```
jakemac53: I think it is important that this happens synchronously to make sure it gets waited on later?

You could verify by adding a test that returns synchronously from `build` after calling this method.

build_barback/lib/src/transformer/transformer.dart
```diff
@@ -137,15 +138,13 @@ class _TransformAssetWriter implements AssetWriter {
...
+    transform.addOutput(barbackAssetFromBytes(id, await bytes));
   }
 
   @override
-  Future writeAsString(build.AssetId id, String contents,
-      {Encoding encoding: UTF8}) {
-    transform.addOutput(barbackAssetFromString(id, contents));
-    return new Future.value(null);
+  Future writeAsString(build.AssetId id, FutureOr<String> contents,
```
jakemac53: ditto

build_barback/lib/src/transformer/transformer.dart
natebosch: The `.complete()` in `BuildStep` will wait for this so I don't think it matters here.

I'm also considering not changing `AssetWriter` and only changing the interface in `BuildStep`. That way most of our implementations won't need to care about it and there are only a few packages which see those directly anyway.

jakemac53: Either way sounds good to me.

natebosch: Reverted this part. AssetWriter doesn't take FutureOr anymore.

PR #272, opened 2017-04-24
Only BuildStep takes FutureOr

natebosch:
Since AssetWriter is mainly used internally there is not as much value
in expanding it's interface, and a much higher cost.

### DIFF
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -85,15 +85,27 @@ class BuildStepImpl implements BuildStep {
   @override
   Future writeAsBytes(AssetId id, FutureOr<List<int>> bytes) {
     _checkOutput(id);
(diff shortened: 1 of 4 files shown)


build/lib/src/builder/build_step_impl.dart
```diff
@@ -85,15 +85,27 @@ class BuildStepImpl implements BuildStep {
   @override
   Future writeAsBytes(AssetId id, FutureOr<List<int>> bytes) {
     _checkOutput(id);
-    var done = _writer.writeAsBytes(id, bytes);
+    Future done;
```
jakemac53: You **may** want to encapsulate this FutureOr handling logic in a function, but its only copied twice for now so up to you.

natebosch: https://github.com/dart-lang/build/pull/273

PR #274, opened 2017-04-24
Remove outputPackage from copyBuilder

natebosch:
This property doesn't work anymore.

- Fix up changelog with some already changed details
- Bump to 0.6.0 since this is a breaking change

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,7 +1,10 @@
-## 0.5.3-dev
+## 0.6.0-dev
 
 - Support build 0.9.0
-  - Rename hasInput to canRead in AssetReader implementations
+  - Rename `hasInput` to `canRead` in `AssetReader` implementations
+  - Replace `declareOutputs` with `buildExtensions` in `Builder` implementations
+- *Breaking* `CopyBuilder` no longer has an `outputPackage` field since outputs
+  can only ever be in the same package as inputs.
 
 ## 0.5.2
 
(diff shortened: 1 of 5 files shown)


build_test/CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
-## 0.5.3-dev
+## 0.6.0-dev
 
 - Support build 0.9.0
-  - Rename hasInput to canRead in AssetReader implementations
+  - Rename `hasInput` to `canRead` in `AssetReader` implementations
+  - Replace `declareOutputs` with `buildExtensions` in `Builder` implementations
+- **Breaking** `CopyBuilder` no longer has an `outputPackage` field since outputs
```
jakemac53: Is this class actually exposed via any public apis?

build_test/lib/src/copy_builder.dart
```diff
@@ -59,13 +55,8 @@ class CopyBuilder implements Builder {
     return {'': outputs};
   }
 
-  AssetId _copiedAssetId(AssetId inputId, int copyNum) {
-    var withExtension = inputId.addExtension(_copiedAssetExtension(copyNum));
-    if (outputPackage == null) {
-      return withExtension;
-    }
-    return new AssetId(outputPackage, withExtension.path);
```
jakemac53: I am confused how this ever worked?

build_test/CHANGELOG.md
natebosch: yes, it's used from `build_runner` without a `/src` import.

build_test/lib/src/copy_builder.dart
natebosch: It used to be legal to do this with `declareOutputs` since we could take another package's source as input and output it into our own package.

natebosch: This is the breakage I was calling out here: https://github.com/dart-lang/build/issues/222#issuecomment-295940412

PR #276, opened 2017-04-24
Update e2e_example to build 0.9.0

natebosch:
- Change Builder implementation to use buildExtensions
- Drop PackagesDirBuilder since it is incompatible with buildExtensions
- Update dependencies


e2e_example/lib/copy_builder.dart
```diff
@@ -10,12 +10,14 @@ class CopyBuilder extends Builder {
   @override
   Future build(BuildStep buildStep) async {
     var id = buildStep.inputId;
-    buildStep.writeAsString(
-        _copiedAssetId(id), await buildStep.readAsString(id));
+    // ignore: unawaited_futures
```
jakemac53: ouch

e2e_example/pubspec.yaml
```diff
@@ -1,16 +1,23 @@
...
 
 dev_dependencies:
-  build_runner: ^0.3.0
+  build_runner: ^0.3.2
 
 transformers:
 - e2e_example:
     $include: 'web/*.txt'
+
+dependency_overrides:
```
jakemac53: Not sure if we actually want these, as it can hide dependency version issues? It does keep us on the latest and greatest though, which also has benefits.

natebosch: It's only until we publish

PR #283, opened 2017-04-27
Refactor watch_impl into more focused methods

natebosch:
- Pull out code to watch for file changes into a separate method.

Improvements over existing code:
- `runWatch` shortened from 132 to 61 lines
- 'updatedInputs' is no longer used for implicit communication between
  the code watching for changes and the code triggering builds, no more
  deep copy
- Change filtering is centralized in the Stream pipeline rather than
  spread through change collection (expected deletes, files not in the
  asset graph), and build triggering (_shouldSkipInput).
- debouncing is declarative on the Stream rather than handled by
  `scheduleBuild`
- Avoiding concurrent builds is declarative on the Stream rather than
  handled by `_nextBuildScheduled`


natebosch:
This mostly came out of me trying to understand more of how build_runner works.

Overall the code is longer and there is a fair bit of complexity in `_Buffer`, but IMO this is still an improvement because the complexity is in code explicitly dealing with Stream stuff rather than spread into code dealing with business logic.

`_Buffer` is pretty close to the same think in RxJS, so potentially we can punt some of this into another package eventually since it should be useful generally.

build_runner/lib/src/generate/change_watcher.dart
```diff
@@ -0,0 +1,96 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:logging/logging.dart';
+import 'package:path/path.dart' as path;
```
jakemac53: lets use `p` as the library prefix

build_runner/lib/src/generate/change_watcher.dart
```diff
@@ -0,0 +1,96 @@
...
+
+    var watcher = directoryWatcherFactory(absolutePackagePath);
+    watchers.add(watcher);
+    listeners.add(watcher.events.listen((WatchEvent e) {
+      var changePath = _normalizeChangePath(
+          e.path, absolutePackagePaths, logger, packageGraph);
+      if (changePath == null) return;
+
+      // Check for ignored paths and immediately bail.
+      if (pathsToIgnore.any((path) => changePath.startsWith(path))) return;
```
jakemac53: Lets change this to use `p.isWithin`? That will handle edge cases better.

build_runner/lib/src/generate/change_watcher.dart
```diff
@@ -0,0 +1,96 @@
...
+
+class AssetChange {
+  final AssetId id;
+  final ChangeType type;
+  AssetChange(this.id, this.type);
+}
+
+Future<Stream<AssetChange>> startFileWatchers(PackageGraph packageGraph,
+    Logger logger, DirectoryWatcherFactory directoryWatcherFactory) async {
+  var ctl = new StreamController<AssetChange>();
```
jakemac53: When I saw this later on I didn't know what it was, lets spell out `controller`.

build_runner/lib/src/generate/change_watcher.dart
```diff
@@ -0,0 +1,96 @@
...
+
+  var packageName = changePathParts[packagesIndex + 1];
+  var packageNode = packageGraph[packageName];
+  if (packageNode == null) {
+    logger.severe('Got update for invalid package: $packageName');
+    return null;
+  }
+  var packagePath = absolutePackagePaths[packageNode];
+  var libPath = path.joinAll(['lib']..addAll(
+      changePathParts.getRange(packagesIndex + 2, changePathParts.length)));
```
jakemac53: Lets change this to just use `skip(packagesIndex + 1)`.

build_runner/lib/src/generate/change_watcher.dart
natebosch: done

build_runner/lib/src/generate/change_watcher.dart
natebosch: done

build_runner/lib/src/generate/change_watcher.dart
natebosch: done

build_runner/lib/src/generate/change_watcher.dart
natebosch: done, skip(packagesIndex + 2)

natebosch:
buffer and debounce are now tested in package:stream_transform and move out of this review.

Any other comments?

PR #292, opened 2017-05-18
EARLY FEEDBACK: New Resolvers implementation for barback

natebosch:
This implementation assumes that inputs won't change during the lifetime
of the Resolvers instance, so we crawl the import graph once per
aggregate transformer rather than once per BuildStep.

Sending for early feedback, this needs further manual testing.

### DIFF
--- a/build_barback/lib/src/transformer/crawl_imports.dart
+++ b/build_barback/lib/src/transformer/crawl_imports.dart
@@ -1,0 +1,131 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
+
+import 'package:analyzer/analyzer.dart' show parseDirectives;
+import 'package:analyzer/src/generated/engine.dart' show TimestampedData;
+import 'package:analyzer/src/generated/source.dart';
+import 'package:build/build.dart';
+import 'package:path/path.dart' as p;
+
+import 'package:analyzer/dart/ast/ast.dart';
+
+/// Crawl Dart files outside of the SDK transitively imported from any library
+/// in [entryPoints] and cache it as a [Source].
+Future<Map<Uri, Source>> crawlImports(
+    Iterable<AssetId> entryPoints, Future<String> read(AssetId id)) async {
+  var roots = entryPoints.where((id) => id.path.endsWith('.dart'));
+  var assets = await _crawlAsync(roots, read, _assetImports);
+  var results = <Uri, Source>{};
+  for (var asset in assets.keys) {
+    results[asset.uri] = _toSource(asset, assets[asset]);
+  }
+  return results;
+}
+
+Iterable<AssetId> _assetImports(AssetId id, String source) {
+  var unit = parseDirectives(source);
+  return unit.directives
+      .where((d) => d is UriBasedDirective)
+      .map((d) => d as UriBasedDirective)
+      .map((d) => d.uri.stringValue)
+      .where((uri) => !uri.startsWith('dart:'))
+      .map((uri) => new AssetId.resolve(uri, from: id));
+}
+
+Source _toSource(AssetId id, String content) => new AssetSource(id, content);
+
+/// Crawl a graph where node values and edges are resolved asynchronously.
+///
+/// Nodes are keyed by type [K] and have values of type [V]. [resolve] finds the
+/// value for a node. [edges] finds the edges leaving a node.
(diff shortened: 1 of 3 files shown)


natebosch:
This doesn't seem to solve the problem with the angular compiler so either
1. This wasn't the problem.
2. This wan't the _only_ problem.
3. This doesn't improve things (enough?)
4. This introduced some new problem.

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+import 'package:analyzer/dart/ast/ast.dart';
+
+/// Crawl Dart files outside of the SDK transitively imported from any library
+/// in [entryPoints] and cache it as a [Source].
+Future<Map<Uri, Source>> crawlImports(
+    Iterable<AssetId> entryPoints, Future<String> read(AssetId id)) async {
+  var roots = entryPoints.where((id) => id.path.endsWith('.dart'));
+  var assets = await _crawlAsync(roots, read, _assetImports);
+  var results = <Uri, Source>{};
+  for (var asset in assets.keys) {
```
jakemac53: nit: `asset` => `id` or `assetId`. Or use `assets.forEach`.

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+  }
+  return results;
+}
+
+Iterable<AssetId> _assetImports(AssetId id, String source) {
+  var unit = parseDirectives(source);
+  return unit.directives
+      .where((d) => d is UriBasedDirective)
+      .map((d) => d as UriBasedDirective)
+      .map((d) => d.uri.stringValue)
```
jakemac53: nit: I would combine these and just do `(d as UriBasedDirective).uri.stringValue`

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+  var roots = entryPoints.where((id) => id.path.endsWith('.dart'));
+  var assets = await _crawlAsync(roots, read, _assetImports);
+  var results = <Uri, Source>{};
+  for (var asset in assets.keys) {
+    results[asset.uri] = _toSource(asset, assets[asset]);
+  }
+  return results;
+}
+
+Iterable<AssetId> _assetImports(AssetId id, String source) {
```
jakemac53: I would add a comment mentioning that this doesn't include `dart:` imports

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+      .where((uri) => !uri.startsWith('dart:'))
+      .map((uri) => new AssetId.resolve(uri, from: id));
+}
+
+Source _toSource(AssetId id, String content) => new AssetSource(id, content);
+
+/// Crawl a graph where node values and edges are resolved asynchronously.
+///
+/// Nodes are keyed by type [K] and have values of type [V]. [resolve] finds the
+/// value for a node. [edges] finds the edges leaving a node.
```
jakemac53: Maybe add something like `Returns all nodes found by crawling [roots]` (it wasn't super clear to me what the return value represented right away).

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+  return results;
+}
+
+/// Synchronously record the the node a [key] has been seen, then start the
+/// work.
+///
+/// The returned Future will complete when the node at [key], and transitively
+/// reachable nodes, have either finished their crawl or are guaranteed to be
+/// completed within another future in [seen].
+Future _visitNode<K, V>(K key, Map<K, Future<V>> seen,
```
jakemac53: In general in Dart code you don't see people modifying arguments very often (like is done with `seen` here).

I would consider creating a class which can contain this state and moving these methods to instance methods. That would also simplify some of the parameter lists.

You could also move `resolve` and `edges` to instance methods that way.

build_barback/lib/src/transformer/crawl_imports.dart
```diff
@@ -0,0 +1,131 @@
...
+    Map<K, Future<V>> seen,
+    FutureOr<V> resolve(K key),
+    FutureOr<Iterable<K>> edges(K key, V value)) async {
+  var value = await resolve(key);
+  var next = await edges(key, value) ?? [];
+  await Future.wait(next.map((k) => _visitNode(k, seen, resolve, edges)));
+  return value;
+}
+
+class AssetSource implements Source {
```
jakemac53: Should this be a private class?

PR #300, opened 2017-06-19
Fix for top-level inference warning.

matanlurey:
(This warning starts appearing in `1.25.0`, but just getting ahead of it)


kevmoo:
@matanlurey FYI – these types of hints are going away w/ 1.25 – right, @leafpetersen ?

PR #305, opened 2017-06-26
Fix log output logic to not overwrite the last line of a multi-line log

kevmoo:


build_runner/lib/src/generate/options.dart
```diff
@@ -105,12 +106,16 @@ void _defaultLogListener(LogRecord record) {
     }
   }
 
-  if (record.level > Level.INFO || !_isPosixTerminal) {
+  var message = lines.join('\n');
```
jakemac53: I don't like doing this join twice, lets see if we can avoid it?

jakemac53: nvm, it would overcomplicate things I think to try and avoid this.

build_runner/lib/src/generate/options.dart
```diff
@@ -105,12 +106,16 @@ void _defaultLogListener(LogRecord record) {
...
+
+  var multiLine = LineSplitter.split(message).length > 1;
```
jakemac53: Probably better to just check for a newline here

natebosch:
Changelog?

PR #307, opened 2017-06-26
Split packages on Travis

kevmoo:


tool/travis.sh
```diff
@@ -7,25 +7,8 @@
...
-pub run test
-popd
-
-pushd build_barback
-pub upgrade
-dartanalyzer --fatal-warnings lib/build_barback.dart
-pub run test
-popd
+dartanalyzer --fatal-warnings .
+echo Any unformatted files?
```
kevmoo: Not really. The next line may or may not print out files.

If it does, this line gives a bit of context – otherwise reading output is a bit confusing.

PR #318, opened 2017-08-01
Move some graph manipulation into asset_graph.dart

natebosch:
Centralizing the logic will make it easier to see how it is used and
update it. This covers most use cases for writing the `needsUpdate`
field - the remaining case is setting it to false after building an
ouptut.

Also clean up some Doc comments

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -107,8 +107,9 @@ class AssetGraph {
         clearNodeAndDeps(output, rootChangeType, parent: node.id);
       }
 
-      // For deletes, prune the graph.
       if (parent == null && rootChangeType == ChangeType.REMOVE) {
+        // This is the root asset, it was removed on disk so it needs to be
+        // removed from the graph.
         remove(id);
       }
       if (node is GeneratedAssetNode) {
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -1,19 +1,22 @@
 // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -77,6 +80,51 @@ class AssetGraph {
...
+      if (node == null) return;
+
+      // Update all outputs of this asset as well.
+      for (var output in node.outputs) {
+        clearNodeAndDeps(output, rootChangeType, parent: node.id);
+      }
+
+      // For deletes, prune the graph.
+      if (parent == null && rootChangeType == ChangeType.REMOVE) {
+        remove(id);
```
jakemac53: Consider adding an additional comment about why this one is being removed (it is the root id that was actually deleted I believe, but that wasn't immediately obvious).

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -77,6 +80,51 @@ class AssetGraph {
...
+      }
+      if (node is GeneratedAssetNode) {
+        node.needsUpdate = true;
+        if (rootChangeType == ChangeType.REMOVE &&
+            node.primaryInput == parent) {
```
jakemac53: If some other input to this action hits this codepath first, then `node.primaryInput` will not be equal to `parent`, but that doesn't mean that `node.primaryInput` wasn't removed as well.... seems like a bug.

Maybe we need a `node.primaryOutputs` so we can remove those more explicitly if `node` is removed instead of this logic?

jakemac53: This could be a followup cl though as I believe it is an existing issue.

natebosch: Yeah I want to add `primaryOutputs` either way. Will take another look at this code path when we add that.

build_runner/lib/src/asset_graph/graph.dart
natebosch: Done

PR #319, opened 2017-08-01
Track primary outputs explicitly

natebosch:
`outputs` is a mix of primary and non-primary outputs which may be
impacted by the asset. Add `primaryOutputs` to track explicitly the
files which may be generated for an input.

- Add primaryOutputs field and append when we run a Builder
- Update serialization and deserialization to pick up outputs
- Bump serialization spec number to invalidate old graphs


natebosch:
Added primaryOutput expectations to tests.

PR #329, opened 2017-08-03
Always update AssetGraph with static information

natebosch:
Instead of build the graph up as builders are run find the structure of
the graph from declared outputs alone. The only changes to the graph
after that should be marking outputs as non-primary inputs are read.

- Remove addIfAbset method and test
- Hide add and remove - there are still public wrappers for these used
  in tests that we should drop soon.
- During graph changes based on disk changes, track whether the file
  that changed was a source and only delete GeneratedAssetNodes if they
  would have come from a deleted source file, not a deleted generated
  file. When a generated file is deleted mark it and it's descendents as
  needing an update. Remove the `seen` set since a deletion of a
  generated file could mark a node as needing and update but a later
  seen deletion of a source file should still remove it.
- Add `updateForSourcs` to fill in the asset graph with new outputs as
  new source files are discovered.
- Add a simple test for the `build` constructor on AssetGraph
- Update AssetGraph construction in tests to use the `build` constructor
  but still create an empty graph.

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -134,17 +134,17 @@ class AssetGraph {
   /// Updates the structure of the graph to account for any new sources and
   /// returns true if the graph has changed.
   bool updateForSources(
-      List<List<BuildAction>> phases, Iterable<AssetId> newSources) {
-    var unseen = newSources.where((s) => !_nodesById.containsKey(s)).toList();
+      List<List<BuildAction>> phases, Iterable<AssetId> sources) {
+    var unseen = sources.where((s) => !_nodesById.containsKey(s)).toSet();
     if (unseen.isEmpty) return false;
-    _addOutputsForSources(phases, unseen.toList());
+    _addOutputsForSources(phases, unseen);
     return true;
   }
 
   void _addOutputsForSources(
-      List<List<BuildAction>> phases, Iterable<AssetId> newSources) {
+      List<List<BuildAction>> phases, Set<AssetId> newSources) {
     newSources.map((s) => new AssetNode(s)).forEach(_add);
-    var allInputs = new Set<AssetId>()..addAll(newSources);
+    var allInputs = new Set<AssetId>.from(newSources);
     var phaseNumber = 0;
     for (var phase in phases) {
       phaseNumber++;
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -5,6 +5,7 @@
 import 'package:build/build.dart';
 import 'package:watcher/watcher.dart';
 
+import '../generate/phase.dart';
 import 'exceptions.dart';
 import 'node.dart';
 
@@ -22,7 +23,7 @@ class AssetGraph {
   /// value if you want incremental rebuilds.
   DateTime validAsOf = new DateTime.fromMillisecondsSinceEpoch(0);
 
-  AssetGraph();
+  AssetGraph._();
 
   /// Part of the serialized graph, used to ensure versioning constraints.
   ///
@@ -37,14 +38,18 @@ class AssetGraph {
           serializedGraph['version'], _version);
     }
 
-    var graph = new AssetGraph();
+    var graph = new AssetGraph._();
     for (var serializedItem in serializedGraph['nodes']) {
-      graph.add(new AssetNode.deserialize(serializedItem));
+      graph._add(new AssetNode.deserialize(serializedItem));
     }
     graph.validAsOf =
         new DateTime.fromMillisecondsSinceEpoch(serializedGraph['validAsOf']);
     return graph;
   }
+
+  factory AssetGraph.build(
+          List<List<BuildAction>> phases, Set<AssetId> sources) =>
+      new AssetGraph._().._addOutputsForSources(phases, sources);
 
   /// Puts this graph into a serializable form.
   Map serialize() => {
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -94,38 +100,79 @@ class AssetGraph {
   /// set of assets that need to be deleted.
   Iterable<AssetId> updateAndInvalidate(Map<AssetId, ChangeType> updates) {
     var deletes = new Set<AssetId>();
-    var seen = new Set<AssetId>();
     void clearNodeAndDeps(AssetId id, ChangeType rootChangeType,
-        {AssetId parent}) {
-      if (seen.contains(id)) return;
-      seen.add(id);
+        {AssetId parent, bool rootIsSource}) {
```
jakemac53: When is the root not an original source? We should not be reporting file change events for generated sources afaik.

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -94,38 +100,79 @@ class AssetGraph {
...
 
     updates.forEach(clearNodeAndDeps);
     return deletes;
   }
 
+  /// Updates the structure of the graph to account for any new sources and
+  /// returns true if the graph has changed.
+  bool updateForSources(
+      List<List<BuildAction>> phases, Iterable<AssetId> newSources) {
+    var unseen = newSources.where((s) => !_nodesById.containsKey(s)).toList();
```
jakemac53: Should it be an error to pass existing sources? If not I would consider changing `newSources` to just `sources`.

jakemac53: Also consider `toSet` instead of `toList`

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -94,38 +100,79 @@ class AssetGraph {
...
+    if (unseen.isEmpty) return false;
+    _addOutputsForSources(phases, unseen.toList());
```
jakemac53: You already called `toList` above, no need to do an extra copy here :).

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -94,38 +100,79 @@ class AssetGraph {
...
+    return true;
+  }
+
+  void _addOutputsForSources(
+      List<List<BuildAction>> phases, Iterable<AssetId> newSources) {
+    newSources.map((s) => new AssetNode(s)).forEach(_add);
+    var allInputs = new Set<AssetId>()..addAll(newSources);
```
jakemac53: nit: `new Set<AssetId>.from(newSources)`

You may just want to make this take a `Set` though and do `toSet` above, since these are internal only you can be a bit more precise with the desired types and remove some copies.

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -94,38 +100,79 @@ class AssetGraph {
...
+      for (var action in phase) {
+        var inputs = allInputs.where(
+            (input) => action.inputSet.globs.any((g) => g.matches(input.path)));
+        for (var input in inputs) {
+          var outputs = expectedOutputs(action.builder, input);
+          phaseOutputs.addAll(outputs);
+          get(input).primaryOutputs.addAll(outputs);
+          for (var output in outputs) {
+            if (contains(output) && get(output) is AssetNode) {
+              _remove(output);
```
jakemac53: A bit confused by this, won't it bash away original srcs with generated asset nodes?

build_runner/test/generate/build_test.dart
```diff
@@ -123,7 +123,7 @@ void main() {
     var cachedGraph = new AssetGraph.deserialize(
         JSON.decode(writer.assets[graphId].stringValue));
 
-    var expectedGraph = new AssetGraph();
+    var expectedGraph = new AssetGraph.build([], new Set());
```
jakemac53: it might be worth adding a `.empty` constructor?

build_runner/lib/src/asset_graph/graph.dart
natebosch: We report deletes, not edits

https://github.com/dart-lang/build/blob/7820ae1b143068a4d75bf35029cbd90a429f690c/build_runner/lib/src/generate/watch_impl.dart#L177

Maybe we should not report deletes either?

build_runner/lib/src/asset_graph/graph.dart
natebosch: toSet makes sense. Done

build_runner/lib/src/asset_graph/graph.dart
natebosch: Sloppy refactoring :) Done

build_runner/lib/src/asset_graph/graph.dart
jakemac53: Oh, actually that makes sense. If somebody deletes a generated output then we should regenerate it. It is a bit of a weird case but I think it makes sense to support.

build_runner/lib/src/asset_graph/graph.dart
natebosch: Taking a Set. I think I still want to make the copy though because they mean different things. If we find a performance issue while profiling large packages we can reconsider.

build_runner/lib/src/asset_graph/graph.dart
natebosch: Yes - if we see a generated file on disk in `build_impl.dart` we don't know that it was generated so we treat it as a source. Here when we see that it's actually generated we overwrite it as a generated node and expect that build_impl will clean it up as appropriate. We're mainly concerned here with getting the structure of the graph right, not with the state of the build or previous builds.

build_runner/test/generate/build_test.dart
natebosch: Perhaps... but hopefully not. I think only tests would use it so I want to take a shot at finding a new pattern for tests that doesn't require manual graph manipulation like this - or maybe subclassing the graph just for tests.

build_runner/lib/src/asset_graph/graph.dart
jakemac53: ah ok, makes sense

build_runner/test/generate/build_test.dart
jakemac53: ok

PR #332, opened 2017-08-05
Shorten _isExpectedDelete

natebosch:
`Set.remove` already does the conditional checking.


natebosch:
Once I saw it I couldn't *not* see it 😀

PR #335, opened 2017-08-07
Drop PhaseGroup and Phase classes

### LINKED ISSUE #333: Remove the `phase` concept
As build_runner's asset graph gets smarter we should need to rely less on having a 'phase' done before running further builders because we can avoid reading or listing the wrong assets based on the graph rather than strictly what exists on disk.

We should replace the `Phase` with a `BuildAction`, and possibly the `PhaseGroup` should just be a `List<BuildAction>`.

I think I've only seen phases which consist of a single build action anyway - but if there are uses cases we think we'd be losing we should take a close look.

natebosch:
Closes #333

The majority of builds are a single action, in the places where there
were multiple actions most were already split into separate phases.

Some builds will be negatively impacted by this in the short term before
we can take advantage of higher parallelism in builds thanks to knowing
the entire asset graph statically.

A very small set of cases may be impacted by needing to more carefully
construt InputSets to 'hide' outputs from other actions that would have
otherwise run in the same phase and been isolated already.


natebosch:
IMO this is a net improvement at the usage sites.

The only reservation I might have is if we aren't confident that every use case can be solved with this slightly more restricted API. I think though that if such a case does come up we could probably add a non-breaking change to allow `List<List<BuildAction>>` again via separate methods like `buildWithPhases`

build_runner/README.md
```diff
@@ -40,11 +37,7 @@ the `CopyBuilder` for now, just know that its a `Builder` which copies files:
 import 'package:build_runner/build_runner.dart';
 
 main() async {
-  /// The [PhaseGroup#singleAction] constructor is a shorthand for:
-  ///
-  ///   new PhaseGroup().newPhase().addAction(builder, inputSet);
-  await build(new PhaseGroup.singleAction(
-      new CopyBuilder('.copy'), new InputSet('my_package', ['lib/*.dart'])));
+  await build([new CopyBuilder('.copy'), 'my_package', ['lib/*.dart']]);
```
jakemac53: I think you are missing the `new BuildAction` here  and below in a few places

build_runner/lib/src/generate/phase.dart
```diff
@@ -5,74 +5,15 @@ import 'package:build/build.dart';
...
-    var phase = new Phase();
-    _phases.add(phase);
-    return phase;
-  }
-
-  /// Adds a [Phase] to this [PhaseGroup].
-  void addPhase(Phase phase) {
-    _phases.add(phase);
-  }
+  BuildAction(this.builder, String package, [List<String> globs = const ['**']])
```
jakemac53: lets change this to a named arg as discussed offline

build_runner/README.md
natebosch: Done

build_runner/lib/src/generate/phase.dart
natebosch: Done

natebosch:
cc @kevmoo

PR #336, opened 2017-08-07
Remove glob arguments to '**/*' in tests

natebosch:
The default value is `['**']` which is not quite the same as `['**/*']`
but close enough for these tests.


natebosch:
Yeah the default argument picks up all files in the package.

PR #337, opened 2017-08-08
Drop phaseOutputIds

natebosch:
This was previously used to limit readable files by phase but this all
goes through the AssetGraph now.


natebosch:
It didn't impact any tests because this code has to impact - we write to this Set but don't read from it anymore. The place we used to read this Set was deleted here: https://github.com/dart-lang/build/pull/330/files#diff-7a296c2aaedb49f2b91b7fa9c81b536cL392 but I did not fully clean up the code in that commit.

PR #349, opened 2017-08-09
Drop `resolvers` argument in build_runner

### LINKED ISSUE #347: Remove `Resolvers` argument to build/watch/serve
It would be easier to remove the `release` method if we are sure we control the entire lifespan of Resolver and Resolvers instances in build_runner

AFAIK these arguments aren't used anyway.

natebosch:
Closes #347

This will make it easier to optimize a new Resolver implementation if we
are confident we control the entire lifespan.


natebosch:
I think it was intended to be useful for consumers but the idea of multiple interchangeable Resolver implementations never really panned out.

PR #350, opened 2017-08-09
Change to async apis in Resolver interface

### LINKED ISSUE #339: Make the apis in Resolver async
Migrating to the AnalysisDriver will require some of the apis that are currently synchronous to instead return a Future.

In anticipation of that change we should start returning Futures wherever we'll need them.

natebosch:
Closes #339

Since we have some churn anyway rename `get*` methods.

- Change the APIs on resolver to return Future.
- Change the BuildStep to return a Resolver synchronously.
- Allow `Resolvers` to keep returning a Future since we likely need to
  keep doing async work setting up a resolver. Add a wrapper to turn a
  `Future<Resolver>` into a `Resolver` - this will continue to work as
  long as *all* the public apis are async.

### DIFF
--- a/build/lib/src/analyzer/resolver.dart
+++ b/build/lib/src/analyzer/resolver.dart
@@ -14,29 +14,30 @@ abstract class Resolver {
   ///
   /// This will be `false` in the case where the file is not Dart source code,
   /// or is a `part of` file (not a standalone Dart library).
-  bool isLibrary(AssetId assetId);
+  Future<bool> isLibrary(AssetId assetId);
 
   /// All libraries accessible from the entry point, recursively.
   ///
   /// **NOTE**: This includes all Dart SDK libraries as well.
-  Iterable<LibraryElement> get libraries;
+  Stream<LibraryElement> get libraries;
 
   /// Returns a resolved library representing the file defined in [assetId].
   ///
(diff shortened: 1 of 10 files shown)


build/lib/src/analyzer/resolver.dart
```diff
@@ -14,29 +14,30 @@ abstract class Resolver {
...
   /// Returns a resolved library representing the file defined in [assetId].
   ///
   /// * Throws [NonLibraryAssetException] if [assetId] is not a Dart library.
-  LibraryElement getLibrary(AssetId assetId);
-
-  /// Finds the first library identified by [libraryName], or null if no
-  /// library can be found.
+  Future<LibraryElement> libraryFor(AssetId assetId);
+  @Deprecated('Use libraryFor')
+  Future<LibraryElement> getLibrary(AssetId assetId);
```
jakemac53: I get that `get*` methods are redundant/annoying but I am not really sure if its worth a breaking change? We don't save any typing (actually add a character below) or provide any additional value with this change.

natebosch: I figured since every single usage of this method needs to get visited anyway this was a good chance to drop my pet peeve

jakemac53: heh good point that they have to be updated anyways... sgtm

PR #354, opened 2017-08-10
Disallow implicit casts

natebosch:
- Disable implicit casts in our analysis options.
- Fix a few bugs that were discovered:
  - Missing assignment to memoized _uri field
  - Incorrect argument types
  - Checking wrong variable in AssetGraph matcher. Fix tests that were
    relying on this matcher being a weaker check than intended.
- Fix newly errors which only needed a cast. The most annoying parts are
  around deserialization. A couple are around APIs in external packages
  that are (unintentionally?) dynamic.
- Fix a couple import ordering lints

### DIFF
--- a/build/lib/src/asset/id.dart
+++ b/build/lib/src/asset/id.dart
@@ -98,7 +98,7 @@ class AssetId implements Comparable<AssetId> {
   /// A `package:` URI suitable for use directly with other systems if this
   /// asset is under it's package's `lib/` directory, else an `asset:` URI
   /// suitable for use within build tools.
-  Uri get uri => _uri ?? path.startsWith('lib/')
+  Uri get uri => _uri ??= path.startsWith('lib/')
       ? new Uri(
           scheme: 'package', path: '$package/${path.replaceFirst('lib/','')}')
       : new Uri(scheme: 'asset', path: '$package/$path');
@@ -110,9 +110,9 @@ class AssetId implements Comparable<AssetId> {
   /// Note that this is intended for communicating ids across isolates and not
   /// for persistent storage of asset identifiers. There is no guarantee of
   /// backwards compatibility in serialization form across versions.
-  AssetId.deserialize(data)
-      : package = data[0],
-        path = data[1];
+  AssetId.deserialize(List data)
+      : package = data[0] as String,
+        path = data[1] as String;
 
   /// Returns `true` if [other] is an [AssetId] with the same package and path.
   @override
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -80,7 +80,8 @@ class BuildStepImpl implements BuildStep {
   @override
   Future writeAsBytes(AssetId id, FutureOr<List<int>> bytes) {
     _checkOutput(id);
-    var done = _futureOrWrite(bytes, (b) => _writer.writeAsBytes(id, b));
+    var done =
+        _futureOrWrite<List<int>>(bytes, (b) => _writer.writeAsBytes(id, b));
     _outputsCompleted = _outputsCompleted.then((_) => done);
     return done;
   }
@@ -89,14 +90,14 @@ class BuildStepImpl implements BuildStep {
   Future writeAsString(AssetId id, FutureOr<String> content,
       {Encoding encoding: UTF8}) {
     _checkOutput(id);
-    var done = _futureOrWrite(
+    var done = _futureOrWrite<String>(
         content, (c) => _writer.writeAsString(id, c, encoding: encoding));
     _outputsCompleted = _outputsCompleted.then((_) => done);
     return done;
   }
 
   Future _futureOrWrite<T>(FutureOr<T> content, Future write(T content)) =>
-      (content is Future<T>) ? content.then(write) : write(content);
+      (content is Future<T>) ? content.then(write) : write(content as T);
 
   /// Waits for work to finish and cleans up resources.
   ///
--- a/build/lib/src/generate/run_builder.dart
+++ b/build/lib/src/generate/run_builder.dart
@@ -24,7 +24,8 @@ Future<Null> runBuilder(Builder builder, Iterable<AssetId> inputs,
     AssetReader reader, AssetWriter writer, Resolvers resolvers,
     {Logger logger, String rootPackage}) async {
   logger ??= new Logger('runBuilder');
-  rootPackage ??= new Set.from(inputs.map((input) => input.package)).single;
+  rootPackage ??=
+      new Set<String>.from(inputs.map((input) => input.package)).single;
   //TODO(nbosch) check overlapping outputs?
   Future<Null> buildForInput(AssetId input) async {
     var outputs = expectedOutputs(builder, input);
--- a/build_runner/lib/src/asset/file_based.dart
+++ b/build_runner/lib/src/asset/file_based.dart
@@ -50,7 +50,7 @@ class FileBasedAssetReader implements RunnerAssetReader {
     for (var file in files) {
       // TODO(jakemac): Where do these files come from???
       if (path.basename(file.path).startsWith('._')) continue;
-      yield _fileToAssetId(file, packageNode);
+      yield _fileToAssetId(file as File, packageNode);
     }
   }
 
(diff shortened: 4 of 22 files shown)


build/lib/src/asset/id.dart
```diff
@@ -110,9 +110,9 @@ class AssetId implements Comparable<AssetId> {
   /// Note that this is intended for communicating ids across isolates and not
   /// for persistent storage of asset identifiers. There is no guarantee of
   /// backwards compatibility in serialization form across versions.
-  AssetId.deserialize(data)
-      : package = data[0],
-        path = data[1];
+  AssetId.deserialize(List data)
```
jakemac53: Can we do `List<String>` to be more specific?

Also we should update the comment as it is not applicable to package:build (I think it was copied from barback)

build/lib/src/builder/build_step_impl.dart
```diff
@@ -80,7 +80,8 @@ class BuildStepImpl implements BuildStep {
   @override
   Future writeAsBytes(AssetId id, FutureOr<List<int>> bytes) {
     _checkOutput(id);
-    var done = _futureOrWrite(bytes, (b) => _writer.writeAsBytes(id, b));
+    var done =
+        _futureOrWrite<List<int>>(bytes, (b) => _writer.writeAsBytes(id, b));
```
jakemac53: Hmm, why can't this be inferred based on the type of `bytes`?

build/lib/src/builder/build_step_impl.dart
```diff
@@ -89,14 +90,14 @@ class BuildStepImpl implements BuildStep {
...
-    var done = _futureOrWrite(
+    var done = _futureOrWrite<String>(
         content, (c) => _writer.writeAsString(id, c, encoding: encoding));
     _outputsCompleted = _outputsCompleted.then((_) => done);
     return done;
   }
 
   Future _futureOrWrite<T>(FutureOr<T> content, Future write(T content)) =>
-      (content is Future<T>) ? content.then(write) : write(content);
+      (content is Future<T>) ? content.then(write) : write(content as T);
```
jakemac53: This `as T` cast seems weird as well... I guess the type inference just isn't smart enough here but it should be able to be? Maybe worth filing a bug on the language team.

build/lib/src/generate/run_builder.dart
```diff
@@ -24,7 +24,8 @@ Future<Null> runBuilder(Builder builder, Iterable<AssetId> inputs,
     AssetReader reader, AssetWriter writer, Resolvers resolvers,
     {Logger logger, String rootPackage}) async {
   logger ??= new Logger('runBuilder');
-  rootPackage ??= new Set.from(inputs.map((input) => input.package)).single;
+  rootPackage ??=
+      new Set<String>.from(inputs.map((input) => input.package)).single;
```
jakemac53: It seems like `Set.from` should be generic, as I would expect this to just work as well?

build_runner/lib/src/asset/file_based.dart
```diff
@@ -50,7 +50,7 @@ class FileBasedAssetReader implements RunnerAssetReader {
     for (var file in files) {
       // TODO(jakemac): Where do these files come from???
       if (path.basename(file.path).startsWith('._')) continue;
-      yield _fileToAssetId(file, packageNode);
+      yield _fileToAssetId(file as File, packageNode);
```
jakemac53: nit: maybe just cast `files` above to an `Iterable<File>`

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -157,4 +157,4 @@ class WatchImpl {
 
 Map<AssetId, ChangeType> _collectChanges(List<List<AssetChange>> changes) =>
     new Map.fromIterable(changes.expand((c) => c),
-        key: (c) => c.id, value: (c) => c.type);
+        key: (AssetChange c) => c.id, value: (AssetChange c) => c.type);
```
jakemac53: ack, this should be inferred as well if the iterable had a generic type in `Map.fromIterable`.

I think its worth noting all these and filing a bug on the core libs to add generic types.

build/lib/src/asset/id.dart
natebosch: Will update the comment. I think `List<String>` would work today be cause `<dynamic>[] is List<String>` passes - but I'm not sure if that will always be the case when the VM starts using strong mode semantics at runtime. I think that what I have is more indicative of how the checks work - each element that gets pulled out of the List needs to be cast to a String because each element could fail at runtime.

On a related note - I wonder if we should pull serialization out of these classes...

build/lib/src/builder/build_step_impl.dart
natebosch: The function literal gets in the way.

natebosch: sent that too soon...

The function literal gets in the way.

It would work to have

```dart
Future _someThing(List<int> arg) async {}
...
var done = _futureOrWrite(bytes, _someThing);
```

But once you introduce that function literal the inference gets confused.

Would also work to do

```dart
var done = _futureOrWrite(bytes, (List<int> b) => _writer.writeAsBytes(id, b));
```

Would you prefer that?

jakemac53: would adding `int` to the arg for the function literal fix it? (that seems more desirable to me)

build/lib/src/builder/build_step_impl.dart
natebosch: I'm hoping the enhanced type promotion stuff Bob was looking at will help here. This is in line with the places where an `is` check lets you treat the variable as if it had the cast, but an `is!` doesn't, and the `else` clause doesn't get type promotion from either.

build/lib/src/asset/id.dart
jakemac53: I think if we turned this into `List<String>` it would force an explicit cast to `List<String>` anywhere that is passing in a `List<dynamic>` right?

jakemac53: In terms of the serialization I think its fine to live in the individual classes, but if you have a separate proposal I am open to a discussion certainly.

build/lib/src/generate/run_builder.dart
natebosch: The signature of `Set.from` is `Set.from(Iterable elements)` - note the implicit `<dynamic>` and that the generic arg on `Set` of `<E>` doesn't appear in the signature. There is nothing for the ctor to use to infer the type for the constructed instance.

The reason is that the `from` constructors are intended to help you get a collection with the reified type you want if you know all the elements are of the correct type, even if the reified type of the input Iterable doesn't match. If the signature were `Set.from(Iterable<E> elements)` then some other usages would get more clunky...

`new Set.from(stuff.where((e) => e is String).map((e) => e as String);`

build_runner/lib/src/asset/file_based.dart
natebosch: Similar to the AssetId.deserialize constructor - the reified type of `files` is `Iterable<FilesystemEntity>` - maybe the downcast is allowed in the `as` though? I think with strong mode runtime semantics we can't do that case.

We what we could do is add the extra `.map((e) => e as File)` though. This is the same kind of thing where the ergonomics of Iterable.where with an `is` check kind of suck.

build/lib/src/builder/build_step_impl.dart
jakemac53: I think I prefer it on the function literal, but I guess I would hope that it could be inferred by the type of bytes, which is known.

Since the type of bytes is T, and the type of the first arg is T, it should only need one or the other to infer both right?

build/lib/src/generate/run_builder.dart
jakemac53: I guess that makes sense, although couldn't it just be `new Set.from(stuff as Iterable<String>)` which isn't really much worse than `new Set<String>.from(stuff)`?

Actually though, why not just use `inputs.map((input) => inputs.package).toSet()`, that does the right thing it looks like.

build_runner/lib/src/asset/file_based.dart
natebosch: This one seems to work at runtime without much consequence so I moved the cast. It might start failing in the VM in the future but we can address it then.

build/lib/src/builder/build_step_impl.dart
natebosch: Moved it to the argument. We can start a thread and see if type inference could handle this case - I'm not sure the specific challenges involved.

build/lib/src/asset/id.dart
natebosch: Yeah today we generally have to cast to `List` but we'd have to cast to `List<String>` if I change the argument - and that cast will fail with strong mode runtime semantics - so we'd have to do `new List<String>.from(values)` when we know that `values` only contains `String` elements... that is really clunky so I prefer the cast at the point where we pull the elements out (that's similar to what the from ctor does - shifts collection level generic to element level)

build/lib/src/generate/run_builder.dart
natebosch: the `as Iterable<String>` would fail with strong mode runtime semantics.

The `.toSet()` approach feels much better - good call!

PR #357, opened 2017-08-14
Allow build_barback 0.4.x from build_test

natebosch:
Currently there exists a version lock trying to upgrade to build 0.10

### DIFF
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -15,7 +15,7 @@ dependencies:
   glob: ^1.1.0
 
 dev_dependencies:
-  build_barback: ^0.2.0
+  build_barback: ^0.4.0
   build_test: ^0.6.0
   test: ^0.12.0
   transformer_test: ^0.2.0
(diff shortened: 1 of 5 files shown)


natebosch:
So this depends on getting build 0.10 for the tests to pass - I *think* that should be OK. The `lib/` stuff should all be backwards compatible still.

natebosch:
Travis won't pass for this one until we get some of these new versions published.

Will separately look at the formatting issue.

build/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   glob: ^1.1.0
 
 dev_dependencies:
-  build_barback: ^0.2.0
+  build_barback: ^0.4.0
```
jakemac53: is 0.2.0 still compatible? If so I would just change it to be >=0.2.0 <0.5.0

natebosch: No it's not, build_barback has a public implementation of the Resolver interface which saw breaking changes.

PR #360, opened 2017-08-15
Disallow reads outside the asset graph

### LINKED ISSUE #359: build_runner can fail after reading an asset outside the asset graph
Bug was introduced here: https://github.com/dart-lang/build/commit/fbf0e26d6f9a4fb82769e22ce3413f1f16498ad6#diff-7a296c2aaedb49f2b91b7fa9c81b536cL538

We [always allow](https://github.com/dart-lang/build/blob/5a11a992b317fec7e6a6d810d8f86455dde033c6/build_runner/lib/src/asset/reader.dart#L34) reads of assets outside the graph.

I think what should happen is we should disallow reads outside the graph and make sure all files from all packages are in the graph.

This caused a failure in [json_serializable](https://github.com/dart-lang/json_serializable/pull/31#issuecomment-322509740)

natebosch:
Fixes #359

We used to rely on the graph being updated if it didn't already have the
source node but now expect the graph to keep it's structure after
initialization. In order to allow reads to any file that would have been
visible before add all packages as sources, not just those that have a
build action. This should only increase the size of the serialized graph
for projects which did not read these files anyway - which excludes
anything using BarbackResolvers and crawling imports.

### DIFF
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_runner
-version: 0.4.1+2
+version: 0.4.0+2
 description: Tools to write binaries that run builders.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_runner
-version: 0.4.1
+version: 0.4.1+2
 description: Tools to write binaries that run builders.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build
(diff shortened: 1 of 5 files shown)


build_runner/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_runner
-version: 0.4.0+1
+version: 0.4.1
```
kevmoo: Should be a +2, right? No new API

natebosch: I can never keep the rules straight...

build_runner/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_runner
-version: 0.4.0+1
+version: 0.4.1+2
```
kevmoo: `0.4.0+2` 😁

natebosch: d'oh

natebosch:
Working on a test now

natebosch:
Added a test that passes with the current code and fails with the old.

PR #363, opened 2017-08-16
Add an `action` argument to `resolveSource`

### LINKED ISSUE #355: We might want different usage for `resolveSource`
The `tearDown` future feels a little clunky to me now that all the methods on `Resolver` are async. I had to add a bunch of completers in https://github.com/dart-lang/source_gen/pull/270 because you used to be able to do some synchronous work with the Resolver and now you can't.

It's possible that if we get off of `code_transformers` this issue might go away... if it doesn't we should consider a signature like:

```dart
Future resolveSource(String source, Future Function(Resolver) action)
```

Usage like:
```dart
setUp(() async {
  await resolveSource(source, (resolver) async {
    libraryElement = await resolver.findLibraryByName('foo');
  });
}
```

natebosch:
Closes #355

After the Resolver apis all became async it became impossible to do any
work with the Resolver instance without also passing a `tearDown`
argument. By default expect the work to be done now inside an `action`
callback and continue to allow the old behavior by returning the
resolver in the action and passing a teardown.

Change `tearDown` from a `Future<Null>` to a `Future` for a more
flexible api. Can become `Future<void>` once the language allows it.

### DIFF
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -38,7 +38,7 @@ import 'package_reader.dart';
 /// Resolver resolver;
 ///
 /// setUpAll(() async {
-///   onTearDown = new Completer<Null>();
+///   resolverDone = new Completer<Null>();
 ///   resolver = await resolveSource('...', (resover) => resolver,
 ///       tearDown: resolverDone.future);
 /// });
@@ -82,7 +82,7 @@ Future<T> resolveSource<T>(
 /// Resolver resolver;
 ///
 /// setUpAll(() async {
-///   onTearDown = new Completer<Null>();
+///   resolverDone = new Completer<Null>();
 ///   resolver = await resolveAsset('...', (resover) => resolver,
 ///       tearDown: resolverDone.future);
 /// });
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -13,35 +13,39 @@ import 'in_memory_writer.dart';
 import 'multi_asset_reader.dart';
 import 'package_reader.dart';
 
-/// Returns a future that completes with a source [Resolver] for [inputSource].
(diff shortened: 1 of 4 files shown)


natebosch:
See https://github.com/dart-lang/source_gen/pull/271 for example upgrade

build_test/lib/src/resolve_source.dart
```diff
@@ -13,97 +13,104 @@ import 'in_memory_writer.dart';
...
+/// argument:
+///
 /// ```dart
-/// Completer<Null> onTearDown;
+/// Completer<Null> resolverDone;
+/// Resolver resolver;
 ///
-/// setUp(() {
+/// setUpAll(() async {
 ///   onTearDown = new Completer<Null>();
```
jakemac53: I think this line should be deleted now?

build_test/test/resolve_source_test.dart
```diff
@@ -12,28 +12,22 @@ import 'package:test/test.dart';
 void main() {
   group('should resolveSource of', () {
     test('a simple dart file', () async {
-      var resolverDone = new Completer<Null>();
-      var resolver = await resolveSource(r'''
+      var libExample = await resolveSource(r'''
```
jakemac53: nit: maybe in one of these examples put the expect inside the `action` callback? It would be good just to confirm that works as we want and also as an additional example of how you can use the resolver for multiple things without having to use the `tearDown` logic.

build_test/lib/src/resolve_source.dart
natebosch: It was renamed to `resolverDone` but we still need this initialized.

build_test/test/resolve_source_test.dart
natebosch: Good call - added an example.

matanlurey:
This looks good, thanks!

Issue #366, opened 2017-08-16
Errors can occur during `buildStep.complete()` which are not caught


natebosch:
If a builder uses the following:

```dart
// intentionally unawaited
buildStep.writeAsString(newId, buildStep.readAsString(oldId));
```

Then the exception doesn't happen in `build()` but in `complete()`. This is not in the try/catch so it bubbles up the the uncaught exception handler and get caught after the build has finished.

natebosch:
Somehow the onError handler is catching two exceptions (they are at least duplicates of eachother)
https://github.com/dart-lang/build/blob/9379cf0622dd861f48e78054bf215bdba900231c/build_runner/lib/src/generate/build_impl.dart#L165-L166

It's not clear to me why though - we do seem to have try/catches around all the work.

natebosch:
Ah, this has to do with how we chain the `_outputsCompleted` Future in build_step_impl. We probably want to use the `Result` class from `package:async`.

PR #367, opened 2017-08-16
Capture write errors before calling `complete()`

natebosch:
Fixes #336

Since errors can occur before we ever call `.then()` on these Futures we
need to capture them as a `Result` to hold the error until we reease it
in `complete()`.

Add a regression test that fails before this change.


natebosch:
Ugh, this fixed #366 - linked the wrong item before

PR #368, opened 2017-08-16
Disallow reads into .dart_tool

### LINKED ISSUE #326: Disallow reading root_package|.dart_tool/** during builds
We'll end up reading from here once we start writing transient files - but those reads will come as `some_other_package|lib/file` or maybe `root_package|lib/generated_file` and show not include `.dart_tool` in the `AssetId.path`.

We currently delete the stored asset graph file before the build which is the only thin that might be in there to read for now - but with upcoming changes we shouldn't allow any reads - otherwise a build could impact future builds.

natebosch:
Fixes #326

.dart_tool will never be considered an input or be available for reads.

Add a regression test which expects to fail when trying to read from
this directory.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.4.1-dev
+
+- Prevent reads into `.dart_tool` for more hermetic builds.
+
 ## 0.4.0+3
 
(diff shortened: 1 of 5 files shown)


natebosch:
Requires https://github.com/dart-lang/build/pull/367 for the tests to pass

Issue #200, opened 2017-02-10
over-zealous globbing of valid inputs at start of each build


jakemac53:
It looks like we are globbing all potential inputs ahead of time, including hidden directories etc for the root package, https://github.com/dart-lang/build/blob/1a51edff2c8dff6cff2fe0c25dda39c79ded47c7/build_runner/lib/src/generate/build_impl.dart#L469.

We should be able to skip this step entirely, and just check if inputs are valid lazily when they are requested? We would need to make sure they haven't been created since the current build started, but that should be pretty trivial.

Primary inputs could be collected lazily as well, when the phase that needs them begins.

I might be missing something else though :).

natebosch:
We should be avoiding `['build', 'package', '.pub']` because of [this](https://github.com/dart-lang/build/blob/46e500c48209add989aed7a9ddd8626f3cc98e7f/build_runner/lib/src/asset/file_based.dart#L23) and now `.dart_tool` because of [this](https://github.com/dart-lang/build/blob/46e500c48209add989aed7a9ddd8626f3cc98e7f/build_runner/lib/src/generate/build_impl.dart#L387)

Filed https://github.com/dart-lang/build/issues/374 to make those consistent.

Since we've move towards an eagerly built static asset graph rather than lazy I'll close this out.

PR #373, opened 2017-08-22
List assets based on asset graph

natebosch:
We can know without hitting disk what files will be readable in a phase
- use this directly rather than globbing from disk and filtering.

### DIFF
--- a/build_runner/lib/src/asset/reader.dart
+++ b/build_runner/lib/src/asset/reader.dart
@@ -24,8 +24,10 @@ class SinglePhaseReader implements AssetReader {
   final Set<AssetId> _assetsRead = new Set();
   final AssetGraph _assetGraph;
   final int _phaseNumber;
+  final String _rootPackage;
 
-  SinglePhaseReader(this._delegate, this._assetGraph, this._phaseNumber);
+  SinglePhaseReader(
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/asset/reader.dart
```diff
@@ -58,6 +60,11 @@ class SinglePhaseReader implements AssetReader {
   }
 
   @override
-  Iterable<AssetId> findAssets(Glob glob) =>
-      _delegate.findAssets(glob).where(_isReadable);
+  Iterable<AssetId> findAssets(Glob glob) => _assetGraph.allNodes
+      .where((n) => n.id.package == _rootPackage)
```
jakemac53: Probably worth filing a bug to optimize the graph for filtering by package, not something you need to address right now though.

jakemac53:
I think I was thinking of https://github.com/dart-lang/build/issues/200 which is different but probably can be closed since we have to glob everything ahead of time to build up the asset graph now anyways

build_runner/lib/src/asset/reader.dart
natebosch: Filed https://github.com/dart-lang/build/issues/375

Issue #378, opened 2017-08-23
Stop checking for outputs in the wrong package so many places


natebosch:
We currently check for outputs going to the wrong place in:

- The [BuildStep](https://github.com/dart-lang/build/blob/4035013631e78723ff6d9ef4665c11144f5d4d42/build/lib/src/builder/build_step_impl.dart#L123-L129)
- The build_runner [build_impl](https://github.com/dart-lang/build/blob/4035013631e78723ff6d9ef4665c11144f5d4d42/build_runner/lib/src/generate/build_impl.dart#L396-L401)
- The build_runner [FileBasedAssetWriter](https://github.com/dart-lang/build/blob/4035013631e78723ff6d9ef4665c11144f5d4d42/build_runner/lib/src/asset/file_based.dart#L99-L102) - (for deletes - this one we probably want to keep for safety as it doesn't go through the other check since it's used directly by build_runner)

We also only allow outputs as [declared by the `buildExtensions`](https://github.com/dart-lang/build/blob/4035013631e78723ff6d9ef4665c11144f5d4d42/build/lib/src/builder/build_step_impl.dart#L130-L132) and this should be the only way we check this going forward. Outputs can only change extensions, which means they can only ever be in the same package as inputs. We should shift everything to do earlier checks that the inputs are only ever for packages where we will be allowed to output.

PR #379, opened 2017-08-24
Remove restrictions on outputs by package

natebosch:
Towards #378

We know that outputs can only be for the same package as inputs and we
check that the only assets written are declared outputs. We don't need
to also pass around a root package or force that all inputs are for the
same package - the build systems which call `runBuilder` will need to do
appropriate checks regardless of whether we duplicate them here.

### DIFF
--- a/build/lib/src/generate/run_builder.dart
+++ b/build/lib/src/generate/run_builder.dart
@@ -14,23 +14,21 @@ import '../builder/builder.dart';
 import '../builder/logging.dart';
 import 'expected_outputs.dart';
 
-/// Run [builder] on each asset in [inputs].
+/// Run [builder] on with asset in [inputs] as the primary input.
 ///
-///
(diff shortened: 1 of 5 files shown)


build/lib/src/generate/run_builder.dart
```diff
@@ -14,23 +14,21 @@ import '../builder/builder.dart';
 import '../builder/logging.dart';
 import 'expected_outputs.dart';
 
-/// Run [builder] on each asset in [inputs].
+/// Run [builder] on with asset in [inputs] as the primary input.
```
grouma: This reads odd. Were you trying to say 'Run [builder] with each asset in [inputs] as the primary input.'

natebosch: yeah, will delete 'on'

natebosch: ... and add 'each'

Issue #384, opened 2017-08-25
Overhaul the serve method - write a Handler backed by the asset graph


natebosch:
We need to be able to do some path translation for assets that would come from locations on disk other than their apparent path.

The [comment](https://github.com/dart-lang/build/blob/dead012b3590c76d7575b24723afb490a04aa3b3/build_runner/lib/src/generate/build.dart#L129-L131) is misleading here. We talk about overriding the `requestHandler` to get a different behavior than a server on `address:port` - but this happens regardless. The "custom behavior" you can override is reading files from disk - you can't change how they are served. We have a bug where we don't properly pass through `requestHandler`, but this is fine - changing that isn't something we'd want to support anyway. If anything we should have our request handler go through 

I think we should change our signature to something like:

```dart
abstract class BuildHandler {
  Stream<BuildResult> get builds;
  Future<Reponse> handle(Request request);
}
BuildHandler buildHandler(...);
```

Usage like:

```dart
import 'package:shelf/shelf.dart';
import 'package:build_runner/build_runner.dart;

import 'phases.dart';

void main() {
  // Here `serve` is coming from package:shelf rather than package:build_runner
  serve(buildHandler(phases).handle, address, port);
}
```

jakemac53:
This generally SGTM, we probably want to change the `builds` stream to actually be a Stream of some other class that wraps a `Future<BuildResult>` or similar so that you can know when builds start and finish as opposed to just when they finish.

Alternatively it could be a `Stream<BuildEvent>` which could be an abstract class with `BuildStart` and `BuildEnd` implementations (`BuildEnd` would then add the `BuildResult` field)? That would potentially give us more flexibility (`BuildStart` for instance could expose a set of invalidated assets).

natebosch:
The `handle` method would take care of blocking requests during builds. Are there other use cases for knowing the start/stop of builds?

I'm on board with either approach if we have a use case for exposing that info.

jakemac53:
One potential use case would be exposing a widget which gave you the current build status in the browser. I think many custom integrations would want to know build information though, and adding in the extension point for more info in the future should help reduce churn.

matanlurey:
Yuuuuuuuge 👍 on `BuildEvent` for custom integrations later on.

PR #386, opened 2017-08-28
Support running builders against other packages

### LINKED ISSUE #216: Meta: Can build_runner support running other pkg's builders?
i.e., If you ever wanted `build` to be a ~100% replacement for `barback`, we'd need this.

This would also make `build_runner` a slightly viable alternative to `bazel_codegen` - as you wouldn't be required to check in the results of code generation (just like bazel). Anyway, just an open question - maybe it's not even possible.

natebosch:
Closes #216

Add an argument to `build` which indicates that generated files should
be written to `.dart_tool/build/generated/<package>/<path>` rather than
in the source directory for the root package. With this argument enabled
all writes, including for the root package, go to the generated
directory. If we want to support running 'local' and 'non-local'
builders in the same build we will need to make this more fine grained.

- Don't surface the `FileSystemException` when trying to glob a
  directory that doesn't exist.
- Add BuildCacheReader and BuildCacheWriter which translates paths for
  generated assets into the special directory.
- Don't enforce that builders run only on the root package when outputs
  are written to the generated cache.
- Plumb the `writeToCache` argument through to `build_impl` for both
  `build` and `watch`. We will need further changes in `serve` to
  support this (#384)

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'dart:mirrors';
 
 import 'package:build/build.dart';
+import 'package:glob/glob.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
 import 'package:watcher/watcher.dart';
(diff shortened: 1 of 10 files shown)


build_runner/lib/src/generate/build_definition.dart
```diff
@@ -216,4 +219,15 @@ class BuildDefinitionLoader {
       }
     }
   }
+
+  Iterable<AssetId> _listGeneratedAssetIds() sync* {
+    var glob = new Glob('$generatedOutputDirectory/**');
+    for (var id in _reader.findAssets(glob)) {
+      var packagePath = id.path.substring(generatedOutputDirectory.length + 1);
```
grouma: This can potentially cause issues with relative paths given where this is run. This path should always be relative to the root.

natebosch: Synced offline. I originally thought this was a problem but since we're going through `_reader.findAssets` it will default to relative to the root package so this will always work no matter our `cwd`.

PR #389, opened 2017-08-29
Avoid attempting deletes in other packages

natebosch:
Some paths were going through `_writer.delete` rather than `_delete`
which would not translate paths into the build cache and so existing
generated files from other packages could try delets in the wrong place.

- Add `rootPackage` handling to `InMemoryRunnerAssetReader` so that
  `fineAssets` can work correcly.
- Use the wrapped `_delete` method which translates paths everywhere in
  `build_impl`.
- Add a test which exercises the case of deleting existing assets in the
  generated directory. This test would have passed anyway since the
  `_deleteFilesByDefault` path must be used in tests (StdIoType won't
  work correctly in tests) and that branch was correctly using
  `_delete`.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Prevent reads into `.dart_tool` for more hermetic builds.
 - Bug Fix: Rebuild entire asset graph if the build script changes.
+- Add `writeToCache` argument to `build` and `watch` which separates generated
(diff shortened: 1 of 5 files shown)


natebosch:
I don't think this change needs an entry since it's fixing a bug from https://github.com/dart-lang/build/pull/386 which hasn't been published... however that commit didn't include a changelog entry so I guess we do need one anway

PR #392, opened 2017-08-31
Add module/cycle finding

### LINKED ISSUE #345: Implement a way to determine which .dart files produce a .summary and .js file
As a first pass for getting DDC working we'll only chunk strongly connect components into modules and let all other dart libraries stand as a library on their own.

Add some code that can take a `Resolver` for a dart library and figure out whether we'll be outputting a module for it - or if not which .dart file will get the module.

natebosch:
Closes #345

We'll start with some code that can be run from a SummaryBuilder or
DDCBuilder inline rather than something which writes a .json file.

- Add `package:build_compilers` which will hold the SummaryBuilder and
  DDCBuilder. Configure Travis to run tests.
- Add `modules.dart` with code to determing the primary library in a
  cycle, and to find the modules for imports.


build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:build/build.dart';
+
+class Module {
+  /// The libraries in the strongly connected cycle.
+  ///
+  /// In most cases without cyclic imports this will contain a single asset.
+  final Set<AssetId> cycle;
```
jakemac53: nit: I would just call this `srcs` so it is more general.

build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
...
+  /// compile this module.
+  ///
+  /// Every import from any of the assets in [cycle] will be covered by the
+  /// modules rooted within this set.
+  final Set<AssetId> dependencies;
+  Module(this.cycle, this.dependencies);
+}
+
+Module defineModule(AssetId from, LibraryElement library) {
+  final cycleUris = library.libraryCycle.map((l) => '${l.source.uri}').toSet();
```
jakemac53: nit: these aren't of type `Uri` so I would name this something else (also below for some vars)

build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
...
+      dependencyModules.map(toAssetId).toSet());
+}
+
+/// Returns whether [library] owns the module for it's strongly connected import
+/// cycle.
+bool isPrimary(LibraryElement library) =>
+    _earliest(library.libraryCycle) == '${library.source.uri}';
+
+String _earliest(List<LibraryElement> libraries) {
+  if (libraries.length < 2) return '${libraries.first.source.uri}';
```
jakemac53: nit: check `== 1`? Possibly add a different `isEmpty` check that throws?

build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
...
+  ///
+  /// In most cases without cyclic imports this will contain a single asset.
+  final Set<AssetId> cycle;
+
+  /// The libraries which would have a `.sum` or `.js` output necessary to
+  /// compile this module.
+  ///
+  /// Every import from any of the assets in [cycle] will be covered by the
+  /// modules rooted within this set.
+  final Set<AssetId> dependencies;
```
jakemac53: nit: As far as I can tell these are only direct dependencies, I would rename it as such. The comment is also misleading because it claims these are all the required inputs but they aren't.

build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:build/build.dart';
+
+class Module {
```
jakemac53: You may also want to add a `primarySource` field here?

build_compilers/test/modules_test.dart
```diff
@@ -0,0 +1,80 @@
...
+
+import 'package:build_compilers/src/modules.dart';
+
+void main() {
+  LibraryElement libCycle;
+  LibraryElement libSecondaryInCycle;
+  LibraryElement libNoCycle;
+  LibraryElement libCycleWithB;
+  LibraryElement libCycleWithA;
+  final assetCycle = makeAssetId('a|lib/a_cycle.dart');
```
jakemac53: nit: rename `assetCycleA` (and `libCycleA` etc)?

build_compilers/lib/src/modules.dart
matanlurey: It might be nice to dartdoc this class, given how overloaded "Module" is sometimes.

build_compilers/test/modules_test.dart
```diff
@@ -0,0 +1,80 @@
...
+  group('isPrimary', () {
+    test('Marks library with no cycle as primary', () async {
+      expect(isPrimary(libNoCycle), true);
+    });
+
+    test('Marks library with lowest alpha sort as primary', () async {
+      expect(isPrimary(libCycle), true);
+    });
+
+    test('Does not mark library with higher alpha sort as primary', () async {
```
jakemac53: nit: dartfmt

build_compilers/lib/src/modules.dart
matanlurey: Is it possible to provide an example where this will/won't represent a cycle?

Perhaps add a bool getter hasCyclicImports?

build_compilers/lib/src/modules.dart
matanlurey: An example for this might be nice too.

build_compilers/lib/src/modules.dart
```diff
@@ -0,0 +1,68 @@
...
+  /// The libraries which would have a `.sum` or `.js` output necessary to
+  /// compile this module.
+  ///
+  /// Every import from any of the assets in [cycle] will be covered by the
+  /// modules rooted within this set.
+  final Set<AssetId> dependencies;
+  Module(this.cycle, this.dependencies);
+}
+
+Module defineModule(AssetId from, LibraryElement library) {
```
matanlurey: Consider making this a factory constructor on `Module`?

natebosch:
Somehow I completely ignored the fact that we do need transitive module dependencies 😢 

We will end up wanting to serialize this to JSON. We're pretty concerned about the performance impact of serializing/deserializing these so often, and we have some ideas on how we might be able to optimize it but we'll wait until we can measure some performance issues - we expect other optimizations like the Resolver are going to buy much more benefit.

build_compilers/lib/src/modules.dart
natebosch: Done

natebosch: Done

build_compilers/lib/src/modules.dart
natebosch: Calling it `sources`

natebosch: Added an example of a cycle. I don't think I'll add `hasCyclicImports` until a use case comes up.

build_compilers/lib/src/modules.dart
natebosch: Renamed as `directDependencies` and and updated the comment.

natebosch: I'm hoping this should be clear based on the new comment and the example on `sources`. Let me know if you think this is still unclear.

build_compilers/lib/src/modules.dart
natebosch: Went back to type `Uri` and I'm sorting by `uri.path` now instead of calling `toString()` on them repeatedly for comparison.

build_compilers/lib/src/modules.dart
natebosch: Using an assert for the empty case.

build_compilers/test/modules_test.dart
natebosch: The interesting bit is that it's a cycle, not that it's in package A. Would prefer to keep this as is unless you feel strongly about it.

build_compilers/test/modules_test.dart
natebosch: Also added more groups so that the names could be a little shorter and removed the unnecessary `async`s

build_compilers/test/modules_test.dart
jakemac53: It just makes it harder to understand the test imo, but I don't feel that strongly. I found myself jumping around this file a lot when reviewing it to understand what was going on etc.

build_compilers/lib/src/modules.dart
```diff
@@ -5,26 +5,55 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
 
+/// A collection of Dart libraries in a strongly connected component and the
```
jakemac53: nit: `Dart files`, it may include part files etc right?

Also we should probably add some testing around part files?

build_compilers/lib/src/modules.dart
```diff
@@ -5,26 +5,55 @@
...
+/// Modules can span pub package boundaries when there are import cycles across
+/// packages.
 class Module {
-  /// The libraries in the strongly connected cycle.
+  /// The library which will be used to reference any library in [sources].
   ///
-  /// In most cases without cyclic imports this will contain a single asset.
-  final Set<AssetId> cycle;
+  /// The assets which are built once per module, such as DDC compiled output or
+  /// Analyzer summaries, will be build for the primary source and not for any
```
jakemac53: This makes it sound as if the other sources aren't included in the summary. Maybe instead say that the summaries/ddc output will be named after the primary source?

build_compilers/lib/src/modules.dart
natebosch: Done

build_compilers/lib/src/modules.dart
natebosch: Filed https://github.com/dart-lang/build/issues/393

I think we should cover part files in a followup.

build_compilers/lib/src/modules.dart
natebosch: Rephrased.

Issue #400, opened 2017-09-07
Signal when builds are happening


natebosch:
Today `watch` and `serve` return a `Stream<BuildResult>`. In https://github.com/dart-lang/build/issues/384 we considered changing it to `Stream<Future<BuildResult>>` or `Stream<BuildEvent>` where the event is either a `BuildStart` or `BuildEnd`.

I think we should also consider

```dart
class BuildResults {
  /// The current build if any is running.
  Future<BuildResult> currentBuild;
  Stream<BuildResult> builds;
}
```

This is most directly satisfies the "block if a build is happening" case and doesn't require a `where` and `map` to get back to `Stream<BuildResult>`. If it turns out to be necessary we can also add a `Stream<Null> buildStarts` without a breaking change.

kevmoo:
`Stream<Future<BuildResult>>` makes me 🤢 – still the plan?

natebosch:
My current PR goes with the class that has a `Future<BuildResult>` for the active build and a  `Stream<BuildResult>`. If you want to advocate instead for `Stream<BuildEvent>` with start and stop events we can also explore that further.

kevmoo:
Just no stream of future :-)

From mobile. Sorry for brevity and/or spelling.

jakemac53:
I like wrapping everything up in a single class which we can return an instance of from the various top level methods. That makes it easier for us to provide some convenience methods and additional information later on.

I would rename the `builds` getter to `buildResults` just so it is more specific and makes more sense later on if/when we do add other streams.

In the future I think `BuildResult` should probably extend a base `BuildEvent` class as opposed to being a separate object because that makes it easier to provide a single `Stream<BuildEvent>`, but that could happen if/when we decide to add a stream of `BuildStart` events.

natebosch:
Once we publish the latest build_runner `watch` will return a class with this information.

PR #401, opened 2017-09-07
Add BuildResults and make WatchImpl private

natebosch:
Towards #400

- Add `BuildResults` to allow signaling when there is an ongoing build.
- Rename `WatchImpl` to the private `_Watch` and expose it only through
  the new top level `runWatch` as a `BuildResult`.
- Remove gates around already running watcher since it's not possible to
  get a handle on the instance to run it twice
- Remove `terminate()` and eagerly pass in a `Future` to signal
  termination.
- Add `_Terminator` class in `build.dart` to hold this signal and remove
  the multiple paths to cancel the listen. Previously either the Future
  returned from terminate, or the end of the Stream, would signal
  completion, now only use the end of the Stream.

### DIFF
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -58,18 +58,12 @@ Future<BuildResult> build(List<BuildAction> buildActions,
       writer: writer,
       logLevel: logLevel,
       onLog: onLog);
+  var terminator = new _Terminator(terminateEventStream);
   var buildImpl = new BuildImpl(options, buildActions);
 
-  /// Run the build!
-  var futureResult = buildImpl.runBuild();
+  var result = await buildImpl.runBuild();
 
-  // Stop doing new builds when told to terminate.
-  _setupTerminateLogic(terminateEventStream, () {
-    new Logger('Build').info('Waiting for build to finish...');
-    return futureResult;
-  }, cancelWhen: futureResult);
-
-  var result = await futureResult;
+  await terminator.cancel();
   await options.logListener.cancel();
   return result;
 }
@@ -110,20 +104,16 @@ Stream<BuildResult> watch(List<BuildAction> buildActions,
       onLog: onLog,
       debounceDelay: debounceDelay,
       directoryWatcherFactory: directoryWatcherFactory);
-  var watchImpl = new WatchImpl(options, buildActions);
+  var terminator = new _Terminator(terminateEventStream);
+  var buildResults =
+      runWatch(options, buildActions, terminator.shouldTerminate);
 
-  var resultStream = watchImpl.runWatch();
-
-  // Stop doing new builds when told to terminate.
-  _setupTerminateLogic(terminateEventStream, () async {
-    await watchImpl.terminate();
-  }, cancelWhen: watchImpl.onTerminated);
-
-  watchImpl.onTerminated.then((_) {
-    options.logListener.cancel();
+  buildResults.builds.drain().then((_) async {
+    await terminator.cancel();
+    await options.logListener.cancel();
   });
 
-  return resultStream;
+  return buildResults.builds;
 }
 
 /// Same as [watch], except it also provides a server.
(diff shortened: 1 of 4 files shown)


natebosch:
Functionally the only difference here is some of the extra logging during termination is removed.

build_runner/lib/src/generate/build.dart
```diff
@@ -159,46 +149,42 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
...
-      exit(2);
-    }
-  });
-
-  cancelWhen?.then((_) {
-    terminateListener.cancel();
-  });
-  return terminateListener;
+class _Terminator {
+  final Future shouldTerminate;
```
jakemac53: nit: add a doc comment to this

build_runner/lib/src/generate/build.dart
```diff
@@ -159,46 +149,42 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
...
+  final StreamSubscription _subscription;
+
+  factory _Terminator(Stream terminateEventStream) {
```
jakemac53: nit: if this supports null values it should probably be an optional arg (if for no other reason than to send the right signal that it isn't required)

build_runner/lib/src/generate/build.dart
```diff
@@ -159,46 +149,42 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
...
-    } else {
-      exit(2);
-    }
-  });
-
-  cancelWhen?.then((_) {
-    terminateListener.cancel();
-  });
-  return terminateListener;
+class _Terminator {
```
jakemac53: Probably worth adding a simple doc comment to this class as well describing how its intended to be used.

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -20,9 +20,13 @@ import 'exceptions.dart';
 import 'options.dart';
 import 'phase.dart';
 
+BuildResults runWatch(
```
jakemac53: nit: Lets add some doc comments (what is `until`?)

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -67,72 +61,50 @@ class WatchImpl {
       _expectedDeletes.add(id);
       if (existingOnDelete != null) existingOnDelete(id);
     };
+    builds = this.runWatch(until);
```
jakemac53: nit: rename the instance method so you don't need `this` to disambiguate for the user (maybe just make it private?)

jakemac53:
Also I think we should restore the logging in some form, you should get some immediate feedback when you hit ctrl-c otherwise you will just hit it again which we don't want users to do in most scenarios (its better if they let the current build finish)

build_runner/lib/src/generate/build_result.dart
```diff
@@ -54,3 +56,8 @@ enum BuildStatus {
 
 /// The type of a build.
 enum BuildType { incremental, full }
+
+abstract class BuildResults {
```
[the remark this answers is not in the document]
natebosch: Going with `BuildState`

build_runner/lib/src/generate/build_result.dart
```diff
@@ -54,3 +56,8 @@ enum BuildStatus {
...
+  Future<BuildResult> get currentBuild;
+  Stream<BuildResult> get builds;
```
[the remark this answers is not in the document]
natebosch: using `buildResults`

build_runner/lib/src/generate/watch_impl.dart
natebosch: Done

build_runner/lib/src/generate/watch_impl.dart
natebosch: Done

build_runner/lib/src/generate/build.dart

build_runner/lib/src/generate/build.dart

build_runner/lib/src/generate/build.dart

natebosch:
Added back some logging to termination.

build_runner/lib/src/generate/build.dart
```diff
@@ -150,26 +149,33 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
...
     await options.logListener.cancel();
   });
 
-  return buildResults.builds;
+  return buildState.buildResults;
 }
 
+/// Fires [shouldTerminate] once a `SIGINT` is intercepted.
+///
+/// The `SIGINT` stream can optionally be replaces with another Stream in the
```
jakemac53: s/replaces/replaced

build_runner/lib/src/generate/build.dart
```diff
@@ -150,26 +149,33 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
...
+/// constructor. [cancel] should be called after work is finished. If multiple
+/// events are receieved on the terminate event stream before work is finished
+/// the process will be terminated with [exit].
 class _Terminator {
+  /// A Future that fires when a signal has been received indicated that builds
```
jakemac53: s/indicated/indicating

build_runner/lib/src/server/server.dart
```diff
@@ -18,14 +18,14 @@ Future<HttpServer> _futureServer;
...
-Future<HttpServer> startServer(BuildResults results, BuildOptions options) {
+Future<HttpServer> startServer(BuildState buildState, BuildOptions options) {
   if (_futureServer != null) {
     throw new StateError('Server already running.');
   }
 
   try {
     blockingHandler = (Request request) async {
-      if (results.currentBuild != null) await results.currentBuild;
+      if (buildState.currentBuild != null) await buildState.currentBuild;
```
jakemac53: I think you can await null, so you probably don't need this if?

build_runner/lib/src/generate/build.dart

build_runner/lib/src/generate/build.dart

build_runner/lib/src/server/server.dart
natebosch: Done.

PR #405, opened 2017-09-11
initial impl of package:scratch_space

jakemac53:
See README.md for full description, but at a high level this is a new package that exposes a `ScratchSpace` class which can be used to manage a temp directory that mirrors assets from the build system, under which you can run arbitrary executables that don't have dart apis, and then copying those outputs back in a way the build system can understand.

### DIFF
--- a/scratch_space/README.md
+++ b/scratch_space/README.md
@@ -1,0 +1,55 @@
+# [![Build Status](https://travis-ci.org/dart-lang/build.svg?branch=master)](https://travis-ci.org/dart-lang/build)
+
+## [`ScratchSpace`][dartdoc:ScratchSpace]
+
+A `ScratchSpace` is a thin wrapper around a temporary directory. The
+constructor takes zero arguments, so making one is as simple as doing
+`new ScratchSpace()`.
+
+### Adding assets to a `ScratchSpace`
+
+To add assets to the `ScratchSpace`, you use the `ensureAssets` method, which
+takes an `Iterable<AssetId>` and an `AssetReader` (for which you should
+generally pass your `BuildStep` which implements that interface).
+
+**Note:** It is important to note that the `ScratchSpace` does not guarantee
+that the version of a file within it is the most updated version, only that
+some version of it exists. For this reason you should create a new
+`ScratchSpace` for each build (within a single build files cannot change, but
+between builds they might).
+
+### Deleting a `ScratchSpace`
+
+When you are done with a `ScratchSpace` you should call `delete` on it to make
+sure it gets cleaned up, otherwise you can end up filling up the users tmp
+directory.
+
+**Note:** You cannot delete individual assets from a `ScratchSpace` today, you
+can only delete the entire thing. If you have a use case for deleting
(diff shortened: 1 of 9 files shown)


scratch_space/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+**Note:** It is important to note that the `ScratchSpace` does not guarantee
+that the version of a file within it is the most updated version, only that
+some version of it exists. For this reason you should create a new
+`ScratchSpace` for each build (within a single build files cannot change, but
+between builds they might).
+
+### Deleting a `ScratchSpace`
+
+When you are done with a `ScratchSpace` you should call `delete` on it to make
```
natebosch: Can we look for a way for this to be done automatically using new build lifecycle apis in package:build? Will we try to do that in a followup?

scratch_space/README.md
```diff
@@ -0,0 +1,55 @@
...
+**Note:** You cannot delete individual assets from a `ScratchSpace` today, you
+can only delete the entire thing. If you have a use case for deleting
+individual files you can [file an issue][tracker]. 
+
+### Getting the actual File objects for a `ScratchSpace`
+
+When invoking an external binary, you probably need to tell it where to look
+for files. There are a few fields/methods to help you do this:
+
+  * `String get packagesDir`: The `packages` directory in the `ScratchSpace`.
```
natebosch: Will we write a `.packages` as well? Do we expect tools to continue to understand the directory style in addition to the manifest style?

scratch_space/lib/src/util.dart
```diff
@@ -0,0 +1,23 @@
...
+
+/// Returns the top level directory in [uri].
+///
+/// Throws an [ArgumentError] if [uri] is just a filename with no directory.
+String topLevelDir(String uri) {
+  var parts = p.url.split(p.url.normalize(uri));
+  String error;
+  if (parts.length == 1) {
+    error = 'The uri `$uri` does not contain a directory.';
+  } else if (parts.first == '..') {
```
natebosch: This would also escape: `/allowed/path/../../../not_allowed`

I think we need `parts.contains('..')`

scratch_space/README.md
jakemac53: Yes, once those are available we will add some helpers here that use them.

scratch_space/README.md
jakemac53: We could certainly... we would have to update it any time you call `ensureAssets` but that wouldn't be difficult.

DDC/Analyzer don't need it though, so I think maybe we could wait and add it as an option later on if somebody asks for it?

natebosch: Makes sense

scratch_space/lib/src/util.dart
jakemac53: There are actually tests for this case :). Since we normalize the uri first we only have to check the first segment.

jakemac53: I wasn't actually testing the basic case so I added a test for that as well

PR #406, opened 2017-09-11
Replace `serve` with `ServeHandler`

### LINKED ISSUE #384: Overhaul the serve method - write a Handler backed by the asset graph

natebosch:
Closes #384

It will now be the responsibility of the calling script to start the
server with shelf.

- Add `BuildHandler` class which implement `BuildState` and adds a shelf
  handler method.
- Add a handler backed by an `AssetReader`.
- Make `WatchImpl` public again and expose the necessary properties for
  creating the new handler.

### DIFF
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -121,8 +121,9 @@ Stream<BuildResult> watch(List<BuildAction> buildActions,
 /// By default a static server will be set up to serve [directory] at
 /// [address]:[port], but instead a [requestHandler] may be provided for custom
 /// behavior.
-Stream<BuildResult> serve(List<BuildAction> buildActions,
+Future<BuildHandler> buildHandler(List<BuildAction> buildActions,
     {bool deleteFilesByDefault,
+    bool writeToCache,
     PackageGraph packageGraph,
     RunnerAssetReader reader,
     RunnerAssetWriter writer,
@@ -137,6 +138,7 @@ Stream<BuildResult> serve(List<BuildAction> buildActions,
     Handler requestHandler}) {
   var options = new BuildOptions(
       deleteFilesByDefault: deleteFilesByDefault,
+      writeToCache: writeToCache,
       packageGraph: packageGraph,
       reader: reader,
       writer: writer,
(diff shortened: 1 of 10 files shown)


natebosch:
Getting some early eyes on this. Will add tests.

build_runner/lib/src/generate/build.dart
```diff
@@ -121,8 +121,9 @@ Stream<BuildResult> watch(List<BuildAction> buildActions,
 /// By default a static server will be set up to serve [directory] at
```
jakemac53: will need to update the docs :)

build_runner/lib/src/server/server.dart
```diff
@@ -4,45 +4,56 @@
...
+        : new AssetId(_rootPackage, p.joinAll(pathSegments));
+    try {
+      if (!await _reader.canRead(assetId)) {
+        return new Response.notFound('Not Found');
+      }
+    } on ArgumentError catch (_) {
+      return new Response.notFound('Not Found');
+    }
+    var bytes = await _reader.readAsBytes(assetId);
+    var headers = {HttpHeaders.CONTENT_LENGTH: '${bytes.length}'};
```
jakemac53: We are going to want to add more headers here as well, to handle some form of caching (probably just etag based?)

build_runner/lib/src/generate/build.dart
```diff
@@ -121,8 +121,9 @@ Stream<BuildResult> watch(List<BuildAction> buildActions,
...
 /// [address]:[port], but instead a [requestHandler] may be provided for custom
 /// behavior.
-Stream<BuildResult> serve(List<BuildAction> buildActions,
+Future<BuildHandler> buildHandler(List<BuildAction> buildActions,
```
jakemac53: I am not sure about the name here, it sounds more like its handling builds than requests :).

build_runner/lib/src/generate/build.dart

build_runner/lib/src/generate/build.dart
natebosch: Change to ServeHandler and dropped this method in favor of having only `watch`

build_runner/lib/src/server/server.dart
natebosch: yeah definitely but I think we can wait until we have some experience using this and add it then. We'll be heavily influence by the types of optimizations we make using the AssetGraph

natebosch:
If we wanted to we *could* make `watch` return a synchronous `ServeHandler` and hid more of the asynchrony underneath since it only exposes asynchronous methods. Let me know if you want me to try it. See `watch_test.dart` for the type of work that needs to be done by clients otherwise if they want to keep a synchronous access to `Stream<BuildResult>`

PR #410, opened 2017-09-12
Add `Resource` and `ResourceManager` classes 

jakemac53:
Fixes https://github.com/dart-lang/build/issues/396

See [README.md](https://github.com/dart-lang/build/pull/410/files#diff-ecbc6c47898b2960c58173697b2fb550R163) for more info.

PR #411, opened 2017-09-12
add info to the README about using a ScratchSpace as a Resource

jakemac53:
Based on https://github.com/dart-lang/build/pull/410

### DIFF
--- a/scratch_space/README.md
+++ b/scratch_space/README.md
@@ -6,17 +6,41 @@ A `ScratchSpace` is a thin wrapper around a temporary directory. The
 constructor takes zero arguments, so making one is as simple as doing
 `new ScratchSpace()`.
 
+In general you actually want to wrap a `ScratchSpace` in a `Resource`, so that
+you can re-use the scratch space across build steps in an individual build.
+This is safe to do since you are not allowed to overwrite files within a build.
+
+This should look something like the following:
+
+```
+final myScratchSpaceResource =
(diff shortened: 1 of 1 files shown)


scratch_space/README.md
```diff
@@ -6,17 +6,41 @@ A `ScratchSpace` is a thin wrapper around a temporary directory. The
 constructor takes zero arguments, so making one is as simple as doing
 `new ScratchSpace()`.
 
+In general you actually want to wrap a `ScratchSpace` in a `Resource`, so that
```
natebosch: [nit] drop 'actually'

scratch_space/README.md
```diff
@@ -6,17 +6,41 @@ A `ScratchSpace` is a thin wrapper around a temporary directory. The
...
+you can re-use the scratch space across build steps in an individual build.
+This is safe to do since you are not allowed to overwrite files within a build.
+
+This should look something like the following:
+
+```
+final myScratchSpaceResource =
```
natebosch: do we want to have this as a top level field within the scratch space package? Do we think most uses would want a communal scratch space, or most uses would want to have their own per builder?

scratch_space/README.md
jakemac53: done

scratch_space/README.md
jakemac53: I opted not to do this for now just because it is less likely to run into issues if you have full control over your own scratch space. If we encourage all users to share one then one badly behaved builder could bork everything.

PR #414, opened 2017-09-12
Refactor checks for file changes in between builds

### LINKED ISSUE #408: When using writeToCache generated assets look like they are missing
We always try to rebuild stuff, even when the assets exist in the cache and we have a cached asset graph.

natebosch:
Fixes #408

Previous approach would not see generated files in the asset cache since
it was using the reader from the options instead of the wrapped reader.
Rather than make the `BuildCacheReader` implement `RunnerAssetReader`
refactor the check for updates to use the already loaded
`currentSources` which does honor the asset cache.


natebosch:
Added two tests, one which used to pass for the non writeToCache case, and one which used to fail and passes now for the writeToCache case.

PR #415, opened 2017-09-12
Try `index.html` if the requested URL is not found

### LINKED ISSUE #413: Default serve handler should have a default document of `index.html`
You should be able to go to a directory and have it load the `index.html` by default.

natebosch:
Fixes #413

- Add tests for AssetHandler
- Fix bug when `packages/` is at the beginning of the path instead of
  nested.
- Pull out a separate `_handle` method operating on the path segments
  try it again with `index.html` appended to the path whenever the first
  request is a 404.

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -43,13 +43,12 @@ class AssetHandler {
 
   AssetHandler(this._reader, this._rootPackage);
 
-  Future<Response> handle(Request request) async {
-    var response = await _handle(request.url.pathSegments);
-    if (response.statusCode == 404) {
-      response = await _handle(
-          new List.from(request.url.pathSegments)..add('index.html'));
+  Future<Response> handle(Request request) {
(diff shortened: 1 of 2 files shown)


natebosch:
This doesn't make dartium work, might need to only do this when the URI ends with a trailing slash.

build_runner/lib/src/server/server.dart
```diff
@@ -43,10 +43,17 @@ class AssetHandler {
 
   AssetHandler(this._reader, this._rootPackage);
 
-  Future<Response> handle(Request request) async {
+  Future<Response> handle(Request request) {
     var pathSegments = request.url.pathSegments;
+    if (request.url.path.endsWith('/')) {
```
jakemac53: Pub does not require this (and I don't  think most servers generally do).

natebosch: synced offline. Dartium doesn't work visiting `host:port/web` but does visiting `host:port/web/`. We should probably stick with this for now and revisit if we find a need.

PR #418, opened 2017-09-12
Ignore file changes under .git

### LINKED ISSUE #417: Running `ls` triggers another build when using watch
There are always 0 outputs, but somehow we're triggering the file watchers even without touching any files. Maybe this isn't a big deal?

natebosch:
Fixes #417

A `git commit` shouldn't trigger a build. We might end up finding other
places where we should be ignoring files, but git is wideley used enough
that we can hardcode support.

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -117,13 +117,16 @@ class WatchImpl implements BuildState {
   /// Checks if we should skip a watch event for this [change].
   bool _shouldProcess(AssetChange change) {
     if (_isCacheFile(change)) return false;
+    if (_isGitFile(change)) return false;
     if (_isEditOnGeneratedFile(change)) return false;
     if (_isExpectedDelete(change)) return false;
     if (_isUnwatchedDelete(change)) return false;
     return true;
   }
 
-  bool _isCacheFile(AssetChange change) => change.id.path.contains(cacheDir);
+  bool _isCacheFile(AssetChange change) => change.id.path.startsWith(cacheDir);
+
+  bool _isGitFile(AssetChange change) => change.id.path.startsWith('.git/');
 
   bool _isEditOnGeneratedFile(AssetChange change) =>
       _assetGraph.get(change.id) is GeneratedAssetNode &&


natebosch:
It's nasty to test that a build *doesn't* happen so it's probably not worth a test with hardcoded waits.

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -117,13 +117,16 @@ class WatchImpl implements BuildState {
...
     if (_isCacheFile(change)) return false;
+    if (_isGitFile(change)) return false;
     if (_isEditOnGeneratedFile(change)) return false;
     if (_isExpectedDelete(change)) return false;
     if (_isUnwatchedDelete(change)) return false;
     return true;
   }
 
-  bool _isCacheFile(AssetChange change) => change.id.path.contains(cacheDir);
+  bool _isCacheFile(AssetChange change) => change.id.path.startsWith(cacheDir);
```
kevmoo: Why the change here? Should we be using pkg/path?

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -117,13 +117,16 @@ class WatchImpl implements BuildState {
...
+
+  bool _isGitFile(AssetChange change) => change.id.path.startsWith('.git/');
```
kevmoo: Not a windows-friendly path.

Better to use pkg/path?

build_runner/lib/src/generate/watch_impl.dart
natebosch: Just realized that we could be more optimal since the cache dir is always at the root of the package. I don't think we need package:path but I could be overly optimistic.

build_runner/lib/src/generate/watch_impl.dart
natebosch: In theory `AssetId` is supposed to already be normalized for us but we might have bugs. If so I'd rather fix them there than here.

natebosch: https://github.com/dart-lang/build/blob/cca7237378ff45c7858158c07c501072c2ffe1f0/build/lib/src/asset/id.dart#L146-L156

PR #422, opened 2017-09-13
Skip builds for changes sources with no outputs

### LINKED ISSUE #47: Should be smarter about running builds
Today if any files in any packages are updated it will run a rebuild, which often ends up being a no-op. Instead we should at least verify that the updates are for files which either exist in the asset graph (ie: they are a dependency) or which match one of the input sets for one of the phases.

natebosch:
Fixes #47


jakemac53:
We should add a test probably as well, could be a followup though.

We are going to need some better way of tracking in tests whether or not a build happened.

natebosch:
Filed https://github.com/dart-lang/build/issues/424 to track adding tests for this.

PR #428, opened 2017-09-14
Add orderedPackages and dependentsOf utilities

natebosch:
As we implement automatic build scripts it's useful to be able to have a
proeorder sort of packages by dependency. Add a method to get the
preorder sort, and a method to filter it to the packages which have a
certain dependency.

### DIFF
--- a/build_runner/lib/src/package_graph/package_graph.dart
+++ b/build_runner/lib/src/package_graph/package_graph.dart
@@ -1,6 +1,8 @@
 // Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
+import 'dart:collection';
 import 'dart:io';
 
 import 'package:path/path.dart' as path;
@@ -116,6 +118,40 @@ class PackageGraph {
 
   /// Shorthand to get a package by name.
   PackageNode operator [](String packageName) => allPackages[packageName];
+
+  /// Finds all packages which depend on [packageName] in postorder by
+  /// dependencies.
+  ///
+  /// See [orderedPackages] for ordering guarantees. The node for [packageName]
+  /// will not be included in the result.
+  Iterable<PackageNode> dependentsOf(String packageName) {
+    if (!allPackages.containsKey(packageName)) return const [];
+    var node = allPackages[packageName];
+    return orderedPackages.where((n) => n.dependencies.contains(node));
+  }
+
+  /// All of the packages in postorder by dependencies.
+  ///
+  /// Depedenencies of a package will come after the package in the result. If
+  /// there is a package cycle the relative position of packages within the
+  /// cycle is non-deterministic. For any two packages for which neither is a
+  /// transitive dependency of the other the relative position of the packages
+  /// within the cycle is non-deterministic. The root package will always be
+  /// last in this list regardless of it's position in a cycle.
+  Iterable<PackageNode> get orderedPackages sync* {
+    var seen = new Set<PackageNode>();
+    var queue = new Queue<PackageNode>();
+    queue.addLast(root);
+    while (queue.isNotEmpty) {
+      var next = queue.last;
+      if (next.dependencies.isEmpty || seen.contains(next)) {
+        yield queue.removeLast();
+      } else {
+        queue.addAll(next.dependencies);
+        seen.add(next);
+      }
+    }
+  }
 
   @override
   String toString() {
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -117,6 +119,40 @@ class PackageGraph {
...
+  /// the other the relative position of the packages within the cycle is
+  /// non-deterministic.
+  Iterable<PackageNode> get orderedPackages sync* {
+    var seen = new Set<PackageNode>();
+    var queue = new Queue<PackageNode>();
+    queue.addLast(root);
+    while (queue.isNotEmpty) {
+      var next = queue.last;
+      if (next.dependencies.isEmpty || seen.contains(next)) {
+        yield queue.removeLast();
```
jakemac53: I think this logic will result in duplicate packages if you have a shared dependency between two packages (for instance if `a` is the root and depends on `b`, and `c`, and both `b` and `c` depend on `d`, then `d` will get yielded multiple times).

To me the recursive logic is easier to follow than this iterative version (especially with the queue and sync* etc), and we shouldn't ever have so many packages that a recursive algorithm could run into issues. That is just my preference though.

```dart
Iterable<PackageNode> get orderedPackages {
  var packages = <PackageNode>[];
  var seen = new Set<PackageNode>();

  void crawlDeps(PackageNode package) {
    seen.add(package);
    for (var dep in package.dependencies) {
      if (seen.contains(dep)) continue;
      crawlDeps(dep);
    }
    packages.add(package);
  }
  crawlDeps(root);
  
  return packages;
}
```

build_runner/test/package_graph/package_graph_test.dart
```diff
@@ -150,6 +150,63 @@ void main() {
           throwsA(anything));
     });
   });
+
+  group('orderedPackages', () {
```
jakemac53: Please add another test for the shared dependencies scenario I listed above.

build_runner/test/package_graph/package_graph_test.dart
```diff
@@ -150,6 +150,63 @@ void main() {
...
+    });
+
+    test('handles cycles', () {
+      var a = new PackageNode('a', '1.0.0', PackageDependencyType.path, null);
+      var b = new PackageNode('b', '1.0.0', PackageDependencyType.path, null);
+      a.dependencies.add(b);
+      b.dependencies.add(a);
+      var graph = new PackageGraph.fromRoot(a);
+      var inOrder = graph.orderedPackages.map((n) => n.name).toList();
+      expect(inOrder, containsAllInOrder(['a', 'b']));
```
jakemac53: nit: technically the order here is non-deterministic so you could just use `contains`.

Also it would probably be good to add another test with a cycle that doesn't start from the root.

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -117,6 +119,40 @@ class PackageGraph {
...
+  /// will not be included in the result.
+  Iterable<PackageNode> dependentsOf(String packageName) {
+    if (!allPackages.containsKey(packageName)) return const [];
+    var node = allPackages[packageName];
+    return orderedPackages.where((n) => n.dependencies.contains(node));
+  }
+
+  /// All of the packages in postorder by dependencies.
+  ///
+  /// Depedenencies of a package will come before the package in the result. If
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -117,6 +119,40 @@ class PackageGraph {
...
+  /// there is a package cycle the relative position of packages within the
+  /// cycle is non-deterministic, ecept that the root package will always come
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/test/package_graph/package_graph_test.dart
natebosch: Done

build_runner/test/package_graph/package_graph_test.dart
natebosch: This is a bug in my implementation. I want to guarantee that the root package is always last, but my implementation doesn't have that guarantee. The recursive solution does though so I kept this test, fixed the order, and updated the test name for clarity.

Also added a non-root cycle test

build_runner/lib/src/package_graph/package_graph.dart
natebosch: yeah, I can fix the bug with duplicates easily but it's not as clear to me how to fix the bug with the root package in a cycle. I switched back to the recursive version and we can hope that we never see deep enough package graphs where it's an issue.

I did keep `sync*` because IMO it's more clear than the closure and `yield*` works nicely with the recursive call.

PR #430, opened 2017-09-14
Use a single ResourceManager per build and improve parallelism

jakemac53:
These are both optimizations that came out of playing around with analyzer summary and dartdevc builders.

- We now only create one `ResourceManager` per build, and pass that in to `runBuilder`. This means that scratch spaces can be shared throughout a build.
- We now run all builders on all primary inputs in a phase at the same time, previously we blocked on each one so we had no parallelism even within a phase.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -72,7 +72,9 @@ class BuildImpl {
   Future<BuildResult> run(Map<AssetId, ChangeType> updates) async {
     var watch = new Stopwatch()..start();
     if (updates.isNotEmpty) await _updateAssetGraph(updates);
-    var result = await _safeBuild();
+    var resourceManager = new ResourceManager();
+    var result = await _safeBuild(resourceManager);
+    await resourceManager.disposeAll();
     if (result.status == BuildStatus.success) {
       _logger.info('Succeeded after ${watch.elapsedMilliseconds}ms with '
           '${result.outputs.length} outputs\n\n');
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -249,9 +254,15 @@ class BuildImpl {
       // Yield the outputs.
       for (var output in wrappedWriter.assetsWritten) {
         (_assetGraph.get(output) as GeneratedAssetNode).wasOutput = true;
-        yield output;
+        streamController.add(output);
       }
     }
+
+    Future.wait(primaryInputs.map(runForInput)).then((_) {
```
natebosch: [nit] should we change the return type to `Future<Iterable>` instead of using a `StreamController`?

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -222,12 +226,13 @@ class BuildImpl {
...
       var skipBuild = !builderOutputs.any((output) =>
           (_assetGraph.get(output) as GeneratedAssetNode).needsUpdate);
-      if (skipBuild) continue;
+      if (skipBuild) return;
       var wrappedReader = new SinglePhaseReader(
           _reader, _assetGraph, phaseNumber, _packageGraph.root.name);
       var wrappedWriter = new AssetWriterSpy(_writer);
       await runBuilder(
-          builder, [input], wrappedReader, wrappedWriter, _resolvers);
+          builder, [input], wrappedReader, wrappedWriter, _resolvers,
```
natebosch: `runBuilder` can handle multiple inputs - is there any way to structure this to use that?

build_runner/lib/src/generate/build_impl.dart
jakemac53: Sure it doesn't seem like we are taking advantage of the stream anyways, I can follow up

build_runner/lib/src/generate/build_impl.dart
jakemac53: I can try that out sure, will follow up

natebosch:
👍

Issue #343, opened 2017-08-08
Add a SummaryBuilder


natebosch:
In the future we might want to run the anlayzer in-process or as a worker. As a first pass maybe write inputs to a temp directory and run the analyzer from there.

matanlurey:
Do you imagine this (and #344) being part of `build_runner` or some other thing?

matanlurey:
On a related note, maybe we need `IpcBuilder` and/or `AsyncWorkerBuilder`. Up to you though :)

natebosch:
Both this and the DDCBuilder will probably start in the `build_runner` package but we could move them elsewhere if we find they have utility in general.

I don't expect we'll ever want to use them with bazel.

PR #431, opened 2017-09-14
Initial summary builder implementation

jakemac53:
There are a lot of performance issues we will need to work out and other details but this is a good start I think and worth submitting as-is (this package isn't published yet anyways).

Closes https://github.com/dart-lang/build/issues/343

### DIFF
--- a/build_compilers/lib/src/modules.dart
+++ b/build_compilers/lib/src/modules.dart
@@ -95,8 +95,7 @@ class Module {
       if (transitiveDeps.contains(next)) continue;
       transitiveDeps.add(next);
       var module = new Module.forLibrary(await resolver.libraryFor(next));
-      modulesToCrawl.addAll(module.directDependencies
-          .where((id) => !transitiveDeps.contains(id)));
+      modulesToCrawl.addAll(module.directDependencies);
     }
     return transitiveDeps;
   }
--- a/build_compilers/lib/src/modules.dart
+++ b/build_compilers/lib/src/modules.dart
@@ -2,8 +2,12 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
+
+import 'summary_builder.dart';
 
 /// A collection of Dart libraries in a strongly connected component and the
 /// modules they depend on.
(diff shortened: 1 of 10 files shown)


build_compilers/lib/src/modules.dart
```diff
@@ -68,6 +83,23 @@ class Module {
...
+    var transitiveDeps = new Set<AssetId>();
+    var modulesToCrawl = directDependencies.toSet();
+    while (modulesToCrawl.isNotEmpty) {
+      var next = modulesToCrawl.last;
+      modulesToCrawl.remove(next);
+      if (transitiveDeps.contains(next)) continue;
+      transitiveDeps.add(next);
+      var module = new Module.forLibrary(await resolver.libraryFor(next));
+      modulesToCrawl.addAll(module.directDependencies
+          .where((id) => !transitiveDeps.contains(id)));
```
grouma: Isn't the where clause handled by line 95?

build_compilers/lib/src/workers.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'dart:io';
+
+import 'package:bazel_worker/driver.dart';
+import 'package:build/build.dart';
+import 'package:cli_util/cli_util.dart' as cli_util;
+import 'package:path/path.dart' as p;
+
+String get _scriptExtension => Platform.isWindows ? '.bat' : '';
+
+const _defaultMaxWorkers = 4;
```
grouma: Maybe base this on a factor of the number of cores, which is readily available through the Platform API.

build_compilers/test/summary_builder_test.dart
```diff
@@ -0,0 +1,68 @@
...
+import 'matchers.dart';
+
+main() {
+  Map<String, dynamic> assets;
+
+  setUp(() {
+    assets = {
+      'b|lib/b.dart': '''final world = 'world';''',
+      'a|lib/a.dart': '''
+import 'package:b/b.dart';
```
grouma: Not a fan of the formatting here. Can you manually fix it up?

build_compilers/lib/src/workers.dart
jakemac53: done

build_compilers/test/summary_builder_test.dart
jakemac53: Indented things, usually I don't like to do that because it means the actual file contents have weird indentation but I guess it doesn't matter in this case.

Issue #436, opened 2017-09-19
Change `findAssets` to work for the current package for the buildStep, not the root package


jakemac53:
When we couldn't output assets in other packages this made sense (they were equivalent), now we need to change the docs and implementations to allow you to glob for whatever package the buildStep is running on (this is what is intuitive, and its all that we can implement with a Transformer).

jakemac53:
I think in order to implement this for `package:build_test` we will have to add a named arg `String package` to `findAssets`. Note that build_runner does already do this for `RunnerAssetReader`, so we might want to add this generally, but we don't really want it exposed on `BuildStep`.

Possibly we can add a `MultiPackageAssetReader` interface or something to `package:build` which both `package:build_runner` and `package:build_test` will implement with their asset readers.

jakemac53:
Working on a PR for this now and it seems like the `MultiPackageAssetReader` does help unify things a bit, so I am going to run with that. It looks like this:

```
/// The same as an `AssetReader`, except that `findAssets` takes an optional
/// argument `package` which allows you to glob any package.
///
/// This should not be exposed to end users generally, but can be used by
/// different build system implementations.
abstract class MultiPackageAssetReader extends AssetReader {
  /// Returns all readable assets matching [glob] under [package].
  ///
  /// Some implementations may require the [package] argument.
  @override
  Iterable<AssetId> findAssets(Glob glob, {String package});
}
```

PR #437, opened 2017-09-20
Add MultiPackageAssetReader to standardize support for globbing non-root packages

jakemac53:
Fixes https://github.com/dart-lang/build/issues/436.

This is unfortunately a breaking change for build_runner and build_test although a very minor one (its really only an internally breaking change but we do expose these apis).

I also snuck in a separate fix for build_runner where it would treat all `GeneratedAssetNode`s as inputs even if `wasOutput` was `false`.

cc @natebosch in case you want to take a look


build_test/CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 0.8.0-dev
+
+- `InMemoryAssetReader`, `MultiAssetReader`, `StubAssetReader` and
+  `PackageAssetReader` now implement the `MultiPackageAssetReader` interface.
+- `testBuilder` now supports `Builder`s that call `findAssets` in non-root
+  packages.
```
grouma: [nit] Indentation seems off here.

build_test/test/multi_asset_reader_test.dart
```diff
@@ -60,5 +60,26 @@ void main() {
       ]);
       expect(assetReader.findAssets(new Glob('lib/*.dart')), [idA, idB]);
     });
+
+    test('should support the `package` arg in `findAssets`', () {
```
grouma: Might not hurt to add a test for when no root package is defined nor is a package passed to the reader.

build_runner/lib/src/asset/reader.dart
```diff
@@ -61,10 +59,13 @@ class SinglePhaseReader implements AssetReader {
...
-      .where((n) => glob.matches(n.id.path))
-      .where((n) =>
-          n is! GeneratedAssetNode ||
-          (n as GeneratedAssetNode).phaseNumber < _phaseNumber)
-      .map((n) => n.id);
+          .where((n) => n.id.package == _primaryPackage)
+          .where((n) => glob.matches(n.id.path))
+          .where((n) {
+        if (n is GeneratedAssetNode) {
+          return n.wasOutput && n.phaseNumber < _phaseNumber;
```
grouma: Can you elaborate on the logic change here?

build_test/test/multi_asset_reader_test.dart
grouma: Looks like with the recent push this scenario has been covered.

build_runner/lib/src/asset/reader.dart
jakemac53: This is the other bug fix I mentioned, basically `GeneratedAssetNode`s can represent potential outputs that were never actually output (you can overdeclare outputs).

Previously calls to `findAssets` would give you these non-existent assets and then it would fail if you tried to read them. This fixes that by checking `wasOutput`.

build_test/CHANGELOG.md
jakemac53: it just looks weird with the quotes

build_runner/lib/src/asset/reader.dart
```diff
@@ -24,10 +22,10 @@ class SinglePhaseReader implements AssetReader {
   final Set<AssetId> _assetsRead = new Set();
   final AssetGraph _assetGraph;
   final int _phaseNumber;
-  final String _rootPackage;
+  final String _primaryPackage;
```
natebosch: [nit] I'm not sure this name is more clear. `PackageGraph` still calls this `root` right? FWIW I think 'root' is pretty clear since it's also the 'workspace root' when we build.

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -189,9 +189,13 @@ class BuildImpl {
   }
 
   Set<AssetId> _inputsForPhase(int phaseNumber) => _assetGraph.allNodes
-      .where((n) =>
-          n is! GeneratedAssetNode ||
-          (n as GeneratedAssetNode).phaseNumber < phaseNumber)
+      .where((n) {
+        if (n is GeneratedAssetNode) {
+          return n.wasOutput == true && n.phaseNumber < phaseNumber;
```
natebosch: are we worried about null values in `wasOutput`? We don't have this in other places... would prefer this without the `== true`

build_test/test/multi_asset_reader_test.dart
jakemac53: this class doesn't directly throw but I added a test that it propagates errors from the wrapped readers (and used this scenario as an example)

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -189,9 +189,13 @@ class BuildImpl {
...
+        } else {
+          return true;
```
natebosch: FWIW I think it's more readable in the old style rather than a boolean return inside a conditional. The repeated `as GeneratedAssetNode` does get noisy so I'm fine with this, but eventually we'll be able to drop the casts if the enhanced type promotion stuff goes through.

build_test/CHANGELOG.md
natebosch: Looks fine to me. I think the quote is just throwing it off visually.

build_runner/lib/src/asset/reader.dart
grouma: Ack. Thanks!

build_test/lib/src/multi_asset_reader.dart
```diff
@@ -9,12 +9,14 @@ import 'package:collection/collection.dart';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 
-/// An [AssetReader] that delegates to multiple other asset readers.
+/// An [MultiPackageAssetReader] that delegates to multiple other asset
```
natebosch: [nit] s/An/A/

PR #438, opened 2017-09-20
Manually create a ResourceManager and pass it to the Builder

nshahan:
Fix issue where using a `ScratchSpace` as a `Resource` in a `BuilderTransformer` would throw the error: "Tried to use a deleted ScratchSpace".

### DIFF
--- a/build_barback/lib/src/transformer/transformer.dart
+++ b/build_barback/lib/src/transformer/transformer.dart
@@ -48,10 +48,14 @@ class BuilderTransformer
         (id) => transform
             .readInputAsString(toBarbackAssetId(id))
             .catchError((_) => null));
-    return Future.wait(inputs.map((input) => _apply(input, transform)));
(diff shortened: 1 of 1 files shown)


build_barback/lib/src/transformer/transformer.dart
```diff
@@ -48,10 +48,14 @@ class BuilderTransformer
         (id) => transform
             .readInputAsString(toBarbackAssetId(id))
             .catchError((_) => null));
-    return Future.wait(inputs.map((input) => _apply(input, transform)));
+    var resourceManager = new ResourceManager();
+    return Future
+        .wait(inputs.map((input) => _apply(input, transform, resourceManager)))
+        .then((_) => resourceManager.disposeAll());
```
natebosch: Do we want to handle the failure case too or are we guaranteed there is no error here? `whenComplete` will fire regardless of whether it completes with a result or an error.

PR #439, opened 2017-09-21
fix a few scenarios where rebuilds were broken

jakemac53:
This fixes two bugs related to the cache directory:

- files in the cache dir that aren't in the serialized asset graph are no longer treated as inputs
- it will no longer try to build over-declared outputs (outputs with `wasOutput == false`) on startup for every build

### DIFF
--- a/build_runner/lib/src/util/constants.dart
+++ b/build_runner/lib/src/util/constants.dart
@@ -11,11 +11,18 @@ final String assetGraphPath = '$cacheDir/$scriptHash/asset_graph.json';
 /// Directories used for build tooling.
 ///
 /// Reading from these directories may cause non-hermetic builds.
-const toolDirs = const ['.dart_tool', 'build', 'packages', '.pub'];
+const toolDirs = const [
+  '.dart_tool',
+  'build',
+  'packages',
+  '.pub',
+  '.git',
+  '.idea'
+];
 
 /// The directory to which assets will be written when `writeToCache` is
 /// enabled.
-const generatedOutputDirectory = '.dart_tool/build/generated';
+const generatedOutputDirectory = '$cacheDir/generated';
 
 /// Relative path to the cache directory from the root package dir.
 const String cacheDir = '.dart_tool/build';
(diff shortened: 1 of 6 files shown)


build_runner/lib/src/util/constants.dart
```diff
@@ -11,11 +11,18 @@ final String assetGraphPath = '$cacheDir/$scriptHash/asset_graph.json';
 /// Directories used for build tooling.
 ///
 /// Reading from these directories may cause non-hermetic builds.
-const toolDirs = const ['.dart_tool', 'build', 'packages', '.pub'];
+const toolDirs = const [
+  '.dart_tool',
+  'build',
+  'packages',
+  '.pub',
+  '.git',
```
grouma: Should this list be configurable some how? Also I'm curious how reading from .idea and .git can cause non-hermetic builds.

build_runner/test/generate/build_definition_test.dart
```diff
@@ -0,0 +1,105 @@
...
+      var packageGraph = new PackageGraph.fromRoot(packageA);
+      options = new BuildOptions(
+          reader: assetReader,
+          writer: assetWriter,
+          packageGraph: packageGraph,
+          logLevel: Level.OFF,
+          writeToCache: true);
+    });
+
+    group('updates', () {
```
grouma: [nit] Using `updates` as the group name makes a lot of these tests read weird. Maybe use, `For build updates`?

build_runner/lib/src/util/constants.dart
jakemac53: Ya these are folders we ignore, we might want to make it configurable in the future or possibly ignore all top level hidden dirs? I can file an issue for that.

build_runner/test/generate/build_definition_test.dart
jakemac53: `updates` is a field on `BuildDefinition` and this is the group of tests that are testing that field

build_runner/lib/src/util/constants.dart
jakemac53: https://github.com/dart-lang/build/issues/440

PR #441, opened 2017-09-21
add a basic graph inspector tool to build_runner

jakemac53:
got a quick start on this - today its interactive but we might want to change that since you don't get much bash niceness but its definitely useful for inspecting graphs and provides a good extension point for future features.

Today this has two commands, `inspect` (for inspecting individual nodes), and `graph` (for inspecting the whole graph). We can rename these later too if we want.

### DIFF
--- a/build_runner/bin/graph_inspector.dart
+++ b/build_runner/bin/graph_inspector.dart
@@ -151,8 +151,8 @@ class GraphCommand extends Command {
 
   @override
   run() {
-    bool showGenerated = argResults['generated'];
-    bool showSources = argResults['original'];
+    var showGenerated = argResults['generated'] as bool;
+    var showSources = argResults['original'] as bool;
     Iterable<AssetNode> nodes;
     if (showGenerated) {
       nodes = assetGraph.allNodes.where((n) => n is GeneratedAssetNode);
@@ -162,12 +162,12 @@ class GraphCommand extends Command {
       nodes = assetGraph.allNodes;
     }
 
-    String package = argResults['package'];
+    var package = argResults['package'] as String;
     if (package != null) {
       nodes = nodes.where((node) => node.id.package == package);
     }
 
-    String pattern = argResults['pattern'];
+    var pattern = argResults['pattern'] as String;
     if (pattern != null) {
       var glob = new Glob(pattern);
       nodes = nodes.where((node) => glob.matches(node.id.path));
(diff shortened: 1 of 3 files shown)


build_runner/bin/graph_inspector.dart
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+import 'dart:async';
```
natebosch: [nit] I'm also a fan of a blank line under the copyright before the first import

build_runner/bin/graph_inspector.dart
```diff
@@ -0,0 +1,203 @@
...
+        abbr: 'p', help: 'Filters nodes to a certain package');
+    argParser.addOption('pattern',
+        abbr: 'm', help: 'glob pattern for path matching');
+  }
+
+  @override
+  run() {
+    var showGenerated = argResults['generated'] as bool;
+    var showSources = argResults['original'] as bool;
+    Iterable<AssetNode> nodes;
```
natebosch: [optional] Do you ever use anything from the node other than `id`? If not we could make this an `Iterable<AssetId>` and some of the lines below could be shorted to `assetGraph.sources` and `assetGraph.outputs`.

build_runner/bin/graph_inspector.dart
```diff
@@ -0,0 +1,203 @@
...
+      nodes = nodes.where((node) => glob.matches(node.id.path));
+    }
+
+    for (var node in nodes) {
+      _listAsset(node.id, stdout, indentation: '  ');
+    }
+  }
+}
+
+AssetId _getIdFromStringUri(String stringUri) {
```
natebosch: [nit] would prefer without `get`. `_idFromString` is clear to me.

build_runner/bin/graph_inspector.dart
jakemac53: done

build_runner/bin/graph_inspector.dart
jakemac53: done

build_runner/bin/graph_inspector.dart
jakemac53: done

PR #446, opened 2017-09-22
Add recording of read assets to InMemoryAssetReader

nshahan:
Gives ability to verify which assets have been read by a Builder.

### DIFF
--- a/build_test/lib/src/in_memory_reader.dart
+++ b/build_test/lib/src/in_memory_reader.dart
@@ -9,9 +9,9 @@ import 'package:glob/glob.dart';
 
 import 'in_memory_writer.dart';
 
-/// An [AssetReader] that records which assets have been read to [readAssets].
+/// An [AssetReader] that records which assets have been read to [assetsRead].
 abstract class RecordingAssetReader implements AssetReader {
-  Iterable<AssetId> get readAssets;
+  Iterable<AssetId> get assetsRead;
 }
 
 /// An implementation of [AssetReader] with primed in-memory assets.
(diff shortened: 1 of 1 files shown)


build_test/lib/src/in_memory_reader.dart
```diff
@@ -9,29 +9,39 @@ import 'package:glob/glob.dart';
 
 import 'in_memory_writer.dart';
 
+/// An [AssetReader] that records which assets have been read to [readAssets].
+abstract class RecordingAssetReader implements AssetReader {
+  Iterable<AssetId> get readAssets;
```
jakemac53: nit: `assetsRead`? `readAssets` sounds like an action :)

build_test/lib/src/in_memory_reader.dart
```diff
@@ -9,29 +9,39 @@ import 'package:glob/glob.dart';
...
   /// Create a new asset reader that contains [sourceAssets].
   ///
   /// May optionally define a [rootPackage], which is required for some APIs.
   InMemoryAssetReader({Map<AssetId, DatedValue> sourceAssets, this.rootPackage})
-      : assets = sourceAssets ?? <AssetId, DatedValue>{};
+      : assets = sourceAssets ?? <AssetId, DatedValue>{},
+        readAssets = new Set<AssetId>();
 
   @override
   Future<bool> canRead(AssetId id) async => assets.containsKey(id);
```
jakemac53: nit: we should probably add these assets as well, calling `canRead` on an asset adds a dependency on it just like `readAs*`

build_test/lib/src/in_memory_reader.dart
nshahan: Yeah English is tough. It's hard to distinguish read (present tense) vs read (past tense). I'll change it.

build_test/lib/src/in_memory_reader.dart
nshahan: Done.

nshahan: Just to clarify, does this mean that you would expect there to be a dependency on a file that returns false from canRead? In an offline conversation, I gathered that this is true.

PR #451, opened 2017-09-23
Add excludes: ... option to BuildAction.

### LINKED ISSUE #186: InputSet should support exclude patterns
The Glob library doesn't support extended glob syntax so there is no way to exclude file patterns natively. We could add an `exclude` optional argument that would allow us to define additional patterns to subtract from the matched files.

Currently InputSet is only a container for a list of globs so we might need to centralize the logic of finding matches including exclusion.

matanlurey:
Closes #186.

/cc @natebosch

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -204,8 +204,7 @@ class BuildImpl {
   /// Gets a list of all inputs matching [inputSet].
   Set<AssetId> _matchingInputs(InputSet inputSet, int phaseNumber) =>
       _inputsForPhase(phaseNumber)
-          .where((input) =>
-              input.package == inputSet.package && inputSet.matches(input.path))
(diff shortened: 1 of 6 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -205,8 +205,7 @@ class BuildImpl {
   Set<AssetId> _matchingInputs(InputSet inputSet, int phaseNumber) =>
       _inputsForPhase(phaseNumber)
           .where((input) =>
-              input.package == inputSet.package &&
-              inputSet.globs.any((g) => g.matches(input.path)))
+              input.package == inputSet.package && inputSet.matches(input.path))
```
natebosch: How about `inputSet.match(input)` and shift the package checking there as well?

matanlurey: Love it. PTAL @natebosch

Issue #456, opened 2017-09-26
Use a PackageWatcher abstraction over Watcher?


matanlurey:
I noticed when in the process of working on #424 that the usage of `Watcher` is quite complex; it needs to map arbitrary file events back to a potential asset changed. This makes tests such as those in `watch_test.dart` already a little opaque:

```dart
FakeWatcher.notifyWatchers(
  new WatchEvent(
    ChangeType.MODIFY, 
    path.absolute('a', 'packages', 'a', 'a.txt')
  )
);
```

What I'm proposing is a `build_runner` abstraction over `Watcher`:

```dart
abstract class PackageWatcher {
  factory PackageWatcher(String path) = _PackageWatcher;

  Stream<AssetId> watchDirectory(String directory);
}
```

This would be much easier to mock out, but also easier (potentially) in the future to support custom file systems, in-memory implementations, or whatever requirement others might have. WDUT?

/cc @jakemac53 @natebosch

jakemac53:
Sounds good to me

natebosch:
Sounds good

PR #458, opened 2017-09-26
Add initial implementation of `PackageWatcher`.

matanlurey:
Purposefully did not try to change any internals of `build_runner` to use it (yet).

Work towards #456.

/cc @natebosch

### DIFF
--- a/build_runner/lib/src/watcher/package_watcher.dart
+++ b/build_runner/lib/src/watcher/package_watcher.dart
@@ -1,0 +1,74 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:build/build.dart';
+import 'package:build_runner/build_runner.dart';
+import 'package:path/path.dart' as p;
+import 'package:watcher/watcher.dart';
+
+/// An event that occurs when an [asset] change occurred due to [type].
+class AssetChange {
+  /// Asset that changed.
+  final AssetId asset;
+
+  /// Type of change that occurred.
+  final ChangeType type;
+
+  const AssetChange(this.asset, this.type);
+
+  @override
+  bool operator ==(o) => o is AssetChange && asset == o.asset && type == o.type;
+
+  @override
+  int get hashCode => asset.hashCode ^ type.hashCode;
+
+  @override
+  String toString() => 'AssetChange {$asset, $type}';
+}
+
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/watcher/package_watcher.dart
```diff
@@ -0,0 +1,74 @@
...
+// Actual (default) implementation that delegates to DirectoryWatcher.
+class _PackageWatcher implements PackageWatcher {
+  static Watcher _defaultWatch(String path) => new DirectoryWatcher(path);
+
+  final _CreateWatcher _createWatcher;
+  final PackageGraph _packageGraph;
+
+  const _PackageWatcher(this._packageGraph,
+      {_CreateWatcher watch: _defaultWatch})
+      : _createWatcher = watch;
```
jakemac53: nit: don't use the default args cause they suck, just check for null in the assignment:

```dart
 const _PackageWatcher(this._packageGraph,
      {_CreateWatcher watch})
      : _createWatcher = watch ?? _defaultWatch;
```

build_runner/lib/src/watcher/package_watcher.dart
```diff
@@ -0,0 +1,74 @@
...
+typedef Watcher _CreateWatcher(String path);
+
+// Actual (default) implementation that delegates to DirectoryWatcher.
+class _PackageWatcher implements PackageWatcher {
+  static Watcher _defaultWatch(String path) => new DirectoryWatcher(path);
+
+  final _CreateWatcher _createWatcher;
+  final PackageGraph _packageGraph;
+
+  const _PackageWatcher(this._packageGraph,
```
jakemac53: Shouldn't this really take a `Package` not a `PackageGraph`? We need to be able to watch packages other than the root one.

matanlurey: In the next PR, I'm making the following change:

```dart
@protected
Stream watchPath(String path);

Stream watchGraph();
```

... which sort of begs this to be called `PackageGraphWatcher`. Does that make sense?

jakemac53: ah ok, so `watchPath` only watches the root, and `watchGraph` watches all packages?

matanlurey: Err, `watchNode(PackageNode node)`, is probably better than `watchPath`.

The idea is `watchGraph()` calls `watchNode(...)` for each tracked dependency/node.

jakemac53: Ah ok cool, sgtm :)

matanlurey:
Closing in favor of a better approach.

Issue #455, opened 2017-09-25
Globbing apis don't properly track dependencies


jakemac53:
Specifically for new files, if you do a glob for "**.dart" for instance you should get reran if any new dart files are added. Today that is not the case because we only track the files that the glob originally matched against.

We may actually not want to track the actual input files at all in this case and just track the globs separately? That should help trim the dependency graph and still gives us all the info we need.

matanlurey:
Interesting enough this is also a bug/lack of feature in `iblaze`/`bazel_watcher`:

```build
dart_library(
  srcs = glob(["lib/**.dart"]),
)
```

... If you**add** `lib/x.dart`, it will not trigger a re-build, as it apparently was expensive to do so at bazel-like level; that being said, pub packages tend to be easier to reason with and maybe we could watch for _all_ changes, and then check if the change matches any glob.

jakemac53:
Ya, luckily we only allow you to glob within the same package so that narrows down the number of actions we have to check against when a new file is added, but it still has the potential to be somewhat expensive.

This is only for new files though, and only matters when you have an existing asset graph so it won't slow down the initial build, and it should be pretty rare that you see a ton of new files otherwise. Potentially changing to a new branch or something where you added a thousand files might suck a bit, but there isn't really any way around that anyways.

PR #460, opened 2017-09-26
Fix dependency tracking for glob apis, and add some more tests

jakemac53:
Fixes https://github.com/dart-lang/build/issues/455

This also doesn't cause rebuilds on file changes, only on adds/deletes (unless you read the files explicitly after globbing). We might want to do something similar for `canRead` at some point as well but that is a separate issue :).

cc @natebosch

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -256,7 +256,9 @@ class BuildImpl {
 
       // Yield the outputs.
       for (var output in wrappedWriter.assetsWritten) {
-        (_assetGraph.get(output) as GeneratedAssetNode).wasOutput = true;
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -256,7 +256,9 @@ class BuildImpl {
 
       // Yield the outputs.
       for (var output in wrappedWriter.assetsWritten) {
-        (_assetGraph.get(output) as GeneratedAssetNode).wasOutput = true;
+        var node = _assetGraph.get(output) as GeneratedAssetNode;
+        node.wasOutput = true;
+        node.globs = wrappedReader.globsRan.toSet();
```
jakemac53: actually reading over this I will need to do this even if it _wasn't_ output so I will follow up with another test and fix

Issue #320, opened 2017-08-01
Checking for existing outputs fails if an intermediate output is deleted


natebosch:
Situation:
- There is a source file `source.dart` and two phases of builders `.dart` -> `.phase1` and `.phase1` -> `.phase2`
- The `.dart_tool` directory does not exist so there is no serialized asset graph
- `source.phase1` does *not* exist on disk, `source.phase2` *does* exist on disk.

Checking for existing outputs will not see a conflict with `source.phase2` because we only feed inputs that exist on disk [here](https://github.com/dart-lang/build/blob/465060fa2b9ab0f576c90c8385b49922e61c5fa2/build_runner/lib/src/generate/build_impl.dart#L318) - not outputs that we expect will exist.

If we run a build we will write `source.phase1` and then get an exception "Cannot overwrite inputs." when trying to write `sources.phase2`.

jakemac53:
Ah ya, that makes sense. Creating the whole asset graph up front should make this easier.

natebosch:
We do attempt this - we just are reading from the wrong state when we grab inputs during the check.

PR #462, opened 2017-09-27
add a regression test for checking existing outputs with missing intermediate output

jakemac53:
closes https://github.com/dart-lang/build/issues/320


jakemac53:
(it looks like this was already fixed by other changes)

Issue #421, opened 2017-09-13
Support per-package or per-entrypoint builders


natebosch:
Capturing a discussion we had:

We need some way to get at least the SDK js file into the root, or into each top-level directory that has an entrypoint like `web/` and `test/`. Another use case is a per-package `module_config` that can do more course-grained module chunking than we get with #392

Some possibilities we discussed:
- Add a new type of `Builder` which get a list of assets in the package to a `declareOutputs` type method.
  - This is tricky to support in bazel, though it is possible if we allow codegen rules to implement custom skylark to mirror the implementation in Dart.
  - We would not be able to be fully config driven like we are today - but maybe that is OK for these exceptions?
  - Possibly only export this from `build_runner` instead of `build`? Our `build_compilers` package could have a dependency on `build_runner` and that would be the only build system to support this.
- Add a new type of `Builder` which only get's the asset IDs of entry points. We need to find entry points anyway to get the lazy version of a build command working (serve is easy lazily). Has the same caveats as above, would not help with per-package module config.
- Add a special `$SDK` package asset ID that can read the js file from the SDK. Would work for the `serve` case, would not help with module config.

PR #465, opened 2017-09-27
Add a PackageBuilder as a normal builder with no primary input

jakemac53:
A `PackageBuilder` also has an additional required getter `String get package`, and an additional method `Iterable<AssetId> declareOutputs()`.

For now I did not allow you even a list of files in `declareOutputs` (since ddc doesn't need them), so this is really just for well known outputs right now (such as `.moduleConfig`). In general I think we would only want to give you a restricted `AssetReader` which has the `canRead` an `findAssets` apis but nothing more, as opposed to giving you the full list always. That way we don't cause unnecessary invalidation for `PackageBuilders` that don't care about the available inputs.

I haven't added any tests yet (although I have been playing around with module computing using this so I know it does work). I will follow up with tests shortly though just wanted to get this out there for initial feedback.

In general there are some awkward things due to certain fields now possibly being null (`buildStep.input` for instance), but I think it is probably better than having the `PackageBuilder` and `Builder` classes have to implement some other base class.... maybe not though idk. If we did move this into the core `build` package we could consider doing that.

fixes https://github.com/dart-lang/build/issues/421


natebosch:
Should the outputs be a better instead of a function? Should it be an iterate of String and contain only paths since it would not be allowed to output an asset in another package?

jakemac53:
> Should the outputs be a getter instead of a function?

As it stands now sure, that is probably better, and we could rename it `outputs`? I think we will want to add in some sort of restricted asset reader as an argument though at which point it would have to change back to a function, but that is fine.

>  Should it be an iterate of String and contain only paths since it would not be allowed to output an asset in another package?

That does make some sense, although it feels a bit inconsistent just because all the other apis use AssetId? It does match the actual desired semantics better I think though so it is probably worth doing.

jakemac53:
I am also scrapping the `package` field on the `PackageBuilder` class, it isn't necessary since that is already captured by the `BuildAction`.

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -222,53 +233,103 @@ class BuildImpl {
...
+    var builderOutputs = outputIdsForBuilder(builder, package);
+
+    if (!_buildShouldRun(builderOutputs)) return <AssetId>[];
+
+    var wrappedReader =
+        new SinglePhaseReader(_reader, _assetGraph, phaseNumber, package);
+    var wrappedWriter = new AssetWriterSpy(_writer);
+
+    var logger = new Logger('runBuilder');
+    var buildStep = new BuildStepImpl(null, builderOutputs, wrappedReader,
```
natebosch: we'll have to be more careful about what we consider breaking changes around this.

build_runner/lib/src/package_builder/package_builder.dart
```diff
@@ -0,0 +1,15 @@
+import 'package:build/build.dart';
+
+abstract class PackageBuilder extends Builder {
```
natebosch: If only we had type unions we wouldn't need the `extends` here...

I wonder if we'd be better off with a hierarchy like:

- `BuildAction`
- `AssetBuildAction extends BuildAction` - basically the implementation we have now and the default ctor for `BuildAction` is a factory that gives this.
- `PackageBuildAction extends BuildAction` - a new implementation that has a `PackageBuilder` instead of a `Builder`.  Then we shift the `is` check and have a more distinct separation between what can or can't be run with `package:build`  code.

Thoughts?

build_runner/lib/src/generate/build_impl.dart
jakemac53: Ya, I had to add two imports to `src` of package:build :(.

build_runner/lib/src/package_builder/package_builder.dart
jakemac53: I think that makes sense, I can work on updating it tomorrow. We can also remove the `inputs` field from the base class so its only on `AssetBuildAction`...which would be a little bit awkward since the `BuildAction` default constructor would be taking an arg for a field that doesn't exist on the interface, but I think it would be fine.

natebosch: Yeah I think I've seen cases where the default constructor instantiates a subclass and passes through arguments the parent doesn't actually use. In practice I don't think it feels very awkward, and we can make a named constructor like `new BuildAction.package()` for the other subclass.

jakemac53: Ok, done.

I did not add a redirecting factory constructor on `BuildAction` for `PackageBuildAction`, but I could if you want. I don't think that is any cleaner than just having people go through the regular `PackageBuildAction` constructor though.

jakemac53:
Added some tests as well now, PTAL

build_runner/lib/src/package_builder/package_builder.dart
natebosch: This still has the `extends` did you find out that it isn't possible to remove that?

I was hoping for `BuildAction` to have no `builder` field, for `AssetBuildAction` to have a `builder` field of type `Builder`, and for `PackageBuildAction` to have a `builder` (or another name) field of type `PackageBuilder`, so that once we do the `is` check on the action type we get a concrete type for the builder we use.

jakemac53: oh - I didn't make it so that PackageBuilder doesn't extend builder but I did make `PackageBuildAction` have a `PackageBuilder`, so you don't need any casts after checking for the type of the `BuildAction`.

I can look at removing the `Builder get builder` getter from `BuildAction` as well and making `PackageBuilder` not extend builder but think that will be more involved.

jakemac53: Ok updated, this turned out to be an easy change.

You still have a normal `BuildStep` in `PackageBuilder.build`, but that would be a larger change to make a `PackageBuildStep`. It might be worth making the change though as I could remove the imports from `src` if we had our own `PackageBuildStepImpl`? But its also a lot of extra code just to clean up that little bit of weirdness :/

natebosch: I think it's fine to have the `BuildStep` argument and src import for now. Since we should have low usage it should be relatively easy to roll a breaking change away from `BuildStep` if we decide we need it. I'm in favor of  moving forward in the mean time.

PR #469, opened 2017-09-29
Add a basic BuildPerformanceTracker

jakemac53:
This tracks the time it took to run each `BuildAction` (phase effectively), and total time for a build.

You can pretty easily use this to roll up the total time per `Builder` type etc, and since I added it to the `BuildResult` you can easily slice and dice things however you want in your builder script after each build.

### DIFF
--- a/build_runner/test/generate/performance_tracker_test.dart
+++ b/build_runner/test/generate/performance_tracker_test.dart
@@ -1,0 +1,66 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:test/test.dart';
+
+import 'package:build/build.dart';
+import 'package:build_runner/build_runner.dart';
+import 'package:build_test/build_test.dart';
+
+main() {
+  group('PerformanceTracker', () {
+    BuildPerformanceTracker tracker;
+
+    setUp(() {
+      tracker = new BuildPerformanceTracker()..start();
+    });
+
+    test('can track start/stop times and total duration', () {
+      tracker.stop();
+      expect(tracker.startTime.microsecondsSinceEpoch,
+          lessThan(new DateTime.now().microsecondsSinceEpoch));
+      expect(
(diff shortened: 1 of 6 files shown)


natebosch:
Haven't looked at the code yet, but a quick note on the name... Can we omit 'Build' since that context should be clear and the name is pretty long?

natebosch:
Also as the client sees it, it shouldn't be a 'Tracker' but the already gathered results. We shouldn't exist any apis that we use to do the tracking

jakemac53:
I can remove `Build`, and make this an internal class, and then create a `PerformanceView` class which only exposes the performance apis and not the actual tracking apis (and just wraps a `Tracker`)?

jakemac53:
Or actually I will just make a base `PerformanceView` interface with the performance information apis, and implement that in the `Tracker`, and just make the public return types be that interface type.

jakemac53:
Actually now that I think about it a bit further I do want to leave `Build` in the name, because there is an overall build performance class, and then an individual action performance class.

I am thinking the new names will be `BuildPerformance` and `BuildActionPerformance`?

natebosch:
Name suggestions and hiding internal apis with an implementing class sound good to me.

build_runner/test/generate/performance_tracker_test.dart
```diff
@@ -0,0 +1,68 @@
...
+      var actionsTotal = new Duration(microseconds: 0);
+      for (var trackedAction in tracker.actions) {
+        actionsTotal += trackedAction.duration;
+        expect(trackedAction.startTime.microsecondsSinceEpoch,
+            greaterThan(last.microsecondsSinceEpoch));
+        expect(trackedAction.stopTime.microsecondsSinceEpoch,
+            greaterThan(trackedAction.startTime.microsecondsSinceEpoch));
+        last = trackedAction.stopTime;
+      }
+      expect(tracker.stopTime.microsecondsSinceEpoch,
```
grouma: Not the biggest fan of how we are validating the logic here. If we intend to expand on the TimeTracker interface we likely should introduce the capability to test with a FakeClock.

jakemac53: Filed https://github.com/dart-lang/build/issues/471

PR #470, opened 2017-09-29
Fix Resource bug with multiple ResourceManagers

jakemac53:
Previously multiple `ResourceManager`s would share `Resource` instances if running at the same time.


jakemac53:
ok, published the `0.10.2+1` release as well now

Issue #459, opened 2017-09-26
Add PackageNode.synthetic (name?)


matanlurey:
```dart
class PackageNode {
  PackageNode.synthetic(String name, Uri location)
    : this(name, null, null, location);
}
```

This allows removing lots of testing code that relies on passing `null`. We can optionally annotate this with `@visibleForTesting` if you're afraid users might think about using it in production code.

jakemac53:
I am ok w/this, or optionally we could create a helper method under test/common/

PR #473, opened 2017-09-30
Add a `noPubspec` constructor for `PackageNode`

matanlurey:
Closes https://github.com/dart-lang/build/issues/459.

### DIFF
--- a/build_runner/lib/src/package_graph/package_graph.dart
+++ b/build_runner/lib/src/package_graph/package_graph.dart
@@ -186,7 +186,7 @@ class PackageNode {
   /// Create a [PackageNode] without any details from a `pubspec.yaml` file.
   ///
   /// This is useful for testing, or in cases where a package may be synthetic.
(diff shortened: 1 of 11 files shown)


build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -179,6 +183,13 @@ class PackageNode {
 
   PackageNode(this.name, this.version, this.dependencyType, this.path);
 
+  /// Create a [PackageNode] without any details from a `pubspec.yaml` file.
+  ///
+  /// This is useful for testing, or in cases where a package may be synthetic.
+  PackageNode.noPubspec(this.name, this.path)
```
jakemac53: Should we make path be a named optional arg as well?  It looks like we still pass null a fair bit.

matanlurey: Sg. Done.

Issue #467, opened 2017-09-28
Infer root for `BuildAction`?


matanlurey:
A typical user bug is copy-pasting a `build.dart` from another package, only to hit an error:

> I am finding source_gen for the first time when attempting to use json_serializable.
> I figured out that I need to create a "build script" and started by copying the one from here
 (https://github.com/dart-lang/json_serializable/blob/master/json_serializable/tool/build_actions.dart). However, when I run dart tool/build.dart I get a the error: `A build action is attempting to operate on a package which is not the root`

I also never really understood `BuildAction` needing a package - I imagine if we want Bazel compatibility ultimately then building _another_ package is not exactly allowed anyway. I realize this might be obsolete (or uninteresting) once we have https://github.com/dart-lang/build/issues/353, but worth filing I think.

At the very least, maybe we should have the error help you out a bit:

```
A build action is attempting to operate on package "foo", but the build script is located in package "bar". It's not valid to attempt to generate files for another package, did you mean to write:

  new BuildAction(..., 'bar')

... instead?
```

jakemac53:
Now with `writeToCache` we do support running builders on other packages. We could consider packing `package` an optional argument though and defaulting it to the root package?

matanlurey:
I'm going to just improve the error message for now!

PR #474, opened 2017-09-30
Fancy error message for a bad root package

matanlurey:
Closes https://github.com/dart-lang/build/issues/467.

Dependent on https://github.com/dart-lang/build/pull/473.


build_runner/lib/src/generate/build_definition.dart
```diff
@@ -60,7 +60,12 @@ class _Loader {
     if (!_options.writeToCache &&
         _buildActions.any((action) =>
```
jakemac53: I think you can just remove this check now right? (the same check effectively happens inside this if now)

build_runner/lib/src/generate/exceptions.dart
```diff
@@ -25,10 +28,16 @@ class UnexpectedExistingOutputsException extends FatalBuildException {
 class InvalidBuildActionException extends FatalBuildException {
   final String _reason;
 
-  const InvalidBuildActionException.nonRootPackage()
+  InvalidBuildActionException.nonRootPackage(BuildAction action, String root)
       : _reason =
-            'A build action is attempting to operate on a package which is not '
-            'the root.';
+            'A build action (${action.builder}) is attempting to operate on '
```
jakemac53: Should we mention the `writeToCache` option as well?

build_runner/lib/src/generate/build_definition.dart
matanlurey: Done.

PR #480, opened 2017-10-03
update graph inspector to output the needsUpdate field

jakemac53:


build_runner/bin/graph_inspector.dart
```diff
@@ -100,6 +100,7 @@ class InspectNodeCommand extends Command {
         ..writeln('  type: ${node.runtimeType}');
 
       if (node is GeneratedAssetNode) {
+        description.writeln('  needsUpdate: ${node.needsUpdate}');
```
kevmoo: Is this just logging inf?

jakemac53: ya, this is just outputting an additional field when you query for a node

Issue #477, opened 2017-10-03
Path dependencies don't capture watch events correctly


jakemac53:
Probably related to the recent watcher changes, you get `AssetChange` events that look like `{asset: angular_components|../angular2_components/lib/src/components/material_dialog/material_dialog.dart, type: modify}`. Those obviously don't trigger a rebuild :).

cc @matanlurey

PR #481, opened 2017-10-03
fix asset change detection with relative paths, and scopeLog time reporting

jakemac53:
fixes https://github.com/dart-lang/build/issues/477

~~still need to add a regression test though and figure out the best way to do that.~~

I also got a bit overzealous and replaced all import prefixes of package:path to be `p` instead of `path` because `path` conflicts way to often with variable names.

### DIFF
--- a/build_runner/lib/src/package_graph/package_graph.dart
+++ b/build_runner/lib/src/package_graph/package_graph.dart
@@ -4,7 +4,7 @@
 
 import 'dart:io';
 
-import 'package:path/path.dart' as path;
+import 'package:path/path.dart' as p;
 import 'package:yaml/yaml.dart';
 
 /// A graph of the package dependencies for an application.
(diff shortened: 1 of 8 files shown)


jakemac53:
Ok, I updated the test to create an actual tmp dir.

It now tests with the original path for the node being provided as both absolute and relative, which covers the actual cases we care about as opposed to trying to spoof relative change events.

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -199,6 +201,12 @@ class PackageNode {
     dependencies: [${dependencies.map((d) => d.name).join(', ')}]''';
 }
 
+/// Converts [path] to an absolute path, returns `null` if given `null`.
+String _toAbsolute(String path) {
```
matanlurey: Silly nit, but make this `static`?

jakemac53: done

build_runner/lib/src/watcher/node_watcher.dart
```diff
@@ -18,6 +18,10 @@ class PackageNodeWatcher {
   final _WatcherStrategy _strategy;
   final PackageNode _node;
 
+  /// The actual watcher instance.
```
natebosch: [nit] what information is this adding?

build_runner/lib/src/watcher/node_watcher.dart
```diff
@@ -40,9 +44,12 @@ class PackageNodeWatcher {
   /// packageA.watch()
   /// ```
   Stream<AssetChange> watch([String relative]) {
```
natebosch: Given that this is now stateful and has new restrictions on how it can be called I wonder if we should refactor a bit. I don't have any concrete suggestions so feel free to to defer and I can take a deeper look later.

Issue #95, opened 2016-03-16
make BuildOptions immutable


kevmoo:
It just...hurts

jakemac53:
I was thinking about exposing this at the top level as well, and splitting it into `BuildOptions`, `WatchOptions`, and `ServeOptions`. Then we could supply a `fromArgs` constructor as well to support some standard command line args.

PR #483, opened 2017-10-05
Add (unused) {Build|Watch}Options.

matanlurey:
Progress towards https://github.com/dart-lang/build/issues/95.

The current implementation (as used) is not quite stateless. I'm not sure if this one should be.

Some open questions:

1. Should log listening/printing be handled through another abstraction/mechanism?
2. Should `BuildOptions` expose creating asset readers/writers, or a single instance?
3. Should the defaults be documented?

### DIFF
--- a/build_runner/lib/src/generate/build_options.dart
+++ b/build_runner/lib/src/generate/build_options.dart
@@ -1,0 +1,86 @@
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
+
+import '../asset/file_based.dart';
+import '../asset/reader.dart';
+import '../asset/writer.dart';
+import '../package_graph/package_graph.dart';
+import 'directory_watcher_factory.dart';
+
+/// Configuration for a given invocation of a `build`.
+class BuildOptions {
+  /// Whether to delete previously generated files before running a build.
+  final bool deleteFilesByDefault;
+
+  /// Whether to write to a cache directory instead of the source directory.
+  ///
+  /// Enabling this option is the only way to allow builders to run against
+  /// packages other than the root.
+  final bool writeToCache;
+
+  /// Dependencies to execute a build against.
+  final PackageGraph packageGraph;
+
+  /// Manages reading files from the file system.
+  final RunnerAssetReader reader;
+
+  /// Manages writing files to the file system.
+  final RunnerAssetWriter writer;
+
+  factory BuildOptions({
+    bool writeToCache,
+    bool deleteFilesByDefault,
+    PackageGraph packageGraph,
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/generate/build_options.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:logging/logging.dart';
+import 'package:meta/meta.dart';
```
jakemac53: This needs to be added to the pubspec.yaml as well, we haven't used it so far

build_runner/lib/src/generate/build_options.dart
```diff
@@ -0,0 +1,86 @@
+import 'package:logging/logging.dart';
```
jakemac53: copyright

build_runner/lib/src/generate/build_options.dart
```diff
@@ -0,0 +1,86 @@
...
+  /// Whether to write to a cache directory instead of the source directory.
+  ///
+  /// Enabling this option is the only way to allow builders to run against
+  /// packages other than the root.
+  final bool writeToCache;
+
+  /// Dependencies to execute a build against.
+  final PackageGraph packageGraph;
+
+  /// Manages reading files from the file system.
```
jakemac53: nit: remove `from the file system` (same for writer)

jakemac53:
> 1. Should log listening/printing be handled through another abstraction/mechanism?

I don't have a strong opinion either way, the previous mechanism does have the flaw of not letting you allow the default behavior _and_ your own handling on top, you can only override the default completely.

> 2. Should BuildOptions expose creating asset readers/writers, or a single instance?

I think just passing in the instance directly was fine before, its easy enough to just create a `PackageGraph` on your own. At the same time I think the factory approach is also fine. These overrides are pretty much exclusively used by internal tests.

> 3. Should the defaults be documented?

That would probably be good :)

natebosch:
Do we expect these classes to be public? If so I think we should consider hiding some of the fields.

Some things like `directoryWatcherFactory` I think only exist to enable testing. Even probably the reader and writer don't really need to be configurable. If we refactor the top level methods like `build()` into a `build()` and `buildInternal()` we could limit the things that should only be configurable for tests.

jakemac53:
The original idea was this would allow you to run in other environments such as the browser, bazel, as well as for testing. Testing is the only one that has actually materialized though, and the reality is we still import and rely on dart:io in a few places internally.

I am in favor of just reducing the public api surface.

matanlurey:
Closing for now, since this was just for some input, and not a priority.

Issue #432, opened 2017-09-15
Proposal: add `bool get optional` field to Builder interface (defaults to false)


jakemac53:
Some builders want to run lazily and only if their output is needed by some other builder or requested directly from a server. This should be able to extend to full builds as well so the outputs will never be generated if not needed by any other part of the build.

One of the primary use cases initially will be for analyzer summaries and ddc, we only want to summarize and/or compile any modules that are actually required by an actual web application. Today we compile for the entire world and it does a lot of extra work.

Note that this is different from `lazy` transformers, which always run if you do `pub build`, which is why I called the field `optional` instead. They are lazy, but they also won't ever be built in any context unless required by some other build action.

I think these actions would still live in a specific phase, so their inputs would be limited to what was available in that phase, they would just delay their execution until later.

cc @natebosch @matanlurey

jakemac53:
@matanlurey I think it might make sense for angular to do run in this mode as well? You could potentially only compile templates for the components that are actually used in the app for instance.

natebosch:
Should this be done with build.yaml and in the BuidAction rather than the builder? This wouldn't have meaning in bazel would it?

jakemac53:
True that it wouldn't be optimizable in bazel, but that should be fine? We don't have to make guarantees around it, just a best effort based on implementation.

I think it would be a lot simpler to have the config be based on a field in your builder implementation than requiring a build.yaml.

matanlurey:
Maybe it would help explaining when this optimization could be used.

From an AngularDart perspective, I don't understand it.

@jakemac53:

> @matanlurey I think it might make sense for angular to do run in this mode as well? You could potentially only compile templates for the components that are actually used in the app for instance.

We only know what components are "used" by one of four signals:

* An import to the file containing them (not always true)
* Use of the component in a `directives: const [ ... ]` list (still not always true)
* Use of the component's `Type` somewhere in the application (still not always true)
* Use of the component's `{Type}NgFactory` (still not always true)

I'm not sure how I'd optimize given these constraints.

jakemac53:
@matanlurey I was thinking essentially you could have a bootstrap builder which runs in the last phase and basically just crawls the import graph asking for a corresponding `.template.dart` file for every import it sees? Maybe I am missing something but I would think that could work. Even just resolving the entrypoint library should actually cause this to happen automatically (assuming it has an import to the .template.dart for your root component).

natebosch:
[...]

PR #487, opened 2017-10-05
Add `isOptional` to `BuildAction`

jakemac53:
Any optional action will be ran lazily when any of its outputs is requested by another action, or if one of its outputs matches one of the primary input extensions of another action.

If no outputs of an optional action are ever requested by another action, then they _will not be built at all_. This is good for things like ddc/summaries/etc which we only want to build for the reachable dart files from any entry point.

Also note that using this setting for any builder that uses a `Resolver` is extremely likely to deadlock because of the current resolver implementation, so you need to be careful how you use it today.

Fixes https://github.com/dart-lang/build/issues/432.


jakemac53:
I will add a changelog entry tomorrow as well

build_runner/lib/src/asset/reader.dart
```diff
@@ -24,9 +25,11 @@ class SinglePhaseReader implements AssetReader {
   final _globsRan = new Set<Glob>();
   final int _phaseNumber;
   final String _primaryPackage;
+  final BuildImpl _buildImpl;
```
natebosch: this dependency cycle makes me nervous.

I've always imagined the communication happening across the `AssetGraph` rather than the `BuildImpl`

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -164,10 +164,9 @@ class AssetGraph {
       Set<AssetId> newSources, String rootPackage) {
     newSources.map((s) => new AssetNode(s)).forEach(_add);
     var allInputs = new Set<AssetId>.from(newSources);
-    var phaseNumber = 0;
-    for (var action in buildActions) {
-      phaseNumber++;
+    for (var phase = 0; phase < buildActions.length; phase++) {
```
natebosch: could we break some of these changes out into separate commits and get them in early?

build_runner/lib/src/asset/reader.dart
jakemac53: `BuildImpl` handles all the actual building, which is what we need to do (whenever somebody tries to read an output of an optional builder). Otherwise `AssetGraph` would probably have to have a `BuildImpl` which creates a similar (if not worse) situation imo.

We could instead pass in a function which lazily builds things (in this case we would just pass `this.runLazyPhaseForInput`), but that doesn't really solve the circular dependency issue... although it does restrict usage of the `BuildImpl` to that one function.

jakemac53:
ok merged the other changes around the phase numbers and serializing `needsUpdate` as part of the graph

build_runner/lib/src/asset/reader.dart
natebosch: I think that would be an improvement since it makes the communication pattern more explicit.

I think it would be nice to shift further towards an AssetGraph or related class for this type of thing but we could also revisit that later.

jakemac53: Ok, done. I also changed `runLazyPhaseForInput` to private since it is no longer directly referenced by external code which is a nice side effect of this change.

build_runner/lib/src/generate/phase.dart
```diff
@@ -7,46 +7,59 @@ import 'package:build/build.dart';
...
 
-/// A "phase" in the build graph, which represents running a [builder] on a
+/// A "phase" in the build graph, which represents running a builder on a
 /// specific [package].
+///
+/// See the [BuildAction] and [PackageBuildAction] implementations.
 abstract class BuildAction {
   String get package;
 
+  /// If `true` then this action runs lazily when another action tries to read
```
natebosch: [nit] "Whether to run lazily when another action..."

Also a blank line following the first sentence. Maybe:

"Whether to run lazily when an output is read.

An optional builder will only run if one of it's outputs is read by a later Builder, or is used as a primary input to a Builder. If no build actions read the output of an optional builder it will never run."

build_runner/lib/src/asset/reader.dart
```diff
@@ -6,6 +6,9 @@ import 'package:glob/glob.dart';
 import '../asset_graph/graph.dart';
 import '../asset_graph/node.dart';
 
+typedef Future RunPhaseForInput(
```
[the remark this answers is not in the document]
jakemac53: ya, I added a `show` to the export in the entrypoint lib :)

build_runner/lib/src/generate/phase.dart
jakemac53: done

build_runner/lib/src/asset/reader.dart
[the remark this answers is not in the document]
natebosch: oops, cmd+f fail on my part.

PR #489, opened 2017-10-06
Serialize module config

jakemac53:
Uses `JsonSerializable`, so this is a decent e2e test of our own products which is good :). Already resulted in a few bugs being filed.

cc @kevmoo

### DIFF
--- a/build_compilers/test/build_test.dart
+++ b/build_compilers/test/build_test.dart
@@ -7,6 +7,7 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
 
 // TODO: Generalize this test and move it to package:build_test. This is copied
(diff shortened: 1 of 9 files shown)


build_compilers/test/build_test.dart
```diff
@@ -0,0 +1,66 @@
...
+  String pkgRoot;
+  try {
+    pkgRoot = _runProc('git', ['rev-parse', '--show-toplevel']);
+    var currentDir = Directory.current.resolveSymbolicLinksSync();
+    if (!p.isWithin(pkgRoot, currentDir)) {
+      throw new StateError('Expected the git root ($pkgRoot) '
+          'to be a parent of the current directory ($currentDir).');
+    }
+  } catch (e) {
+    print("Skipping this test – git didn't run correctly");
```
natebosch: [nit] don't mix quote types.

jakemac53: This is just because its using a `'` inside the string, which is allowed

kevmoo: There's a lint for that 😄

natebosch: oops, missed the `'`. Carry on, nothing to see here.

PR #491, opened 2017-10-06
make a ModuleBuilder which serializes out module config files

jakemac53:
This is now also consumed by the summary and dartdevc builders.

### DIFF
--- a/build_compilers/test/util.dart
+++ b/build_compilers/test/util.dart
@@ -1,3 +1,7 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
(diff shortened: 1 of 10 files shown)


build_compilers/test/util.dart
```diff
@@ -0,0 +1,20 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:build_test/build_test.dart';
+
+/// Forwards to [testBuilder], and adds all output assets to [assets].
+Future<Null> testBuilderAndAddAssets(
```
natebosch: [nit] I think `collect` tells us more than `add`. How about `testBuilderCollectAssets`.

jakemac53: done

PR #499, opened 2017-10-10
speed up matchInputs by lazily building inputs simultaneously

jakemac53:


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -221,18 +221,18 @@ class BuildImpl {
   Future<Set<AssetId>> _matchingInputs(InputSet inputSet, int phaseNumber,
```
natebosch: I wonder if we need a different name for this method...

jakemac53: It does return a set of matching inputs still, but its true the name doesn't indicate that it is doing a lot of work potentially.

jakemac53:
Renamed the description

Issue #498, opened 2017-10-10
Add ddc bootstrap builder


jakemac53:
We compile ddc modules today, but we don't have the bootstrapping code in yet.

For SDK resources we need to figure out how we want to reference them, these include the following assets under the sdk:

- lib/dev_compiler/amd/dart_sdk.js
- lib/dev_compiler/amd/require.js
- lib/dev_compiler/web/dart_stack_trace_mapper.js
- lib/dev_compiler/web/ddc_web_compiler.js

We can either create a special `$sdk` package (or similar), or possibly copy these into the `build_compilers` package using a `PackageBuilder` and reference them from there. The main issue with the build_compilers strategy is we don't have a way of tracking changes to those files, which might be a deal breaker.

jakemac53:
https://github.com/dart-lang/build/pull/501 adds the `$sdk` package

Issue #502, opened 2017-10-12
Add some integration tests for package:build_compilers


jakemac53:
The easiest way to do this is probably just to create a fake package with some actual tests, and run those using the build_runner server and the --pub-serve argument to pub run test. Alternatively we could try just using the build command from build_runner and running with --precompiled.

This is currently blocked on being able to run ddc tests at all, which is blocked on having a builder that can bootstrap dart entrypoints for testing, and probably also having the ability to serve a subdirectory (so that test/ isn't in the uri) to match pub serve.

jakemac53:
closing this now, we have some reasonable coverage of the basic scenarios now

PR #503, opened 2017-10-12
Add a DevCompilerBootstrapBuilder

jakemac53:
Most of this is just copied from what we do in pub.

I also added source maps and error files for the DevCompilerBuilder since I noticed those were missing when testing.

The current test is very pretty basic, I added https://github.com/dart-lang/build/issues/502 to cover making some integration tests which will be much more representative.

Closes https://github.com/dart-lang/build/issues/498

### DIFF
--- a/build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
+++ b/build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
@@ -31,7 +31,7 @@ class DevCompilerBootstrapBuilder extends Builder {
   };
 
   @override
-  Future build(BuildStep buildStep) async {
+  Future<Null> build(BuildStep buildStep) async {
     var dartEntrypointId = buildStep.inputId;
     var isAppEntrypoint = await _isAppEntryPoint(dartEntrypointId, buildStep);
     if (!isAppEntrypoint) return;
@@ -102,8 +102,8 @@ class DevCompilerBootstrapBuilder extends Builder {
     }));
   }
 
-  /// Returns whether or not [dartId] is an app entrypoint (basically, whether or
-  /// not it has a `main` function).
+  /// Returns whether or not [dartId] is an app entrypoint (basically, whether
+  /// or not it has a `main` function).
   Future<bool> _isAppEntryPoint(AssetId dartId, AssetReader reader) async {
     assert(dartId.extension == '.dart');
     // Skip reporting errors here, dartdevc will report them later with nicer
--- a/build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
+++ b/build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
@@ -7,7 +7,7 @@ import 'dart:convert';
 
 import 'package:analyzer/analyzer.dart';
 import 'package:build/build.dart';
-import 'package:path/path.dart' as _p;
+import 'package:path/path.dart' as _p; // ignore: library_prefixes
 
 import 'dev_compiler_builder.dart';
 import 'module_builder.dart';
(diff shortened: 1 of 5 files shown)


build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
```diff
@@ -0,0 +1,272 @@
...
+import 'package:path/path.dart' as _p; // ignore: library_prefixes
+
+import 'dev_compiler_builder.dart';
+import 'module_builder.dart';
+import 'modules.dart';
+
+/// Alias `_p.url` to `p`.
+_p.Context get p => _p.url;
+
+final String bootstrapJsExtension = '.dart.bootstrap.js';
```
jacob314: nit: a bit funky that these start with a leading dot.

jakemac53: Ya that is really just because of how `changeExtension` on `AssetId` works, it won't put in the leading dot.

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
```diff
@@ -0,0 +1,272 @@
...
+  final buildExtensions = {
+    '.dart': [
+      bootstrapJsExtension,
+      jsEntrypointExtension,
+      jsEntrypointSourceMapExtension
+    ],
+  };
+
+  @override
+  Future build(BuildStep buildStep) async {
```
jacob314: Future -> Future<Null>

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
```diff
@@ -0,0 +1,272 @@
...
+
+    // The name of the entrypoint dart library within the entrypoint JS module.
+    //
+    // This is used to invoke `main()` from within the bootstrap script.
+    //
+    // TODO(jakemac53): Sane module name creation, this only works in the most
+    // basic of cases.
+    //
+    // See https://github.com/dart-lang/sdk/issues/27262 for the root issue
+    // which will allow us to not rely on the naming schemes that dartdevc uses
```
jacob314: > 80 chars here and in a few other comments.

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
```diff
@@ -0,0 +1,272 @@
...
+  el.async = false;
+  el.src = "/packages/\$sdk/dev_compiler/amd/require.js";
+  el.setAttribute("data-main", _currentDirectory + "$bootstrapModuleName");
+  document.head.appendChild(el);
+})();
+''';
+
+/// JavaScript snippet to determine the directory a script was run from.
+final _currentDirectoryScript = r'''
+var _currentDirectory = (function () {
```
jacob314: wow... this snippet has now been copied an impressive # of places. Maybe it should go in the DDC package.

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
jacob314: nm. I see existing code in this script using the same style so no need to change.

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
jakemac53: done

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
jakemac53: this one is actually good but I fixed a different one

build_compilers/lib/src/dev_compiler_bootstrap_builder.dart
jakemac53: Ya, it probably should, as well as some of this other stuff. Not worth blocking on though imo.

PR #509, opened 2017-10-13
add a test package bootstrap builder to build_test

jakemac53:
Partly unblocks https://github.com/dart-lang/build/issues/502


natebosch:
Is this where we are going to want this long term?

Do we have a plan for what dependencies are going to look like for end clients?

jakemac53:
I think this would be an OK place, but we could see about adding it to the test package directly as well

PR #510, opened 2017-10-13
Updates to serve handler, now provides a factory for handlers of specified top level dirs

jakemac53:
`ServeHandler.handle` has been replaced with `Handler ServeHandler.handleFor(String rootDir)`. This allows you to create separate handlers per directory you want to serve, which maintains pub serve conventions and allows interoperation with `pub run test --pub-serve=$PORT` (which requires you to be serving the `test` dir).

More work towards https://github.com/dart-lang/build/issues/502 (and general unblocking of running tests)

cc @kevmoo @matanlurey

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -29,9 +29,17 @@ class ServeHandler implements BuildState {
   @override
   Stream<BuildResult> get buildResults => _state.buildResults;
 
-  Future<Response> handle(Request request) async {
+  Handler handlerFor(String rootDir) {
+    if (p.url.split(rootDir).length != 1) {
+      throw new ArgumentError.value(
+          rootDir, 'rootDir', 'Only top level directories are supported');
+    }
+    return (Request request) => _handle(request, rootDir);
+  }
+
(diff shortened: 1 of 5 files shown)


jakemac53:
@natebosch going to merge for now, but feel free to leave a post-submit review and I will address changes.

build_runner/lib/src/server/server.dart
```diff
@@ -43,9 +51,14 @@ class AssetHandler {
 
   AssetHandler(this._reader, this._rootPackage);
 
-  Future<Response> handle(Request request) {
-    var pathSegments = request.url.pathSegments;
-    if (request.url.path.endsWith('/')) {
+  Future<Response> handle(Request request, String rootDir) {
+    var pathSegments = <String>[];
+    if (rootDir != null) {
```
natebosch: when would this be null? If we don't expect it to ever be null we could add an `assert` instead of a conditional.

jakemac53: originally I had this as an optional arg, so ya I will send a pr to clean it up :)

PR #513, opened 2017-10-13
use linked summaries that are available when computing other linked summaries

jakemac53:
Previously we always used unlinked summaries, but this speeds things up a bit since there is less redundant linking that is done (about 10-20% speedup on the examples I tried).

Generally this means that for modules in other packages you will get linked summaries, and modules within the same package you will get the unlinked ones.

@stereotype441 I assigned you since you are probably most familiar with these apis, is it fine to mix summary types like this? It seems to work but I wanted to confirm with you :).


stereotype441:
I don't see any problem with combining linked and unlinked summaries in this way, and I don't see any problems in the code, so lgtm.

Note, however, that I'm not really familiar with this code base, so you might consider getting a second review from someone else :)

jakemac53:
updated it

PR #515, opened 2017-10-14
Fix dartdoc reference

kevmoo:


kevmoo:
I'm working on it w/ the dartdoc folks.

natebosch:
Is it possible to run the analyzer with the `--fatal-lints` flag on Travis? Would it be sensible to do so?

kevmoo:
We could...

Issue #500, opened 2017-10-11
Files added after you start `watch` aren't servable until you restart the build


jakemac53:
I think we ignore the watch notifications for these since they aren't in the asset graph, but we should be adding them to the asset graph so that they can be served.

PR #518, opened 2017-10-16
Add an integration-y test for serving.

matanlurey:
Including a skipped test for https://github.com/dart-lang/build/issues/500 to fix 😄 

I tried to fix it, but there is a lot of moving parts for me. So here's a test instead!

Let me know if you want to call this `server_functional_test` or something instead.

### DIFF
--- a/build_runner/test/server/serve_integration_test.dart
+++ b/build_runner/test/server/serve_integration_test.dart
@@ -1,0 +1,111 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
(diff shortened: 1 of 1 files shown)


build_runner/test/server/serve_integration_test.dart
```diff
@@ -0,0 +1,111 @@
...
+      writer: writer,
+      logLevel: Level.ALL,
+      directoryWatcherFactory: (path) => new FakeWatcher(path),
+    ));
+    handler = server.handlerFor('web');
+
+    nextBuild = new Completer<Null>();
+    subscription = server.buildResults.listen((_) {
+      if (!nextBuild.isCompleted) {
+        nextBuild.complete();
```
jakemac53: should this just assign `nextBuild` to a `new Completer<Null>()` here? That would probably make us less likely to use it incorrectly if we add more tests here :).

matanlurey:
Thanks!

PR #519, opened 2017-10-16
Update e2e_example to be simpler and run an integration test.

jakemac53:
Start towards https://github.com/dart-lang/build/issues/502


e2e_example/LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright 2016, the Dart project authors. All rights reserved.
```
natebosch: 2017?

e2e_example/tool/build.dart
```diff
@@ -1,12 +1,66 @@
...
 
 Future main() async {
-  await build(phases);
+  var graph = new PackageGraph.forThisPackage();
+  var buildActions = <BuildAction>[
+    new BuildAction(new TestBootstrapBuilder(), graph.root.name,
+        inputs: ['test/**_test.dart']),
+  ];
+
+  void crawlPackageAndAddBuilder(
```
natebosch: I think it would be nice to use the `orderedPackages` and `dependentsOf` from the `PackageGraph` rather than a custom function here - I think this is what the generated scripts will end up using.

See for example: https://gist.github.com/natebosch/3f17d779a068dcf04f0b0ccbcb00edb9#file-serve-dart for using `dependentsOf` with angular.

e2e_example/pubspec.yaml
```diff
@@ -1,23 +1,28 @@
 name: e2e_example
+description: A web app that uses AngularDart Components
```
natebosch: Looks like it doesn't use the components yet? I think we should wait to update the description for the commit which starts using them.

e2e_example/LICENSE
jakemac53: done

e2e_example/pubspec.yaml
jakemac53: done, cruft cause I started with angular stagehand but later simplified

e2e_example/tool/build.dart
jakemac53: done, using `orderedPackages`

jakemac53:
heh, looks like I will have to go about installing chrome as well, probably because with mono_repo we aren't using the standard travis tasks.

jakemac53:
@kevmoo ya I filed https://github.com/dart-lang/mono_repo/issues/18, can deal with local hacks for now

PR #523, opened 2017-10-17
Correctly collapse change events

jakemac53:
Fixes https://github.com/dart-lang/build/issues/500

Previously we would just use the last event that came in which in the case of ADD and then MODIFY caused us to miss the ADD.

Also fixed up the integration test to properly remove added files.

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -142,7 +142,7 @@ class AssetGraph {
     }
 
     // Delete all the invalidated assets, then remove them from the graph. This
-    // order is important because some `AssetReader`s throw if the id is not in
+    // order is important because some `AssetWriter`s throw if the id is not in
     // the graph.
     await Future.wait(idsToDelete.map(delete));
     idsToRemove.forEach(_remove);
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:collection';
 
 import 'package:build/build.dart';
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -103,7 +104,17 @@ class BuildImpl {
   Future<Null> _updateAssetGraph(Map<AssetId, ChangeType> updates) async {
     var deletes = _assetGraph.updateAndInvalidate(
         _buildActions, updates, _packageGraph.root.name);
-    await Future.wait(deletes.map(_delete));
+    await Future.wait(deletes.map((node) {
```
jakemac53: I actually struggled a fair bit to come up with anything clean here - its harder than it seems and this was the best I could come up with.

If you have a better idea that doesn't involve a huge refactor let me know :).

jakemac53: Actually there is one other option that I just thought of, we could change `updateAndInvalidate` to take in a `Future<Null> delete(AssetId id)` method, which it would call _before_ actually removing the node from the graph (and we would pass `_delete` here. That also means `updateAndInvalidate` would become async but I think that is fine, and it removes any of this hackery.

jakemac53: Updated it to use the new logic, definitely cleaner IMO (at least we don't have to duplicate the cache location logic).

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -134,10 +141,11 @@ class AssetGraph {
       }
     }
 
-    deletes.addAll(allNodes
-        .where((n) => n is GeneratedAssetNode && n.needsUpdate)
-        .map((n) => n.id));
-    return deletes;
+    // Delete all the invalidated assets, then remove them from the graph. This
```
natebosch: this is unfortunate... We're getting some pretty tight coupling here. It is nicer than the previous version.

Did you mean to mention AssetWriter rather than AssetReader? I don't understand why a reader would be involved with a delete.

I don't have any specific alternatives to offer, but we can take another look later.

jakemac53: Updated to `AssetWriter` (although technically `AssetReader` does also require things to be in the graph).

I definitely agree that this isn't ideal, but I couldn't come up with a better solution that doesn't require a much larger refactor which isn't worth going into at this time.

jakemac53:
I will work on a unit test for `updateAndInvalidate` as well, it looks like that doesn't have any unit tests today.

build_runner/lib/src/asset_graph/graph.dart
natebosch: After some more thought I'm less concerned about this than I was, I think this is likely to be fine for the foreseeable future.

jakemac53:
@natebosch ok I added a unit test, which did also lead to a change where I am now priming the `outputs` with `primaryOutputs` earlier on (when the asset graph is built up). Previously this didn't happen until inside the actual build.

PR #525, opened 2017-10-17
clean up analysis errors

jakemac53:
Stop excluding e2e_example and add some ignore comments.

### DIFF
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -54,7 +54,7 @@ class BuildStepImpl implements BuildStep {
       Iterable<AssetId> expectedOutputs,
       this._reader,
       this._writer,
-      @deprecated this._rootPackage,
+      @deprecated this._rootPackage, // ignore: deprecated_member_use
       this._resolvers,
       this._resourceManager)
       : _expectedOutputs = expectedOutputs.toSet();
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -49,14 +49,8 @@ class BuildStepImpl implements BuildStep {
 
   final ResourceManager _resourceManager;
 
-  BuildStepImpl(
-      this.inputId,
-      Iterable<AssetId> expectedOutputs,
-      this._reader,
-      this._writer,
-      @deprecated this._rootPackage, // ignore: deprecated_member_use
-      this._resolvers,
-      this._resourceManager)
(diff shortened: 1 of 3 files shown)


build/lib/src/builder/build_step_impl.dart
```diff
@@ -54,7 +54,7 @@ class BuildStepImpl implements BuildStep {
       Iterable<AssetId> expectedOutputs,
       this._reader,
       this._writer,
-      @deprecated this._rootPackage,
+      @deprecated this._rootPackage, // ignore: deprecated_member_use
```
jakemac53: Unless we are actually planning on cleaning this up? I think its necessary though.

natebosch: I think we probably should clean it up - but it would be a breaking because we now reference `BuildStepImpl` in `build_runner`.

We don't read this field anymore do we?
https://github.com/dart-lang/build/pull/379/files#diff-4544b82ad6d6dd0948c77a13cae96758

jakemac53: We reference it in a few places in this file to do checks that you aren't trying to read invalid files etc.

natebosch: In that case I'd say remove the `@deprecated`

natebosch: or maybe we intended to infer it from `this.inputId.package`?

jakemac53: I don't think its always the same as `this.inputId.package`, it should be the root package in the graph. I can just remove the `@deprecated`.

jakemac53:
@natebosch updated to just remove the deprecated annotation

build/lib/src/generate/run_builder.dart
```diff
@@ -29,16 +29,22 @@ Future<Null> runBuilder(Builder builder, Iterable<AssetId> inputs,
...
     var outputs = expectedOutputs(builder, input);
     if (outputs.isEmpty) return;
-    var buildStep = new BuildStepImpl(input, outputs, reader, writer,
-        rootPackage ?? input.package, resolvers, resourceManager);
+    var buildStep = new BuildStepImpl(
+        input,
+        outputs,
+        reader,
+        writer,
+        rootPackage ?? input.package, // ignore: deprecated_member_use
```
natebosch: do we still need this ignore?

jakemac53: updated

Issue #524, opened 2017-10-17
File watching doesn't start until after the first build is complete


jakemac53:
Because of this any edits during the first build will not be captured.

We should really set up all the file watchers before even starting the first build, otherwise we can get in a bad state.

natebosch:
yeah I think I broke this behavior in https://github.com/dart-lang/build/pull/404

natebosch:
I think what we want is to undo https://github.com/dart-lang/build/pull/331 so that we can `await` and be sure the watchers are running before moving on to do more work.

PR #527, opened 2017-10-18
schedule file watchers before the first build completes

jakemac53:
Fixes https://github.com/dart-lang/build/issues/524

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -73,47 +73,73 @@ class WatchImpl implements BuildState {
   /// Runs a build any time relevant files change.
   ///
   /// Only one build will run at a time, and changes are batched.
-  Stream<BuildResult> _run(BuildOptions options, List<BuildAction> buildActions,
-      Future until) async* {
-    var fatalBuild = new Completer();
-    var firstBuild = new Completer<BuildResult>();
-    currentBuild = firstBuild.future;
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -73,54 +73,81 @@ class WatchImpl implements BuildState {
...
+  ///
+  /// File watchers are scheduled synchronously.
+  Stream<BuildResult> _run(
+      BuildOptions options, List<BuildAction> buildActions, Future until) {
+    var fatalBuildCompleter = new Completer();
+    var firstBuildCompleter = new Completer<BuildResult>();
+    currentBuild = firstBuildCompleter.future;
+    var controller = new StreamController<BuildResult>();
+
+    // We need this inside `doBuild`, but we won't actually create it until
```
natebosch: FWIW I think this would be easier to follow with the old structure and a `Future<Stream>` returned from `watch` but I don't have strong feelings about it.

jakemac53: We would still need to define the `doBuild` function before a `BuildImpl` actually exists though, unless we jumped through some major hoops. The issue is that `BuildImpl.create()` actually schedules the first build, and even if it didn't we still want to start the watchers before we create the `BuildDefinition` to eliminate race conditions. The only thing it would allow us to do is eliminate the async closure that actually creates the build.

Issue #526, opened 2017-10-18
Builds can be inconsistent if files change during a build


jakemac53:
Today we don't read in files at the very start of a build, or cache files at all, which means that if a file changes during a build any actions after that point will see the new version of the file which is different than the version other builders saw _in the same build_.

We can fix this one of two ways:

- Lazy file caching: this way all builders within a build would see the file in the same state, but it is possible that we would schedule an additional build that isn't necessary. 
- Add eager file caching: read all source files upfront, before each build. This would be a bit slower but we actually are going to want to do this anyways once we start using file hashes to detect changes.

jakemac53:
Fwiw, from some really basic testing adding the lazy caching strategy appears to provide a significant performance boost for angular_components_example. With angular + ddc + summaries the initial build goes from ~140 seconds to ~100 seconds (obviously still to slow but a nice improvement).

PR #530, opened 2017-10-18
Add back basic file caching

jakemac53:
Fixes https://github.com/dart-lang/build/issues/526, well enough at least (It guarantees that all builders see the same version of a file within a build).

We could still end up scheduling an extra build that isn't necessary (if the file was first read during a build after the change happened), but that is less of an issue and could be resolved using asset hashes.

This also has the nice benefit of speeding up builds, especially the larger builds. Initial build for angular_components_example goes from ~140s for me to ~83s.

I will follow up with some unit tests.

### DIFF
--- a/build_runner/lib/src/asset/cache.dart
+++ b/build_runner/lib/src/asset/cache.dart
@@ -8,44 +8,55 @@ import 'dart:convert';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 
+/// An [AssetReader] that wraps another [AssetReader] and caches all results
+/// from it.
+///
+/// Assets are cached until [invalidate] is invoked.
+///
+/// Does not implement [findAssets].
 class CachingAssetReader implements AssetReader {
+  /// Cached results of [canRead].
   final _canReadCache = <AssetId, Future<bool>>{};
-  final _contentCache = <AssetId, Future<dynamic>>{};
-  final AssetReader _wrapped;
 
-  CachingAssetReader(this._wrapped);
+  /// Cached results of [readAsBytes].
+  final _bytesContentCache = <AssetId, Future<List<int>>>{};
+
+  /// Cached results of [readAsString], per [Encoding] type used.
+  ///
+  /// These are computed and stored lazily using [readAsBytes].
+  final _stringContentCache = <AssetId, Map<Encoding, Future<String>>>{};
+
+  /// The [AssetReader] to delegate all reads to.
+  final AssetReader _delegate;
+
+  CachingAssetReader(this._delegate);
 
   @override
   Future<bool> canRead(AssetId id) =>
-      _canReadCache.putIfAbsent(id, () => _wrapped.canRead(id));
+      _canReadCache.putIfAbsent(id, () => _delegate.canRead(id));
 
   @override
-  Stream<AssetId> findAssets(Glob glob, {String package}) =>
+  Stream<AssetId> findAssets(Glob glob) =>
       throw new UnimplementedError('unimplemented!');
 
   @override
-  Future<List<int>> readAsBytes(AssetId id) => _contentCache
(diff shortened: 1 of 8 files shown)


natebosch:
What is the impact on memory usage?

build_runner/lib/src/asset/cache.dart
```diff
@@ -0,0 +1,62 @@
...
+  /// These are computed and stored lazily using [readAsBytes].
+  final _stringContentCache = <AssetId, Map<Encoding, Future<String>>>{};
+
+  /// The [AssetReader] to delegate all reads to.
+  final AssetReader _delegate;
+
+  CachingAssetReader(this._delegate);
+
+  @override
+  Future<bool> canRead(AssetId id) =>
```
natebosch: how do we account for files that get added and can be read later? Is this cache per-phase?

jakemac53:
In regards to memory usage, previously for angular_components_example it would use up to 1.4GB. Now it uses about 2.4GB, so close to double, but manageable.

We could save some of that by not storing the bytes and strings separately, but caching the strings actually resulted in a significant speedup in my testing (saved about 20 seconds), which seems worth while.

I think we can likely reclaim some memory when we write the new Resolver as well.

build_runner/lib/src/asset/cache.dart
jakemac53: We still wrap this in the `SinglePhaseReader`, which handles checking phase numbers etc. This just wraps the `FileBasedAssetReader` (or whatever reader you pass in to watch/build), which is always just hitting the filesystem directly.

natebosch: does that mean that new files can never appear readable? Does the SinglePhaseReader no call through for `canRead`?

natebosch: I think we should maybe only cache the `true` values and call through the delegate for everything else here.

jakemac53: Ah, I didn't think that we ever called through to the underlying reader for this in the `SinglePhaseAssetReader`, but it looks like we do.

It seems like we should just be going directly to the `AssetGraph`, and checking if the `GeneratedAssetNode.wasOutput` for generated nodes. For source nodes this logic should hold fine and is actually what we want. We don't want this to return a different answer for two actions because the user added a new file during the build for instance.

natebosch: I'm a little bit nervous about having this cached like this and depending on it being wrapped in an implementation that won't call it... Can we do the safer implementation, or change it to throw not implemented?

jakemac53: I agree this seems a bit tricky, but in reality it is not bad imo. This class does not in fact depend on being wrapped in a different implementation. It just does exactly what it claims to, which is cache results until you call invalidate. This is important in order to maintain the same state throughout a build.

The only special logic involved is in the `SinglePhaseAssetReader`. That class is already pretty special, and uses a lot more knowledge about the asset graph and state of the world to block reads, run builds, and return different results based on phases etc. This is just another case for it, where for generated assets it does not call the wrapped reader, but instead looks at the asset graph directly to determine if a generated file exists.

Throwing for `canRead` isn't an option because it does need to be able to work (for source assets). The other behavior of not caching for `canRead` is not actually safer, its explicitly creating a potential for inconsistent builds (if you create a new source file during a build).

jakemac53:
I am going to add at least a test for calling `canRead` on a generated asset before its actually available, and then later on once it is available, to make sure we don't regress that use case.

I will also add one for creating a new file and deleting a file during a build to make sure we don't regress those use cases either.

jakemac53:
merging for now, we can follow up later if you have more ideas but I think the tests should alleviate concerns around `canRead`.

PR #532, opened 2017-10-19
Library root fix for DDC

jakemac53:
Changes the library root to be the top level dir of the package in which a module resides, not the full dirname of the module.

This fixes an issue where relative imports inside of test/web that reached from a subdir to a parent dir (inside the same top level dir) would give an error.


jakemac53:
What do you know... this revealed a bug via our e2e tests :)

Apparently editing a library dependency inside of a non-lib folder causes an error. I think the issue is that the summary ends up with the full absolute path to the files, but that includes the old scratch space directory which no longer exists. We might be able to use --url-mapping or something to resolve it.

jakemac53:
Going to start skipping the e2e_tests I guess for now :(.

PR #537, opened 2017-10-23
Move BuildConfig from package:dazel

natebosch:
This is mostly a copy of the code from `dazel` with a few modifications:

- Shuffle `pubspec.dart` under the same src diretory as the other files
  rather than `bazelify`
- Rename `DartLibrary` as `BuildTarget` and don't implement any parent
  interface.
- Rename `DartBuilderBinary` as `BuilderDefinition` and don't implement
  any parent interface
- Fix some strong mode errors by adding explicit casts.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -13,9 +13,14 @@ import 'pubspec.dart';
 /// The parsed values from a `build.yaml` file.
 class BuildConfig {
   /// Supported values for the `platforms` attribute.
-  static const _allPlatforms = const [_vmPlatform, _webPlatform];
+  static const _allPlatforms = const [
+    _vmPlatform,
+    _webPlatform,
+    _flutterPlatform,
+  ];
   static const _vmPlatform = 'vm';
   static const _webPlatform = 'web';
+  static const _flutterPlatform = 'flutter';
 
   static const _targetOptions = const [
     _builders,
@@ -74,14 +79,14 @@ class BuildConfig {
   /// The default config if you have no `build.yaml` file.
   BuildConfig.useDefault(Pubspec pubspec,
       {bool includeWebSources: false,
-      bool enableDdc: true,
+      List<String> platforms: const [],
       Iterable<String> excludeSources: const []})
       : packageName = pubspec.pubPackageName {
     var sources = ["lib/**"];
     if (includeWebSources) sources.add("web/**");
     dartLibraries[packageName] = new BuildTarget(
         dependencies: pubspec.dependencies,
-        enableDdc: enableDdc,
+        platforms: platforms,
         isDefault: true,
         name: packageName,
         package: pubspec.pubPackageName,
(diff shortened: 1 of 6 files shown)


build_config/lib/src/build_config.dart
```diff
@@ -0,0 +1,344 @@
...
+    }
+  }
+
+  final String packageName;
+
+  /// All the `builders` defined in a `build.yaml` file.
+  final dartBuilderBinaries = <String, BuilderDefinition>{};
+
+  /// All the `targets` defined in a `build.yaml` file.
+  final dartLibraries = <String, BuildTarget>{};
```
jakemac53: I wonder if we actually still want this?

build_config/lib/src/build_config.dart
```diff
@@ -0,0 +1,344 @@
...
+
+import 'package:path/path.dart' as p;
+import 'package:yaml/yaml.dart';
+
+import 'pubspec.dart';
+
+/// The parsed values from a `build.yaml` file.
+class BuildConfig {
+  /// Supported values for the `platforms` attribute.
+  static const _allPlatforms = const [_vmPlatform, _webPlatform];
```
jakemac53: Note that in bazel we now support 'flutter' as a platform as well I think

build_config/lib/src/pubspec.dart
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
jakemac53: Should we consider publishing this as a separate package?

build_config/lib/src/build_config.dart
natebosch: We probably don't need it for `build_runner` but I think we still need it for `dazel` and my plan was to eventually rip out the implementation from there and wire this one in.

I should rename these though...

build_config/lib/src/build_config.dart
matanlurey: @grouma for FYI.

grouma: How should we keep this in sync with: https://github.com/dart-lang/rules_dart/blob/6bc0a5777089ec61bace0b3576b82c1edc0692e7/dart/build_rules/common/constants.bzl?

natebosch: We end up with tight version constraints between `dazel` and `rules_dart`. We'll have to keep the version of `build_config` in mind as well. We don't have a good story around preventing breakage because of things like this other than frequent testing.

build_config/lib/src/pubspec.dart
jakemac53: I have seen several community members ask for a package like this, but at the same time I don't know that we really want to support it.

natebosch: I think I'd be happy to switch this package over to use something we publish later - I don't think I want to try to use this as the starting point for a new package though.

jakemac53: sgtm

Issue #533, opened 2017-10-19
Editing library dependency under same non-lib dir causes build failure


jakemac53:
It looks like summaries for non-lib modules have absolute uris to the files, which point at a tmp dir that only exists for a single build.

We will need to change the uris in the summaries inside of `_analyzerSourceArgsForModule` in summary_builder.dart, and then also change dev_compiler_builder.dart to know about the new uris for those files (probably using --url-mapping).

jakemac53:
cc @vsmenon @jmesserly I think I am taking the right approach to this but I haven't actually been able to get it to work.

For example, we have a file structure like:

test/
  hello_world_test.dart
  common/
    message.dart

When compiling I get this error:

```
[error] Target of URI doesn't exist: 'common/message.dart'. (/tmp/build_compilersJtITsk/test/hello_world_test.dart, line 12, col 8)
[error] Undefined name 'message'. (/tmp/build_compilersJtITsk/test/hello_world_test.dart, line 24, col 41)

Please fix all errors before compiling (warnings are okay).
```

I am currently passing this argument:

--url-mapping=file:///tmp/build_compilersJtITsk/test/common/message.dart,/test/common/message.dart

The summary for `test/common/message.dart` looks like this:

```
{
  version: 1.0,
  linkedLibraryUris: ["file:///test/common/message.dart"],
  unlinkedUnitUris: ["file:///test/common/message.dart"],
  library "file:///test/common/message.dart": {
    dependencies[0]: {uri: "file:///test/common/message.dart"},
    dependencies[1]: {
      parts[0]: "dart:core/annotations.dart",
      parts[1]: "dart:core/bool.dart",
      parts[2]: "dart:core/comparable.dart",
      parts[3]: "dart:core/date_time.dart",
      parts[4]: "dart:core/double.dart",
      parts[5]: "dart:core/duration.dart",
      parts[6]: "dart:core/errors.dart",
      parts[7]: "dart:core/exceptions.dart",
      parts[8]: "dart:core/expando.dart",
      parts[9]: "dart:core/function.dart",
      parts[10]: "dart:core/identical.dart",
      parts[11]: "dart:core/int.dart",
      parts[12]: "dart:core/invocation.dart",
      parts[13]: "dart:core/iterable.dart",
      parts[14]: "dart:core/iterator.dart",
      parts[15]: "dart:core/list.dart",
      parts[16]: "dart:core/map.dart",
      parts[17]: "dart:core/null.dart",
[...]

PR #542, opened 2017-10-24
Fix non lib sub libraries

jakemac53:
Fixes https://github.com/dart-lang/build/issues/533

This changes how we run ddc/analyzer workers to be much closer to how bazel runs them. The scratch space now persists across builds (although we clean it up manually), which means ddc and analyzer can now run directly under that directory (similar to how it runs at bazel root).

I also changed all usages of --uri-mapping to match exactly what we do in bazel as much as I could.

I snuck in a minor cleanup of `computeTransitiveDependencies` as well so it returns `Module`s now instead of the primary ids, which means we can use its getters to reference the related ids to that module instead of doing manual `changeExtension` hacks.


jakemac53:
filed https://github.com/dart-lang/build/issues/543

Issue #543, opened 2017-10-24
Add `beforeExit` callback to `Resource` api


jakemac53:
The `build_compilers` package needs this to finish cleaning up its scratch space.

This should be invoked when the final build is done, right before the process exits. It should get called after the final call to `dispose`.

PR #545, opened 2017-10-24
add beforeExit callback to Resource api

jakemac53:
Fixes https://github.com/dart-lang/build/issues/543

I am not super happy with the fact that each implementation has to worry about this.... but I don't see a way around it :(.

### DIFF
--- a/build/lib/src/resource/resource.dart
+++ b/build/lib/src/resource/resource.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 typedef FutureOr<T> CreateInstance<T>();
 typedef FutureOr DisposeInstance<T>(T instance);
+typedef FutureOr BeforeExit();
 
 /// A [Resource] encapsulates the logic for creating and disposing of some
 /// expensive object which has a lifecycle.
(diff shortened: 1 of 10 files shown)


build/lib/src/resource/resource.dart
```diff
@@ -23,10 +24,16 @@ class Resource<T> {
   /// disposed.
   final DisposeInstance<T> _userDispose;
 
+  /// Optional method which is called before the process is going to exit. This
```
natebosch: [nit] blank line following first sentence. Similar in other doc comments.

build/lib/src/resource/resource.dart
```diff
@@ -53,8 +60,16 @@ class Resource<T> {
 class ResourceManager {
   final _resources = new Set<Resource>();
 
+  /// The [Resource]s that we need to call `beforeExit` on. We have to hang on
+  /// to these forever, but they should be small in number, and we don't hold
+  /// on to the instances.
```
natebosch: "we don't hold on to the instances" - this isn't clear to me, could you rephrase to make it more clear we mean the instances of the created resource, not the `Resource` instance?

build/lib/src/resource/resource.dart
jakemac53: done

build/lib/src/resource/resource.dart
jakemac53: done

Issue #536, opened 2017-10-20
Invalidate assets if the package location changes


jakemac53:
Because we are just using timestamps it is possible to change a dependency and not have the build system notice, because the timestamp of the files in the new location (generally in the pubcache) is older than the old location.

We should just encode in the asset graph (or a different file next to the asset graph) the resolved locations of all packages, and invalidate assets accordingly.

Alternatively we could just invalidate the whole build any time the `.packages` file changes but that would be a bit drastic most likely.

jakemac53:
I am going to fix this instead by changing our asset invalidation over to use content hash based invalidation, which is generally much safer and will unblock other things as well.

Issue #412, opened 2017-09-12
Add caching headers from the AssetHandler


natebosch:
After moving away from the static resource handler we lost caching. If we start using hashes or similar for AssetGraph invalidation we should easily be able to reuse it for etag based client side caching.

matanlurey:
And I imagine we could test this by firing XHRs at the handler and asserting caches?

Issue #371, opened 2017-08-17
Invalidate outputs based on better than timestamp


natebosch:
Today a changed file can cascade to any output which may end up rebuilding way more than is optimal. If we track some type of content hash we can get away with pruning the build graph more eagerly.

jakemac53:
This will also help for correctness, so I am going to go ahead and tackle it.

See https://github.com/dart-lang/build/issues/536 for an example.

jakemac53:
Actually I am only going to tackle invalidating _inputs_, not also invaliding outputs based on whether files actually changed (at least right now).

PR #547, opened 2017-10-26
Use content hashes to invalidate assets instead of last modified timestamps

jakemac53:
All `AssetNode`s now have a `Digest` field (from package:crypto). These are computed eagerly, and must be available for any node which has a corresponding asset that exists. It is also serialized as a part of the asset graph (base64 encoded for json compatibility for now).

There is now an additional interface `DigestAssetReader` which adds the `Future<Digest> digest(id)` method to a normal asset reader, and the `RunnerAssetReader` interface now requires that interface to be implemented.

`AssetGraph.build` is now a static method which returns a `Future<AssetGraph>`, and requires a `DigestAssetReader`. It eagerly reads in digests for all source assets.

MD5 is the current hash function used, because collisions are not a concern in this case (we don't use the hashes for lookups, only to compare new/old values for a single file). Bazel also uses md5 today, as well as the analyzer for summaries, so this is also consistent with those (although they may change it is still kind of nice). It is also significantly faster than sha1 in practice.

This does complicate some of the tests a bit unfortunately (search this pr for `computeHash` in tests....), but its a necessary evil imo. I don't see a way to simplify it significantly.

I didn't yet completely remove the `lastModified` method, it is still used when checking if a build script itself has been updated. I wanted to keep the scope of this as small as I could, so I will follow up with a different pull request to update that as well, and remove `lastModified` entirely.

Partially fixes https://github.com/dart-lang/build/issues/536
Unblocks https://github.com/dart-lang/build/issues/371 and https://github.com/dart-lang/build/issues/412


jakemac53:
@natebosch after thinking a bit more on this I think it might make sense to change this field on `AssetNode` to be a `Future<Digest>`, and compute it lazily for at least some assets (specifically for assets that have no outputs).

The main reasoning be that you might be serving large assets like images, and we likely don't want to be hashing those on startup.

natebosch:
I'm very in favor of setting the interface to `Future` to allow laziness

build_runner/lib/src/asset/reader.dart
```diff
@@ -2,18 +2,34 @@ import 'dart:async';
...
 import 'package:build/build.dart';
+import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
 import '../asset_graph/graph.dart';
 import '../asset_graph/node.dart';
 
 typedef Future RunPhaseForInput(int phaseNumber, AssetId primaryInput);
 
+abstract class DigestAssetReader implements AssetReader {
+  /// Asynchronously compute the digest for [id].
```
natebosch: should we document expected behavior of this call if the Asset does not exist?

jakemac53:
> Should we have any new tests for this?

I will at least add some unit tests for the MD5DigestReader, but for the most part we don't need any additional tests I don't think (already updated the asset graph tests, and the e2e tests aren't affected).

natebosch:
> but for the most part we don't need any additional tests I don't think

Do we have a test for the case in #536 ?

jakemac53:
Also just fyi in terms of numbers, for angular_components_example:

- Creating a new asset graph takes ~1 second, doesn't noticeably affect the first build time
- Updating asset graph on incremental startup takes ~750ms
- Ram usage isn't noticeably affected

So, actually surprisingly little negative affect for angular_components_example. As mentioned earlier I could imagine a case where this would try to compute hashes of very large files which would not be good, but we can address that by making the hash be computed lazily for nodes with no output.

jakemac53:
> Do we have a test for the case in #536 ?

Don't think so (otherwise it would be failing), I can add something.

jakemac53:
@natebosch just fyi after playing around with a few options I don't think I want to go the `Future<Digest>` route, because that means every `AssetNode` needs a reference to a `DigestReader`.

Instead I changed the field from `digest` to `lastKnownDigest`, and changed the docs to say it may be null if the node has no outputs. As indicated this field will now only eagerly be computed for nodes that actually have outputs. It will be lazily computed and assigned if the asset is read by any build step as well (to ensure its consistent with the state in which it was last read).

natebosch:
SGTM

jakemac53:
Actually - this doesn't fully fix https://github.com/dart-lang/build/issues/536 because the last modified logic is still used for build script update detection.

Given that I don't think its worth adding that test as a part of this PR (any reasonable one would have to be skipped anyways).

build_runner/lib/src/asset/cache.dart
```diff
@@ -36,6 +38,9 @@ class CachingAssetReader implements AssetReader {
       _canReadCache.putIfAbsent(id, () => _delegate.canRead(id));
 
   @override
+  Future<Digest> digest(id) => _delegate.digest(id);
```
jakemac53: Note that I removed caching of digests from here - they are cached on `AssetNode`s already so this would be duplicative.

build_runner/lib/src/asset/reader.dart
jakemac53: done

jakemac53:
Ok @natebosch this is actually ready for a review now :)

build_runner/lib/src/asset_graph/serialization.dart
```diff
@@ -121,6 +131,28 @@ class _AssetGraphSerializer {
   }
 }
 
+/// Used to serialize the type of a node using an int.
+enum _NodeTypeId { Source, Synthetic, Generated }
```
natebosch: [nit] I'm not sure 'Id' is useful in these enum names

jakemac53: done

Issue #548, opened 2017-10-26
Should be able to recover from failed builds


jakemac53:
I tried writing an e2e test today for creating an error and then fixing it, looks like this got broken at some point though.

PR #550, opened 2017-10-27
Lots of bug fixes relating to failed builds and recovery from them

jakemac53:
- Adds a new type of asset node, a `SyntheticAssetNode`. These are created if a build step calls `canRead` for an asset that doesn't exist. Those files still need to be tracked so if they are added later on that build step gets invalidated.
- Changes `scopeLog` to `scopeLogAsync`, since its always used in an async context, and we now run in an error handling zone that logs errors to `logger.severe`. Any call to `scopeLogAsync` is now guaranteed to eventually complete with exactly one error or value, and will not leak unhandled async errors.
- Updates the call to `runBuilder` to pass in a logger whose name is `$builder on $input`, so you get more context for errors.
- Fixes the `computeTransitiveModules` to check that modules exist before trying to read them, and warn if they don't. Eventually we probably want this to fail the build as well, but only on some equivalent of "--mode=release". For dev time we probably don't want any builds to fail so the app can serve errors in a friendly fashion.
- Fixes a bug in `updateAndInvalidate` where it would delete nodes that shouldn't be deleted.

Fixes https://github.com/dart-lang/build/issues/548.

### DIFF
--- a/build_compilers/lib/src/dev_compiler_builder.dart
+++ b/build_compilers/lib/src/dev_compiler_builder.dart
@@ -40,9 +40,9 @@ class DevCompilerBuilder implements Builder {
     try {
       await createDevCompilerModule(module, buildStep);
     } on DartDevcCompilationException catch (e) {
-      log.warning('Error compiling ${module.jsId}:\n$e');
       await buildStep.writeAsString(
           buildStep.inputId.changeExtension(jsModuleErrorsExtension), '$e');
+      log.severe('', e);
     }
   }
 }
--- a/build/lib/src/builder/logging.dart
+++ b/build/lib/src/builder/logging.dart
@@ -9,7 +9,27 @@ const Symbol logKey = #buildLog;
 /// Will be `null` when not running within a build.
 Logger get log => Zone.current[logKey] as Logger;
 
-T scopeLog<T>(T fn(), Logger log) => runZoned(fn, zoneSpecification:
-        new ZoneSpecification(print: (self, parent, zone, message) {
-      log.info(message);
-    }), zoneValues: {logKey: log});
+/// Runs [fn] in an error handling [Zone].
+///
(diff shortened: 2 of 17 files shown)


build_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -40,9 +40,9 @@ class DevCompilerBuilder implements Builder {
     try {
       await createDevCompilerModule(module, buildStep);
     } on DartDevcCompilationException catch (e) {
-      log.warning('Error compiling ${module.jsId}:\n$e');
       await buildStep.writeAsString(
           buildStep.inputId.changeExtension(jsModuleErrorsExtension), '$e');
+      log.severe('', e);
```
kevmoo: Empty string? Weird?

jakemac53: This is just logging the error - it feels a bit weird for sure but there is already additional context provided by the loggers name (it is named based on the current builder and primary input).

The logger handles consistently formatting the error already as well - so that ends up being nicer than passing the error as the message.

kevmoo: ack

kevmoo: ...why no stack? Not useful?

jakemac53: Ya - we don't have a true stack trace here. This error is originating from a worker, and contains the entire message that the worker sent back - which may or may not contain a useful stack trace already.

jakemac53:
> Would it be sensible to add any unit tests specifically around replacing a SyntheticAssetNode with a real one? We probably want to make sure it works for it to get replaced with either a source or generated file.

Ya, this is effectively being tested with the ddc error recovery e2e test, but some unit tests would be good as well, I can work on that.

jakemac53:

PR #551, opened 2017-10-27
serve build errors directly on build failures

jakemac53:
Looks like this for now, we can make it pretty later if we want :D

screenshot from 2017-10-27 14 34 29

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -45,14 +45,8 @@ class ServeHandler implements BuildState {
   FutureOr<Response> _handle(Request request, String rootDir) async {
     await currentBuild;
     if (_lastBuildResult.status == BuildStatus.failure) {
-      return new Response(200,
-          body: '<h1>Build failed!</h1>'
-              '<p><strong>Error:</strong></p>'
-              '<p><strong style="color: red">'
-              '${_htmlify(_lastBuildResult.exception)}'
-              '</strong></p>'
-              '<p><strong>Stack Trace:</strong></p>'
-              '<p>${_htmlify(_lastBuildResult.stackTrace)}</p>',
+      return new Response(HttpStatus.INTERNAL_SERVER_ERROR,
+          body: _htmlErrorPage(_lastBuildResult),
           headers: {
             HttpHeaders.CONTENT_TYPE: 'text/html',
           });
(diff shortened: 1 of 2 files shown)


jakemac53:
cc @kevmoo

I think in general we might want to go this route for build errors, even at development time. It also means we could drop the `.errors` files so the asset graphs are smaller, and we can drop the error handler logic from the require.js bootstrapping code.

build_runner/lib/src/server/server.dart
```diff
@@ -39,8 +44,24 @@ class ServeHandler implements BuildState {
 
   FutureOr<Response> _handle(Request request, String rootDir) async {
     await currentBuild;
+    if (_lastBuildResult.status == BuildStatus.failure) {
+      return new Response(200,
```
natebosch: should this be 200? We don't want to indicate a server error?

build_runner/lib/src/server/server.dart
```diff
@@ -39,8 +44,24 @@ class ServeHandler implements BuildState {
...
+          body: '<h1>Build failed!</h1>'
```
natebosch: I think it could be worth pulling this out into a separate method, otherwise it takes up more space than it earns in this one.

build_runner/lib/src/server/server.dart
jakemac53: Ya I went back and forth on this a bit but it looks like chrome/firefox at least will render the page even if the response status is 500.

Updated to `HttpStatus.INTERNAL_SERVER_ERROR`.

build_runner/lib/src/server/server.dart
jakemac53: done

build_runner/lib/src/server/server.dart
```diff
@@ -39,8 +44,26 @@ class ServeHandler implements BuildState {
 
   FutureOr<Response> _handle(Request request, String rootDir) async {
     await currentBuild;
+    if (_lastBuildResult.status == BuildStatus.failure) {
+      return new Response(HttpStatus.INTERNAL_SERVER_ERROR,
+          body: _htmlErrorPage(_lastBuildResult),
+          headers: {
+            HttpHeaders.CONTENT_TYPE: 'text/html',
```
natebosch: [optional] This might take up fewer lines without the comma

jakemac53:

Issue #552, opened 2017-10-28
Include more info when throwing `UnexpectedOutputException`


kevmoo:
https://github.com/dart-lang/build/blob/d5dd5ef8e4579384cc06c32d7bf857e3e6139408/build/lib/src/builder/build_step_impl.dart#L138

It'd be nice to include the set of expected outputs – even just for debugging.

PR #556, opened 2017-10-30
Add expectedOutputs to UnexpectedOutputException

jakemac53:
Fixes https://github.com/dart-lang/build/issues/552

### DIFF
--- a/build/lib/src/builder/exceptions.dart
+++ b/build/lib/src/builder/exceptions.dart
@@ -5,9 +5,11 @@ import '../asset/id.dart';
 
 class UnexpectedOutputException implements Exception {
   final AssetId assetId;
+  final Iterable<AssetId> expectedOutputs;
 
-  UnexpectedOutputException(this.assetId);
+  UnexpectedOutputException(this.assetId, {this.expectedOutputs});
 
   @override
-  String toString() => 'UnexpectedOutputException: $assetId';
+  String toString() => 'UnexpectedOutputException: $assetId'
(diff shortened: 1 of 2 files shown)


build/lib/src/builder/exceptions.dart
```diff
@@ -5,9 +5,11 @@ import '../asset/id.dart';
 
 class UnexpectedOutputException implements Exception {
   final AssetId assetId;
+  final Iterable<AssetId> expectedOutputs;
```
natebosch: [nit] can we call the `expected`? I think "Outputs" should be obvious in context.

build/lib/src/builder/exceptions.dart
```diff
@@ -5,9 +5,11 @@ import '../asset/id.dart';
...
 
-  UnexpectedOutputException(this.assetId);
+  UnexpectedOutputException(this.assetId, {this.expectedOutputs});
 
   @override
-  String toString() => 'UnexpectedOutputException: $assetId';
+  String toString() => 'UnexpectedOutputException: $assetId'
+      '${expectedOutputs == null ? '' : '\nExpected one of: $expectedOutputs'}';
```
natebosch: [nit] "Expected only:" - "one of" might imply that we don't expect multiple, which we do allow.

build/lib/src/builder/exceptions.dart
jakemac53: done

build/lib/src/builder/exceptions.dart
jakemac53: done

jakemac53:

PR #561, opened 2017-11-02
Drop last modified apis, update build script checker to use digests

jakemac53:
Lots of small changes to the test infrastructure as well here:

- `DatedValue`, `DatedString`, and `DatedBytes` are all gone.
- Added the `decodedMatches` matcher.
- Added ability to skip the build script check for testing purposes.
- Moved build script update tests to integration tests.


natebosch:
Travis found some issues.

build_runner/lib/src/changes/build_script_updates.dart
```diff
@@ -3,58 +3,140 @@
...
-import 'package:path/path.dart' as path;
+import 'package:path/path.dart' as p;
 
 import '../asset/reader.dart';
+import '../asset_graph/graph.dart';
 import '../generate/options.dart';
 
 class BuildScriptUpdates {
-  final RunnerAssetReader _reader;
+  final AssetGraph _graph;
```
natebosch: can we get away with passing what is necessary as arguments to `hasBeenUpdated`?

build_runner/lib/src/changes/build_script_updates.dart
```diff
@@ -3,58 +3,140 @@
...
-    var urisInUse = currentMirrorSystem().libraries.keys;
-    var updateTimes = await Future.wait(urisInUse.map(_uriUpdateTime));
-    return updateTimes.any((u) => u.isAfter(time));
+  Iterable<Uri> get _urisForThisScript => currentMirrorSystem().libraries.keys;
+
+  /// Checks if the current running program has been updated, based on previous
+  /// known digests in [_graph].
+  Future<bool> hasBeenUpdated() async {
+    return (await Future.wait(_urisForThisScript.map(_wasUpdated)))
+        .any((wasUpdated) => wasUpdated);
```
natebosch: [nit] can this be `.contains(true)`?

build_runner/lib/src/changes/build_script_updates.dart
```diff
@@ -3,58 +3,140 @@
...
+
+  /// Checks whether [uri] has been updated.
+  ///
+  /// If it can't tell due to an unsupported uri, then this always returns
+  /// `true`.
+  Future<bool> _wasUpdated(Uri uri) async {
+    AssetId id;
+    try {
+      id = _idForUri(uri);
+      // `null` but no error is a signal to skip this uri.
```
natebosch: call this out in the Doc comment

build_runner/lib/src/changes/build_script_updates.dart
```diff
@@ -3,58 +3,140 @@
...
+      return true;
+    }
+    if (node.lastKnownDigest == null) {
+      // This could happen if a new import was added, or we weren't able to
+      // initially compute the last known digest.
+      return true;
+    } else {
+      var currentDigest = await _reader.digest(id);
+      if (node.lastKnownDigest == currentDigest) return false;
+      return true;
```
natebosch: [nit] Return a condition rather thatn `if (condition) return false;`

```dart
else {
  return node.lastKnownDigest != await _reader.digest(id);
}
```

build_runner/lib/src/changes/build_script_updates.dart
jakemac53: We could, but we would have to pass it through a few different functions - not sure what we would gain?

build_runner/lib/src/changes/build_script_updates.dart
jakemac53: yep! done

build_runner/lib/src/changes/build_script_updates.dart
jakemac53: done

build_runner/lib/src/changes/build_script_updates.dart
jakemac53: done

PR #564, opened 2017-11-02
add logging for the time to check the build script for updates

jakemac53:
This is taking up a good amount of time actually (~300ms on each build for angular_components_example).

I will file a separate issue to speed this up (incremental builds shouldn't really need to do this at all, they can just look at the file changes coming from the watcher).


jakemac53:
added https://github.com/dart-lang/build/issues/565

PR #569, opened 2017-11-02
Update build_test dependency from build_barback

natebosch:


natebosch:
This should unbork the build_barback part of Travis. Other stuff still won't work.

PR #570, opened 2017-11-03
Fix build_compilers deps

natebosch:


natebosch:
This should unblock publishing, and along with the fix in build_barback should make Travis green

natebosch:
build_compilers isn't published yet. This is resolving the deps before publishing 0.0.1

Issue #566, opened 2017-11-02
Assert during build in checked mode


kevmoo:
w/ json_serializable at https://github.com/dart-lang/json_serializable/commit/1ebc3799adeb396bb03c22cfd4131954d3b423d7

```
~ json_serializable/json_serializable/ > dart --checked tool/build.dart
[INFO] BuildDefinition: Reading cached asset graph completed, took 28ms
[INFO] BuildDefinition: Building new asset graph completed, took 57ms
[INFO] BuildDefinition: Recording initial build script state completed, took 313ms
[INFO] Build: Checking for stale files completed, took 3ms
[SEVERE] Build: Failed after 8196ms
'package:build_runner/src/generate/build_impl.dart': Failed assertion: line 325 pos 9: 'inputNode.primaryOutputs.containsAll(builderOutputs)': json_serializable|test/test_files/json_literal.g.dart missing primary outputs json_serializable|test/test_files/json_literal.g.non_nullable.dart, json_serializable|test/test_files/json_literal.g.nullable.dart
package:build_runner  BuildImpl.run
```

PR #573, opened 2017-11-03
Fix asset graph structure bug with `writeToCache == false` and pre-existing outputs

jakemac53:
fixes https://github.com/dart-lang/build/issues/566

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -101,16 +101,16 @@ class AssetGraph {
   /// `primaryOutput`s.
   ///
   /// Returns a [Set<AssetId>] of all removed nodes.
-  Set<AssetId> _remove(AssetId id, {Set<AssetId> removedSet}) {
-    removedSet ??= new Set<AssetId>();
+  Set<AssetId> _remove(AssetId id, {Set<AssetId> removedIds}) {
+    removedIds ??= new Set<AssetId>();
     var node = get(id);
-    if (node == null) return removedSet;
-    removedSet.add(id);
+    if (node == null) return removedIds;
+    removedIds.add(id);
     for (var output in node.primaryOutputs) {
-      _remove(output, removedSet: removedSet);
+      _remove(output, removedIds: removedIds);
     }
     _nodesByPackage[id.package].remove(id.path);
-    return removedSet;
+    return removedIds;
   }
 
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -97,8 +97,21 @@ class AssetGraph {
     }));
   }
 
-  /// Removes the node representing [id] from the graph.
-  AssetNode _remove(AssetId id) => _nodesByPackage[id.package].remove(id.path);
+  /// Removes the node representing [id] from the graph, and all of its
+  /// `primaryOutput`s.
+  ///
+  /// Returns a [Set<AssetId>] of all removed nodes.
+  Set<AssetId> _remove(AssetId id, {Set<AssetId> removedSet}) {
```
natebosch: [nit] how about calling it `removed` or `removedIds` since the type is already `Set`

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -262,23 +272,33 @@ class AssetGraph {
...
+    return removed;
   }
 
   @override
   String toString() => allNodes.toList().toString();
 
   // TODO remove once tests are updated
   void add(AssetNode node) => _add(node);
-  AssetNode remove(AssetId id) => _remove(id);
+  Set<AssetId> remove(AssetId id) => _remove(id);
```
natebosch: Are we going to have to keep this? May as well drop the private version if we have an exact forwarding public method.

build_runner/lib/src/asset_graph/graph.dart
natebosch: Could we get away without communicating through an argument if we make this return Iterable and use `sync*`?

jakemac53: done

jakemac53: If we return an iterable then this won't actually run unless somebody reads from the iterable.

build_runner/lib/src/asset_graph/graph.dart
jakemac53: See the TODO above, at some point I would like to delete these, it just isn't a priority right now.

jakemac53:
You could consider making a goldens test in the e2e_example dir or test/fixtures dir, but it might be premature and just cause a lot of churn.

PR #574, opened 2017-11-03
use the pool package to limit open file handles

jakemac53:
This does not cause any noticeable regression in build times on angular_components_example.

Fixes https://github.com/dart-lang/build/issues/572

### DIFF
--- a/build_runner/lib/src/asset/file_based.dart
+++ b/build_runner/lib/src/asset/file_based.dart
@@ -14,7 +14,7 @@ import '../package_graph/package_graph.dart';
 import 'reader.dart';
 import 'writer.dart';
 
-/// Pool for async file writes, we don't want to use too many file handles.
+/// Pool for async file operations, we don't want to use too many file handles.
 final _descriptorPool = new Pool(32);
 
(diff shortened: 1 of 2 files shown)


jakemac53:
Actually, this looks like it results in a pretty consistent 5-10% speedup. I will play around a bit with other numbers (the 32 number comes from pub).

build_runner/lib/src/asset/file_based.dart
```diff
@@ -8,11 +8,15 @@ import 'dart:io';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 import 'package:path/path.dart' as path;
+import 'package:pool/pool.dart';
 
 import '../package_graph/package_graph.dart';
 import 'reader.dart';
 import 'writer.dart';
 
+/// Pool for async file writes, we don't want to use too many file handles.
```
natebosch: [nit] "Pool for async file operations"

This also wraps reads

jakemac53: done

jakemac53:
Going to stick with 32 for now, I do see very slight slowdowns if I reduce the number a lot (although even with a pool size of 1 it is still faster than it was before).

PR #575, opened 2017-11-03
Outline binary to generate and run build script

natebosch:
This script does not do anything useful yet - demonstrate finding
builders, writing a script to disk, and running it with `dart`.

- Add a binary ad bin/serve.dart
- Create a PackageGraph and find packages which have a `build.yaml`
- Write a script to `.dart_tool/build/build.dart` and execute it

### DIFF
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -1,0 +1,41 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:build_runner/build_runner.dart';
+import 'package:build_config/build_config.dart';
+
+const _scriptDir = '.dart_tool/build';
+const scriptLocation = '$_scriptDir/build.dart';
+
+Future<Null> ensureBuildScript() async {
+  var scriptFile = new File(scriptLocation);
+  // TODO - how can we invalidate this?
+  if (scriptFile.existsSync()) return;
+  var scriptDir = new Directory(_scriptDir);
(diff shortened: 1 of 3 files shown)


natebosch:
Super trivial for now - I figure I can move in very small steps and identify problems as early as possible.

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -0,0 +1,41 @@
...
+import 'package:build_config/build_config.dart';
+
+const _scriptDir = '.dart_tool/build';
+const scriptLocation = '$_scriptDir/build.dart';
+
+Future<Null> ensureBuildScript() async {
+  var scriptFile = new File(scriptLocation);
+  // TODO - how can we invalidate this?
+  if (scriptFile.existsSync()) return;
+  var scriptDir = new Directory(_scriptDir);
```
jakemac53: nit: you can replace this line and the next with `scriptFile.createSync(recursive: true)`.

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -0,0 +1,41 @@
...
+  if (!scriptDir.existsSync()) scriptDir.createSync(recursive: true);
+  await scriptFile.writeAsString(await _generateBuildScript());
+}
+
+Future<String> _generateBuildScript() async {
+  var result = new StringBuffer();
+  var packageGraph = new PackageGraph.forThisPackage();
+  var buildConfigs =
+      await Future.wait(packageGraph.orderedPackages.map(_packageBuildConfig));
+  result.writeln('void main() {');
```
jakemac53: Should we use code_builder? If so it might make sense to start that sooner than later.

natebosch: I'm not sure - we should definitely investigate it.

build_runner/lib/src/build_script_generate/build_script_generate.dart

PR #578, opened 2017-11-06
Upgrade to code_builder 2.1

natebosch:
Adds `Library` and an alias for `File` which is a little less awkward to
use along with `dart:io`.


natebosch:
Ah, maybe we shouldn't do this upgrade yet. `code_builder` only supports the 2.0 SDK at this version and we still support the stable SDK...

jakemac53:
Oh, hmmm. We are going to have to only support the 2.x sdk at some point anyways... not sure how I feel about that.

natebosch:
Yeah I think we'll have to be 2.x only at some point but I don't think it makes sense to add the restriction any sooner than necessary. Will hold off on this for now.

PR #579, opened 2017-11-06
Add integration tests for build script updates, and fix existing bug

jakemac53:
I also broke out a bunch of existing utils in the integration test into a re-usable library, most of the changes here are relating to that.


e2e_example/test/common/utils.dart
```diff
@@ -0,0 +1,178 @@
...
+Directory _toolDir = new Directory(p.join('.dart_tool', 'build'));
+
+Process _process;
+Stream<String> _stdErrLines;
+Stream<String> _stdOutLines;
+
+/// Runs the build script in this package, and waits for the first build to
+/// complete.
+///
+/// By default, this ensures a brand new build by deleting the `.dart_tool`
```
natebosch: [optional] deleting something by default doesn't match typical user behavior or my expectations (even if it's more commonly what we'd do in an integration test). Consider reversing the default.

e2e_example/test/common/utils.dart
```diff
@@ -0,0 +1,178 @@
...
+    _stdErrLines.listen((line) => print('StdErr: $line'));
+  }
+
+  await Future.wait(extraExpects.map((cb) async => await cb()));
+  await nextSuccessfulBuild;
+}
+
+/// Kills the current build script.
+///
+/// By default this will clean up the `.dart_tool` directory, but you can
```
natebosch: same as above

e2e_example/test/common/utils.dart
```diff
@@ -0,0 +1,178 @@
...
+      .split('\n')
+      .where((line) => line.startsWith('??'))
+      .map((line) => line.split(' ')[1])
+      .map((path) {
+    var parts = p.split(path);
+    assert(parts[0] == 'e2e_example');
+    return p.joinAll(parts.skip(1));
+  });
+  for (var path in untracked) {
+    Process.runSync('rm', [path]);
```
natebosch: can we use `git clean` command for this?

e2e_example/test/common/utils.dart
jakemac53: done

e2e_example/test/common/utils.dart
jakemac53: done

e2e_example/test/common/utils.dart
jakemac53: ah nice, didn't know about that one, done

jakemac53:
pub.dartlang.org is down which is causing the travis failure, it passes locally though so I am going to go ahead and submit this.

Issue #565, opened 2017-11-02
Speed up the build script update checks for incremental builds


jakemac53:
Today it computes the digests for each incremental build, which takes ~300ms on the angular_components_example app. This doesn't scale with overall app size, just size of the build script itself, but it would still be good to get it faster.  

One obvious option would be just using the updates from the watcher to check if any files were invalidated for incremental builds.

PR #580, opened 2017-11-06
Speed up build script invalidation, and reduce duplicate code.

jakemac53:
We now track the assets that are part of the build script and compare
those against any known updates coming from the already established
patterns. This reduces the time to check for updates to the
build script to typically report as 0ms (its re-using all the work we already did instead of redoing it).

Fixes https://github.com/dart-lang/build/issues/565


build_runner/lib/src/changes/build_script_updates.dart
```diff
@@ -9,61 +9,65 @@ import 'package:build/build.dart';
...
-import '../asset/reader.dart';
 import '../asset_graph/graph.dart';
 import '../generate/options.dart';
 
 class BuildScriptUpdates {
-  final AssetGraph _graph;
-  final String _rootPackage;
-  final DigestAssetReader _reader;
-  final _logger = new Logger('BuildScriptUpdates');
+  final Set<AssetId> _allAssetIds;
```
natebosch: This isn't very descriptive to me. How about `buildScriptAssets` or `buildScriptDependencies`?

jakemac53: This is already in a `BuildScriptUpdates` class so I think that context already exists, but I did change it to `_allSources` which should make it more clear?

natebosch: `_allSources` SGTM

PR #582, opened 2017-11-07
run tests in verbose mode, and try to ensure we shut down build servers

jakemac53:


jakemac53:
@natebosch pushed another commit which should hopefully resolve the actual flaky tests

Issue #353, opened 2017-08-09
Generate a `build.dart` automatically


natebosch:
Will need to consider dependencies, build.yaml, etc.

matanlurey:
Note from #429 - let's make sure that we can re-run the script when the generator changes.

natebosch:
After https://github.com/dart-lang/build/pull/754 and https://github.com/dart-lang/build/pull/750 and a few other pulls this is now working.

PR #583, opened 2017-11-07
Generate a real serve script

natebosch:
Towards #353

- Instantate builders discovered in `build.yaml` and apply them to
  direct dependencies
- Call `watch` from `build_runner` and set up `serve` from `shelf`
- Always regenerate the script for now until we figured out a better
  appraoch for invalidation

Open issues:
- Does not run DDC
- Does not honor manually configured `build.yaml` to apply a builder
- Does not handle Builder arguments

### DIFF
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -61,8 +61,6 @@ Method _findBuildActions(Iterable<BuildConfig> buildConfigs) =>
           ..addAll(_addBuildActions(buildConfigs))
           ..add(refer('actions').returned.statement))));
 
-//Expression _instantiatedBuilders(Iterable<BuildConfig> buildCongifs) => null;
-
 Method _main() => new Method((b) => b
   ..name = 'main'
   ..modifier = MethodModifier.async
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -15,29 +15,136 @@ const scriptLocation = '.dart_tool/build/build.dart';
 Future<Null> ensureBuildScript() async {
   var scriptFile = new io.File(scriptLocation);
   // TODO - how can we invalidate this?
-  if (scriptFile.existsSync()) return;
+  //if (scriptFile.existsSync()) return;
   scriptFile.createSync(recursive: true);
   await scriptFile.writeAsString(await _generateBuildScript());
 }
 
+final _packageGraph = new TypeReference((b) => b
+  ..symbol = 'PackageGraph'
+  ..url = 'package:build_runner/build_runner.dart');
+
+final _buildAction = new TypeReference((b) => b
+  ..symbol = 'BuildAction'
+  ..url = 'package:build_runner/build_runner.dart');
+
+final _buildActions = new TypeReference((b) => b
+  ..symbol = 'List'
+  ..types.add(_buildAction));
+
 Future<String> _generateBuildScript() async {
   var packageGraph = new PackageGraph.forThisPackage();
   var buildConfigs =
-      await Future.wait(packageGraph.orderedPackages.map(_packageBuildConfig));
-  var printStatements = buildConfigs
-      .where((config) => config.builderDefinitions.isNotEmpty)
-      .map((config) => refer('print').call([
-            literalString('I would run: \'${config.packageName} - '
-                '${config.builderDefinitions.keys.toList()}')
-          ]).statement);
-  final library = new File((b) => b.body.addAll([
-        new Method.returnsVoid((b) => b
-          ..name = 'main'
-          ..body = new Block((b) => b..statements.addAll(printStatements)))
-      ]));
-  final emitter = new DartEmitter();
+      (await Future.wait(packageGraph.orderedPackages.map(_packageBuildConfig)))
+          .where((c) => c.builderDefinitions.isNotEmpty);
+  final library = new File(
+      (b) => b.body.addAll([_findBuildActions(buildConfigs), _main()]));
+  final emitter = new DartEmitter(new Allocator.simplePrefixing());
   return new DartFormatter().format('${library.accept(emitter)}');
 }
+
+Method _findBuildActions(Iterable<BuildConfig> buildConfigs) =>
+    new Method((b) => b
+      ..name = '_buildActions'
+      ..requiredParameters.add(new Parameter((b) => b
+        ..name = 'graph'
+        ..type = _packageGraph))
+      ..returns = _buildActions
+      ..body = new Block((b) => b
+        ..statements.addAll([
+          _buildActions.newInstance([]).assignVar('actions').statement,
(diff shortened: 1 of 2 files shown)


natebosch:
The generated script for a package with a single dependency exposing a Builder:

```dart
import 'package:build_runner/build_runner.dart' as _1;
import 'package:some_builder/build/some_builder.dart' as _2;
import 'package:shelf/shelf_io.dart' as _3;

List<_1.BuildAction> _buildActions(_1.PackageGraph graph) {
  var actions = new List<_1.BuildAction>();
  var args = <String>[];
  var buildersForsome_builder = [_2.someBuilder([])];
  graph.dependentsOf('some_builder').map((n) => n.name).forEach((p) => actions
      .addAll(buildersForsome_builder.map((b) => new _1.BuildAction(b, p))));
  return actions;
}

main() async {
  var actions = _buildActions(new _1.PackageGraph.forThisPackage());
  var handler = await _1.watch(actions, writeToCache: true);
  var server = await _3.serve(handler.handlerFor('web'), 'localhost', 8000);
  await handler.buildResults.drain();
  await server.close();
}
```

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -15,30 +15,135 @@ const scriptLocation = '.dart_tool/build/build.dart';
 Future<Null> ensureBuildScript() async {
   var scriptFile = new io.File(scriptLocation);
   // TODO - how can we invalidate this?
-  if (scriptFile.existsSync()) return;
+  //if (scriptFile.existsSync()) return;
   scriptFile.createSync(recursive: true);
   await scriptFile.writeAsString(await _generateBuildScript());
```
jakemac53: can we wrap this in a `logTimedAsync` so we can see how long its taking? If its fast I am actually fine w/doing it on startup for every build. Since we are using content hashes now it should still be valid for incremental builds.

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -15,30 +15,135 @@ const scriptLocation = '.dart_tool/build/build.dart';
...
 
+final _packageGraph = new TypeReference((b) => b
+  ..symbol = 'PackageGraph'
+  ..url = 'package:build_runner/build_runner.dart');
+
+final _buildAction = new TypeReference((b) => b
+  ..symbol = 'BuildAction'
+  ..url = 'package:build_runner/build_runner.dart');
+
+final _buildActions = new TypeReference((b) => b
```
jakemac53: It is a bit confusing how there is also a method called `_buildActions`, maybe call this `_buildActionsType`?

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -15,30 +15,135 @@ const scriptLocation = '.dart_tool/build/build.dart';
...
+      ..returns = _buildActions
+      ..body = new Block((b) => b
+        ..statements.addAll([
+          _buildActions.newInstance([]).assignVar('actions').statement,
+          // TODO - Pass actual arguments
+          literalList([], new TypeReference((b) => b..symbol = 'String'))
+              .assignVar('args')
+              .statement,
+        ]
+          ..addAll(_addBuildActions(buildConfigs))
```
jakemac53: it is a bit hard to tell what this corresponds to, I would consider using regular function syntax and removing some of the cascades.

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -15,30 +15,135 @@ const scriptLocation = '.dart_tool/build/build.dart';
...
+      refer('server').property('close').call([]).awaited.statement
+    ])));
+
+Iterable<Code> _addBuildActions(Iterable<BuildConfig> configs) {
+  var statements = <Code>[];
+  for (var config in configs) {
+    var varName = 'buildersFor${config.packageName}';
+    statements.addAll([
+      _packageBuilders(config, varName),
+      refer('graph')
```
jakemac53: consider naming this `packageGraph` to disambiguate

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -15,30 +15,135 @@ const scriptLocation = '.dart_tool/build/build.dart';
...
+          .property('dependentsOf')
+          .call([literalString(config.packageName)])
+          .property('map')
+          .call([
+            new Method((b) => b
+              ..requiredParameters.add(new Parameter((b) => b..name = 'n'))
```
jakemac53: nit: n => p?

I think the fact that its a `package` is more relevant than the fact that its a `node`.

build_runner/lib/src/build_script_generate/build_script_generate.dart
natebosch: It's not fast. 2-3 seconds as currently implemented. I'm not sure why it's so slow.

natebosch: I'm going to hold off for another commit for this.

build_runner/lib/src/build_script_generate/build_script_generate.dart
natebosch: I split out the TypeReference definitions into a separate file imported with a prefix of `types`.

build_runner/lib/src/build_script_generate/build_script_generate.dart
jakemac53: sure, that is fine w/me

build_runner/lib/src/build_script_generate/build_script_generate.dart

build_runner/lib/src/build_script_generate/build_script_generate.dart

build_runner/lib/src/build_script_generate/build_script_generate.dart
natebosch: done

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -10,35 +10,131 @@ import 'package:build_runner/build_runner.dart';
 import 'package:code_builder/code_builder.dart';
 import 'package:dart_style/dart_style.dart';
 
+import 'types.dart' as types;
+
 const scriptLocation = '.dart_tool/build/build.dart';
 
 Future<Null> ensureBuildScript() async {
   var scriptFile = new io.File(scriptLocation);
   // TODO - how can we invalidate this?
```
jakemac53: do we have a bug for this?

natebosch: no, it's track in an internal design doc but not on github.

should we have a separate milestone for the script generation or keep putting them in the M0 milestone? We'll probably start collecting quite a few.

jakemac53: I think we should keep them under M0, since it does fall under that

natebosch: https://github.com/dart-lang/build/issues/585

PR #584, opened 2017-11-07
Track inputs for GeneratedAssetNodes

jakemac53:
This fixes a bug where the set of dependencies for a given generated asset could never be reduced except if those files were deleted.

It also partially unblocks https://github.com/dart-lang/build/issues/371.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -411,12 +411,38 @@ class BuildImpl {
       SingleStepReader reader, AssetWriterSpy writer) async {
     // Reset the state for each output, setting `wasOutput` to false for now
     // (will set to true in the next loop for written assets).
+    //
+    // Also updates the `inputs` set for each output, and the `outputs` sets for
+    // all inputs.
     for (var output in declaredOutputs) {
-      (_assetGraph.get(output) as GeneratedAssetNode)
+      var node = _assetGraph.get(output) as GeneratedAssetNode;
+      node
         ..needsUpdate = false
         ..wasOutput = false
         ..lastKnownDigest = null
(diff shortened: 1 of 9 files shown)


build_runner/test/generate/build_test.dart
```diff
@@ -473,94 +475,38 @@ void main() {
...
-        'a|lib/b.txt.copy': 'b',
-        'a|lib/c.txt': 'c',
-        'a|$assetGraphPath': JSON.encode(graph.serialize()),
-      }, outputs: {
-        'a|web/a.txt.copy': 'a',
-        'a|lib/c.txt.copy': 'c',
-      }, writer: writer);
-    });
-
-    test('invalidates generated assets based on graph age', () async {
```
jakemac53: note that I deleted this test, it wasn't applicable any more and should have been deleted during the migration to content hashes

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -411,12 +411,38 @@ class BuildImpl {
...
         ..needsUpdate = false
         ..wasOutput = false
         ..lastKnownDigest = null
         ..globs = reader.globsRan.toSet();
+
+      // Update dependencies that don't exist any more.
+      var removedInputs = node.inputs.difference(reader.assetsRead);
+      node.inputs.removeAll(removedInputs);
+      for (var input in removedInputs) {
+        // TODO: special type of dependency here? This means the primary input
```
natebosch: IMO it's fine to always treat the primary input like it was read. The edge cases shouldn't break anything, at worst builds are slightly less optimal

jakemac53: ya, thats pretty much my same line of thinking for now.

Issue #585, opened 2017-11-07
Figure out how and when to invalidate the generated `build.dart`


natebosch:
For now we'll just always rebuild - but it might not be very efficient.

The script is generated without knowledge of the AssetGraph or content hashes. The inputs are the PackageGraph and `build.yaml` files (along with build_runner behavior) - perhaps we can write a version of those to disk and parse it - but it's also not clear how much of the time goes in to recomputing those vs generating the code...

jakemac53:
@natebosch this looks like it is fast enough now that we can just regenerate it each time and rely on the normal build script invalidation to catch updates to it?

natebosch:
Reopening - when we shifted more of the `build.yaml` parsing from script generation time to build time this file got a lot more stable. I think we can reasonably check the `builders` section of the local or path dependency `build.yaml` files, along with the hosted versions of other packages.

natebosch:
@jakemac53 - this changed a lot after moving to snapshots. Think there is anything more to do here? There might be a UX improvement if we more eagerly invalidate the snapshot based on a hash of some information rather than waiting until it runs and figures out on it's own that it needs to be regenerated...

jakemac53:
I think we can close this, this existing workflow is pretty decent and this shouldn't need to be invalidated to often.

PR #586, opened 2017-11-07
Log time to generate build script

natebosch:
Gathering data for #585 

- Move default log listener into its own library and make it public
- Set up logging in the entrypoint
- Surround generating the build script with `logTimedAsync`


natebosch:
It's actually not as slow as I thought, ~350ms for a very small build. We can wait and see how it scales.

PR #592, opened 2017-11-08
Allow the generated entrypoint in the AssetGraph

### LINKED ISSUE #587: Incremental builds don't work with build script in `.dart_tool`
We exclude `.dart_tool/` from the AssetGraph and since the build script is there it can't be used to detect build script changes so incremental builds are not attempted.

We could potentially always include the script entrypoint in the asset graph (override the exclude for just that file) but it would mean relative imports don't work to other files in `.dart_tool/` - this is probably an acceptable tradeoff since we control the script contents.

natebosch:
Fixes #587

Moves the entrypoint into a subdirectory in case we end up needing to
generate multiple files with relative imports. Add an exception for the
entrypoint directory to allow those files into the asset graph.

### DIFF
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -12,9 +12,10 @@ import 'package:dart_style/dart_style.dart';
 import 'package:logging/logging.dart';
 
 import '../logging/logging.dart';
+import '../util/constants.dart';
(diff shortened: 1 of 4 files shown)


natebosch:
Added a test that ensures it shows up in the AssetGraph.

natebosch:
😢 

Now we're generating files against the entrypoint
https://github.com/dart-lang/build/issues/594

PR #595, opened 2017-11-09
Exclude tool directories by default

### LINKED ISSUE #594: Don't use the generated entrypoint as a source

natebosch:
Fixes #594

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -202,16 +202,9 @@ class _Loader {
     var inputSets = _options.packageGraph.allPackages.values.map((package) =>
         new InputSet(package.name, package.includes,
             excludes: package.excludes));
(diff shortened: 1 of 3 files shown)


natebosch:
It sucks that we need to hardcode this behavior in different places...

Technically we could limit this to exclude only `entryPointDir` here since that is the only one that leaks into the AssetGraph now, but excluding all of `toolDirs` is slightly safer against other ways we might break this in the future.

natebosch:
Ok I made this a bit deeper overhaul.

- Drop `_isValidInput` since it's now redundant against the default excludes on `InputSet` along with the fact that we were already limiting to `package.includes` which covers the `lib/**` case.
- Allow the `InputSet` to filter out the entrypoint directory (since we never want to consider that anywhere else) and add it back in explicitly

jakemac53:
LGTM, we should add a regression test for this as well though

PR #597, opened 2017-11-09
Add builder factories for build_compilers

natebosch:
This will make the usage of build_compilers look more like other places
using `build.yaml` and may allow us to configure it similarly in the
future.

The pattern I'd like to establish is for builder factories to go in
`lib/build/builders.dart`.

Add const constructors for all the builders since they have no state.

### DIFF
--- a/build_compilers/pubspec.yaml
+++ b/build_compilers/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_compilers
-version: 0.0.2-dev
+version: 0.1.0-dev
 description: Builder implementations wrapping Dart compilers.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build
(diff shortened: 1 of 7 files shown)


build_compilers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_compilers
-version: 0.0.1
+version: 0.0.2-dev
```
jakemac53: Why the breaking change bump?

natebosch: uhh, I guess cause I'm bad at semver...

Should this be a +1?

natebosch: I think it's cause I normally don't see `+1-dev` since we most often use plus releases for "this is something small I'll release right away"

jakemac53: I would also be fine with a 0.1.0

PR #606, opened 2017-11-14
Use a combined inputs hash to check if we need to run a build

jakemac53:
This was more nuanced than I expected, and I ended up needing to clean up a few other things in the process. Highlights are as follows:

- Changed summaries to semantic summaries (we were actually relying on the full ones before, whoops!)
- Fixed some behavior around replacing `SyntheticAssetNode`s with real ones. Previously we would remove all the outputs from the graph and clean up their inputs sets which wasn't correct (we want to retain all those edges and nodes, and just swap out the node in question).
- When invalidating nodes, we no longer pre-emptively delete them. In fact, I never explicitly delete them in this cl I just rely on the asset writer to overwrite them (unless their primary input is deleted).
- The inputs set for `GeneratedAssetNode`s is now ordered to ensure the combined md5 hash will always be the same for the same inputs.

Fixes https://github.com/dart-lang/build/issues/371

### DIFF
--- a/build_compilers/lib/src/summary_builder.dart
+++ b/build_compilers/lib/src/summary_builder.dart
@@ -76,7 +76,7 @@ Future createUnlinkedSummary(Module module, BuildStep buildStep,
   request.arguments.addAll([
     '--build-summary-only',
     '--build-summary-only-unlinked',
-    '--build-summary-output=${summaryOutputFile.path}',
+    '--build-summary-output-semantic=${summaryOutputFile.path}',
     '--strong',
   ]);
 
@@ -117,17 +117,17 @@ Future createLinkedSummary(Module module, BuildStep buildStep,
   var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
 
   var allAssetIds = new Set<AssetId>()
+    // TODO: Why can't we just add the unlinked summary?
+    // That would help invalidation.
     ..addAll(module.sources)
     ..addAll(transitiveLinkedSummaryDeps)
     ..addAll(transitiveUnlinkedSummaryDeps);
   await scratchSpace.ensureAssets(allAssetIds, buildStep);
   var summaryOutputFile = scratchSpace.fileFor(module.linkedSummaryId);
   var request = new WorkRequest();
-  // TODO(jakemac53): Diet parsing results in erroneous errors in later steps,
-  // but ideally we would do that (pass '--build-summary-only-diet').
   request.arguments.addAll([
     '--build-summary-only',
-    '--build-summary-output=${summaryOutputFile.path}',
+    '--build-summary-output-semantic=${summaryOutputFile.path}',
     '--strong',
   ]);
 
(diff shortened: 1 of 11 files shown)


build_compilers/lib/src/summary_builder.dart
```diff
@@ -76,7 +76,7 @@ Future createUnlinkedSummary(Module module, BuildStep buildStep,
   request.arguments.addAll([
     '--build-summary-only',
     '--build-summary-only-unlinked',
-    '--build-summary-output=${summaryOutputFile.path}',
+    '--build-summary-output-semantic=${summaryOutputFile.path}',
```
natebosch: At least a few of these changes feel like they could be justified in isolation and should be relatively easy to move to a new PR.

Can we submit this one separately?

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -66,7 +66,9 @@ class AssetGraph {
     var existing = get(node.id);
     if (existing != null) {
       if (existing is SyntheticAssetNode) {
-        _remove(existing.id);
+        // Don't call _remove, that transitively removes primary outputs. We
```
natebosch: can we rename to `_removeSubtree` or something like that to make this clear?

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -74,13 +76,26 @@ class GeneratedAssetNode extends AssetNode {
...
   /// to generate it.
-  final Set<AssetId> inputs;
+  ///
+  /// This needs to be an ordered set because we compute combined input digests
+  /// using this later on.
+  final SplayTreeSet<AssetId> inputs;
+
+  /// A digest combining all digests of all previous inputs.
+  ///
+  /// This is used to determine if a node really needs to be output or not.
```
natebosch: [optional] I might rephrase this, the `Builder` still decides whether to output an asset, we decide whether to run the build.

"Used to determine whether all the inputs to a build step are identical to the previous run indicating that the previous output is still valid."

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -329,14 +339,20 @@ class BuildImpl {
...
         _reader,
         _assetGraph,
         phaseNumber,
         input.package,
         (phase, input) => _runLazyPhaseForInput(phase, input, resourceManager));
+
+    if (!await _buildShouldRun(builderOutputs, wrappedReader)) {
+      return <AssetId>[];
+    }
+    // We may have read some inputs in the call to `_buildShouldRun`, we want
```
natebosch: should we not be going through the wrapped reader if we don't want to track this?

Or is that the only way for the lazy assets to work?

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -443,6 +515,10 @@ class BuildImpl {
         assert(inputNode != null, 'Asset Graph is missing $input');
         inputNode.outputs.add(output);
       }
+
+      // And finally compute the combined digest for all inputs.
```
natebosch: some of these comments are "what" with no "why" - should we be refactoring to make the code easier to understand?

build_compilers/lib/src/summary_builder.dart
kevmoo: Do we get a speedup w/ this?

jakemac53: I will break this out into its own cl and test in isolation, its mostly about invalidating later build steps but it should also improve the time to read and write summaries

build_runner/lib/src/generate/build_impl.dart
jakemac53: broke out some parts of this into appropriately named functions so its easier to follow

PR #608, opened 2017-11-14
Prepare to publish build_config

natebosch:
- Copy the LICENSE into this directory since pub uses it during publish
- Add a CHANGELOG
- Drop the `-dev` version


build_config/LICENSE
```diff
@@ -0,0 +1,26 @@
+Copyright 2016, the Dart project authors. All rights reserved.
```
jakemac53: should we update this to 2017?

natebosch: Done

PR #612, opened 2017-11-15
Replace orderedPackages with orderedPackageCycles

### LINKED ISSUE #601: Decide what order BuildActions should run in - builder preference or package preference
With our manual build script we have been preferring to apply builders across packages, then move on to the next builder.
For example: https://github.com/dart-lang/angular/blob/6fc43bb170683432f4532877d54abb3ccbc676c6/_tests/tool/build.dart#L76-L80

If we have the builders `[b1, b2]` then `b1` will run everywhere before `b2` runs and can never read any file that *any* `b2` produces - including from dependencies.

This is different from pub and different from bazel. In both of these places builders run in order within a target, but assume they can read any asset produced by later builders in dependencies.

Which behavior should we prefer?

In order to have DDC operate while allowing package cycles we need to use this behavior.

natebosch:
Closes #601

It isn't be feasible to require fully non-cyclic modules specified
statically. Builders will run in package-order generally, but within a
cycle they will run in builder order. This ordering most closely matches
bazel with some magic happenign only within package cycles.

- Add a utility `stronglyConnectedComponents` implementing Tarjan's
  algorithm.
- Deprecate `orderedPackages` and `dependentsOf` since the new approach
  with better information will be preferred.
- Replace previous `orderedPackages` implementaiton with a view on top
  of the ordered cycles.
- Add top-level method `applyBuilders` to make it easier to get the
  `BuildActions` necessary to run builders with our newly decided
  ordering.


natebosch:
Hmm, I just realized that `applyBuilders` still isn't quite going to do what we want since we'd still apply all the filtered builders, and then the non-filtered builders.

In practice this probably isn't going to matter, and is closer to what we want than where we are today, but we probably need a different signature to handle this fully.

I think what we will need is a signature like:

```dart
class BuilderApplication {
  final Builder builder;
  final String filterToDependentsOf;
}


List<BuildAction> applyBuilders(
    List<List<PackageNode>> packages, List<BuilderApplication> builders) {}
```

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -133,6 +135,7 @@ class PackageGraph {
   ///
   /// See [orderedPackages] for ordering guarantees. The node for [packageName]
   /// will not be included in the result.
+  @Deprecated('Use `applyBuilders` to filter `orderedPackageCycles`')
```
kevmoo: Mention deprecations in the changelog...

build_runner/lib/src/package_graph/dependency_ordering.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'dart:math' show min;
+
+/// Finds the strongly connected components of an ordered graph using Tarjan's
+/// algorithm.
+///
+/// [V] is the type of values in the graph vertices. [K] must be a type suitable
+/// for using as a Map or Set key, and [key] must provide a consistent key for
+/// ever vertex. [successors] should return the children for a given vertex.
+///
+/// [nodes] must contain at least a root of every tree in the graph if there are
```
jakemac53: Consider renaming `nodes` to `rootNodes` or something that indicates better what is required.

Or, just make this take a single root node. Does that not work for all our use cases?

build_runner/lib/src/package_graph/dependency_ordering.dart
```diff
@@ -0,0 +1,57 @@
...
+/// disjoint subgraphs but it may contain all nodes.
+List<List<V>> stronglyConnectedComponents<K, V>(
+    Iterable<V> nodes, K Function(V) key, Iterable<V> Function(V) successors) {
```
jakemac53: nit: `children` instead of `successors`? that is the more common term afaik.

build_runner/lib/src/package_graph/dependency_ordering.dart
```diff
@@ -0,0 +1,57 @@
...
+  final result = <List<V>>[];
+  final lowLinks = <K, int>{};
+  final indexes = <K, int>{};
+  final onStack = new Set<K>();
+
+  var index = 0;
+  var toVisit = new Queue<V>();
+
+  void strongConnect(V node) {
+    indexes[key(node)] = index;
```
jakemac53: nit: store `key(node)` and `key(next)` below in local variables instead of calling them repeatedly which isn't guaranteed to be cheap

build_runner/lib/src/package_graph/dependency_ordering.dart
```diff
@@ -0,0 +1,57 @@
...
+    lowLinks[key(node)] = index;
+    index++;
+    toVisit.addLast(node);
```
jakemac53: nit: The naming of this variable is weird, you are currently visiting it so adding it to `toVisit` seems wrong. Consider `lastVisited` or something like that.

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -147,8 +150,25 @@ class PackageGraph {
...
+
+  /// Collects packages into strongly connected subcomponents and orders those
+  /// components in reverse dependency order.
+  ///
+  /// A cycle is defined as a collection of packages where each package is
+  /// present in the transitive dependency graph of each other package.
+  ///
+  /// Cycles are ordered in reverse dependency order, so if package `A` depends
+  /// on package `B`, and they are not in a cycle, package `A` will exist in a
+  /// later sublist than `B`. Ordering within a cycle is non-deterministic,
```
jakemac53: We should be able to make it deterministic pretty easily by just sorting alphabetically within cycles?

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -147,8 +150,25 @@ class PackageGraph {
...
-      _orderedPackages(root, new Set<PackageNode>());
+      orderedPackageCycles.expand((c) => c);
+
+  /// Collects packages into strongly connected subcomponents and orders those
+  /// components in reverse dependency order.
+  ///
+  /// A cycle is defined as a collection of packages where each package is
+  /// present in the transitive dependency graph of each other package.
+  ///
+  /// Cycles are ordered in reverse dependency order, so if package `A` depends
```
jakemac53: `Cycles are ordered in reverse dependency order` I don't understand what you mean by this - maybe you really mean `components`? That seems duplicative with the first sentence though.

build_runner/test/package_graph/package_graph_test.dart
```diff
@@ -217,6 +217,96 @@ void main() {
     });
   });
 
+  group('orderedPackageCycles', () {
+    test('with two sub trees', () {
+      var a = new PackageNode('a', '1.0.0', PackageDependencyType.path, null);
```
jakemac53: It might be worth adding some sort of utility to create these package structures in a more visual way such as a map?

For instance this could look like:

```dart
var graph = createPackageGraph(/*name of root node*/ 'a', {
  'a': {
    'dependencies': ['left1', 'left2'],
  },
  'left1': {
    'dependencies': ['left2'],
  },
  'left2': {
    'dependencies': [],
  },
  'right1': {
    'dependencies': ['right2'],
  },
  'right2': {
    'dependencies': [],
  },
});
```

Or we could provide in memory pubspecs or something, idk just a thought :)

build_runner/test/package_graph/package_graph_test.dart
```diff
@@ -217,6 +217,96 @@ void main() {
...
+      var right1 =
+          new PackageNode('right1', '1.0.0', PackageDependencyType.pub, null);
+      var right2 =
+          new PackageNode('right2', '1.0.0', PackageDependencyType.pub, null);
+      a.dependencies.addAll([left1, right1]);
+      left1.dependencies.add(left2);
+      right1.dependencies.add(right2);
+      var graph = new PackageGraph.fromRoot(a);
+      var inOrder = graph.orderedPackageCycles.map((c) => c.map((p) => p.name));
+      expect(
```
jakemac53: nit: for all of these tests add an expect for the total length of `inOrder`.

build_runner/lib/build_runner.dart
```diff
@@ -11,5 +11,6 @@ export 'src/generate/performance_tracker.dart'
     show BuildPerformance, BuildActionPerformance;
 export 'src/generate/phase.dart';
 export 'src/package_builder/package_builder.dart' show PackageBuilder;
+export 'src/package_graph/apply_builders.dart';
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:build/build.dart';
+
+import '../generate/phase.dart';
+import 'package_graph.dart';
+
+bool _alwaysTrue(_) => true;
+
+bool Function(PackageNode) _dependencyFilter(String dependency) =>
```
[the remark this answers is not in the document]
natebosch: Done keeping it private so we don't need to expose more symbols than we already are. I also made the `filter` field private since it shouldn't necessarily be useful elsewhere and we want to focus on the named constructors.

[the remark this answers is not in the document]
natebosch: Went with `_hasDependency`

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+    (p) => p.dependencies.any((d) => d.name == dependency);
+
+bool Function(PackageNode) _isPackage(String package) =>
```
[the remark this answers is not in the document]
natebosch: Removed everywhere.

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+    (p) => p.name == package;
+
+bool Function(PackageNode) _packagesContains(Set<String> packages) =>
```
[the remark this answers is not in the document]
natebosch: How about `_isInSet`?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+    (p) => packages.contains(p.name);
+
+class BuilderApplication {
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+  final Builder builder;
+
+  /// Determines which package need [builder] applied.
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+  final bool Function(PackageNode) filter;
+
+  /// Apply [builder] to all packages in the dependency graph.
+  const BuilderApplication.allPackage(this.builder) : filter = _alwaysTrue;
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+
+  /// Apply [builder] to the pacakges which have a direct dependency on the
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+      : filter = _dependencyFilter(dependency);
+
+  /// Apply [builder] to a single package in the graph.
+  BuilderApplication.singlePackage(this.builder, String package)
+      : filter = _isPackage(package);
+
+  /// Apply [builder] to a set of explicitly listed packages.
+  BuilderApplication.onPackages(this.builder, Set<String> packages)
+      : filter = _packagesContains(packages);
+}
```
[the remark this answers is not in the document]
natebosch: I'd like to hold off until we have a concrete use case. There's already a workaround of examining the `packageGraph` manually and dumping all relevant packages into a `Set`.

As I mentioned above I made the `_filter` field private and so if we can avoid exposing details of what this does I think the named constructors are the most descriptive way to use this.

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+
+/// Creates a [BuildAction] to apply each builder in [builders] to each package
+/// in [packages] such that all builders are run for earlier subcomponents
+/// before moving on to later packages.
+///
+/// For example, given the builders `[b1, b2]` and package components
+/// `[[p1], [p2, p3], [p4]]` where `p2` and `p3` represent a dependenency cycle
```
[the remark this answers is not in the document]
natebosch: Yeah that's a good call. If we're doing that should we consider removing `orderedPackageCycles` from `PackageGraph` and keeping that internal too?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -0,0 +1,61 @@
...
+  /// Apply [builder] to a single package in the graph.
+  BuilderApplication.singlePackage(this.builder, String package)
+      : filter = _isPackage(package);
+
+  /// Apply [builder] to a set of explicitly listed packages.
+  BuilderApplication.onPackages(this.builder, Set<String> packages)
+      : filter = _packagesContains(packages);
+}
+
+/// Creates a [BuildAction] to apply each builder in [builders] to each package
```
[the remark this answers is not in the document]
natebosch: Yeah I really struggled with names here. "cycle" and "subcomponent" aren't super obvious on their own.

Will try to rephrase this taking into consideration the new signature discussed below.

build_runner/lib/src/package_graph/dependency_ordering.dart
natebosch: Was attempting to match the wikipedia article as close as possible. Same reason I used `successors` below.

I think it doesn't hurt to still take multiple roots, and the algorithm is tolerant to taking all nodes in the graph which is the general way it's used. Our specific case let's us get away with only passing one root.

build_runner/lib/src/package_graph/dependency_ordering.dart
natebosch: Done

build_runner/lib/src/package_graph/dependency_ordering.dart
natebosch: Done

build_runner/lib/src/package_graph/dependency_ordering.dart
natebosch: Done.

build_runner/lib/src/package_graph/package_graph.dart
natebosch: Done

build_runner/lib/src/package_graph/package_graph.dart
natebosch: I'm not sure it would be worthwhile to spend any time sorting. We can add that if we find a use case.

build_runner/test/package_graph/package_graph_test.dart
natebosch: definitely worth adding later. For now this was copy/pasted from our existing tests of `orderedDependencies`

build_runner/test/package_graph/package_graph_test.dart
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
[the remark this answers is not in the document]
jakemac53: Ya I think so, this new api gives you effectively the same functionality while hiding some of the internals which is good I think.

build_runner/lib/src/package_graph/package_graph.dart
jakemac53: The issue I am concerned about here is the ordering will actually have a real effect, so this makes builds non-deterministic.

Specifically if the packages try to read generated outputs from each other, for one of them `canRead` might return false. Which one that is depends on the cycle ordering.

build_runner/lib/src/package_graph/apply_builders.dart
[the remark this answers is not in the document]
jakemac53: It is easy enough to add later, so I think I am fine w/this.

We will need a `BuilderApplication.transitiveDependentsOf` I think for angular - technically anything which transitively depends on angular might need codegen because of exports and provider lists. Right @matanlurey?

build_runner/lib/src/package_graph/dependency_ordering.dart
jakemac53: I don't think matching the wiki article is important, especially in terms of variable names. For instance it uses single letter names and other things which don't match dart readability guidelines. I would do what makes the most sense in our context (Dart, package:build).

Fwiw, I also really dislike the `lowLinks` name :).

natebosch: I was also considering whether we should create a `package:graphs` to include this, and maybe `asyncCrawl` from `build_runner`. I looked at the implementation of this algorithm in `pub` and I think it'd be pretty easy to drop in this method.
I like having these types of complex-but-generally-useful things partitioned off into separate packages rather than re-implement them everywhere depending on certain details of the use case.

@kevmoo - any thoughts on whether a new package would be worth the effort? So far 2 use cases for this function, and another function that would fit nicely in the package. It would mean a new dependency in the SDK but could be worthwhile.

jakemac53: +1 for putting this in a separate package (pub uses this for dartdevc as well, not sure if that's what you were referring to)

kevmoo: are you asking for a 'graph' package?

natebosch: Yes, probably a 'graphs' package since 'graph' is already taken (but dead?)

I'll write up a 1-pager and share it with you internally.

kevmoo:
We could always ask to take it :-)

natebosch:
I realized that this still doesn't cover all use cases.

We need to add some kind of support for the `inputs` and `excludes` globs.

natebosch:
Ok, I think the new syntax covers the use cases we know about today, and it opens up a path to removing `BuildAction` from the public surface area.

Take a look at `e2e_example/tool/build.dart` to see what it looks like. I think it's pretty straightforward and it's an easy migration from the current approach.

build_runner/lib/src/package_graph/dependency_ordering.dart
natebosch: Hmm, there is an implementation of this algorithm already published in package:collection.

https://www.dartdocs.org/documentation/collection/1.14.3/collection/stronglyConnectedComponents.html

It has an API which is slightly less easy to use for this case, since we don't already have exactly a `Map<T, Iterable<T>>`. It wouldn't be quite drop in for pub either, while the signature I chose could be.

build_runner/CHANGELOG.md
```diff
@@ -3,9 +3,9 @@
 - Add an `enableLowResourcesMode` option to `build` and `watch`, which will
   consume less memory at the cost of slower builds. This is intended for use in
   resource constrained environments such as Travis.
-- Add `applyBuilders` and `PackageGraph.orderedPackageCycles`. After finding a
-  list of Builders to run, use these tools to apply them in the correct order
-  across packages.
+- Add `applyBuilders`. After finding a list of Builders to run, and defining
+  which packages need them applied, use this tools to apply them in the correct
```
jakemac53: s/tools/tool

build_runner/lib/src/generate/phase.dart
```diff
@@ -30,9 +30,9 @@ abstract class BuildAction {
   /// Runs [builder] on [package] with [inputs] as primary inputs, excluding
   /// [excludes]. Glob syntax is supported for both [inputs] and [excludes].
   factory BuildAction(Builder builder, String package,
-      {List<String> inputs = const ['**'],
-      List<String> excludes = const [],
-      bool isOptional}) {
+      {List<String> inputs, List<String> excludes, bool isOptional}) {
+    inputs ??= const ['**'];
+    excludes ??= const [];
```
jakemac53: This default should be handled by `InputSet` (since its an optional arg there and this is just forwarding it)

e2e_example/tool/build.dart
```diff
@@ -13,27 +13,18 @@ import 'package:shelf/shelf_io.dart' as shelf_io;
 
 Future main() async {
   var graph = new PackageGraph.forThisPackage();
-  var buildActions = <BuildAction>[
-    new BuildAction(new TestBootstrapBuilder(), graph.root.name,
+  var builders = <BuilderApplication>[
```
jakemac53: Technically we shouldn't need the explicit type here I don't think?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -5,59 +5,68 @@
...
-/// dependency on some other package.
-List<BuildAction> applyBuilders(Iterable<Iterable<PackageNode>> packages,
-        Iterable<BuilderApplication> builders) =>
-    packages
-        .expand((cycle) => builders.expand((b) => cycle
-            .where(b._filter)
-            .map((p) => new BuildAction(b.builder, p.name))))
+/// dependency on some other package by choosing the appropriate
+/// [BuilderApplication].
+List<BuildAction> applyBuilders(
```
jakemac53: This name doesn't make so much sense any more imo, maybe `createBuildActions` or something like that? The 2nd arg should probably be renamed `builderApplications` as well.

The main thing I think is confusing is the difference between `apply` and `applyBuilders`, it isn't clear from the names how they should be used and in what order.

build_runner/lib/src/generate/phase.dart
natebosch: Done.

e2e_example/tool/build.dart
natebosch: Done

build_runner/CHANGELOG.md
natebosch: Done

build_runner/lib/src/package_graph/apply_builders.dart
natebosch: `createBuildActions` SGTM

build_runner/CHANGELOG.md
```diff
@@ -3,8 +3,8 @@
 - Add an `enableLowResourcesMode` option to `build` and `watch`, which will
   consume less memory at the cost of slower builds. This is intended for use in
   resource constrained environments such as Travis.
-- Add `applyBuilders`. After finding a list of Builders to run, and defining
-  which packages need them applied, use this tools to apply them in the correct
+- Add `createBuildActios`. After finding a list of Builders to run, and defining
```
jakemac53: s/createBuildActios/createBuildActions

build_runner/lib/src/generate/phase.dart
```diff
@@ -31,8 +31,6 @@ abstract class BuildAction {
   /// [excludes]. Glob syntax is supported for both [inputs] and [excludes].
   factory BuildAction(Builder builder, String package,
       {List<String> inputs, List<String> excludes, bool isOptional}) {
-    inputs ??= const ['**'];
```
jakemac53: we still need this default here since its passed to a non-optional arg not forwarded to an optional one

natebosch:
Travis is failing due to other reasons. Tested with `mono_repo presubmit`.

PR #613, opened 2017-11-15
move initial asset graph updates into BuildDefinition.load

jakemac53:
- Fixes a bug where if between builds your build script changed such that it imports a new asset that didn't exist in the graph before you would get an error like `BuildScriptUpdates: e2e_example|tool/foo.dart was not found in the asset graph, incremental builds will not work. This probably means you don't have your dependencies specified fully in your pubspec.yaml`.
  - Now the asset graph is updated with new sources before that check so you won't see the error.
- Allows us to remove the updates and conflictingAssets fields from BuildDefinition which were always a bit weird.

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
 
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
@@ -30,13 +31,6 @@ final _logger = new Logger('BuildDefinition');
 class BuildDefinition {
   final AssetGraph assetGraph;
 
-  /// Assets which have changed since the cached asset graph was created.
-  final Map<AssetId, ChangeType> updates;
-
-  /// Assets which should be generated but already exist on disk and can't be
-  /// proven to be from the last build.
-  final Set<AssetId> conflictingAssets;
-
   final DigestAssetReader reader;
   final RunnerAssetWriter writer;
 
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/build_definition.dart
```diff
@@ -239,4 +246,63 @@ class _Loader {
       yield new AssetId(package, path);
     }
   }
+
+  /// Handles cleanup of pre-existing outputs for initial builds (where there is
+  /// no cached graph).
+  Future<Null> _initialBuildCleanup(Set<AssetId> conflictingAssets,
```
natebosch: one of my goals with `BuildDefinition` was to have a more clear separation between figuring out what work needs to happen and doing the work.

I'm somewhat torn on this change.

jakemac53: We are already doing a fair bit of work, because we are tracking the asset graph here. Because of that we have to check the build script for updates here (because we might have to throw away the asset graph). In order to do that properly we have to update the asset graph as well so it contains any new sources.

In terms of the `_initialBuildCleanup` I think that still makes sense here as well as its a part of setting up the initial state for the build. The state of the world is not valid until those files have been cleaned up. This also allows us to garbage collect the `updates` and `conflictingOutputs` sets which previously never could be.

We are still not doing any actual building as a part of this change, we are only setting up the state of the world such that it is ready for a build to happen.

natebosch: Should we look at a new name for `load`? Maybe `prepareWorkspace` or something?

jakemac53: Sure, I would be fine w/that

PR #614, opened 2017-11-15
break up prepareWorkspace into more targeted functions to make it easier to follow

jakemac53:
I would break this up even further except we can't return multiple values from a function and building a class just to wrap up a `BuildScriptUpdates` and an `AssetGraph` seems overkill.

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -143,11 +143,11 @@ class _Loader {
 
   /// If `_options.writeToCache` is `true` then this returns the all the sources
   /// found in the cache directory, otherwise it returns an empty set.
-  Future<Set<AssetId>> _findCacheDirSources() async {
+  Future<Set<AssetId>> _findCacheDirSources() {
     if (_options.writeToCache) {
-      return await _listGeneratedAssetIds().toSet();
+      return _listGeneratedAssetIds().toSet();
     }
-    return new Set<AssetId>();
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/generate/build_definition.dart
```diff
@@ -165,17 +128,89 @@ class _Loader {
...
+        }
+      }
+    }
+  }
+
+  /// If `_options.writeToCache` is `true` then this returns the all the sources
+  /// found in the cache directory, otherwise it returns an empty set.
+  Future<Set<AssetId>> _getCacheDirSources() async {
+    if (_options.writeToCache) {
+      return await _listGeneratedAssetIds().toSet();
```
natebosch: [nit] no need for `await` here. Also prefer a `get` method to a method prefixed wight 'get'.

Could be

```dart
Future<Set<AssetId>> get _cacheDirSources => _options.writeToCache
    ? _listGeneratedAssetIds().toSet()
    : new Future.value(new Set<AssetId>());
```

or

```dart
Future<Set<AssetId>> get _cacheDirSources async => _options.writeToCache
    ? _listGeneratedAssetIds().toSet()
    : new Set<AssetId>();
```

jakemac53: A getter doesn't fit with the style of the rest of the file, I renamed it to `_findCacheDirSources` which is more in line with `_findInputSources` and keeps the style consistent. It also indicates more that real work is being done.

Issue #559, opened 2017-11-01
Validate that current build actions match the cached graphs build actions


jakemac53:
Today you can accept command line args and change the structure of the build graph. Or, more subtly, a pub upgrade could change the package graph structure, resulting in a change to the structure of the build graph. Either of these things should result in the previous build being thrown away.

We would need to serialize the previous build actions in some form in order to verify this.

PR #615, opened 2017-11-16
Verify build actions against previous ones

jakemac53:
Currently using the `runtimeType` to verify builders - I don't know that `Object.hashCode` would be consistent across runs (or even if it is today, I don't see anything guaranteeing that will always be the case).

This technically allows you to change a builder implementation to one by the same name in a different file and we wouldn't detect that, but I think that is fine.

Closes https://github.com/dart-lang/build/issues/559

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'dart:convert';
 
 import 'package:build/build.dart';
+import 'package:convert/convert.dart';
 import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
 import 'package:watcher/watcher.dart';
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -336,3 +343,29 @@ class AssetGraph {
   void add(AssetNode node) => _add(node);
   Set<AssetId> remove(AssetId id) => _removeRecursive(id);
 }
+
+/// Computes a [Digest] for [buildActions] which can be used to compare one set
+/// of [BuildAction]s against another.
+Digest computeBuildActionsDigest(Iterable<BuildAction> buildActions) {
```
natebosch: private?

jakemac53: This is used from build_definition as well, I could move it elsewhere?

jakemac53: We could also pass the list of build actions into the `deserialize` method and do the validation there but it feels weird.

natebosch: ah fair enough. I'm fine if it's public then.

jakemac53:
cc @eernstg just as an fyi this should help with package:reflectable, you should no longer need to remove manually the `.dart_tool` directory.

jakemac53:
Travis is having issues, but I ran a local mono_repo presubmit which was green

PR #621, opened 2017-11-16
fix bug where nodes werent synchronously updated

jakemac53:
Previously there was a gap of time between when we marked a node as `wasOutput = false` and `needsUpdate = false`, and then came back later and marked `wasOuput = true` for the nodes that were really output.

Now we do all updates to nodes synchronously to ensure they are always in a valid state.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -11,6 +11,7 @@ import 'package:build/src/builder/logging.dart';
 import 'package:build_barback/build_barback.dart' show BarbackResolvers;
 import 'package:convert/convert.dart';
 import 'package:crypto/crypto.dart';
+import 'package:glob/glob.dart';
 import 'package:logging/logging.dart';
 import 'package:stack_trace/stack_trace.dart';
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -433,26 +433,26 @@ class BuildImpl {
...
-        ..needsUpdate = false
-        ..wasOutput = false
-        ..lastKnownDigest = null
-        ..globs = reader.globsRan.toSet();
 
+      var allInputs = reader.assetsRead;
+      if (node.primaryInput != null) allInputs.add(node.primaryInput);
+      inputsDigest ??= await _computeCombinedDigest(node.inputs, reader);
+
+      // **IMPORTANT**: All updates to `node` must be synchronous. With lazy
```
kevmoo: ...meaning one should change all of the properties in one go?

natebosch:
changelog?

build_runner/lib/src/generate/build_impl.dart
jakemac53: yes

PR #623, opened 2017-11-16
Add `toAny` PackageFilter

natebosch:
Usage would look like:

```dart
apply(templateCompiler(args),
    toAny([toPackage('angular'), toDependentsOf('angular')]));
```


jakemac53:
we should get some unit tests going for these as well - or at least track adding some in a bug if you want to wait until the api stabilizes

natebosch:
Filed https://github.com/dart-lang/build/issues/627

natebosch:
We had discussed offline renaming this to `toAll` so I went ahead with that name.

PR #630, opened 2017-11-17
Deprecate inputs and excludes from PackageNode

natebosch:
The only use case for these was that we were depending on them being set
appropriately for the root and non-root packages. We had previously been
handling this in two places and it was changed to only be configued in
the PackageNode. Adding it back to BuildDefinition will allow us to
deprecate and remove the fields in the PackageGraph.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Deprecations
 
 - Deprecate `PackageGraph.orderedPackages` and `PackageGraph.dependentsOf`.
+- Deprecate `PackageNdoe.includes` and `PackageGraph.excludes`.
 
 ### Internal Improvements
 
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -272,8 +272,11 @@ class _Loader {
   /// Returns the set of original package inputs on disk.
   Future<Set<AssetId>> _findInputSources() async {
     var inputSets = _options.packageGraph.allPackages.values.map((package) =>
-        new InputSet(package.name, package.includes,
-            excludes: package.excludes));
+        new InputSet(
+            package.name,
+            package.name == _options.packageGraph.root.name
+                ? const ['**']
+                : const ['lib/**']));
     var sources = (await _listAssetIds(inputSets).toSet())
       ..addAll(await _options.reader
           .findAssets(new Glob('$entryPointDir/**'))
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -271,17 +271,26 @@ class _Loader {
 
   /// Returns the set of original package inputs on disk.
   Future<Set<AssetId>> _findInputSources() async {
-    var inputSets = _options.packageGraph.allPackages.values.map((package) =>
-        new InputSet(
-            package.name,
-            package.name == _options.packageGraph.root.name
-                ? const ['**']
-                : const ['lib/**']));
+    var inputSets =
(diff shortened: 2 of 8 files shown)


natebosch:
Hmm, this was added to allow limiting what we read from the `$sdk` package... Will need to rethink this.

natebosch:
We'll also need to rethink how this works when we get around to considering modules. A 'package' shouldn't really have globs like this, but a 'module' might.

natebosch:
@jakemac53 - this is ready for review

As we discussed offline we're hardcoding the SDK and non-root globs in `build_definition.dart` for now. This was already where we were handling this until we added the SDK package for DDC.

Once we get around to looking at modules we can revisit how this works, but it probably still won't make sense to encode this information at the PackageNode level.

build_runner/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 ### Deprecations
 
 - Deprecate `PackageGraph.orderedPackages` and `PackageGraph.dependentsOf`.
+- Deprecate `PackageNdoe.includes` and `PackageGraph.excludes`.
```
jakemac53: `PackageNode` (for both)

build_runner/lib/src/generate/build_definition.dart
```diff
@@ -271,16 +271,28 @@ class _Loader {
...
           .toSet());
     return sources;
   }
 
+  /// Only allow reads into `lib/` for non-root packages.
+  ///
+  /// As an optimization, also limit the SDK package to the files necessary for
+  /// DDC.
+  InputSet _inputSetForPackage(PackageNode package) {
+    List<String> includes = package.name == r'$sdk'
```
jakemac53: Lets move `$sdk` to a constant somewhere, and consider exposing it (build_compilers uses it as well).

I am fine with this being a followup though if you prefer.

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -191,20 +191,20 @@ class PackageNode {
...
   /// The glob patterns to exclude from this package.
   ///
   /// Any files matching these pattern will be excluded from the asset graph.
-  final List<String> _excludes;
-  Iterable<String> get excludes => _excludes;
+  @deprecated
+  final Iterable<String> excludes;
 
   PackageNode(this.name, this.version, this.dependencyType, String path,
       {Iterable<String> includes, Iterable<String> excludes})
```
jakemac53: These should be deprecated as well now right? Does passing them have any effect?

build_runner/lib/src/package_graph/package_graph.dart
```diff
@@ -191,20 +191,20 @@ class PackageNode {
   ///
   /// Only the files matching these patterns will end up in the asset graph,
   /// unless they also match a pattern in [excludes].
-  final List<String> _includes;
-  Iterable<String> get includes => _includes;
+  @deprecated
+  final Iterable<String> includes;
```
jakemac53: Are these used at all any more?

natebosch: They are part of the public API.

There is very likely not any other code relying on this, but removing it completely would be a breaking change.

build_runner/lib/src/package_graph/package_graph.dart
natebosch: Only insomuch as if some other code is reading these they are still tracked.

jakemac53: My point though is that this is actually a breaking change - since it no longer has an affect on the build and it used to.

natebosch: Ah, good point. Maybe we need to hold this until we start pushing through the breaking changes.

PR #631, opened 2017-11-17
Change BuildApplication to user builder factories

### LINKED ISSUE #625: Align new BuilderApplication with BuilderDefinition
As of now we instantiate the `Builder` and once it hits `creatBuildActions` the config is opaque - we can't see the original name it had from `build.yaml`.

As we start to look at handling module definitions and manually configured builders, as well as overriding builder arguments, we'll need to be able to decide when to ignore a `BuilderApplication` in favor of the manual config.

For a start we could consider

```dart
class BuildApplication {
  final List<Builder Function(List<String>)> builderFactories;
  final String builderName;
  final String vendingPackage;

  final PackageFilter filter;
}
```

The interfaces we choose here could have an impact on our options for solving https://github.com/dart-lang/build/issues/596

natebosch:
Fixes #625

We'll eventually need to support overriding what the `BuildApplication`
says with per-package or per-module config, including passing different
arguments to each `Builder` instance. We need to pass through the
`BuilderFactory` functions and add the necessary keys to relate a config
with the builder.

In order to support the old use case with less noise add `applyToRoot`
with some hacks to plumb through support for checking which package node
is the root. Once we add an `isRoot` field to `PackageNode` we can solve
this with a normal `PackageFilter`.

For now this will end up constructing more `Builder` instances than is
necessary, we can add caching by unique sets of arguments as an
enhancement.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -7,8 +7,7 @@ analyzer:
     dead_code: error
     override_on_non_overriding_method: error
   exclude:
-    - "bazel-*"
-    - "**/goldens/**"
+    - "test/goldens/generated_build_script.dart"
 linter:
   rules:
     # Errors
--- a/build_runner/lib/src/package_graph/apply_builders.dart
+++ b/build_runner/lib/src/package_graph/apply_builders.dart
@@ -33,8 +33,8 @@ typedef Builder BuilderFactory(List<String> args);
 /// Apply [builder] to the root package.
 BuilderApplication applyToRoot(Builder builder,
         {List<String> inputs, List<String> excludes}) =>
-    new BuilderApplication._('', '', [(_) => builder], null,
-        inputs: inputs, excludes: excludes);
+    new BuilderApplication._('', '', [(_) => builder], (_) => false,
+        inputs: inputs, excludes: excludes, applyToRoot: true);
 
 /// Apply each builder from [builderFactories] to the packages matching
 /// [filter].
--- a/build_runner/lib/src/package_graph/apply_builders.dart
+++ b/build_runner/lib/src/package_graph/apply_builders.dart
@@ -28,29 +28,58 @@ PackageFilter toPackages(Set<String> packages) =>
 PackageFilter toAll(Iterable<PackageFilter> filters) =>
     (p) => filters.any((f) => f(p));
 
-/// Indcates that [builder] should be run against all packages matching
+typedef Builder BuilderFactory(List<String> args);
+
+/// Apply [builder] to the root package.
+BuilderApplication applyToRoot(Builder builder,
+        {List<String> inputs, List<String> excludes}) =>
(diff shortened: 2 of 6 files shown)


build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -28,29 +28,58 @@ PackageFilter toPackages(Set<String> packages) =>
...
+BuilderApplication applyToRoot(Builder builder,
+        {List<String> inputs, List<String> excludes}) =>
+    new BuilderApplication._('', '', [(_) => builder], null,
+        inputs: inputs, excludes: excludes);
+
+/// Apply each builder from [builderFactories] to the packages matching
 /// [filter].
 ///
 /// If the builder should only run on a subset of files within a package pass
 /// globs to [inputs] or [excludes];
```
jakemac53: I think we lost any public docs on the `isOptional` arg during this transition, it would be good to add something back since the behavior is a bit nuanced.

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -65,13 +94,37 @@ class BuilderApplication {
...
-    stronglyConnectedComponents<String, PackageNode>(
-            [packageGraph.root], (node) => node.name, (node) => node.dependencies)
-        .expand((cycle) => builderApplications.expand((b) => cycle
-            .where(b.filter)
-            .map((p) => new BuildAction(b.builder, p.name,
-                inputs: b.inputs,
-                excludes: b.excludes,
-                isOptional: b.isOptional))))
-        .toList();
+    Iterable<BuilderApplication> builderApplications, List<String> args) {
```
jakemac53: These args aren't always going to be the same for all packages/builderApplications... how do we handle that?

natebosch: We'll deal with that once we add support for modules and manual config.

The important thing here was just that we can deal with it without a breaking signature change.

analysis_options.yaml
```diff
@@ -7,8 +7,7 @@ analyzer:
     dead_code: error
     override_on_non_overriding_method: error
   exclude:
-    - "bazel-*"
-    - "**/goldens/**"
+    - "test/goldens/generated_build_script.dart"
```
jakemac53: Do we want to exclude this for sure? We do get some analysis today which might be useful to know if we broke things. We could add ignore for file comments to get rid of the hints.

natebosch: Synced offline. Eventually yes I think we will, but for now the `// ignore_for_file:` doesn't seem to work for me.

Once we stop supported the stable SDK we can bump our min version of code_builder and be able to start generating this file without any lints.

build_runner/lib/src/package_graph/apply_builders.dart
natebosch: Done

PR #635, opened 2017-11-20
Add `auto_apply` option to build.yaml

natebosch:
The DDC builders should not be applied automatically, and most package
won't have a dependency on `build_compilers`.


jakemac53:
It doesn't look like this is being used yet - I assume that is in a followup?

natebosch:
Yes, I'll start reading this field in the build script generator next.

natebosch:
https://github.com/dart-lang/build/pull/637

PR #637, opened 2017-11-21
Only add builders with auto_apply: True

natebosch:


jakemac53:
LGTM, but we should add a goldens test and/or unit test

natebosch:
Added a builder that shouldn't show up in the golden to validate behavior.

Issue #639, opened 2017-11-21
Allow build_compilers to be entirely config driven


natebosch:
In https://github.com/dart-lang/build/pull/638 some knowledge of DDC is hardcoded in the build script generator. We could plausibly make the generator completely agnostic to this if we allow our config to be more expressive.

- [x] Change `auto_apply` from a boolean to an enum (string in the config) to allow  at least: `'dependents'`, `'all_packages'`, `'root_package'`, `'manual'`, where 'manual' is the default
- [x] Add a field `requires_extension` which says that the builder needs any builder which outputs `.dart` files to have already run. This allows us to make at least enough ordering decision for DDC to work.
- [x] Add a `defaults: generate_for:` so that the DDC bootstrap builder can say it wants to apply to `['web/**.dart', 'test/**.browser_test.dart']`
- [x] Add `is_optional` config so the modules/summaries/ddc steps can be optional. This is hardcoded today
- [x] Add support for using manual builder config
- [ ] ~Add default config for `input` in the root package - act like it had a `build.yaml` specifying this.~

This should be enough for DDC and Dart2JS. With the way the test Builder is implemented today I don't think it uses a Resolver so it should be fine to put it anywhere in the order.

natebosch:
@jakemac53 - I have a version of this working with everything except `default_inputs`. That one makes me a little nervous - are we OK with hardcoding `package:test` knowledge into `build_compilers` so that it builds `test/**.browser_test.dart`?

Should we allow overriding the `web/**.dart` default? This is an argument to the BuildAction, not the Builder so this wouldn't fit in `BuilderOptions`...

natebosch:
We decided to punt on `default_inputs` since it's more likely to cause problems than solve them.

We can make the majority of DDC stuff be driven by config, but the `DevCompilerBootstrapBuilder` will need to be hardcoded magic in the entrypoint generator for now.

natebosch:
build_compilers can be completely config driven now. The build script generator does not hardcode anything as of https://github.com/dart-lang/build/pull/750

I've also run a proof of concept with sass_builder and angular driven by config.

PR #640, opened 2017-11-21
Add required_inputs config option

natebosch:
Towards #639

We want to be able to specify that `DDC` needs all builders which
generate Dart files to have already run before it can run. We won't have
a full order solver, but this will us allow us to solve the two use
cases we have seen so far: all `.dart` before a compiler, and all `.css`
before the Angular stylesheet compiler.

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -25,12 +25,12 @@ the following keys:
   merged `buildExtensions` maps from each `Builder` in `builder_factories`.
 - **auto_apply**: Optional. Whether to apply this builder automatically to
   packages which have a dependency to this package. Defaults to `False`.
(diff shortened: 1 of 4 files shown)


build_config/README.md
```diff
@@ -25,6 +25,12 @@ the following keys:
   merged `buildExtensions` maps from each `Builder` in `builder_factories`.
 - **auto_apply**: Optional. Whether to apply this builder automatically to
   packages which have a dependency to this package. Defaults to `False`.
+- **required_inputs**: Optional. If a Builder must see every input with one or
```
grouma: Should probably state that this is a list of extensions.

Issue #642, opened 2017-11-22
Element.documentationComment is not reachable.


matanlurey:
Original issue (moved here): https://github.com/dart-lang/source_gen/issues/279

We use an `AnalysisContext` with `preseveComments = false`, maybe for perf? Not clear.

/cc @natebosch

Issue #628, opened 2017-11-17
Bundle breaking changes for build_runner 0.7.0


natebosch:
These are up for discussion. Collecting them all here so we don't forget something before publishing.

- [x] Drop `PackageNode.dependencyType`.  Nothing reads this field. We could potentially use it to set up fewer directory watchers, but we aren't today.
- [x] Drop `PackageNode.version`. Nothing reads this field.
- [x] Drop `PackageGraph.orderedDependencies` and `PackageGraph.dependentsOf` - We want most use cases for these to shift into `createBuildActions` and stuff inside `build_runner` can use `stronglyConnectedComponents`.
- [x] Stop exporting `InputSet` - shouldn't be any use case
- [x] Stop exporting `BuildAction` - we want `BuilderApplication` to replace this.
- [x] Drop `PackageNode.includes` and `PackageNode.excludes`.
- [x] Add `PackageNode.isRoot` - this would be breaking if we want to add assertions that the PackageGraph has exactly one node reporting `isRoot = true` at `PackageGraph.root`.
- [x] Set minimun SDK to 2.0.0: #645 
- [x] Remove `writeToCache` - #647 
- [x] Update to latest `build_barback` to pick up #643 
- [x] Add a whitelist to the directories that can be considered sources: #666

jakemac53:
I would argue that we should consider not dropping `dependencyType` and `version`, and instead start actually using them to speed up startup.

natebosch:
How would we use `version`?

jakemac53:
I didn't realize how much code we have surrounding this actually - go ahead and delete the version/dependency type

natebosch:
I think we've hit all the breaking changes. A quick scan through the dartdoc looks like we've got it where we want for now - at least for the public interfaces

PR #643, opened 2017-11-22
Configure analysis options.

matanlurey:
Closes https://github.com/dart-lang/build/issues/642.

This ... might be tough, if we have folks relying on the current behavior. I'm assuming this should probably be `0.5.0`, but I'm also open for considering the current behavior as a bug since it (a) wasn't using strong mode and (b) was unintentionally different than `bazel_codegen`.

Added two tests as well.

/cc @kevmoo


natebosch:
I think we do want to release this as a breaking change, and we'll want to wait for build_runner 0.7.0 to pull it to that package since most users only see it through that dependency.

We'll also want to check the impact on performance. We're probably buying back more performance by not analyzing method bodies than we're losing preserving comments so I bet we'll come out ahead on this change overall.

matanlurey:
@natebosch Do you want me to still land this PR to MASTER?

I'll update the CHANGELOG and pubspec accordingly.

natebosch:
Let's hold off on landing until next week, I'll try to get a rough idea of the performance early next week and then we can merge.

We're blocked on publishing the breaking change version of build_runner for a little while so we can't take advantage of this until then.

matanlurey:
Sg. Do we have any performance benchmarks as part of this repo?

natebosch:
No formal benchmarks. My plan was to pick a medium size build and do a few runs with `time` just to get a rough idea.

matanlurey:
Sg. `gwt_mail_sample` might be a good target to convert and benchmark.

jakemac53:
LGTM assuming no significant perf issues

natebosch:
I ran a few builds on angular_components_example before and after this change. I didn't see any change in the overall build time.
If there is any difference here it's likely dominated by other factors. I think we can land this as a breaking change.

matanlurey:
@natebosch Do you want me to land it today?

natebosch:
Anytime is fine once the pubspec is updated. We won't be able to take advantage in build_runner until the next breaking release. I added a note to #628 to keep track

matanlurey:
Great. I'll land it sometime this afternoon.

matanlurey:
@natebosch How should I land this?

I need to add the following to be able to run `pub get`:

```yaml
dependency_overrides:
  # TODO: Remove after a version is published that supports 0.5.0.
  build_test:
    path: ../build_test
```

We might want to add a RELEASE.md or something for the dirty details of updating these.

matanlurey:
Also, https://github.com/dart-lang/pub/issues/1745.

natebosch:
Yeah we typically add the `dependency_override` sections as necessary to get Travis to pass, then right before publishing we remove them and expect Travis to fail until everything is published.

It's not a fun process but without an atomic submit we don't have much better.

I think in this case it would also be fine to preemptively publish `build_test` with a constraint of <0.6.0

matanlurey:
That might be cleanest. I'll send you a PR for that first.

natebosch:
build_test is published allowing 0.5.0

The pub get should work now with a version bump on build_barback.

matanlurey:
@natebosch PTAL again before I merge.

Issue #651, opened 2017-11-25
e2e-example doesn't work on Windows machine


yury-yufimov:
build_runner has some bugs for running on Windows machine. When I try to execute e2e-example's build.dart, I receive following error:
```
[WARNING] BuildScriptUpdates: e2e_example|../../../../../C:/projects/tmp/build/e2e_example/tool/build.dart was not found in the asset graph, incremental builds will not work.
 This probably means you don't have your dependencies specified fully in your pubspec.yaml.
```
Some investigation shows, that
https://github.com/dart-lang/build/blob/52801f53655bcb0b6c66f8c6af8c881b6b4d1078/build_runner/lib/src/changes/build_script_updates.dart#L70
receives uri with path like "/C:/projects/tmp/build/e2e_example/tool/build.dart" (starting with slash!), which can't be handled correctly. Removing of this slash locally fixes the problem, but I think it needs more complex investigation for correct fix.

matanlurey:
We realistically need some sort of windows-based testing for this repository (sorry, most testing and developing has been on unix - OSX, Linux - machines). Let's keep this issue open for tracking this specific fixes.

jakemac53:
We also need to fix up the tests so they will run on windows but we can track that separately. I know today a lot of them fail because the tests expect some linux specific behavior.

Issue #652, opened 2017-11-27
Add AppVeyor integration for Windows testing


matanlurey:
At minimum we should be able to run the `e2e_example` suite as a sniff test.

I can take a shot at this.

PR #654, opened 2017-11-27
Try adding appveyor.

matanlurey:
Just experimenting.

Work towards https://github.com/dart-lang/build/issues/652.

Fixes https://github.com/dart-lang/build/issues/651.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,17 +5,19 @@ branches:
 version: '1.0.{build}'
 
 install:
-  - choco install dart-sdk-dev
+  - ps: wget https://gsdview.appspot.com/dart-archive/channels/dev/raw/latest/sdk/dartsdk-windows-x64-release.zip -OutFile dart-sdk.zip
+  - cmd: echo "Unzipping dart-sdk..."
+  - cmd: 7z x dart-sdk.zip -o"C:\tools" -y > nul
+  - set PATH=%PATH%;C:\tools\dart-sdk\bin
+  - set PATH=%PATH%;%APPDATA%\Pub\Cache\bin
   - RefreshEnv
   - cd e2e_example
-  - C:\tools\dart-sdk\bin\pub.bat get && exit 0
+  - pub get && exit 0
 
 build: off
 
 test_script:
-  - C:\tools\dart-sdk\bin\pub.bat run test
+  - pub run test -j 1
 
 cache:
-  - C:\ProgramData\chocolatey\bin
-  - C:\ProgramData\chocolatey\lib
   - C:\Users\appveyor\AppData\Roaming\Pub\Cache
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,12 +8,12 @@ install:
   - choco install dart-sdk-dev
   - RefreshEnv
   - cd e2e_example
-  - pub get && exit 0
+  - C:\tools\dart-sdk-dev\bin\pub.bat get && exit 0
 
 build: off
 
 test_script:
-  - pub run test
+  - C:\tools\dart-sdk-dev\bin\pub.bat run test
 
 cache:
   - C:\ProgramData\chocolatey\bin
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,12 +8,12 @@ install:
   - choco install dart-sdk-dev
   - RefreshEnv
   - cd e2e_example
-  - C:\tools\dart-sdk\bin\pub.bat get && exit 0
+  - pub.bat get && exit 0
 
 build: off
 
 test_script:
-  - C:\tools\dart-sdk\bin\pub.bat run test
+  - pub.bat run test
 
 cache:
   - C:\ProgramData\chocolatey\bin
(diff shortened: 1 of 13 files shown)


matanlurey:
I think I need @kevmoo to allow the webhook. Or something.

jakemac53:
nice! we should add mono-repo support as well

kevmoo:
@matanlurey just enabled it. Trying 'kicking' it with a new push?

jakemac53:
@matanlurey see https://ci.appveyor.com/project/kevmoo/build/build/1.0.2, not sure if we need to actually call `pub.bat` or just `pub` on windows?

appveyor.yml
```diff
@@ -0,0 +1,21 @@
+branches:
+  only:
+    - master
+
+version: '1.0.{build}'
+
+install:
+  - choco install dart-sdk-dev
```
jakemac53: I don't thing chocolatey gets updated very often but that is probably ok? I think it's the easiest way to install so the trade-off is probably worth it.

matanlurey: Yeah. If we end up needing an edge version we can swap it.

jakemac53:
Ok cool, at least this is running now :). Not sure why the path is different than what choco is thinking it installed it to....

matanlurey:
@jakemac53: Choco always installs both dev and non-dev in that folder it seems.

matanlurey:
It looks like we need `pub.bat` on windows in our testing scripts/etc too.

matanlurey:
Other AppVeyor scripts in dart-lang don't use the full path though: 
* https://github.com/dart-lang/linter/blob/master/appveyor.yml
* https://github.com/dart-lang/pub_cache/blob/master/appveyor.yml
* https://github.com/dart-lang/dartdoc/blob/master/appveyor.yml

jakemac53:
Looks like they are manually adding the right path https://github.com/dart-lang/linter/blob/master/appveyor.yml#L9

jakemac53:
At this point i see exactly the same result here as I do when running locally on windows so that is a good sign

matanlurey:
@jakemac53: I'm going to use what linter/dartdoc does for now.

matanlurey:
Interesting, the `PATH` changes don't seem to effect the manual runs of pub.

@jakemac53 WDUT?

jakemac53:
Not sure exactly, but I see the same behavior on my actual windows machine as well.

matanlurey:
Alright. Do you want to spin the wheels a bit or should I land this as-is and wait for windows-specific fixes? @jakemac53

jakemac53:
Ideally we would land this in a green state so future pull requests aren't red, I think its worth putting in a bit of effort towards that first. I can push changes to this branch today and see how far I can get.

matanlurey:
Sg.

jakemac53:
Ah the perils of working from my personal windows box lol, I obviously am fine with my commits being contributed to this project :)

matanlurey:
I see some non-fatal errors so far:

```
StdErr: FileSystemException: Deletion failed, path = 'C:\Users\appveyor\AppData\Local\Temp\1\build_compilersd6f8f12a-d3c8-11e7-81d9-00155d57f23e' (OS Error: The process cannot access the file because it is being used by another process.
```

@jakemac53

jakemac53:
Yep, I didn't have as much time to look into this as I had hoped today but I did resolve a couple issues. That one is harder, but I do think I know what is going on. It has to do with the ordering of the `Resource`s getting cleaned up. In this case the scratch space trying to delete itself while the worker processes are still running in the temp dir.

jakemac53:
@natebosch it looks like this is going to all be passing now, do you want to make a quick pass over the changes before we submit?

build_runner/lib/src/util/constants.dart
```diff
@@ -6,7 +6,13 @@ import 'dart:io';
 import 'package:crypto/crypto.dart';
 
 /// Relative path to the asset graph from the root package dir.
-final String assetGraphPath = assetGraphPathFor(Platform.script.path);
+final String assetGraphPath = () {
```
natebosch: [optional] Pretty minor but I think the conditional operator improves readability slightly because it draws attention to the thing that is different (toFilePath() vs path) and clearly show the thing that is the same (assetGraphPathFor)

```dart
final String assetGraphPath = () {
  return assetGraphPathFor(Platform.script.scheme == 'file'
      ? Platform.script.toFilePath()
      : Platform.script.path);
}();
```

build_compilers/lib/src/scratch_space.dart
```diff
@@ -21,4 +30,36 @@ final scratchSpaceResource = new Resource<ScratchSpace>(() {
...
+  var numTries = 0;
+  while (true) {
+    numTries++;
+    if (numTries > 3) {
+      _logger
+          .warning('Failed to clean up temp dir ${scratchSpace.tempDir.path}.');
+      return;
+    }
+    try {
+      // The scratch space throws on `delete` if it thinks it was already
```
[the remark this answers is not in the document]
jakemac53: filed https://github.com/dart-lang/build/issues/656

build_runner/lib/src/util/constants.dart
jakemac53: done

jakemac53:
Thanks for kicking the tires on this @matanlurey !

Issue #636, opened 2017-11-20
Change BuilderFactory to take a BuilderOptions argument instead of List<String>


natebosch:
Within bazel we have a dependency on this interface so this would be breaking for bazel, but we have a window to change this where it won't be breaking for build_runner since that isn't published yet.

The UX for configuring arguments in both BUILD files and starting soon in `build.yaml` is nicest when it is a dict/yaml map rather than `--argument=value` strings. We already parse out the dict and convert it into `--` args, only to later reparse those with an `ArgParser` in most cases.

It might be nicer to support an interface like:

```dart
class BuilderOptions {
  bool get isDebug; // We'd need to decide what this means for bazel
  Map<String, dynamic> get config; // Specific to the Builder we're passing this to
}

typedef Builder BuilderFactory(BuilderOptions options);
```

For most Builder authors this is likely satisfactory and won't require using `ArgParser`.

@matanlurey - thoughts?

matanlurey:
Sg!

natebosch:
Need to nail down which fields we'd need.

Some possibilities:

- `isDebug` - Angular has this flag today but no other builders I'm aware of do. Should we promote this to a generally available thing in bazel? Should this mean something in `build_runner`?
- `isOutlineMode` - Angular is definitely the only thing using this. It's a pseudo bazel supported thing in that its a flag outside of angular arguments, but ends up getting merged in with angular args. I don't think this would ever be meaningful in `build_runner`... Maybe we should move this to something more specific to angular and not include it in the general options?

We could also start with with just `config` and use the class as a future proof place we can put other general options if and when they come up.

matanlurey:
I think you could keep both of these Angular-specific for now.

natebosch:
We've got this updated everywhere.

PR #658, opened 2017-11-29
Switch to BuilderFactory from package:build

natebosch:
Towards #636 

Drop `args` handling in `createBuildActions` for now. We should be able
to add this back in a non-breaking change later when we support creating
real `BuilderOptions` instances.


matanlurey:
Nice to see AppVeyor running :D

PR #659, opened 2017-11-29
Change autoApply to an enum

natebosch:
Towards #639

- Add `toNoneByDefault` package filter so that all `BuilderDefinitions`
  can be present in the `BuilderApplications` which will allow us to
  apply them manually later.
- Continue to manually apply the DDC builders in the generated build
  script until. This does mean that the builders show up twice but in
  one case they won't be applied.

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -27,7 +27,7 @@ the following keys:
   automatically to applied. Defaults to `'none'` The possibilities are:
   - `"none"`: Never apply this Builder unless it is manually configured
   - `"dependents"`: Apply this Builder to the package with a direct dependency
-    on the package exposign the builder.
+    on the package exposing the builder.
   - `"all_packages"`: Apply this Builder to all packages in the transitive
     dependency graph.
   - `"root_package"`; Apply this Builder only to the top-level package.
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -23,8 +23,14 @@ the following keys:
 - **build_extensions**: Required. A map from input extension to the list of
   output extensions that may be created for that input. This must match the
   merged `buildExtensions` maps from each `Builder` in `builder_factories`.
-- **auto_apply**: Optional. Whether to apply this builder automatically to
-  packages which have a dependency to this package. Defaults to `False`.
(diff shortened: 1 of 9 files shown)


build_config/README.md
```diff
@@ -23,8 +23,14 @@ the following keys:
 - **build_extensions**: Required. A map from input extension to the list of
   output extensions that may be created for that input. This must match the
   merged `buildExtensions` maps from each `Builder` in `builder_factories`.
-- **auto_apply**: Optional. Whether to apply this builder automatically to
-  packages which have a dependency to this package. Defaults to `False`.
+- **auto_apply**: Optional. The packages which should have this builder
+  automatically to applied. Defaults to `'none'` The possibilities are:
+  - `"none"`: Never apply this Builder unless it is manually configured
+  - `"dependents"`: Apply this Builder to the package with a direct dependency
+    on the package exposign the builder.
```
jakemac53: exposing*

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -153,3 +146,24 @@ Expression _applyBuilder(BuilderDefinition definition, Expression toExpression,
...
+      return refer('toDependentsOf', 'package:build_runner/build_runner.dart')
+          .call([literalString(definition.package)]);
+    case AutoApply.allPackages:
+      return refer('toAllPackages', 'package:build_runner/build_runner.dart')
+          .call([]);
+    case AutoApply.rootPackage:
+      return refer('toPackage', 'package:build_runner/build_runner.dart')
+          .call([literalString(packageGraph.root.name)]);
+  }
+  throw 'Unreachable';
```
jakemac53: typically I prefer to handle this by throwing a more specific error in the `default` case... I can see an argument for this as well though since it allows us to get an analyzer error if we are missing one of the values. I think its worth throwing a `StateError` with a better message still though.

matanlurey: ArgumentError is pretty good:

PR #661, opened 2017-11-29
Add builder ordering

natebosch:
Towards #639

Allows Builders to be applied in an order satisfiying their requirements

- Use the `reuired_inputs` configuration to find all builders which may
  output a required file and treat them as a builder dependency.
- Use `stronglyConnectedComponents` to order builders by dependency.
  Consider a cycle to be a failure case.
- Add `auto_apply` to the DDC builder. Now that we can guarantee it is
  run in the correct order we don't need to add it manually at the end.

The `ddc_bootstrap` builder is still applied manually since we don't
have a solution to specify the input globs.

### DIFF
--- a/build_runner/lib/src/build_script_generate/builder_ordering.dart
+++ b/build_runner/lib/src/build_script_generate/builder_ordering.dart
@@ -20,7 +20,7 @@ Iterable<BuilderDefinition> findBuilderOrder(
 
 List<BuilderDefinition> _findOrder(Iterable<BuilderDefinition> builders) {
   Iterable<BuilderDefinition> dependencies(BuilderDefinition parent) =>
-      builders.where((child) => _hasDependency(parent, child));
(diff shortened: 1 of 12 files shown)


build_runner/lib/src/build_script_generate/builder_ordering.dart
```diff
@@ -0,0 +1,40 @@
...
+      builders, _builderKey, dependencies);
+  return components.map((component) {
+    if (component.length > 1) {
+      throw new ArgumentError('Required input cycle for ${component.toList()}');
+    }
+    return component.single;
+  }).toList();
+}
+
+bool _hasDependency(BuilderDefinition parent, BuilderDefinition child) {
```
jakemac53: nit: consider `_hasInputDependency` or something so its a bit more clear what a dependency means in this context

natebosch: Done

PR #663, opened 2017-11-30
Bump max version of build_barback from build_test

natebosch:


natebosch:
After I publish this we should be able to make Travis green for https://github.com/dart-lang/build/pull/643

PR #667, opened 2017-11-30
Add build_runner:create_merged_dir command

jakemac53:
This is experimental only, primarily just to get angular off of a git dependency for travis builds.

Given the path to a build script it will read the corresponding output graph and create a merged output directory based on those sources.

I also included some of the other fixes from the `build_dir` branch to get --precompiled working with tests.

### DIFF
--- a/build_runner/bin/create_merged_dir.dart
+++ b/build_runner/bin/create_merged_dir.dart
@@ -1,0 +1,131 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:args/args.dart';
+import 'package:build/build.dart';
+import 'package:path/path.dart' as p;
+
+import 'package:build_runner/src/asset/build_cache.dart';
+import 'package:build_runner/src/asset/file_based.dart';
+import 'package:build_runner/src/asset_graph/graph.dart';
+import 'package:build_runner/src/asset_graph/node.dart';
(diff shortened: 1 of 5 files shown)


build_runner/bin/create_merged_dir.dart
```diff
@@ -0,0 +1,131 @@
...
+  var outputDir = new Directory(parsedArgs['output-dir'] as String);
+  if (outputDir.existsSync()) {
+    stdout.writeln('Deleting existing output dir `$outputDir`');
+    outputDir.deleteSync(recursive: true);
+  }
+  outputDir.createSync(recursive: true);
+
+  var scriptFile = new File(scriptPath);
+  if (!scriptFile.existsSync()) {
+    throw new ArgumentError(
```
natebosch: Can we move this and other checks above where we delete the old output?

build_runner/bin/create_merged_dir.dart
```diff
@@ -0,0 +1,131 @@
...
+    rootDirs.add(parts.first);
+  }
+
+  var reader = new BuildCacheReader(new FileBasedAssetReader(packageGraph),
+      assetGraph, packageGraph.root.name);
+
+  for (var node in assetGraph.allNodes) {
+    if (node is SyntheticAssetNode) continue;
+    if (node is GeneratedAssetNode && !node.wasOutput) continue;
+    if (node.id.path == '.packages') continue;
```
natebosch: should we be excluding this from ever getting to the asset graph?

jakemac53: Yes actually, because its never really valid once we do a build.

Filed https://github.com/dart-lang/build/issues/668

build_runner/bin/create_merged_dir.dart
jakemac53: done

PR #669, opened 2017-11-30
Prepare to publish build_config 0.1.1

natebosch:


natebosch:
Where possible, going in order of what is safe.

Issue #617, opened 2017-11-16
Dart2js support


jakemac53:

matanlurey:
I need this for something I'm hacking on, and I'm willing to contribute here!

... but if you're planning on waiting for CFE/Kernel first, then it's alright.

WDUT @jakemac53?

jakemac53:
We don't need to wait on CFE/Kernel I don't think but this is going to be a bit complicated because we want users to be able to switch back and forth between ddc/dart2js with the same build script. We could chat offline or I can add details here for some of our ideas.

matanlurey:
Ah I see.

I was planning on just writing the `Dart2JSBuilder` (similar to the `DartDevCBuilder`), without plumbing it through the rest of the system (and just using this builder directly for my own needs). If that gets you more towards your goal of #617, then I'm happy to contribute it.

If you think you'll need something more clever, I can hold off.

matanlurey:
Though, I could see value in just getting rid of the differentiation in the future:

```dart
abstract class DartJSBuilder implements Builder {
  // DDC
  factory DartJSBuilder() => ...

  // Dart2JS
  factory DartJSBuilder.release() => ...
}
```

... though you might have something more complicated in mind here too :)

jakemac53:
So my current thinking on this is essentially that we should have a single builder which knows how to build both DDC apps (the current DevCompilerBootstrapBuilder) and Dart2js apps.

You would switch between the two modes based on configuration - and could override the current setting with a flag. This would use the standard configuration strategy based on `BuilderOption`s that @natebosch is are starting to add support for now.

Whenever the configuration for a builder changes (we would need to serialize the previous configuration as a part of the asset graph), we would invalidate all the outputs of that builder.

This has some nice properties:

- All the ddc modules are always a part of the build graph, since they are built with a separate builder. 
  - No penalty for switching back and forth between compilers, ddc modules are compiled lazily (so not at all in dart2js mode) and previous cached versions will still exist when you switch back.
- It is only the bootstrap files at the top (*.dart.js and related) that get invalidated when you switch compilers.
- Users can configure their default compilers and easily override with a flag for one-off situations.

matanlurey:
I like all of this!

Only one question:

> Users can configure their default compilers and easily override with a flag for one-off situations.

How do we intend to deal with a package like so:

```
web/
  main.dart     // <-- Compile me with Dart2JS
test/
  app_test.dart // <-- Compile and run me with DDC
```

[...]

PR #680, opened 2017-12-02
Refactor workers in BuildCompilers.

matanlurey:
Work towards https://github.com/dart-lang/build/issues/617.

Let me know if you'd like any changes, or to make it more `bazel_worker` specific again, but this looked like the least amount of boilerplate for creating reusable workers without getting _too_ abstract. It also looks like it might be useful to integrate external tools in the future.

(Happy to add tests, even though this mostly just wraps everything for now)

### DIFF
--- a/build_compilers/lib/src/dev_compiler_builder.dart
+++ b/build_compilers/lib/src/dev_compiler_builder.dart
@@ -64,9 +64,9 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
   var jsOutputFile = scratchSpace.fileFor(module.jsId);
   var libraryRoot = '/${p.split(p.dirname(module.jsId.path)).first}';
   var sdkSummary = p.url.join(sdkDir, 'lib/_internal/ddc_sdk.sum');
-  var request = new WorkRequest();
+  var request = <String>[];
 
-  request.arguments.addAll([
+  request.addAll([
     '--dart-sdk-summary=$sdkSummary',
     '--modules=amd',
     '--dart-sdk=${sdkDir}',
@@ -79,22 +79,22 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
   ]);
 
   if (debugMode) {
-    request.arguments.addAll([
+    request.addAll([
       '--source-map',
       '--source-map-comment',
       '--inline-source-map',
     ]);
   } else {
-    request.arguments.add('--no-source-map');
+    request.add('--no-source-map');
   }
 
   // Add the default analysis_options.
   await scratchSpace.ensureAssets([defaultAnalysisOptionsId], buildStep);
-  request.arguments.add(defaultAnalysisOptionsArg(scratchSpace));
+  request.add(defaultAnalysisOptionsArg(scratchSpace));
 
   // Add all the linked summaries as summary inputs.
   for (var id in transitiveSummaryDeps) {
-    request.arguments.addAll(['-s', scratchSpace.fileFor(id).path]);
+    request.addAll(['-s', scratchSpace.fileFor(id).path]);
   }
 
   // Add URL mappings for all the package: files to tell DartDevc where to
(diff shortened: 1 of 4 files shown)


build_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -64,9 +64,9 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
   var jsOutputFile = scratchSpace.fileFor(module.jsId);
   var libraryRoot = '/${p.split(p.dirname(module.jsId.path)).first}';
   var sdkSummary = p.url.join(sdkDir, 'lib/_internal/ddc_sdk.sum');
-  var request = new WorkRequest();
+  var request = <String>[];
```
natebosch: [nit] rename to 'arguments' or similar

build_compilers/lib/src/workers.dart
```diff
@@ -9,76 +9,134 @@ import 'dart:math' show min;
...
-    _analyzerWorkersAreDoneCompleter?.future ?? new Future.value(null);
-Completer<Null> _analyzerWorkersAreDoneCompleter;
+/// On windows some tools have to be executed with the `.bat` extension.
+final String _scriptExtension = Platform.isWindows ? '.bat' : '';
 
-/// Completes once the dartdevc workers have been shut down.
-Future<Null> get dartdevcWorkersAreDone =>
-    _dartdevcWorkersAreDoneCompleter?.future ?? new Future.value(null);
-Completer<Null> _dartdevcWorkersAreDoneCompleter;
+/// Returns the path to the SDK.
```
natebosch: [nit] use a noun phrase

build_compilers/lib/src/workers.dart
```diff
@@ -9,76 +9,134 @@ import 'dart:math' show min;
...
+    log.warning('Invalid value for $_maxWorkersEnvVar environment variable, '
+        'expected an int but got `$value`. Falling back to default value '
+        'of $_defaultMaxWorkers.');
+    return _defaultMaxWorkers;
+  },
+);
+
+/// Encapsulates running an external tool in a sandbox as a compiler.
+///
+/// TODO: Consider making this public in build_compilers(?) in the future.
```
natebosch: I think if we were to do that we'd need to hide details like `WorkResponse`

build_compilers/lib/src/workers.dart
```diff
@@ -9,76 +9,134 @@ import 'dart:math' show min;
...
 
-/// Resource for fetching the current [BazelWorkerDriver] for dartdevc.
-final dartdevcDriverResource = new Resource<BazelWorkerDriver>(
-    () => _dartdevcDriver, beforeExit: () async {
-  await _dartdevcDriver?.terminateWorkers();
-  _dartdevcWorkersAreDoneCompleter.complete();
-  _dartdevcWorkersAreDoneCompleter = null;
-  __dartdevcDriver = null;
-});
+class _DartDevCWorker extends BatchWorker {
```
natebosch: Could we try a pattern without `extends`? If we don't think we'll need to `@override` other methods in the class I really like when we're precise about what behavior can possibly be changed.

```dart
typedef Future<Process> StartProcess(String workingDirectory);

class BatchWorker {
  final StartProcess _startProcess;
  ...
  BatchWorker(this._startProcess);
}

final dartDevcWorker = new BatchWorker((workingDirectory) => Process.start(
    _defaultPath, const ['--persistent_worker'],
    workingDirectory: workignDirectory));
```

build_compilers/lib/src/summary_builder.dart
```diff
@@ -70,8 +70,8 @@ Future createUnlinkedSummary(Module module, BuildStep buildStep,
   await scratchSpace.ensureAssets(module.sources, buildStep);
 
   var summaryOutputFile = scratchSpace.fileFor(module.unlinkedSummaryId);
-  var request = new WorkRequest();
-  request.arguments.addAll([
+  var request = <String>[];
```
jakemac53: same as nate said above, lets rename this `args` or similar

build_compilers/lib/src/workers.dart
jakemac53: We could change it to a `ProcessResult`, and we could make something that wraps a work response and implements that

matanlurey:
I'm not sure this is the right approach anymore since dart2js does not support the same batch mode as the other compilers. We can always resurrect if I'm wrong :)

Issue #666, opened 2017-11-30
Figure out how we can ignore the 'deploy' directory during builds


natebosch:
Once we implement a `build` or `deploy` command to get full merged output we're going to need to solve the feedback cycles where  the `build/` directory will suddenly look like new sources.

We've discussed a number of options, and they all have drawbacks:

# Hardcode `build/` as the directory and don't allow overriding

## Pros
- We can easily know to ignore it and it's always valid

## Cons
- Doesn't allow a different workflow like deploying to outside the package
- Doesn't allow different build outputs for different modes like production vs debug

# Require static config for directories other than `build/`

## Pros
- We can know statically what to ignore

## Cons
- If you change the config but don't delete the old directory it again looks like a source
- Need to define a schema for the config and stick with it - will we allow different outputs for different build modes?

# Record a list of all directories we've built to and ignore them forever

## Pros
- Handles making new choices without deleting the old directory

## Cons
- Can easily get lost if `.dart_tool/` is deleted.
- Doesn't allow changing your mind and putting source files in a directory that used to be a deploy directory.

# Read `.gitignore` and don't treat anything ignored as a source

## Pros
- Should work most of the time for most people without additional intervention

## Cons
- Would we support `.hgignore`?
- Although not recommended some people might want to build from sources that aren't published.
- Doesn't help for local development before creating a repo

# Only allow building to outside the package

## Pros
- Saner and safer strategy overall, even outside of build considerations

## Cons
- We've already trained our users with `pub` that mixing deploy directory into the package is the right thing to do
- If an IDE is open to the package root people might want to see the deploy directory in their editor

# Require whitelisting the directories which can have sources

We'd start wth `lib/`, `web/`, `bin/` to start, allow adding other directories with static config.

## Pros
- Safe and narrow - opt in to risk rather rather than opt out
- Encourages standard practices
- Should work most of the time for most people without additional intervention

## Cons
- Need to figure out the config

natebosch:
@kevmoo for comment

After writing it out I think I'm most in favor of the whitelist.

jakemac53:
I think I like the whitelist approach as well, and the defaults should work for most people, although we should add `benchmark` and `example`

matanlurey:
👍  for whitelist.

This is something we can always re-think if there is a huge influx of users with more needs.

natebosch:
I'm even inclined to start with a hardcoded whitelist and wait for demand to allow configuring extra source directories.

matanlurey:
[...]

PR #682, opened 2017-12-04
Whitelist main package dirs

jakemac53:
Fixes https://github.com/dart-lang/build/issues/666

I also deleted the `toolDirs` constants since we don't need them any more, and replaced the handling for the entry point script with a new type of node - `InternalAssetNode`, which is similar to a `SourceAssetNode` but is never used as a primary input.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 ### Breaking Changes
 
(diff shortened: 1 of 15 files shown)


jakemac53:
@natebosch updated and fixed the issue we talked offline about regarding edits to the generated entrypoint, PTAL

Issue #653, opened 2017-11-27
Add OSX testing to Travis


matanlurey:
There is a little bit of a cost, as we'd need to make sure all the shell scripts work on OSX.

But like https://github.com/dart-lang/build/issues/652, this is likely long-term important.

PR #691, opened 2017-12-07
Add OSX support to travis.

matanlurey:
(Not ready to merge yet)

Fixes https://github.com/dart-lang/build/issues/653.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,8 +112,3 @@ branches:
 cache:
  directories:
    - $HOME/.pub-cache
-
-before_install:
- - "export DISPLAY=:99.0"
- - "sh -e /etc/init.d/xvfb start"
- - "t=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done"
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,3 +112,8 @@ branches:
 cache:
  directories:
    - $HOME/.pub-cache
+
+before_install:
+ - "export DISPLAY=:99.0"
+ - "sh -e /etc/init.d/xvfb start"
+ - "t=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 # Created with https://github.com/dart-lang/mono_repo
 language: dart
 
+# https://docs.travis-ci.com/user/multi-os/
+os:
+  - linux
+  - osx
+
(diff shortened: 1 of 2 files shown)


jakemac53:
Looks like the only issue is the new chrome addon change isn't working on the linux boxes... any ideas?

jakemac53:
Ah, looks like the chrome addon doesn't automatically start xvfb - added that part back to the before_install section per https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI and it looks like it is going to be green now.

matanlurey:
Woohoo!

jakemac53:
Looks like there is a bit of a backlog for travis mac builds due to https://www.traviscistatus.com/incidents/1vd05j8y0yfq but that should be a short term thing

jakemac53:
hmm actually this looks like a common issue https://www.traviscistatus.com/#month. They don't appear to ever run more than 180 mac jobs and the queue often gets pretty long... not sure if that will be a blocker for us or not. We can try it out for a while and see how annoying it gets in any case.

natebosch:
Should we limit it to the e2e tests on mac?

jakemac53:
See https://blog.travis-ci.com/2017-09-22-macos-update as well talking about this issue

jakemac53:
Limiting to e2e tests won't help us much (they will still be queued on the main queue). However we could do it just as good travis citizens since the macos builders are limited.

matanlurey:
There sounds fine to limit to e2e!

jakemac53:
Oh and of course the xfvb commands fail on osx lol.

Also, given the time it takes to queue these jobs it seems prudent try and see if we can run them only on commits to master, and not for pull requests.

matanlurey:
@jakemac53 Want to just add the `branches: [master]` and call it done?

matanlurey:
After the xfvb fix, of course. We can conditionally run that on != OSX.

natebosch:
> Want to just add the `branches: [master]` and call it done?

We have the config to only run on master, the problem is that it's still too slow for pull requests.
Ideally I think what we'd want is to run on linux for pull requests, and on mac with a cron or something. That way we aren't slowed down getting stuff merged, but we notice within a day if we do break mac.

matanlurey:
It looks like it took at most 4 minutes per task with e2e_example taking ~1.5 minutes:
https://travis-ci.org/dart-lang/build/jobs/313169150

Why don't we start with just e2e_example for OSX like Windows?

natebosch:
The problem is that the osx tasks take a long time to get started. Once they start they run quick enough

natebosch:
There is too much delay before the builds even get started. I don't think we want to enable this until Travis has lower queue times for OSX

Issue #647, opened 2017-11-22
Allow mixed-mode building


natebosch:
We need to allow running Builders that output to source, as well as to the generated cache directory.

- [ ] #626 - How will outputting to source impact ordering? The Builders that output to source should probably all be run before those writing to the cache. They also likely shouldn't be allowed to use `required_inputs`
- [x] Store `isHidden` on `GeneratedAssetNode`. Use this to determine where to read from
- [x] Add `hiddenOutput` to `BuildAction`. We may want to hide this until `BuildAction` is removed from the public interface.
- [x] Support setting `hiddenOutput` through `BuilderApplication`.
- [x] Support some configuration for `hide_outputs` in `build.yaml`
- [x] Deprecate then drop `writeToCache` argument, use the per-action setting instead

jakemac53:
SGTM, the only part I am not 100% sure on is enforcing ordering between source/cache dir outputs. I think the source dir outputs should typically go first but I think we will likely need the flexibility to support it either way.

natebosch:
IMO a Builder outputting to source directory should not be able to read generated outputs. I will need to add that restriction to my prototype.

If a to-source builder can't read the outputs of a to-hidden builder, but reads can go in the other direction, I don't think it would ever make sense to run a to-hidden builder first.

jakemac53:
Ideally transient files would always be generated in the hidden dir I think though, which would mean that to-source builders can't use transient files at all.

natebosch:
I think we can handle transient files differently and still allow reads to them. We can have an enum with `transient`, `hidden`, or `sourceDir` maybe?

jakemac53:
That could work.. but maybe its fine to just not allow it.

To me it makes sense that source dir outputs can't read other hidden dir outputs because the implication is that you are baking in some knowledge about how you currently see the world and shipping a file that depends on that, which could actually be different for some other user consuming your package. In that sense maybe it also makes sense to not allow reading of transient files... or maybe just not reading transient files from outside of your package but even those could end up with built-in dependencies on things that might change.

PR #692, opened 2017-12-07
Support mixed-mode building

natebosch:
Towards #647

- Add hideOutput to BuildAction
- Add hideOutput to BuilderApplication and forward when creating actions
- Add isHidden to GeneratedAssetNode and serialize it
- Update the BuildCacheReader/Writer to check per-node whether to
  translate to the build cache.
- Always wrap with the BuildCacheReader/Writer since they can choose not
  to translate as appropriate.
- When passed the arg `writeToCache` update all the actions to set
  `hideOutput`
- Add support for tagging an expected output with `$$` and update the
  way we translate expected output locations to use this tag. We can
  support having only some assets hidden in the tests.
- Switch from `writeToCache` to `hideOutputs` on BuildActions in tests
- Remove text references to `writeToCache`

Remaining work:
- Drop `writeToCache` argument entirely
- Add Builder config for this option and update build script generator

### DIFF
--- a/build_runner/lib/src/asset_graph/node.dart
+++ b/build_runner/lib/src/asset_graph/node.dart
@@ -7,6 +7,7 @@ import 'dart:collection';
 import 'package:build/build.dart';
 import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
+import 'package:meta/meta.dart';
 
 /// A node in the asset graph which may be an input to other assets.
 abstract class AssetNode {
@@ -91,12 +92,22 @@ class GeneratedAssetNode extends AssetNode {
   /// this node.
   final AssetId builderOptionsId;
 
-  GeneratedAssetNode(this.phaseNumber, this.primaryInput, this.needsUpdate,
-      this.wasOutput, AssetId id, this.builderOptionsId,
-      {Digest lastKnownDigest,
-      Set<Glob> globs,
-      Iterable<AssetId> inputs,
-      this.previousInputsDigest})
+  /// Whether the asset should be placed in the build cache.
+  final bool isHidden;
+
+  GeneratedAssetNode(
+    this.phaseNumber,
+    this.primaryInput,
+    this.needsUpdate,
+    this.wasOutput,
+    AssetId id,
+    this.builderOptionsId, {
+    Digest lastKnownDigest,
+    Set<Glob> globs,
+    Iterable<AssetId> inputs,
+    this.previousInputsDigest,
+    @required this.isHidden,
+  })
       : this.globs = globs ?? new Set<Glob>(),
         this.inputs = inputs != null
             ? new SplayTreeSet.from(inputs)
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -135,7 +135,7 @@ class _Loader {
         _onDelete);
   }
 
-  /// Checks that the [_buildActions] are valid based on whethere they are
+  /// Checks that the [_buildActions] are valid based on whether they are
   /// written to the build cache.
   void _checkBuildActions() {
     final root = _options.packageGraph.root.name;
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -120,13 +120,13 @@ class _Loader {
           () => _initialBuildCleanup(
               conflictingOutputs,
               _options.deleteFilesByDefault,
-              _maybeWrapWriter(_options.writer, assetGraph)));
+              _wrapWriter(_options.writer, assetGraph)));
     }
 
     return new BuildDefinition._(
         assetGraph,
-        _maybeWrapReader(_options.reader, assetGraph),
-        _maybeWrapWriter(_options.writer, assetGraph),
+        _wrapReader(_options.reader, assetGraph),
+        _wrapWriter(_options.writer, assetGraph),
         _options.packageGraph,
         _options.deleteFilesByDefault,
         new ResourceManager(),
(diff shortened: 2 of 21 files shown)


kevmoo:
Really?!?!?

natebosch:
Turned out to not be terribly hard

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -91,12 +92,22 @@ class GeneratedAssetNode extends AssetNode {
...
+    this.primaryInput,
+    this.needsUpdate,
+    this.wasOutput,
+    AssetId id,
+    this.builderOptionsId, {
+    Digest lastKnownDigest,
+    Set<Glob> globs,
+    Iterable<AssetId> inputs,
+    this.previousInputsDigest,
+    @required this.isHidden,
```
jakemac53: nit: I would just move this up with the other required args - if we want to go with @required we should do it for pretty much all the fields at once imo (same elsewhere)

build_runner/lib/src/generate/build_definition.dart
```diff
@@ -135,15 +135,14 @@ class _Loader {
         _onDelete);
   }
 
-  /// Checks that the [_buildActions] are valid based on the
-  /// `_options.writeToCache` setting.
+  /// Checks that the [_buildActions] are valid based on whethere they are
```
jakemac53: whether

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -50,15 +51,17 @@ Future<BuildResult> build(List<BuildAction> buildActions,
...
-      writeToCache: writeToCache,
+      deleteFilesByDefault: deleteFilesByDefault ?? writeToCache,
       packageGraph: packageGraph,
       reader: reader,
       writer: writer,
       logLevel: logLevel,
       onLog: onLog,
       skipBuildScriptCheck: skipBuildScriptCheck,
       enableLowResourcesMode: enableLowResourcesMode);
+  if (writeToCache ?? false) {
```
jakemac53: `== true` is more intuitive imo

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -50,15 +51,17 @@ Future<BuildResult> build(List<BuildAction> buildActions,
     bool skipBuildScriptCheck,
     bool enableLowResourcesMode}) async {
   var options = new BuildOptions(
-      deleteFilesByDefault: deleteFilesByDefault,
-      writeToCache: writeToCache,
+      deleteFilesByDefault: deleteFilesByDefault ?? writeToCache,
```
jakemac53: I would add an extra todo here or something to make sure we clean it up.

Just so I understand correctly - setting `writeToCache: true` effectively just makes all your build actions hidden (maintains old behavior?)

build_runner/test/asset_graph/graph_test.dart
```diff
@@ -86,7 +86,8 @@ void main() {
                 makeAssetId(), md5.convert(UTF8.encode('test')));
 
             var generatedNode = new GeneratedAssetNode(0, node.id, g % 2 == 1,
-                g % 2 == 0, makeAssetId(), builderOptionsNode.id);
+                g % 2 == 0, makeAssetId(), builderOptionsNode.id,
+                isHidden: false);
```
jakemac53: lets make this alternate back and forth like the others, maybe `g % 3 == 0`?

build_runner/test/common/test_phases.dart
```diff
@@ -121,19 +117,31 @@ void checkBuild(BuildResult result,
...
   if (exceptionMatcher != null) {
     expect(result.exception, exceptionMatcher);
   }
 
-  var mapAssetIds = writeToCache
-      ? (AssetId id) => new AssetId(
+  final unhiddenOutputs = <String, dynamic>{};
+  final unhiddenAssets = new Set<AssetId>();
+  for (final id in outputs?.keys ?? const []) {
+    if (id.startsWith(r'$$')) {
```
jakemac53: probably worth adding a comment to this function to call out this behavior

build_runner/test/generate/build_test.dart
```diff
@@ -226,7 +226,7 @@ void main() {
           throwsA(anything));
     });
 
-    group('with `writeToCache: true`', () {
+    group('with `hideOutput: true`', () {
```
jakemac53: Probably worth adding an additional test to this file that mixes hideOutput settings for different build actions

build_runner/lib/src/asset_graph/node.dart
natebosch: non-named boolean args are tough at the call sites.

Not all the named args here should be required but I can go through and find the ones that should be and mark them - and see if any of the non-named args should become named.
Can I follow up with that in a separate PR?

build_runner/lib/src/generate/build_impl.dart
natebosch: yes, the setting maintains behavior. I think this is covered by the TODO to remove the argument.

build_runner/lib/src/generate/build_impl.dart
natebosch: `== true` can cover either null or non-boolean types. I like `?? false` when I'm explicitly only worried about null but I know its a boolean.

I don't have strong feelings though, and this is going to go away soon so I changed it :)

build_runner/test/asset_graph/graph_test.dart
natebosch: Done

build_runner/lib/src/generate/build_definition.dart
natebosch: Done

build_runner/test/common/test_phases.dart
natebosch: Done

build_runner/test/generate/build_test.dart

build_runner/lib/src/asset_graph/node.dart
natebosch: https://github.com/dart-lang/build/pull/695

If you feel it should go out together I can merge it in here.

jakemac53: Whichever is easiest

PR #696, opened 2017-12-08
Only delete conflicting sources outside the cache

natebosch:
Since we restrict reads based on the AssetGraph we don't care whether
there are conflicts inside the build cache, we won't read them until
they have ben overwritten by the current build.

We should never have a situation where an 'internal source' conflicts
with an output since nothing in an 'internal' directory could ever be a
primary input.

- When looking for conflicts use inputSources rather than allSources
  since these are the only deletes we care about.
- Don't compute the `allSources` Set up front. Delay building the set
  until we're about to use it in `_findSourceUpdates`. On the first
  entry where the AssetGraph is null we avoid building it entirely.

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -114,10 +114,14 @@ class _Loader {
         final conflictsInDeps = assetGraph.outputs
             .where((n) => n.package != _options.packageGraph.root.name)
             .where(inputSources.contains)
-            .length;
-        _logger.warning('There are $conflictsInDeps sources in dependencies '
-            'that will be ignored since they are overwritten by '
-            'generated assets');
+            .toSet();
+        // TODO(https://github.com/dart-lang/build/issues/706) - stop special
+        // casing this conflict
+        conflictsInDeps
+            .remove((new AssetId('node_preamble', 'lib/preamble.js')));
+        if (conflictsInDeps.isNotEmpty) {
+          throw new UnexpectedExistingOutputsException(conflictsInDeps);
+        }
       });
 
       await logTimedAsync(
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -107,8 +107,17 @@ class _Loader {
             internalSources, _options.packageGraph.root.name, _options.reader);
         buildScriptUpdates =
             await BuildScriptUpdates.create(_options, assetGraph);
-        conflictingOutputs =
-            assetGraph.outputs.where(inputSources.contains).toSet();
+        conflictingOutputs = assetGraph.outputs
+            .where((n) => n.package == _options.packageGraph.root.name)
+            .where(inputSources.contains)
+            .toSet();
+        final conflictsInDeps = assetGraph.outputs
+            .where((n) => n.package != _options.packageGraph.root.name)
+            .where(inputSources.contains)
+            .length;
+        _logger.warning('There are $conflictsInDeps sources in dependencies '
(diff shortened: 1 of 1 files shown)


natebosch:
Only attempting to delete files in the source tree of the root package. Added a warning when there are files in dependencies which will be hidden by generated files.

An example is there is both a `preamble.dart` and `preamble.js` file in `package:node_preamble`

build_runner/lib/src/generate/build_definition.dart
```diff
@@ -110,8 +107,17 @@ class _Loader {
...
-            assetGraph.outputs.where(allSources.contains).toSet();
+        conflictingOutputs = assetGraph.outputs
+            .where((n) => n.package == _options.packageGraph.root.name)
+            .where(inputSources.contains)
+            .toSet();
+        final conflictsInDeps = assetGraph.outputs
+            .where((n) => n.package != _options.packageGraph.root.name)
+            .where(inputSources.contains)
+            .length;
+        _logger.warning('There are $conflictsInDeps sources in dependencies '
```
jakemac53: Should we log this as severe or possibly even fail the build? This means we are effectively allow you to overwrite sources otherwise...

natebosch: I'm not sure we'd want to fail the build. If somebody publishes a `file.js` next to to a `file.dart` they'd break their downstream customers. How would we let people get out of this situation?

Since most packages depend on `test`, and `test` depends on `node_preamble` every single build is going to hit this case.

jakemac53: By allowing it we are breaking one of the primary contracts we supposedly enforce though. Presumably the published js file has some real use and we will be overwriting it. 

Imo the solution is to either rename the file in the offending package or give people a way to rename the ddc file but that would be very difficult.

natebosch: We could potentially pick a more 'namespaced' extension and call these files `.ddc.js` to make the issue less likely.

bazel allows you to have files in the source tree which are not treated as sources and which you can effectively overwrite. We aren't introducing any non-determinism. The only version of the files the build will ever see is the generated version.

With the way our discussions around things like package cycles have gone I think we're going to need to look for a solution which doesn't involve upstream package owners needing to care about the details of your build.

We could potentially look at some pattern where you need to opt into 'hiding' source files with a whitelist - but it worries me to say that every single customer will need that whitelist, because based on this package every single customer would.

jakemac53: We could use a more namespaced ddc js file - that would require custom module config for all modules in require.js, but that should not cause any major issues. With other module systems it might be harder - so it somewhat ties us to the more customizable module loaders. Or we would need more options from ddc with respect to module naming.

Either way I don't think the solution is to effectively allow overwriting source files (especially only in dependencies).

natebosch: I think the warning is an improvement over our current situation where we `deleteFilesByDefault` and just ignores the delete in non-root packages and we end up ignoring the file anyway.

How about we move forward with this approach for now and file an issue to figure out a stricter approach later?

natebosch: https://github.com/dart-lang/build/issues/705

jakemac53: sure sgtm

natebosch: I rethought this a bit, and going back to our goals of starting as strict as possible and only loosening things when we find real uses cases I think we can hardcode this one known common conflict for now and restrict others. WDYT?

PR #697, opened 2017-12-08
Add `build_to` Builder config option

natebosch:
Towards #647

The value is defaulted based on the `auto_apply` option, since for
certain values the only possibility is `True`.

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -40,11 +40,14 @@ the following keys:
 - **is_optional**: Optional, boolean. Specifies whether a Builder can be run
   lazily, such that it won't execute until one of it's outputs is requested by a
   later Builder. This option should be rare. Defaults to `False`.
-- **hide_output**: Optional, boolean. Specifies whether the outputs produced by
-  the Builder should go to the build cache rather than the source tree. Only
-  builders which hide output may run on primary inputs outside the root
-  package. Defaults to `False`, unless `auto_apply` is set to either
-  `"all_packages"` or `"dependents"` in which case it defaults to `True`.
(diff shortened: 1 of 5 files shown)


natebosch:
`build_to_cache`?

grouma:
Doesn't it always populate the cache? If not, the build_to_cache works.

natebosch:
We now support mixing builders which write to the cache (angular, ddc) with builders that write directly to the source tree (built_value, etc). Only the files which are generated end up in the build cache, we don't write source files there.

I'm a little torn on the naming. I agree that `build_to_cache` is a more direct map to what will happen. But `hide_output` is, I think, a more direct map to the goal the builder author has...

natebosch:
After a quick discussion I think I might change this to an enum-like value with two choices:

`build_to: source` and `build_to: cache`

Thoughts?

grouma:
I prefer the enum approach.

jakemac53:
I like the enum approach as well

natebosch:
Made it an enum `BuildTo` with options `source` and `cache`. PTAL

build_config/README.md
```diff
@@ -40,6 +40,11 @@ the following keys:
 - **is_optional**: Optional, boolean. Specifies whether a Builder can be run
   lazily, such that it won't execute until one of it's outputs is requested by a
   later Builder. This option should be rare. Defaults to `False`.
+- **hide_output**: Optional, boolean. Specifies whether the outputs produced by
```
jakemac53: update this note and chanegelog

PR #704, opened 2017-12-12
Add `--assume-tty` flag

natebosch:
We have no way to fool a subprocess into thinking it's running in a
terminal, add a flag to behave that way anyway.

- Add a flag `--assume-tty` and pass it when the outer layer appears to
  be in a TTY.
- Switch to `packag:io/ansi.dart` escape codes for colors.
- Force `ansiOutputEnabled` based on `--assume-tty`.
- When checking whether we can ask for input, read `assumeTty`.


build_runner/lib/src/generate/build.dart
```diff
@@ -83,6 +83,7 @@ Future<BuildResult> build(List<BuildAction> buildActions,
 ///  typically cause a shutdown).
 Future<ServeHandler> watch(List<BuildAction> buildActions,
         {bool deleteFilesByDefault,
+        bool assumeTty,
```
jakemac53: we should plumb this through `build` as well

build_runner/lib/src/logging/std_io_logging.dart
```diff
@@ -5,27 +5,22 @@
...
-    color = _yellow;
+    color = yellow;
   } else {
-    color = _red;
+    color = red;
   }
-  var header = '${_isPosixTerminal ? '\x1b[2K\r' : ''}'
-      '$color[${record.level}]$_endColor ${record.loggerName}: '
+  final level = color.wrap('[${record.level}]');
+  var header = '${ansiOutputEnabled ? '\x1b[2K\r' : ''}'
```
jakemac53: Note that I think `ansiOutputEnabled` is actually using different logic than `_isPosixTerminal` was but it probably doesn't matter (I think it might be more correct anyways).

build_runner/lib/src/generate/build.dart

build_runner/lib/src/logging/std_io_logging.dart
natebosch: Yeah it's definitely different. I'm not sure it's more correct (it failed on my terminal where the old logic succeeded https://github.com/dart-lang/sdk/issues/31606) but it's at least more standard in Dart...

PR #707, opened 2017-12-12
Drop writeToCacheArgument

natebosch:
Toward #647

- Remove the argument everywhere
- Use `apply` rather than `applyToRoot` from the manual `build.dart`
  file in `e2e_example` since it allows specifying `hideOutput`.


e2e_example/tool/build.dart
```diff
@@ -14,8 +14,11 @@ import 'package:shelf/shelf_io.dart' as shelf_io;
 Future main() async {
   var graph = new PackageGraph.forThisPackage();
   var builders = [
-    applyToRoot(new TestBootstrapBuilder(), inputs: ['test/**_test.dart']),
-    applyToRoot(new ThrowingBuilder()),
+    apply('e2e_example', 'throwing_builder', [(_) => new ThrowingBuilder()],
```
jakemac53: is `applyToRoot` gone as well?

also this should probably be setting `hideOutput: true` even though it never outputs anything

jakemac53: Oh sorry missed that part in the description - we should add `hideOutput: true` to `applyToRoot` though right?

Actually - shouldn't it only exist for `applyToRoot`?

natebosch: `applyToRoot` isn't gone, but doesn't allow setting `hideOutput`.

I'd be fine adding that config if we think we'll need it, but `applyToRoot` was our sort of backwards-compatible approach for people who are used to using manual scripts with `built_value` and others, so in theory no one who is likely to use it should need `hideOutput`.

natebosch: `apply` is generally what we'll have with builders configured by `build.yaml`. `applyToRoot` should only be used in cases where they were using a manual build script with previous versions of `build_runner`.

jakemac53: Ah ok - what I was really getting at is that calling `apply` on a non-root package with `hideOutput: false` shouldn't be allowed (right?). I don't know what we do today in that case.

natebosch: Today we do all the defaulting in `build_config` and expect the `apply` call to be correct. We don't throw until you try to call `watch` or `build` when we check the `BuildAction` instances.

It would, in theory, be allowed to call with `toDependentsOf('somePackage')` without setting `hideOutput: true` as long as only the root package has the dependency.

We could also file an issue to do more validation and defaulting here.

jakemac53: sgtm

Issue #709, opened 2017-12-13
Add resolveSources (plural) to pkg/build_test


matanlurey:
https://github.com/dart-lang/build/blob/ad197837da3f45e49b9c1079447392f575b2a734/build_test/lib/src/resolve_source.dart#L56-L58

This is needed to write tests that will assume multiple on-disk files, instead of just a single one.

A simple `Map<String, String>` probably suffices.

/cc @natebosch PTAL. If agreed I'd like to add this asap for Angular.

natebosch:
SGTM. We should use the same format for the keys as we do in other places: `package_name|path`.

Do we want to deprecate and remove `resolveSource` or keep it as a special case of `resolveSources({inputId: inputSource}, ...)`?

matanlurey:
I honestly like it for the simplified case, and it would be a good candidate to just be short-hand for `{inputId: inputSource}`. Does that make sense?

natebosch:
works for me.

matanlurey:
One snag, currently you can "just use the real asset" by not providing `inputSource`.

That makes `Map<String, String>` a little weird, because it's not clear how to express "provide this asset, but use the real on-disk source". I use this feature quite a bit for things like Angular.

Alternatives:

1. Provide a more typed `Map<AssetId, AssetContents>`, where you can do:

```dart
resolveSources({
  new AssetId('angular', 'lib/angular.dart'): AssetContents.useAssetReader,
  new AssetId('test_lib', 'lib/test_lib.dart'): new AssetContents(r'''
    import 'package:angular/angular.dart';

    ...
  '''),
})
``` 

2. Add some sort of enum-like string to indicate "use real source":

```dart
resolveSources({
  'angular|lib/angular.dart': useAssetReader /* top-level const String */ ,
  'test_lib|lib/test_lib.dart': r''' ... ''',
})
```

3. Start pivoting to some `test_descriptor`-based extension/API instead.

natebosch:
I like option 2. Since this is test code I generally favor terseness over closing type holes.

matanlurey:
Sounds good. Easy enough to pivot if this becomes some super-used package/pattern.

PR #711, opened 2017-12-13
Start supporting resolveSources (plural).

matanlurey:
First part of https://github.com/dart-lang/build/issues/709.

I included a strong-mode fix for `InMemoryAssetReader`, which was trying to treat a `Map<AssetId, dynamic>` _as_ a `Map<AssetId, List<int>>`, which will fail in Dart 2.0.

### DIFF
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -148,14 +148,14 @@ Future<T> _resolveAssets<T>(
   final assetReader = new PackageAssetReader(syncResolver, rootPackage);
   final resolveBuilder = new _ResolveSourceBuilder(action, tearDown);
   final inputAssets = <AssetId, String>{};
-  await Future.forEach(inputs.keys, (String rawAssetId) async {
+  await Future.wait(inputs.keys.map((String rawAssetId) async {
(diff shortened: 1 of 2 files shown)


build_test/lib/src/resolve_source.dart
```diff
@@ -97,32 +115,61 @@ Future<T> resolveSource<T>(
...
+  String rootPackage,
+  FutureOr<T> action(Resolver resolver), {
+  PackageResolver resolver,
+  Future<Null> tearDown,
+}) async {
+  final syncResolver = await (resolver ?? PackageResolver.current).asSync;
+  final assetReader = new PackageAssetReader(syncResolver, rootPackage);
+  final resolveBuilder = new _ResolveSourceBuilder(action, tearDown);
+  final inputAssets = <AssetId, String>{};
+  await Future.forEach(inputs.keys, (String rawAssetId) async {
```
natebosch: Is there a reason we need to do this serially instead of allowing multiple file reads to start?

matanlurey: Nope! Changed to `Future.wait`!

matanlurey:
Ignoring pub flake and merging.

PR #714, opened 2017-12-13
Drop `BuildAction` from the public API

natebosch:
- Change `build` and `watch` to take a `List<BuilderApplication>`.
- Don't export `createBuildAction` or `BuildAction`.
- Change most tests to use `BuildApplication`.
- Delete `PackageBuilder` and `PackageBuildAction` altogether since this
  was unused and we will want a different API for this if an when we
  support it using the `BuilderApplication` style argument.


build_runner/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Breaking Changes
 
+- Remove `BuildAction`. Change `build` and `watch` to take a
```
jakemac53: nit: s/Remove/Removed and s/Change/Changed.

PR #717, opened 2017-12-14
Add InMemoryAssetReader.shareAssetCache constructor

natebosch:
When we started copying asset rather than continuing to use the same Map
instance it broke an undocumented behavior that we relied on in some
tests - that we could automatically keep new written assets reflected by
the writer.

Keep the strong mode types accurate and add a new contstructor which
solves this use case specificall.


build_runner/test/common/in_memory_reader.dart
```diff
@@ -12,9 +12,15 @@ import 'package:glob/glob.dart';
 /// Workaround class for mixin application limitations, if
 /// [InMemoryRunnerAssetReader] extends [InMemoryAssetReader] directly then it
 /// can't use mixins ([Md5DigestReader] in this case).
+///
+/// Classes with Mixings can't call a `super` constructor with named arguments.
```
matanlurey: with mixins

PR #719, opened 2017-12-14
Add `CopyBuilder.builds` stream

natebosch:
We have some tests which need to wait for setup to be finished. The
current pattern of `await new Future(() {})` is unsafe if the setup may
take extra trips around the event loop. Instead get a concrete signal of
the the Builder has started doing work.

### DIFF
--- a/build_runner/test/generate/watch_test.dart
+++ b/build_runner/test/generate/watch_test.dart
@@ -208,13 +208,14 @@ void main() {
 
       test('rebuilds on file updates during first build', () async {
         var blocker = new Completer<Null>();
-        var buildAction =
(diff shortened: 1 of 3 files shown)


natebosch:
Yeah in #400 we had discussed `Stream<Future<BuildResult>>` and similar options which would solve this use case. Since I technically only need the first build I think I *could* use `BuildState.currentBuild` for this particular situation... Do you want me to investigate that route?

jakemac53:
I think this is fine for now - we can investigate later if it comes up
again.

natebosch:
Heres what the alternative looks like: https://github.com/dart-lang/build/pull/722

I actually think I like that a bit better. Will close this for now.

PR #721, opened 2017-12-14
Allow constructing BuildConfig with known deps

natebosch:
Rather than reparse the Pubspec add a constructor that takes know
package name and dependencies.

Don't export `Pubspec` class and delete `BuildConfigSet` since they
don't have use cases outside this package.


build_config/lib/src/build_config.dart
```diff
@@ -58,18 +58,34 @@ class BuildConfig {
   static const _isOptional = 'is_optional';
   static const _buildTo = 'build_to';
 
-  /// Returns a parsed [BuildConfig] file in [path], if one exists.
+  /// Returns a parsed [BuildConfig] file in [path], if one exist, otherwise a
+  /// default config.
   ///
-  /// Otherwise uses the default setup.
-  static Future<BuildConfig> fromPackageDir(Pubspec pubspec, String path,
+  /// [path] must be a directory chich contains a `pubspec.yaml` file and
```
jakemac53: which

build_config/lib/src/build_config.dart
```diff
@@ -58,18 +58,34 @@ class BuildConfig {
...
+    final pubspec = await Pubspec.fromPackageDir(path);
+    return fromBuildConfigDir(
+        pubspec.pubPackageName, pubspec.dependencies, path,
+        includeWebSources: includeWebSources);
+  }
+
+  /// Returns a parsed [BuildConfig] file in [path], if one exists, otherwise a
+  /// default config.
+  ///
+  /// [path] maybe be a directory which contains a `build.yaml`.
```
jakemac53: I would reword this a bit, maybe "[path] should be the path to a directory which may contain a `build.yaml`"

jakemac53: We could also consider checking that this is a Directory and throwing a nice error if not

build_config/lib/src/build_config.dart

build_config/lib/src/build_config.dart
natebosch: I think for now I'll avoid the extra check. I'm not sure if we're depending on allowing bad input. I can follow up with a later change.

build_config/lib/src/build_config.dart
```diff
@@ -74,7 +74,7 @@ class BuildConfig {
   /// Returns a parsed [BuildConfig] file in [path], if one exists, otherwise a
   /// default config.
   ///
-  /// [path] maybe be a directory which contains a `build.yaml`.
+  /// [path] shouldd the path to a directory which may contain a `build.yaml`.
```
jakemac53: should be

natebosch: You'd think typing wouldn't be the hard part...

PR #724, opened 2017-12-15
Add equality and hashCode on BuildAction

natebosch:
This will let us do further refactoring within BuildAction and InputSet
without impacting other usages.

Also add more detail to the BuildAction toString.

### DIFF
--- a/build_runner/lib/src/generate/phase.dart
+++ b/build_runner/lib/src/generate/phase.dart
@@ -79,7 +79,7 @@ class BuildAction {
       other is BuildAction &&
           // Risky, but we don't want to force all Builders to implement a sane
           // hashcode/equals
-          other.builder.runtimeType == builder.runtimeType &&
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/generate/phase.dart
```diff
@@ -63,5 +64,28 @@ class BuildAction {
...
+    final settings = <String>[];
+    if (isOptional) settings.add('optional');
+    if (hideOutput) settings.add('hidden');
+    var result = '$builder on $inputSet';
+    if (settings.isNotEmpty) result += ' $settings';
+    return result;
+  }
+
+  @override
+  bool operator ==(Object other) =>
```
jakemac53: This should be doing some sort of check for the builder as well - same for the hashCode. Previously this was just checking `runtimeType.toString().hashCode`, and I would like to maintain at least that much.

natebosch: ugh, how did I miss that!

PR #725, opened 2017-12-15
Hide the InputSet in BuildAction

natebosch:
This will make it easier to refactor the way we discover sources for an
action so that we can make it match both a `target` input set, as well
as an optional `filter` for a particular builder within a target.

- Make `InputSet` an abstract class with a `matches` method.
- Make `BuildAction` implement `InputSet` so that the particular way
  assets can match is hidden.
- Make the default implementation of `InputSet` a private class that has
  the include/exclude behavior.
- Rename some named arguments to be a closer match to the `glob` utility
  from bazel.


natebosch:
The diff here will get a lot smaller once https://github.com/dart-lang/build/pull/723 and https://github.com/dart-lang/build/pull/724 are merged.

build_runner/lib/src/generate/input_set.dart
```diff
@@ -6,67 +6,66 @@ import 'package:build/build.dart';
 import 'package:collection/collection.dart';
 import 'package:glob/glob.dart';
 
-final List<Glob> _defaultInclude = new List.unmodifiable([new Glob('**')]);
+/// A set of file paths to match inputs to for a builder.
```
jakemac53: We should update this - it doesn't really match what the class represents any more.... in fact the class should probably be renamed as well - `InputMatcher`?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -131,8 +131,8 @@ Iterable<BuildAction> _createBuildActionsForBuilderInCycle(
       .where(builderApplication.filter)
       .map((p) => new BuildAction(b(options), p.name,
           builderOptions: options,
-          inputs: builderApplication.inputs,
-          excludes: builderApplication.excludes,
+          include: builderApplication.inputs,
```
jakemac53: we should probably also update the names of the fields on `builderApplication`

natebosch: https://github.com/dart-lang/build/issues/726

build_runner/lib/src/generate/input_set.dart
natebosch: Done

PR #734, opened 2017-12-18
Clean up previous generated files before rebuilding

jakemac53:
Also added `buildInputs` stream to `CopyBuilder` which is similar to some previous efforts but allows waiting for specific files to be built which is more explicit.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -286,6 +286,9 @@ class BuildImpl {
     if (!await _buildShouldRun(builderOutputs, wrappedReader)) {
       return <AssetId>[];
     }
+
+    await _cleanUpStaleOutputs(builderOutputs);
+
     // We may have read some inputs in the call to `_buildShouldRun`, we want
     // to remove those.
     wrappedReader.assetsRead.clear();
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -341,6 +344,18 @@ class BuildImpl {
     }
   }
 
+  /// Deletes any of `outputs` which previously were output.
```
natebosch: Surround `outputs` with square braces instead of backticks?

build_test/lib/src/copy_builder.dart
```diff
@@ -30,6 +30,12 @@ class CopyBuilder implements Builder {
   /// are passed which do not match the input extension then [build] will throw.
   final String inputExtension;
 
+  /// A stream of all the [BuildStep.inputId]s that are seen.
+  ///
+  /// Events are added at the top of the [build] method.
+  final _buildInputsController = new StreamController<AssetId>();
```
natebosch: should this be a broadcast stream?

build_runner/lib/src/generate/build_impl.dart
jakemac53: done

build_test/lib/src/copy_builder.dart
jakemac53: good call, done

PR #739, opened 2017-12-18
Add `BuildConfig` unnamed constructor

natebosch:
- Separate out parsing into a separate library to reduce noise
- Add a default constructor with named args for the fields.


build_config/lib/src/parse.dart
```diff
@@ -0,0 +1,301 @@
...
+const _builderFactories = 'builder_factories';
+const _import = 'import';
+const _buildExtensions = 'build_extensions';
+const _target = 'target';
+const _autoApply = 'auto_apply';
+const _requiredInputs = 'required_inputs';
+const _isOptional = 'is_optional';
+const _buildTo = 'build_to';
+
+BuildConfig parse(
```
jakemac53: nit: this is a bit generic for a top level method... `parseFromYaml` or `parseConfig`?

PR #740, opened 2017-12-18
Add BuilderConfig class

natebosch:
- Add a `BuilderConfig` class to hold per-target builder configuration
- Move `generate_for` from per-target to per-builder
- Change the `builders` value from a `List<String|Map>` to a `Map`
- Add a Matcher for `BuilderConfig` and map of configs
- Add some `toString` implementations to make test failures a little
  easier to fix
- Update test yaml and built values to match new syntax.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:build/build.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
 
@@ -167,8 +168,10 @@ class BuildTarget {
 
   final Iterable<String> sources;
 
-  /// A map from builder name to the configuration used for this target.
-  final Map<String, Map<String, dynamic>> builders;
+  /// A map from builder key to the configuration used for this target.
+  ///
+  /// Builder keys are in the format `"$package|$builder"`.
+  final Map<String, BuilderConfig> builders;
 
   /// The platforms supported by this target.
   ///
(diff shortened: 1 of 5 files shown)


build_config/lib/src/build_config.dart
```diff
@@ -179,18 +182,59 @@ class BuildTarget {
...
+        'builders': builders
+      }.toString();
+}
+
+/// The configuration a particular [BuildTarget] applies to a Builder.
+///
+/// Build targets may have builders applied automatically based on
+/// [BuilderDefinition.autoApply] and may override with more specific
+/// configuration.
+class BuilderConfig {
```
jakemac53: `BuilderConfig` and `BuilderOptions` are quite similar names... maybe `BuildTargetConfig`?

build_config/lib/src/build_config.dart
```diff
@@ -179,18 +182,59 @@ class BuildTarget {
...
+    this.isEnabled,
+    this.generateFor,
+    this.options: const BuilderOptions(const {}),
+  });
+
+  @override
+  String toString() => {
+        'isEnable': isEnabled,
+        'generateFor': generateFor,
+        'options': options?.config
```
jakemac53: This shouldn't be null right (we have a default)

build_config/lib/src/build_config.dart
```diff
@@ -179,18 +182,59 @@ class BuildTarget {
...
+
+  /// Sources to use as inputs for this Builder in glob format.
+  ///
+  /// This is always a subset of the `include` argument in the containing
+  /// [BuildTarget].
+  ///
+  /// May be `null`, in which case it should fall back on `sources`.
+  final Iterable<String> generateFor;
+
+  final BuilderOptions options;
```
jakemac53: lets add a comment here, mentioning it comes from the `options` argument

build_config/lib/src/build_config.dart
natebosch: Went with `TargetBuilderConfig` as discussed

build_config/lib/src/build_config.dart

build_config/lib/src/build_config.dart

PR #743, opened 2017-12-18
Allow omitting 'default' if there is 1 target

natebosch:


build_config/lib/src/parse.dart
```diff
@@ -111,6 +111,20 @@ BuildConfig parseFromYaml(
         name: packageName,
         package: packageName,
         sources: sources);
+  } else if (buildTargets.length == 1 &&
+      !buildTargets.values.single.isDefault) {
+    // Allow omitting `isDefault` if there is exactly 1 target.
+    var nonDefaultTarget = buildTargets.values.single;
+    buildTargets[buildTargets.keys.single] = new BuildTarget(
```
jakemac53: I worry a bit about this getting out of date if we add new optional args..... what if we add a constructor to `BuildTarget` which takes an existing one and lets you override things? At least it would be more colocated with the main constructor so we would be more likely to update it.

PR #745, opened 2017-12-19
Use extensions to order builders

natebosch:
We will be removing the `inputs` argument to `apply` and move to
`BuildTarget` to filter inputs to Builders. In each of these tests we're
mostly using the `inputs` to limit what previous outputs get read which
is not a typical use case for a build target. A closer match to this is
having intermedaite file extensions.

Where possible switch to the simpler `applyToRoot` rather than `apply`
with arguments we expect to be ignored.


natebosch:
Pushing this early so that a future PR is less noisy.

Issue #710, opened 2017-12-13
Proposal: Move the `digest` method to the `AssetReader` abstract class


jakemac53:
cc @natebosch wdyt?

Today in `build_runner` we have a `DigestAssetReader` class which is used internally for computing digests. It adds the `Future<Digest> digest(AssetId id)` method, which today is used primarily for tracking changes to files.

I would like to move that to the abstract `AssetReader` class - including a default implementation of roughly `md5.convert(await readAsBytes(id))`.

## Why?

Builders could make use of this - specifically any Builder using the bazel worker protocol to talk to a subprocess can use it to efficiently provide a list of inputs and their digests. The analyzer cli today will use that if available to optimize loading of only the changed summaries, and ddc should likely be updated to do the same.

`Resource`s could also use this to do other forms of smart caching/invalidation. For instance the `ScratchSpace` package could use it to do less copying, today it starts over at the start of each build since it doesn't have efficient access to this information.

## Risks

The main risk here is we are opening the door a bit to non-hermetic builds. You have to trust anybody using these Digests to do the right thing. Bazel already does allow this though - and I think it is a reasonable tradeoff.

## When?

I would like to roll this into the other breaking changes in v0.7.0.

natebosch:
Do we anticipate other implementations than `md5.convert(await readAsBytes(id))`?

> I would like to roll this into the other breaking changes in v0.7.0.

This would be breaking in `package:build` v0.12.0, not breaking in `build_runner`

matanlurey:
This sounds reasonable :).

Ultimately we might want tooling to help users find if builds are non-hemertic (imagine a "checked" mode for build_runner), similar to the tooling @davidmorgan build around Bazel internally, but this sounds good to me!

jakemac53:
> Do we anticipate other implementations than md5.convert(await readAsBytes(id))?

Yes - in bazel_codegen we should use the input hashes provided by blaze when in worker mode.

jakemac53:
> This would be breaking in package:build v0.12.0, not breaking in build_runner

Oh yes this would additionally be a breaking change in build_runner though because we expose `DigestAssetReader` today.

natebosch:
It would be good to have a better idea of what benefit we'll see. I think this should be fine for us, my only concern as we discussed offline is that this increases the surface area of `BuildStep` with an 'experts only' method which is easy to misuse and will require a lot of cautionary documentation.

jakemac53:
[...]

Issue #689, opened 2017-12-06
Remove some of the `is` checks against AssetNode subtypes


natebosch:
Knowledge of the node types and their behavior is spread out a bit today - we should try to cut that back.

Filing as a tracking issue so we can clean up here and there.

Some cases should stay - asset graph serialization and inspection for instance. Hopefully most of the rest we can drop.

natebosch:
At least some of these have been addressed. We've talked about potentially doing a bigger refactor that splits the AssetGraph into a 'static' and 'dynamic' pieces, but for now I don't think there is more to do on this issue.

PR #746, opened 2017-12-19
Add magic placeholder assets under lib, web, and test

jakemac53:
These assets are available to use as primary inputs, but cannot be read (and will give an `AssetNotFoundException` if you do try to read them). They are represented by a new type of node, `PlaceholderAssetNode`.

Note that this is a potentially breaking change for builders which have over-broad input extensions (looking at you `CopyBuilder`). These builders may attempt to read the input and throw an exception, but I don't think this is a major issue since it should be a very uncommon case. It may however break existing tests for people using `CopyBuilder` (it broke a bunch of ours).

I also added a bunch of boolean getters to describe functionality of the different asset node types, and replaced is checks where possible with those.

Closes https://github.com/dart-lang/build/issues/710

### DIFF
--- a/build_runner/bin/create_merged_dir.dart
+++ b/build_runner/bin/create_merged_dir.dart
@@ -72,7 +72,7 @@ Future main(List<String> args) async {
       assetGraph, packageGraph.root.name);
 
   for (var node in assetGraph.allNodes) {
-    if (node is SyntheticAssetNode) continue;
+    if (!node.isReadable) continue;
     if (node is GeneratedAssetNode && !node.wasOutput) continue;
     if (node.id.path == '.packages') continue;
     var assetPaths = <String>[];
--- a/build_runner/lib/src/asset_graph/node.dart
+++ b/build_runner/lib/src/asset_graph/node.dart
@@ -27,7 +27,33 @@ abstract class AssetNode {
   /// exist.
   Digest lastKnownDigest;
 
+  /// Whether or not this node was an output of this build.
+  bool get isGenerated => false;
+
+  /// Whether or not this asset can be read.
+  bool get isReadable => true;
+
+  /// Whether or not this node can be used as a primary or secondary input.
+  bool get isValidInput => true;
+
+  /// Whether or not changes to this node will have any effect on other nodes.
+  ///
+  /// Be default, if we haven't computed a digest for this asset and it has no
+  /// outputs, then it isn't interesting.
+  ///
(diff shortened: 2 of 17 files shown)


natebosch:
This also addresses some of #689

build_runner/bin/create_merged_dir.dart
```diff
@@ -72,7 +72,7 @@ Future main(List<String> args) async {
       assetGraph, packageGraph.root.name);
 
   for (var node in assetGraph.allNodes) {
-    if (node is SyntheticAssetNode) continue;
+    if (!node.isReadable) continue;
     if (node is GeneratedAssetNode && !node.wasOutput) continue;
```
natebosch: I'm curious why this isn't covered by `isReadable`.

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -27,8 +27,34 @@ abstract class AssetNode {
   /// exist.
   Digest lastKnownDigest;
 
+  /// Whether or not this node was an output of this build.
+  bool get isGenerated => false;
+
+  /// Whether or not this asset can be read.
```
natebosch: "Whether or not this asset type can be read.

This does not indicate whether the node is readable at this moment in time."

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -27,8 +27,34 @@ abstract class AssetNode {
...
+  bool get isReadable => true;
+
+  /// Whether or not this node can be used as a primary or secondary input.
```
natebosch: Expand this, mention that some internal assets can be read but can't be used as primary inputs.

natebosch: what is a secondary input?

jakemac53:
working on changelog updates as well now

build_runner/bin/create_merged_dir.dart
natebosch: Ah we discussed this and I forgot to delete the comment. Ignore :)

build_runner/lib/src/asset_graph/node.dart

build_runner/lib/src/asset_graph/node.dart

jakemac53: any input that isn't primary (something you accessed through `buildStep`)

jakemac53: updated this to only mention primary inputs as that was actually the intention behind it

PR #748, opened 2017-12-19
Allow omitting `sources` and `dependencies`

natebosch:
A common use case will be a single target for a package which is only
used to configure builders. To make this case easier set sane defaults
for dependencies and sources assuming a single target.

### DIFF
--- a/build_config/lib/src/parse.dart
+++ b/build_config/lib/src/parse.dart
@@ -292,7 +292,7 @@ List<String> _readListOfStringsOrThrow(
     {Iterable<String> defaultValue,
     Iterable<String> validValues,
     bool allowNull: false}) {
-  var value = options[option] ?? defaultValue.toList();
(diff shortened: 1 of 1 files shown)


build_config/lib/src/parse.dart
```diff
@@ -53,8 +53,8 @@ BuildConfig parseFromYaml(
     parseFromMap(packageName, dependencies,
         loadYaml(configYaml) as Map<String, dynamic>);
 
-BuildConfig parseFromMap(String packageName, Iterable<String> dependencies,
-    Map<String, dynamic> config) {
+BuildConfig parseFromMap(String packageName,
+    Iterable<String> packageDependencies, Map<String, dynamic> config) {
```
grouma: Let's add an assert to validate that this is not null so it won't blow up on the toList call below.

PR #750, opened 2017-12-19
Split out BuildActions by BuildTarget

natebosch:
This allows builders to be targeted more specifically within a package.

- Add a `TargetGraph` which mirrors the `PackageGraph` but my have more
  fine grained nodes.
- Remove `inputs` and `excludes` from `BuilderApplication`. This is now
  controlled by the `BuildTarget`s within a package.
- Add `defaultGenerateFor` to `BuildeApplication` since builders can
  suggest reasonable defaults to reduce the boilerplat config for most
  packages.
- Make `createBuildActions` return a `Future` since it reads
  `build.yaml` files.
- Honor the default `generate_for` when generating build scripts.
- Add `auto_apply` and `generate_for` for the DDC bootstrap builder.
  Remove the hardcoded default `apply` call from the generated build
  script.

### DIFF
--- a/build_runner/lib/src/generate/input_matcher.dart
+++ b/build_runner/lib/src/generate/input_matcher.dart
@@ -12,6 +12,8 @@ abstract class InputMatcher {
   bool matches(AssetId input);
   factory InputMatcher({Iterable<String> include, Iterable<String> exclude}) =
       _GlobInputMatcher;
+  factory InputMatcher.allOf(Iterable<InputMatcher> matchers) =>
+      new _MultiMatcher(matchers.toList());
 }
 
 class _GlobInputMatcher implements InputMatcher {
@@ -61,12 +63,33 @@ class _GlobInputMatcher implements InputMatcher {
   bool operator ==(Object other) =>
       identical(this, other) ||
       (other is _GlobInputMatcher &&
-          _deepEquals.equals(_patterns(include), _patterns(exclude)) &&
+          _deepEquals.equals(_patterns(include), _patterns(other.include)) &&
           _deepEquals.equals(_patterns(exclude), _patterns(other.exclude)));
 
   @override
   int get hashCode =>
       _deepEquals.hash([_patterns(include), _patterns(exclude)]);
+}
+
+class _MultiMatcher implements InputMatcher {
+  final List<InputMatcher> delegates;
+
+  _MultiMatcher(this.delegates);
+
+  @override
+  matches(AssetId input) => delegates.every((d) => d.matches(input));
+
+  @override
+  String toString() => 'All of $delegates';
+
+  @override
+  bool operator ==(Object other) =>
+      identical(this, other) ||
+      (other is _MultiMatcher &&
+          _deepEquals.equals(delegates, other.delegates));
+
+  @override
+  int get hashCode => _deepEquals.hash(delegates);
 }
 
 final _deepEquals = const DeepCollectionEquality();
--- a/build_runner/lib/src/package_graph/apply_builders.dart
+++ b/build_runner/lib/src/package_graph/apply_builders.dart
@@ -2,11 +2,15 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+
 import 'package:build/build.dart';
+import 'package:build_config/build_config.dart';
 import 'package:graphs/graphs.dart';
 
 import '../generate/phase.dart';
 import 'package_graph.dart';
+import 'target_graph.dart';
 
 typedef bool PackageFilter(PackageNode node);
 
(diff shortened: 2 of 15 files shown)


build_runner/lib/src/generate/input_matcher.dart
```diff
@@ -61,14 +63,35 @@ class _GlobInputMatcher implements InputMatcher {
...
-          _deepEquals.equals(_patterns(include), _patterns(exclude)) &&
+          _deepEquals.equals(_patterns(include), _patterns(other.include)) &&
           _deepEquals.equals(_patterns(exclude), _patterns(other.exclude)));
 
   @override
   int get hashCode =>
       _deepEquals.hash([_patterns(include), _patterns(exclude)]);
 }
 
+class _MultiMatcher implements InputMatcher {
```
jakemac53: nit: add a comment indicating this takes the intersection of the delegate matchers

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -34,33 +38,28 @@ PackageFilter toAll(Iterable<PackageFilter> filters) =>
 PackageFilter toRoot() => (p) => p.isRoot;
 
 /// Apply [builder] to the root package.
-BuilderApplication applyToRoot(Builder builder,
-        {List<String> inputs, List<String> excludes}) =>
-    new BuilderApplication._('', '', [(_) => builder], toRoot(),
-        inputs: inputs, excludes: excludes);
+BuilderApplication applyToRoot(Builder builder) =>
+    new BuilderApplication._('', '', [(_) => builder], toRoot());
```
jakemac53: I know this hasn't changed but I am a bit curious as to why we don't need unique builder keys for these builder applications?

build_runner/lib/src/package_graph/target_graph.dart
```diff
@@ -0,0 +1,52 @@
...
+  /// All [TargetNode]s indexed by `"$packageName:$targetName"`.
+  final Map<String, TargetNode> allModules;
+
+  TargetGraph._(this.allModules);
+
+  static Future<TargetGraph> forPackageGraph(PackageGraph packageGraph,
+      {Map<String, BuildConfig> overrideBuildConfig}) async {
+    overrideBuildConfig ??= const {};
+    final modules = <String, TargetNode>{};
+    for (final package in packageGraph.allPackages.values) {
```
jakemac53: nit: consider `Future.wait` to speed this up a bit

build_runner/test/generate/build_test.dart
```diff
@@ -59,7 +60,8 @@ void main() {
         await testBuilders([
           apply('', '', [(_) => new CopyBuilder(extension: '1')], toRoot(),
```
jakemac53: should this have a real builder key as well? do we do any validation to make sure we don't have duplicate builder keys, or does it not matter?

build_runner/test/generate/build_test.dart
```diff
@@ -94,22 +96,44 @@ void main() {
       test('multiple mixed build actions', () async {
```
jakemac53: nit: append " with custom build config" to the test name

e2e_example/tool/build.dart
```diff
@@ -13,10 +13,10 @@ Future main(List<String> args) async {
   var builders = [
     apply('e2e_example', 'throwing_builder', [(_) => new ThrowingBuilder()],
         toRoot(),
-        hideOutput: true),
+        hideOutput: true, defaultGenerateFor: const ['test/**']),
```
jakemac53: I don't think this needs to be added?

jakemac53: looks like the cause of one of the travis failures actually

build_runner/lib/src/generate/input_matcher.dart
natebosch: Added as a comment on the factor constructor above.

build_runner/lib/src/package_graph/apply_builders.dart
natebosch: I'm assuming that they don't need additional configuration from build.yaml so there is nothing to line up the key with. We can revisit this later if we need to.

build_runner/lib/src/package_graph/target_graph.dart
natebosch: Will look at optimizations in a followup.

build_runner/test/generate/build_test.dart
natebosch: We don't need to match up with any config so we shouldn't need it. Duplicate keys don't matter for now.

build_runner/test/generate/build_test.dart

e2e_example/tool/build.dart
natebosch: ugh, I meant to add this on the TestBootstrapBuilder.

natebosch:
Was missing the `build.yaml` I had for `build_test` when I updated the golden. Expecting Travis to pass this time. PTAL

PR #752, opened 2017-12-19
Add InputSet in build_config

natebosch:
This gives a more standard way to express include/exclude pairs and a
more flexible way to parse them.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -12,6 +12,37 @@ import 'package:path/path.dart' as p;
 import 'parse.dart';
 import 'pubspec.dart';
 
+/// A glob filter on files inputs or sources.
+///
+/// Takes a list of strings in glob format for [include] and [exclude]. Matches
+/// the `glob()` function in skylark.
+class InputSet {
+  /// The globs to include in the set.
+  ///
+  /// May be null or empty which means every possible path (like `'**'`).
+  final List<String> include;
+
+  /// The globs as a subset of [include] to remove from the set.
+  ///
+  /// May be null or empty which means every path in [include].
+  final List<String> exclude;
+  const InputSet({this.include, this.exclude});
+
+  @override
+  String toString() {
+    final result = new StringBuffer()..write('paths matching $include');
+    if (include == null || include.isEmpty) {
+      result.write('any path');
+    } else {
+      result.write('paths matching $include');
+    }
+    if (exclude != null && exclude.isNotEmpty) {
+      result.write(' except $exclude');
+    }
+    return '$result';
+  }
+}
+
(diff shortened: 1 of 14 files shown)


build_config/lib/src/build_config.dart
```diff
@@ -12,6 +12,37 @@ import 'package:path/path.dart' as p;
 import 'parse.dart';
 import 'pubspec.dart';
 
+/// A glob filter on files inputs or sources.
```
jakemac53: nit: wording is a bit off

build_config/lib/src/build_config.dart
```diff
@@ -12,6 +12,37 @@ import 'package:path/path.dart' as p;
...
+  /// The globs to include in the set.
+  ///
+  /// May be null or empty which means every possible path (like `'**'`).
+  final List<String> include;
+
+  /// The globs as a subset of [include] to remove from the set.
+  ///
+  /// May be null or empty which means every path in [include].
+  final List<String> exclude;
+  const InputSet({this.include, this.exclude});
```
jakemac53: nit: newline above

build_config/lib/src/build_config.dart
```diff
@@ -12,6 +12,37 @@ import 'package:path/path.dart' as p;
...
+
+  @override
+  String toString() {
+    final result = new StringBuffer()..write('paths matching $include');
```
jakemac53: I don't think you want the `write` call here? (looks like its handled better right below)

build_config/lib/src/parse.dart
```diff
@@ -322,3 +315,33 @@ List<String> _readListOfStringsOrThrow(
...
+  }
+
+  if (value is Map) {
+    final invalidOptions = value.keys.toList()
+      ..removeWhere((k) => k == 'include' || k == 'exclude');
+    if (invalidOptions.isNotEmpty) {
+      throw new ArgumentError('Got invalid options `$invalidOptions` '
+          'for $option. Only "include", and "exclude" are allowed');
+    }
+    final include = value.containsKey('include')
```
jakemac53: You could simplify these a bit:

```dart
final include = (value['include'] as List<String>)?.toList()
```

build_config/lib/src/build_config.dart

build_config/lib/src/build_config.dart

build_config/lib/src/build_config.dart
natebosch: yup, broken refactor. Fixed.

build_config/lib/src/parse.dart
natebosch: The suggested `as` won't work in Dart 2 semantics so I went with the future-proof approach.

PR #757, opened 2017-12-20
Allow overriding `build.yaml` for other packages

natebosch:
While we're rolling this out, and likely into the future, it's handy to
be able to hack on another package's `build.yaml` without a path
override to a local copy of the package. Read files matching
`$packageName.build.yaml` and use them instead of the `build.yaml`
coming from the package.

- Add `build_config_overrides.dart` with a function to look for override
  config files in the current directory and parse them into a map.
- When generating the build script and creating build actions, first
  look for overrides and use them over the package provided config.

### DIFF
--- a/build_runner/lib/src/package_graph/build_config_overrides.dart
+++ b/build_runner/lib/src/package_graph/build_config_overrides.dart
@@ -1,0 +1,29 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:build_config/build_config.dart';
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/package_graph/build_config_overrides.dart
```diff
@@ -0,0 +1,29 @@
...
+import 'package:glob/glob.dart';
+import 'package:path/path.dart' as p;
+
+import 'package_graph.dart';
+
+Future<Map<String, BuildConfig>> findBuildConfigOverrides(
+    PackageGraph packageGraph) async {
+  final configs = <String, BuildConfig>{};
+  final configFiles = await new Glob('*.build.yaml').list().toList();
+  for (final file in configFiles) {
```
jakemac53: nit: keep the stream and use `await for ...`

natebosch: Done

PR #758, opened 2017-12-20
Rename `serve` binary to `build_runner`

natebosch:
Pass through arguments rather than always assuming `serve` command.

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -11,13 +11,13 @@ import 'package:logging/logging.dart';
 import 'package:build_runner/src/build_script_generate/build_script_generate.dart';
 import 'package:build_runner/src/logging/std_io_logging.dart';
 
-Future<Null> main() async {
+Future<Null> main(List<String> args) async {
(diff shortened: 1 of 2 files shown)


build_runner/bin/build_runner.dart
```diff
@@ -11,13 +11,13 @@ import 'package:logging/logging.dart';
...
-Future<Null> main() async {
+Future<Null> main(List<String> args) async {
   var logListener = Logger.root.onRecord.listen(stdIOLogListener);
   await ensureBuildScript();
   var dart = Platform.resolvedExecutable;
-  final args = [scriptLocation, 'serve'];
-  if (stdioType(stdin) == StdioType.TERMINAL) args.add('--assume-tty');
-  var buildRun = await new ProcessManager().spawn(dart, args);
+  final innerArgs = [scriptLocation]..addAll(args);
+  if (stdioType(stdin) == StdioType.TERMINAL) innerArgs.add('--assume-tty');
```
jakemac53: nit: technically we should check the existing args for this or --no-assume-tty before adding this

natebosch: checking for any argument with `assume-tty`

PR #760, opened 2017-12-20
Remove `default` target and normalize names

natebosch:
- Remove `default` target, it now always matches the package name.
- Normalize target names in deps, and builder names in builder
  configuration. By normalizing at parse time we can avoid having to
  repeat the normalization logic at usage sites.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -92,6 +92,7 @@ class BuildConfig {
             .map((dep) => normalizeTargetKeyUsage(dep, packageName))
             .toSet(),
         package: packageName,
+        key: defaultTarget,
         sources: const InputSet(),
       )
     };
@@ -119,7 +120,11 @@ class BuildConfig {
 }
 
 class BuilderDefinition {
+  /// The package which provides this Builder.
   final String package;
+
+  /// A unique key for this Builder in `'$package|$builder'` format.
+  final String key;
 
   /// The names of the top-level methods in [import] from args -> Builder.
   final List<String> builderFactories;
@@ -156,11 +161,12 @@ class BuilderDefinition {
   final TargetBuilderConfigDefaults defaults;
 
   BuilderDefinition({
-    this.builderFactories,
-    this.buildExtensions,
-    this.import,
-    this.package,
-    this.target,
+    @required this.package,
+    @required this.key,
+    @required this.builderFactories,
+    @required this.buildExtensions,
+    @required this.import,
+    @required this.target,
     this.autoApply,
     this.requiredInputs,
     this.isOptional,
@@ -206,6 +212,9 @@ class BuildTarget {
 
   final String package;
 
+  /// A unique key for this target in `'$package:$target'` format.
+  final String key;
+
   final InputSet sources;
 
   /// A map from builder key to the configuration used for this target.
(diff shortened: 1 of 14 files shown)


build_config/lib/src/key_normalization.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+String normalizeBuilderKey(String builderKey, String packageName) {
+  if (builderKey.startsWith('|')) return '$packageName$builderKey';
```
jakemac53: might as well break this logic out into a function that takes a separator?

build_config/lib/src/build_config.dart
```diff
@@ -170,6 +167,19 @@ class BuilderDefinition {
...
+        'target': target,
+        'autoApply': autoApply,
+        'import': import,
+        'builderFactories': builderFactories,
+        'buildExtensions': buildExtensions,
+        'requiredInputs': requiredInputs,
+        'isOptional': isOptional,
+        'buildTo': buildTo,
+        'defaults': defaults,
+      }.toString();
```
jakemac53: I don't think this actually prints out that nicely (doesn't put each key/value on its own line). Consider JSON.encode or something like that?

build_config/lib/src/key_normalization.dart
```diff
@@ -2,13 +2,25 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-String normalizeBuilderKey(String builderKey, String packageName) {
+String normalizeBuilderKeyDefinition(String builderKey, String packageName) {
```
jakemac53: I am a bit confused by all of these and exactly what they correspond to - could you add comments with a quick example of how they correspond to different usages in the yaml?

build_config/lib/src/key_normalization.dart
```diff
@@ -2,13 +2,25 @@
...
-String normalizeTargetKey(String targetKey, String packageName) {
+String normalizeTargetKeyDefinition(String targetKey, String packageName) {
+  if (targetKey.startsWith(':')) return '$packageName$targetKey';
+  if (!targetKey.contains(':')) return '$packageName:$targetKey';
+  return targetKey;
+}
+
+String normalizeTargetKeyUsage(String targetKey, String packageName) {
   if (targetKey.startsWith(':')) return '$packageName$targetKey';
   if (!targetKey.contains(':')) return '$targetKey:$targetKey';
```
jakemac53: specifically, why is this one `$targetKey:$targetKey` and not `$packageKey:$targetKey`?

build_config/lib/src/build_config.dart
natebosch: I don't think it has much impact. This is mostly to make test failures easy to parse and we should (hopefully) rarely need it use it

build_config/lib/src/key_normalization.dart
natebosch: Done

build_config/lib/src/key_normalization.dart
natebosch: Added examples.

Issue #508, opened 2017-10-13
Consider a way of providing outputs which are only produced if the file doesn't already exist


jakemac53:
The primary use case for this today is during test bootstrapping. We want to automatically create an `_test.html` file for each `_test.dart` file, but only if it doesn't already exist.

The only options today are:

- force users to always create the html file
- never allow users to create the html file
- add configuration for whether or not to create the html file

natebosch:
We'll have to be careful how we do this - it won't work nicely without `buildToCache`. I think with that flag though we can probably detect the situation and avoid running a `Builder` that claims it will overwrite an existing source.

jakemac53:
Ya, this is tricky potentially to handle in bazel as well although it could be done.

jakemac53:
Ok - we talked offline and are going to run with roughly the following:

- Add an optional arg to `apply` that enables you to declare overlapping outputs, and delays any errors until you actually try to output a pre-existing source.
  - Only allowed if you are also outputting to cache
  - When building the asset graph if we see an output that conflicts with a source file (and you are using this option), we leave the `SourceAssetNode` node in place and do not create a `GeneratedAssetNode`
  - If you ever try and actually write an output which has an existing `SourceAssetNode` then we fail at that point
- Add a corresponding config option to build_config, and use that in the autogen script

This would be opt-in only, as it effectively delays errors to build time instead of asset graph creation time. In general we would prefer to keep errors early.

jakemac53:
@natebosch 

So I have this working but I realized a weird edge case in the process... lets say you have a builder which outputs a file `foo.txt`, and then another builder with `allowDeclaredOutputConflicts` set to `true` that outputs that same file (and comes afterwords). If the first builder doesn't actually output `foo.txt` then the second builder will do the "right thing" and check `canRead`, which will return `false`, and go ahead and try to output that file.

This should theoretically work but in reality we will (or at least should) throw an exception because the `phaseNumber` for the `GeneratedAssetNode` won't match the 2nd builders phase - it will match the first builders phase.

Some potential solutions:

- Do not allow this situation at all - essentially only allow you to declare overlapping outputs for source files but not other generated files.
- Add support somehow to `GeneratedAssetNode` - possibly turn `phaseNumber` into an Iterable.
- Add a new node type for these nodes, where the generating `phaseNumber` is allowed to change and possibly also has a list of potential generating phase numbers.

For now I think I will just drop support for conflicting outputs with other generated sources, lmk what you think though please.

[...]

PR #764, opened 2017-12-21
Add `allowDeclaredOutputConflicts` field to `BuilderApplication`

jakemac53:
Part of https://github.com/dart-lang/build/issues/508, just doesn't expose yaml config yet.

When true we allow your output extensions to overlap with original source files, but not generated files. You are still responsible in your builder for not outputting these files if they already exist. Overlapping generated files are not allowed (see discussion in https://github.com/dart-lang/build/issues/508).

Also added `allowedOutputsFilter` to `runBuilder` and `expectedOutputs` to support restricting the expected outputs of a builder to a subset of the default.


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -350,18 +355,23 @@ class AssetGraph {
...
+      @required bool allowDeclaredOutputConflicts}) {
     var removed = new Set<AssetId>();
     for (var output in outputs) {
       // When any outputs aren't hidden we can pick up old generated outputs as
       // regular `AssetNode`s, we need to delete them and all their primary
       // outputs, and replace them with a `GeneratedAssetNode`.
       if (contains(output)) {
         var node = get(output);
-        if (node is GeneratedAssetNode) {
+        if (allowDeclaredOutputConflicts) {
```
natebosch: should this also check whether it's a source node?

jakemac53: This is actually an old check that I am removing in my next commit to fix a few tests, but good catch!

jakemac53:
Closing this for now - a lot of complications came up that will be much easier to solve after we restructure the asset graph so more logic lives in the nodes themselves.

One of the specifically difficult problems is around primary outputs - the primary outputs for a node are assumed today to be fixed but if a node changes from being a `SourceAssetNode` to a `GeneratedAssetNode` then its primary outputs could change because the `GeneratedAssetNode` will not be available as a primary input until after the phase that generates it. This means swapping a nodes type can have serious ramifications to the asset graph including deleting nodes or adding nodes. This also means that a lot of edges in the graph have to be updated, which has to be done quite carefully today since you have to manually update the edges in each direction.

Instead we are going to go for a simpler solution for now which targets package:test specifically (which was the main use case anyways). We will never generate html files for tests automatically, and allow package:test to generate them, which I am pretty certain it will do on its own if they don't exist.

PR #765, opened 2017-12-21
Prepare to publish build_runner v0.7.0

natebosch:
- Update README to focus on the auto generated path. Move old
  documentation to a 'legacy' section.
- Remove references to `BuildAction` in favor of `BuilderApplication`.
  The dartdoc links are broken now but will start working if and when
  the dartdoc site can generate docs for this package.
- Remove old unused README links.
- Add missing entry to the changelog.
- Drop dev_dependencies

### DIFF
--- a/build_runner/README.md
+++ b/build_runner/README.md
@@ -54,8 +54,8 @@ Builder to decide whether the build needs to be customized. If it does you may
 also provide a `build.yaml` with the configuration. See the
 `package:build_config` README for more information on this file.
 
-Run `pub run build_runner serve` to start up a development server compiling with
-DDC.
+Run `pub run build_runner serve` to start up a development server. To have web
+code compiled with DDC add a `dev_dependency` on `build_compilers`.
 
 ### Inputs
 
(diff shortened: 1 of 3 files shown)


build_runner/README.md
```diff
@@ -48,99 +47,15 @@ dev_dependencies:
...
-```
-
-> This time, all the `*.html` files will be created first, and since the next
-> `BuildAction` is waiting for `*.html` inputs, it will _wait_ for them to be
-> created, and then create the `*.png` (screenshot) files.
-
-**NOTE**: Any time you change your build script (or any of its dependencies),
-the next build will be a full rebuild. This is because the system has no way
-of knowing how that change may have affected the outputs.
+Run `pub run build_runner serve` to start up a development server compiling with
```
jakemac53: we should probably mention the other commands here as well?

natebosch: Synced offline about this. These commands are in a strange state depending on who the target user is. We have a lot of open questions to answer and we'll wait to expand the docs until we have more answers.

Issue #626, opened 2017-11-16
Decide how to order the 'auto' builders from multiple packages


natebosch:
As a for instance, Sass should always run before angular. If both of these were applied automatically, how would we know to run sass first?

If a package exposes multiple builders (not builderFactories), we need to pick an order for those as well. In most cases this likely shouldn't matter since ordering dependencies should be indicated by ordering the builderFactories within a logical builder.

natebosch:
https://github.com/dart-lang/build/pull/661 started on this.

jakemac53:
Should we just close this and re-open if we need additional config? I don't think we have plans to do anything extra right now.

PR #767, opened 2017-12-21
Update Builder ordering to account for `BuildTo`

### LINKED ISSUE #647: Allow mixed-mode building

natebosch:
Final change to close #647 and #626

In general it doesn't make sense for a Builder which is writing to the
source tree to be dependent on an asset which isn't in the source tree.


natebosch:
Oops, I thought this had gone in already.

We probably will want to release a 0.7.1 with this.

PR #768, opened 2017-12-22
Filter out non-hidden actions outside the root

natebosch:
Previously this would be an error, but we want to be able to handle
cases like a `build.yaml` specifying `enable: true` for a Builder that
writes to the source tree, but then also being used as a dependency.

This is a small change for manual build scripts where instead of a
helpful exception an action would just not run.

- Allow BuildTo.source for non-root auto_apply in `build_config`
- Default `BuildTo` to `cache` always rather than depending on the value
  of `AutoApply`. Anything that writes to the source tree should be
  explicitly called out.
- Default `hideOutput` for most BuilderApplications to true to match the
  defaultin of `BuildTo`

### DIFF
--- a/build_runner/test/generate/build_error_test.dart
+++ b/build_runner/test/generate/build_error_test.dart
@@ -11,29 +11,6 @@ import '../common/common.dart';
 import '../common/package_graphs.dart';
 
 void main() {
-  test('fail with a nice error if the root package is not right', () async {
-    String error;
-    try {
-      await testBuilders(
-        [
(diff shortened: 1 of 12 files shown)


build_runner/test/generate/build_error_test.dart
```diff
@@ -11,29 +11,6 @@ import '../common/common.dart';
 import '../common/package_graphs.dart';
 
 void main() {
-  test('fail with a nice error if the root package is not right', () async {
```
jakemac53: or should we keep this test and just expect no outputs?

PR #769, opened 2017-12-22
Update dev_compiler_builder to use the new API

sigmundch:
This will make the branch comply with the changes in https://dart-review.googlesource.com/c/sdk/+/31283


jakemac53:
I guess we don't run travis on merges to the non-master branch, although it probably doesn't matter for now since we don't have explicit tests for the kernel workflow yet.

PR #770, opened 2017-12-22
Add `generateFor` argument to `applyToRoot`

natebosch:


natebosch:
This is useful for json_serializable so I don't need to wrap all the custom Builders up in factories so I can use `apply` instead of `applyToRoot`.

natebosch:
Added `isOptional` and `hideOutput` which brings it to parity with the optional args in `apply`

Issue #387, opened 2017-08-28
Add a mode to build to an output directory


natebosch:
Should include result of DDC. Might also need the input Dart files laid out in a `packages/` directory or similar for source maps

matanlurey:
How is this different than `writeToCache`?

natebosch:
The output directory would include copying source files. Today the cache is only generated files

jakemac53:
https://github.com/dart-lang/build/pull/667 add the `pub run build_runner:create_merged_dir --script=my/dart/script.dart --output-dir=my-output-dir` command but that is currently experimental and not what we want long term most likely.

Anybody who is following along and wants a stopgap solution can use it today though.

matanlurey:
Just to clarify, do we expect more work here for the MVP? Should we boot?

jakemac53:
I think it would be good to keep as a stretch goal? It would improve the testing story if we could get it done.

natebosch:
The solution we'll go with is an `--output-dir` flag for the `build`, `watch`, and `serve` commands.

Closing in favor of #810

PR #775, opened 2017-12-22
Add a `test` command

jakemac53:
This is ugly under the covers - it does roughly this:

- runs a build
- runs the create_merged_dir command
- creates missing html files for package:test manually
- runs pub run test with --precompiled

However, this gets things working with a mix of custom html files and not.

Unfortunately this does break running tests with --pub-serve if you don't have custom html files written (see the skipped test), but I think that is fine.


build_barback/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.5.0+1
+
+- Stop casting `BuildStep` to `BuildStepImpl` and remove `src` import from
```
natebosch: 👍

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -200,3 +209,106 @@ class _ServeCommand extends _WatchCommand {
...
+  }
+
+  /// Creates a merged directory for the current build script.
+  Future<String> _createMergedDir() async {
+    var tmpDir = await Directory.systemTemp.createTemp('build_runner_test');
+    var tmpDirPath = tmpDir.absolute.uri.toFilePath();
+    var scriptLocation = Platform.script.path;
+    var process = await new ProcessManager().spawn('pub', [
+      'run',
+      'build_runner:create_merged_dir',
```
natebosch: can you file an issue to make this a better supported thing?

natebosch: I guess https://github.com/dart-lang/build/issues/387 covers it :)

jakemac53:
looks like there is some minor issue happening on windows - will likely investigate that after the break

Issue #352, opened 2017-08-09
Add some type of locking so multiple build/watch can't step on eachother


natebosch:

jakemac53:
For now this is probably enough, but we need to come up with a design for the testing story as well, and that will require running a separate test command most likely while a regular build/serve are happening?

natebosch:
Yeah I think eventually we'll want to make them cooperate across processes somehow

matanlurey:
A simple mechanism could be a `build.lock`-file for now, no?

jakemac53:
yes for now something like that would be an improvement

jakemac53:
Moving to M1, you have to be a bit careful about how you use watch/serve together with test today (among other scenarios) but solving this properly is going to take some work.

aaronlademann-wf:
@jakemac53 can you elaborate on what you mean by 

> a bit careful about how you use watch/serve together with test today

What I'm seeing right now (and the core team that works on https://github.com/Workiva/over_react are trying to figure out is... is it actually possible to be serving, lets say - the `example/` dir, and then run tests via `pub run build_runner test` - without destroying the asset graph?

Any guidance / current workarounds would be very appreciated and welcome.

cc/ @robbecker-wf @evanweible-wf @corwinsheahan-wf @greglittlefield-wf

jakemac53:
@aaronlademann-wf right now the only safe thing to do is kill the other server before running tests - which is obviously not a great solution.

However, @grouma has been actively working on the build_daemon which solves this problem (there is already a `daemon` command actually, and you can use the build_daemon package to create a client if you want to play with it). It is already integrated into the webdev package at HEAD, but is still a little ways off from being released as we are actively working through all the edge cases. We will integrate it with the built-in build_runner commands once it is ready for prime time as well.

This is a fairly fundamental change and there will be some restrictions - for instance you won't be able to provide different build flags from different clients (it will ask you to kill the existing server in that case before running tests).

jakemac53:
Assigning Gary as he has been the one working on this, I forgot we actually had an issue filed :).

PR #780, opened 2017-12-23
Implement a simple "build.lock" file.

matanlurey:
Fixes https://github.com/dart-lang/build/issues/352.

Let me know if you think we should add anything else (tests, etc). I did a couple of tests locally with the `e2e_example` package, and it seems to work pretty well.


matanlurey:
Boy this is tricky. I might revert all changes but the lock file implementation itself until I can get either of your help. Let me know.

jakemac53:
Yes, this is trickier than it appears - even if you properly clean up the lockfile after the builds then you still might miss the case where somebody kills the task manually in the middle of a build. I also would prefer the lockfile to be in the `.dart_tool` directory - but that would be harder to clean up on errors.

One option that comes to mind is never actually trying to delete the lockfile at all. Instead you just put the pid of the current build task in it. Then when another build tries to run it sees if there is an active task that looks like a build script (the hard part...) with that pid. If not it replaces the file with its own pid and does a build and otherwise it errors or possibly waits to start a build.

Some other thoughts:

- We would like to be able to run tests while you have a server going - today the test command always does a build which would error if there was a server running since it would have the lockfile.
  - We could possibly run tests in --pub-serve mode in this case, if we add similar hacks into the server to serve the html files that the test package expects.
  - The test command could possibly skip the build in this case, and assume everything is up to date.
  - Or possibly the file could contain some additional information about if a build is currently happening.
  - This stuff all only works well in the context of full builds - which likely makes it not a good path for the future where you can serve/build specific directories/files.
- We could only take ownership of the lockfile during an actual build - so watch/serve would release it after each build. This would get complicated though because their state would not get updated if other builds happened so you would end up doing extra work.
- Another option could be to run more as a build server - the lockfile would contain a host/port of an active server to schedule builds with or something. This gets a lot more complicated but would likely result in a better experience overall... except that it would be difficult to debug.

Issue #215, opened 2017-02-24
Should consider log.severe as a build failure


natebosch:
The bazel and barback runners for a Builder consider a log at level severe to indicate a build failure and won't try to continue the build. For consistency we should do this in `build_runner` as well.

matanlurey:
+1

matanlurey:
Thought about this today - it should probably be a configurable flag. For example in https://github.com/dart-lang/source_gen/pull/209 I added a warning that I'd _like_ to make a severe, but I'd worry that would hurt dev-cycle for some users if severe broke the build in build_runner.

WDUT?

jakemac53:
We could change it so that we only fail builds if a builder throws an exception? Otherwise logging as severe would be allowed. Whatever we do though I think its important to be consistent across environments.

matanlurey:
I thought about this more and agree with you 👍

natebosch:
I'd be happy with adding a `--no-fail-on-severe`. I think it makes sense for failing to be the default.

jakemac53:
Should we close this now or is there future work planned?

jakemac53:
Or move it to M1?

natebosch:
I'm good with closing this. We can probably improve the UX but I think the biggest win we'll get is with a synchronous logging API and we're not likely to do that in M0

PR #781, opened 2017-12-23
Add a `--fail-on-severe` flag, but keep it as `false`

matanlurey:
Work towards https://github.com/dart-lang/build/issues/215.

We can flip this to `true` after better testing/validation this is a good idea.

I purposefully didn't update the CHANGELOG/pubspec yet, will do before I merge.


build_runner/lib/src/generate/options.dart
```diff
@@ -43,12 +49,24 @@ class BuildOptions {
...
+    }
+
     Logger.root.level = logLevel;
 
     overrideAnsiOutput(assumeTty == true || ansiOutputEnabled, () {
-      logListener = Logger.root.onRecord.listen(onLog ?? stdIOLogListener);
+      onLog ??= stdIOLogListener;
+      logListener = Logger.root.onRecord.listen((r) {
+        if (r.level == Level.SEVERE) {
+          severeLogHandled = true;
```
jakemac53: do we ever reset this (for watch mode)?

jakemac53:
It might be better to fail the whole build early for this case, although doing so is a bit... tricky so this is probably safer.

build_runner/lib/src/generate/options.dart
matanlurey: Yup, right here:

https://github.com/dart-lang/build/pull/781/files#diff-7a296c2aaedb49f2b91b7fa9c81b536cR89

matanlurey: And again in `watch_impl.dart`.

matanlurey:
Yeah I thought about that, but we don't have a great lifecycle right now.

If we can refactor to have a more clear `start/end` it will be cleaner to implement both this and lock files.

build_runner/lib/src/generate/options.dart
```diff
@@ -31,10 +33,14 @@ class BuildOptions {
   // For testing only, skips the build script updates check.
   bool skipBuildScriptCheck;
 
+  // For internal use only, flipped when a severe log occurred.
+  bool severeLogHandled = false;
```
natebosch: This runs counter to the goal in https://github.com/dart-lang/build/issues/95 and makes me a little nervous.

Any particular reason we are using this rather than throwing at the point of the log call?

jakemac53: We can (and should) move the boolean to the BuildImpl class to remove the addition of dynamic state to BuildOptions. 

The logs are coming from a stream so unless that becomes synchronous  failing in the listener is sketchy.

matanlurey: Do you still want me to land this with a TODO or hold?

natebosch: I'm fine with the change for now. We can revisit in the linked issue.

Issue #782, opened 2017-12-24
Support placeholders (web/$web$) more generally


matanlurey:
For example, in `build_test` and I imagine `bazel_codegen`?

jakemac53:
bazel_codegen does support them, not sure about build_test (didn't add anything explicit but the way it's implemented it might just work, I can't remember for sure)

PR #788, opened 2017-12-24
Add support for special placeholder files in build_test.

matanlurey:
Closes https://github.com/dart-lang/build/issues/782.


matanlurey:
@jakemac53:

> Technically attempting to read these assets is supposed to throw, and they shouldnt show up in globs either, but I don't know how important that is to support fully.

I'm happy to add that to at least the build_test package before committing, though it does mean changing the `expectedOutputs` function, which I'm not sure what the impact would be.

build_test/lib/src/test_builder.dart
```diff
@@ -119,6 +119,15 @@ Future testBuilder(
     inputIds.add(id);
   });
 
+  var allPackages = reader.assets.keys.map((a) => a.package).toSet();
+  for (var pkg in allPackages) {
+    for (var dir in const ['lib', 'web', 'test']) {
+      var asset = new AssetId(pkg, '$dir/\$$dir\$');
+      reader.cacheBytesAsset(asset, const []);
+      inputIds.add(asset);
```
jakemac53: What if we only do this and just remove the line above that adds the asset to the reader? I think that might give the desired behavior.

matanlurey: Sure. I'll tackle this before merging.

matanlurey: Done.

jakemac53:
I was just thinking about this and I think it is actually a breaking change, at least for build_runner?

That isn't a huge issue, but we should release it as such.

matanlurey:
@jakemac53:

> I think it is actually a breaking change, at least for build_runner

Can clarify what you mean here? You mean we need a newer version of BR as a lower constraint?

jakemac53:
> Can clarify what you mean here? You mean we need a newer version of BR as a lower constraint?

What I meant is this is a breaking change in the build_test package in general because there are now additional inputs that will be matched. If you were expecting certain outputs before and had an overzealous builder (like CopyBuilder), you will either have new outputs, or get runtime errors when you try to read these files. Both would cause your tests to fail.

matanlurey:
> What I meant is this is a breaking change in the build_test package

Oh ok. I agree. Do we want to ship `0.9.3` as-is? I can put this behind a flag and not enable yet.

Or we can skip to `0.10.0`.

/cc @natebosch

jakemac53:
I don't think we need to ship the current version, we can just delete `CopyBuilder` and release that with this change.

matanlurey:
PTAL @jakemac53 at the CHANGELOG before I merge.

PR #789, opened 2017-12-24
WIP: validate SDK versions

kevmoo:
Let's be explicit about the real required SDK

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,32 +8,26 @@ dist: trusty
 
 dart:
   - dev
+  - dev/release/2.0.0-dev.14.0
+  - dev/release/2.0.0-dev.13.0
+  - dev/release/2.0.0-dev.12.0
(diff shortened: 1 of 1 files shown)


natebosch:
That's a lot of versions to test. What value are we getting out of all that work?

kevmoo:
Figuring where and why we have the 2.0 sdk constraint

natebosch:
Bumped that in anticipation of using the latest code_builder

kevmoo:
Let's hold back bumps as.long as possible. 1.24 folks want json serial

natebosch:
@matanlurey - any thoughts on this?

matanlurey:
This is just not realistic. We need to focus on our MVP vision, not hypothetical customers.

My vote is latest dev release only until we are stable and shipping 2.0.

PR #798, opened 2018-01-03
Whitelist sources of all root package targets

### LINKED ISSUE #797: Allow configurable root package source whitelist
In #666 we hardcoded this. As it turns out `angular` is using a non-whitelisted directory - `test_files`.

We think this is sensible and we want to allow this use case. Question is how:

1. Read the `generate_for` globs (or maybe `sources` globs?) in the `build.yaml` targets section for the root package and automatically whitelist all of them
2. Add a new top level key in `build.yaml` like `source_whitelist` which is a list of Strings (as globs)
   - If we do this should they append to the defaults or override them?
3. Add a command line argument? Something else?

I'm leaning towards option 1. Any thoughts @jakemac53 @matanlurey

natebosch:
Closes #797

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -70,9 +70,9 @@ Future<ServeHandler> watch(
   var terminator = new Terminator(terminateEventStream);
 
   overrideBuildConfig ??= await findBuildConfigOverrides(options.packageGraph);
-  final targetGraph = await TargetGraph.forPackageGraph(options.packageGraph);
-  final buildActions = await createBuildActions(targetGraph, builders,
+  final targetGraph = await TargetGraph.forPackageGraph(options.packageGraph,
       overrideBuildConfig: overrideBuildConfig);
+  final buildActions = await createBuildActions(targetGraph, builders);
 
   var watch = runWatch(environment, options, buildActions,
       terminator.shouldTerminate, targetGraph.rootPackageConfig);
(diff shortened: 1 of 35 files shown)


jakemac53:
It seems like it would make more sense to whitelist the inputs for all targets in the build.yaml? All the sources that exist in a target should probably exist in the asset graph, regardless of whether they are used as a primary input or not.

jakemac53:
And then maybe we just set the input glob for the default target to be what the current whitelist is?

natebosch:
Switched to whitelisting the target source globs from the root package. PTAL

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -64,10 +65,12 @@ Future<ServeHandler> watch(
   var terminator = new Terminator(terminateEventStream);
 
   overrideBuildConfig ??= await findBuildConfigOverrides(options.packageGraph);
-  final buildActions = await createBuildActions(options.packageGraph, builders,
+  final targetGraph = await TargetGraph.forPackageGraph(options.packageGraph);
```
jakemac53: Should we add `targetGraph` to `options` so we don't have to recompute this?

natebosch: I could. I'd either need to change all the file IO to sync so we can synchronously create the TargetGraph along with the options, or change the way we get options to a static async method.

natebosch:
As discussed offline I switched over to storing the whitelist on the `BuildOptions`

If this looks good I can add some tests around it.

natebosch:
Added test for builds on files outside the hardcoded whitelist, as well as watching file edit/create/delete.

PTAL

Issue #777, opened 2017-12-23
build_runner doesn't pick up new dependencies


matanlurey:
For example I added `browser: any`, ran `pub get`.

I had to re-run the `serve` command so that `packages/browser/dart.js` could be found.

jakemac53:
Oh hmm that makes sense actually because the package location actually changed as a result. I guess the best we can do is watch the .packages file and update the package graph or something but that might be difficult.

natebosch:
We could be pessimistic and treat a change to `.packages` like a change to the build script and throw everything away and start from scratch.

It should be possible to update the PackageGraph and end up treating these as add/change events, but I'm not sure it's worth the complexity.

jakemac53:
> We could be pessimistic and treat a change to .packages like a change to the build script and throw everything away and start from scratch.

We can do a more incremental build if you just restart the script though - maybe we just kill the script with a message asking you to restart it?

When running through wrapper scripts we could auto-restart it for you.

matanlurey:
@natebosch:
> We could be pessimistic and treat a change to .packages like a change to the build script and throw everything away and start from scratch.

But you wouldn't do that if you manually restarted the script, right?

@natebosch:
> It should be possible to update the PackageGraph and end up treating these as add/change events ... but I'm not sure it's worth the complexity.

This is one of the things users complained the most about with the current pub/transformers workflow, having to manually restart the serve/watch script. Is it possible to somehow treat the `.packages` as a "required" input that invalidates dependencies?

The way I look at it, we _should_ tackle this, though if the solution is "have a wrapper script do it" I'm OK with that too. But maybe just notifying the user that a restart is required is good enough for the MVP.

jakemac53:
>This is one of the things users complained the most about with the current pub/transformers workflow, having to manually restart the serve/watch script. Is it possible to somehow treat the .packages as a "required" input that invalidates dependencies?

We will be a bit better in many cases because at least we should start up faster, but I agree it isn't ideal.

Handling this in a wrapper script also has the advantage that it can restart the build when the script itself is changed. A restart will always be required in that case unless we attempted to hot reload or something but that seems sketchy and the benefit would be minimal.

matanlurey:
Will this invalidate the entire build? For example if I add the `browser` package, but no other versions change, hypothetically you should be able to keep the cache for everything, right?

@jakemac53:

> A restart will always be required in that case unless we attempted to hot reload or something but that seems sketchy and the benefit would be minimal.

[...]

PR #802, opened 2018-01-03
kill the watch script if the .packages file changes

jakemac53:
Fixes https://github.com/dart-lang/build/issues/777

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -8,6 +8,7 @@ import 'package:build_config/build_config.dart';
 import 'package:build_runner/src/watcher/asset_change.dart';
 import 'package:build_runner/src/watcher/graph_watcher.dart';
 import 'package:build_runner/src/watcher/node_watcher.dart';
+import 'package:crypto/crypto.dart';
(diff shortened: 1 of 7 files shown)


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -230,6 +232,12 @@ class WatchImpl implements BuildState {
   /// Checks if we should skip a watch event for this [change].
   bool _shouldProcess(AssetChange change) {
     assert(_assetGraph != null);
+    if (_isRootPackagesFile(change.id)) {
+      _fatalBuildCompleter.complete();
```
natebosch: Is there a way we can structure this without adding side effects to this method?

jakemac53: I am looking into it, but without setting up an additional watcher it is not clear what the best way would be.

jakemac53: I moved it up earlier so at least it isn't so hidden inside of `_shouldProcess`, lmk what you think.

natebosch: I like the new location inside the `asyncMap` call.

PR #806, opened 2018-01-03
Update binary to point to kernel-summary-worker

sigmundch:


vsmenon:
lgtm - that works for me

PR #807, opened 2018-01-03
Split BuildOptions into BuildEnvironment

### LINKED ISSUE #803: Figure out a clearer picture for `BuildOptions`
Today we have `BuildOptions` and `BuildDefinition` which serve related purposes. `BuildOptions` is a mix of things which are clearly options, like `failOnSevere`, and things which are not really options but have to do with the environment in which the build is running, like how to read assets and how to log. Some things like `assumeTty` sort of fall in between. In practice we use these 'environment' overrides for testing. In theory we'd like the ability to use these to run in non-io environments like the web.

It came up in #798 that we'd like to have the `TargetGraph` and the `rootPackageWhitelist` along with the options (or maybe the `BuildDefinition`?). During this discussion it was pretty confusing how we'd make decisions about this type of thing.

I'd like to split it into a `BuildOptions` for the things which are options, and `BuildEnvironment` for the things related to the mechanical aspects of the build side effects (like writing files or logging).

natebosch:
Fixes #803

Takes some of the `dart:io` specific pieces and hides them behind and
interface. The visible difference is a change in the way the 'delete
conflicts' prompt is shown to the user.

- Add `BuildEnvironment` interface, and an IO backed implementation as
  well as one that have individual pieces overridden.
- Remove anything that defaults to io implementations from
  `BuildOptions`.

### DIFF
--- a/build_runner/lib/src/environment/build_environment.dart
+++ b/build_runner/lib/src/environment/build_environment.dart
@@ -1,0 +1,38 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+
+import '../asset/reader.dart';
+import '../asset/writer.dart';
+import '../generate/directory_watcher_factory.dart';
+
+/// Utilities to interact with the environment in which a build is running.
+///
+/// All side effects and user interaction should go through the build
+/// environment. An IO based environment can write to disk and interact through
+/// stdout/stdin, while a theoretical web or remote environment might interact
+/// over HTTP.
+abstract class BuildEnvironment {
+  RunnerAssetReader get reader;
+  RunnerAssetWriter get writer;
+  DirectoryWatcherFactory get directoryWatcherFactory;
+
+  void onLog(LogRecord record);
+
+  /// Prompt the user for input.
+  ///
+  /// The message and choices are displayed to the user and the index of the
+  /// chosen option is returned.
+  ///
+  /// If this environmment is non-interactive (such as when running in a test)
+  /// this method should throw [NonInteractiveBuildException].
+  Future<int> prompt(String message, List<String> choices);
+}
+
+/// Thrown when the build attempts to prompt the users but no prompt is
+/// possible.
+class NonInteractiveBuildException implements Exception {}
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/environment/build_environment.dart
```diff
@@ -0,0 +1,38 @@
...
+  void onLog(LogRecord record);
+
+  /// Prompt the user for input.
+  ///
+  /// The message and choices are displayed to the user and the index of the
+  /// chosen option is returned.
+  ///
+  /// If this environmment is non-interactive (such as when running in a test)
+  /// this method should throw [NonInteractiveBuildException].
+  Future<int> prompt(String message, List<String> choices);
```
jakemac53: Why not return the actual choice?

natebosch:
Once we're willing to take breaking changes and expose the environment as an argument this can open us up to some simplifications in the tests. For example we could have a `TestEnvironment` which bundles up the in-memory asset reader and writer, as well as capturing of logs, and then change the method to take the environment rather than individual arguments.

build_runner/lib/src/environment/overridable_environment.dart
```diff
@@ -0,0 +1,58 @@
...
+  @override
+  RunnerAssetWriter get writer => _writer ?? _default.writer;
+
+  @override
+  DirectoryWatcherFactory get directoryWatcherFactory =>
+      _directoryWatcherFactory ?? _default.directoryWatcherFactory;
+
+  @override
+  void onLog(LogRecord record) {
+    if (_onLog != null) {
```
jakemac53: `(onLog ?? _default.onLog)(record)`? I don't know though lol, this is probably fine as is :D

build_runner/lib/src/environment/overridable_environment.dart
```diff
@@ -0,0 +1,58 @@
...
+      _onLog(record);
+    } else {
+      _default.onLog(record);
+    }
+  }
+
+  @override
+  Future<int> prompt(String message, List<String> choices) =>
```
jakemac53: A bit weird that we don't actually expose this one, but probably fine for now.

build_runner/lib/src/environment/build_environment.dart
natebosch: We'd have to match for equality afterward - the choice itself has no meaning to the caller beyond it's index anyway since it's a human readable string.

I had thought about doing something like `Future<T> prompt<T>(String message, List<T> choices);` which would let us do something useful with the choice afterward, but it doesn't seem necessary.

build_runner/lib/src/environment/overridable_environment.dart
natebosch: Could also be `_onLog?.call(record) ?? _default.onLog(record)` :)

as much as I like minimizing lines like this I think either one masks the intent too much

build_runner/lib/src/environment/overridable_environment.dart
natebosch: yeah, I figured we have an easy point to extend this if we do want to allow overriding this one, but since we don't do it today no point in adding it.

Issue #804, opened 2018-01-03
Use dart2js in batch mode


jakemac53:
Without that it gets a bit insane today even with a small number of tests.

PR #809, opened 2018-01-04
Use batch mode for dart2js and only compile one app at a time

jakemac53:
Fixes https://github.com/dart-lang/build/issues/804

### DIFF
--- a/build_web_compilers/lib/src/workers.dart
+++ b/build_web_compilers/lib/src/workers.dart
@@ -3,6 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
+import 'dart:convert';
 import 'dart:io';
 import 'dart:math' show min;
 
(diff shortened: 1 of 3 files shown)


build_web_compilers/lib/src/workers.dart
```diff
@@ -83,4 +90,125 @@ final dartdevcDriverResource = new Resource<BazelWorkerDriver>(
...
+  Dart2JsBatchWorker(this._spawnWorker);
+
+  Future<Dart2JsResult> compile(List<String> args) async {
+    var job = new _Dart2JsJob(args);
+    _workQueue.add(job);
+    if (!_queueIsActive) _startWorkQueue();
+    return job.result;
+  }
+
+  void _startWorkQueue() {
```
natebosch: [[optional] Is there any duplication we can factor out?

jakemac53:
Waiting on appveyor to see how long it takes to run the tests, but locally it takes almost a minute to compile them all so I think we should likely leave the timeout in.

build_web_compilers/lib/src/workers.dart
jakemac53: not really... other than adding the lines to the buffer but the logic is so trivial its not worth breaking out imo

PR #812, opened 2018-01-04
Add human readable Duration printing

### LINKED ISSUE #784: Consider representing durations more user-friendly
For example instead of `Build: Succeeded after 19644ms with 980 outputs`

➡️ `Build: Succeeded after 19.6s with 980 outputs`

natebosch:
Closes #784

Since the build_runner use case is more constrained than a general
human-friendly duration printer we can use a very simple implementation.

- Add `humanReadable` with a simple implementation that fits our use
  case.
- Replace usage of `package:millisecond` with this method.
- Use this method in other places where we were printing milliseconds.

### DIFF
--- a/build_runner/lib/src/logging/human_readable_duration.dart
+++ b/build_runner/lib/src/logging/human_readable_duration.dart
@@ -9,7 +9,7 @@
 ///
 /// Always attempts 2 'levels' of precision. Will show hours/minutes,
 /// minutes/seconds, seconds/tenths of a second, or milliseconds depending on
-/// the largets level that needs to be displayed.
+/// the largest level that needs to be displayed.
 String humanReadable(Duration duration) {
   if (duration < const Duration(seconds: 1)) {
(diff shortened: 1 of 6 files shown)


build_runner/lib/src/logging/human_readable_duration.dart
```diff
@@ -0,0 +1,28 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+/// Returns a human readable string for a duration.
+///
+/// Handles durations that span up to hours - this will not be a good fit for
+/// durations that are longer than days.
+///
+/// Always attempts 2 'levels' of precision. Will show hours/minutes,
+/// minutes/seconds, seconds/tenths of a second, or milliseconds depending on
+/// the largets level that needs to be displayed.
```
nshahan: nit: largets --> largest

natebosch: Done

Issue #810, opened 2018-01-04
Add --output=dir flag for all commands 


jakemac53:
This will create a merged directory after each build at the specified path.

PR #813, opened 2018-01-04
Add --output flag which creates a merged output dir

jakemac53:
This also always does the magic for package:test to create html files where they are missing.

Closes https://github.com/dart-lang/build/issues/810.


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -54,11 +52,15 @@ class _SharedOptions {
 
   final bool enableLowResourcesMode;
 
+  /// Path to the merged output directory, if applicable.
```
natebosch: [nit] can we be more explicit about what 'applicable' means?

`/// Path to the merged output directory, or null if no directory should be created.`

PR #816, opened 2018-01-05
Add a KernelSummaryBuilder and kernel mode for the DevCompilerBuilder

jakemac53:
This doesn't yet have tests or set up the proper config in build.yaml but we can follow up with that.


build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -23,7 +24,10 @@ const jsSourceMapExtension = '.ddc.js.map';
 
 /// A builder which can output ddc modules!
 class DevCompilerBuilder implements Builder {
-  const DevCompilerBuilder();
+  final bool useKernel;
```
sigmundch: probably not worth it right now, but I wonder whether it would help to simply do a DDK builder separate from the DDC builder instead (since so much of the code has special casing for the ddk case.)

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -88,42 +102,75 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
...
 
-  // Add the default analysis_options.
-  await scratchSpace.ensureAssets([defaultAnalysisOptionsId], buildStep);
-  request.arguments.add(defaultAnalysisOptionsArg(scratchSpace));
-
   // Add all the linked summaries as summary inputs.
   for (var id in transitiveSummaryDeps) {
-    request.arguments.addAll(['-s', scratchSpace.fileFor(id).path]);
+    if (useKernel) {
+      var path = p.url.relative(scratchSpace.fileFor(id).path,
```
sigmundch: (optional as well): to make sharing of logic more clear, it might help to rewrite this as:
```dart
  var summaryPath = scratchSpace.fileFor(id).path;
  if (useKernel) {
    summaryPath = path.url.relative(summaryPath, from: ...);
  }
  request.arguments.addAll(['-s', summaryPath]);
```

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -88,42 +102,75 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
...
+
   // And finally add all the urls to compile, using the package: path for
   // files under lib and the full absolute path for other files.
   request.arguments.addAll(module.sources.map((id) {
     var uri = canonicalUriFor(id);
     if (uri.startsWith('package:')) {
       return uri;
     }
-    return new Uri.file('/${id.path}').toString();
+    if (useKernel) {
```
sigmundch: similarly:
```dart
var urlString = '${id.path}';
if (!useKernel) urlString = "${new Uri.file('/$urlString')}";
return urlString;
```

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -88,42 +102,75 @@ Future createDevCompilerModule(Module module, BuildStep buildStep,
...
   }));
 
-  var dartdevc = await buildStep.fetchResource(dartdevcDriverResource);
-  var response = await dartdevc.doWork(request);
+  WorkResponse response;
+  if (useKernel) {
+    // Make sure to clean up the .packages file.
+    assert(packagesFile != null);
+    try {
+      var dartdevk = await buildStep.fetchResource(dartdevkDriverResource);
```
sigmundch: similarly:
```dart
try {
  ...
} finally {
  if (useKernel) do-cleanup?
}
```

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: The advantage of this route is we can allow you to switch modes by changing the config option for the builder, which is a bit easier to do.

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: done, used ternary though

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: done, used ternary though

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: done

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:io';
```
natebosch: Are we considering this to be part of the 'external' behavior? Could we look for any other ways to lay out the code such that we aren't importing both io and package: build in the same library? Hopefully we'd have clear io utilities and some way to be sure they are only touching the scratch space...

jakemac53: In this case we actually are explicitly not touching the shared scratch space, we need to create a custom `.packages` file for each module (really, its each package).

Filed https://github.com/dart-lang/build/issues/818 with one possible approach.

natebosch: Makes sense. The issue looks good.

natebosch: Makes sense. The issue looks good.

Issue #826, opened 2018-01-09
prompt doesn't work for manual build scripts


jakemac53:
not sure why yet, need to do investigation

PR #827, opened 2018-01-09
Fix prompt when assumeTty is false

jakemac53:
Fixes https://github.com/dart-lang/build/issues/826.

I added a basic test but I couldn't think of how to reasonable test this exact scenario actually, lmk if you have any ideas. The real bug was when you are connected to a terminal but don't pass --assume-tty, but we can't replicate that with package:test in a way that I can think of.

### DIFF
--- a/build_runner/lib/src/environment/io_environment.dart
+++ b/build_runner/lib/src/environment/io_environment.dart
@@ -32,7 +32,7 @@ class IOEnvironment implements BuildEnvironment {
   final bool _verbose;
 
   IOEnvironment(PackageGraph packageGraph, bool assumeTty, {bool verbose})
-      : _isInteractive = assumeTty ?? _canPrompt(),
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/environment/io_environment.dart
```diff
@@ -32,7 +32,7 @@ class IOEnvironment implements BuildEnvironment {
   final bool _verbose;
 
   IOEnvironment(PackageGraph packageGraph, bool assumeTty, {bool verbose})
-      : _isInteractive = assumeTty ?? _canPrompt(),
+      : _isInteractive = assumeTty ? true : _canPrompt(),
```
natebosch: `assumeTty || _canPrompt()`?

jakemac53: actually needed to go with `assumeTty == true || _canPrompt()` because we pass nulls for this today....

PR #829, opened 2018-01-09
prepare for release of build_runner and build_web_compilers

jakemac53:


natebosch:
code_builder 3.0.0 was just published. We might want to upgrade here.

jakemac53:
updated to the latest code_builder, looks like we weren't broken but will wait for travis to confirm

PR #833, opened 2018-01-11
Add TestBuilder

natebosch:
The CopyBuilder has been hacked on long enough that it has some issues:
- No point of extensibility other than subclassing or adding new
  features directly.
- It isn't always clear from the constructor arguments what the behavior
  will be.
- inputExtension and extension are specified different - include the `.`
  for the input but not the output.

The TestBuilder replacement instead takes Function arguments to override
behavior and serves as a way to reduce boilerplate around the Builder
class with the same default behavior as the CopyBuilder.

- Mark CopyBuilder as deprecated
- Add TestBuilder with some quality of life help similar to CopyBuilder
  and allow overriding behavior by passing Function arguments.
- Add some utilities to give a nicer syntax for common behavior
  overrides.
- Update tests in `build_runner` to use the TestBuilder

### DIFF
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -39,3 +39,7 @@ dev_dependencies:
   package_resolver: ^1.0.2
   test: ^0.12.0
   test_descriptor: ^1.0.0
+
+dependency_overrides:
+  build_test:
+    path: ../build_test
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -35,7 +35,7 @@ dependencies:
   yaml: ^2.1.0
 
 dev_dependencies:
-  build_test: ^0.9.4
+  build_test: ^0.9.5
   package_resolver: ^1.0.2
   test: ^0.12.0
   test_descriptor: ^1.0.0
--- a/build_runner/test/generate/serve_test.dart
+++ b/build_runner/test/generate/serve_test.dart
@@ -20,8 +20,6 @@ import '../common/package_graphs.dart';
 void main() {
   group('ServeHandler', () {
     InMemoryRunnerAssetWriter writer;
-    CopyBuilder copyBuilder;
-    BuilderApplication copyABuildApplication;
 
     setUp(() async {
       _terminateServeController = new StreamController();
@@ -30,9 +28,6 @@ void main() {
 # Fake packages file
 a:file://fake/pkg/path
 ''');
-
-      copyBuilder = new CopyBuilder();
-      copyABuildApplication = applyToRoot(copyBuilder);
     });
 
     tearDown(() async {
@@ -42,7 +37,7 @@ a:file://fake/pkg/path
 
     test('does basic builds', () async {
       var handler = await createHandler(
-          [copyABuildApplication], {'a|web/a.txt': 'a'}, writer);
+          [applyToRoot(new TestBuilder())], {'a|web/a.txt': 'a'}, writer);
       var results = new StreamQueue(handler.buildResults);
       var result = await results.next;
       checkBuild(result, outputs: {'a|web/a.txt.copy': 'a'}, writer: writer);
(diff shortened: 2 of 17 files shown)


build_runner/test/generate/serve_test.dart
```diff
@@ -56,11 +51,15 @@ a:file://fake/pkg/path
     });
 
     test('blocks serving files until the build is done', () async {
+      var blockContainer = <Future>[];
```
jakemac53: nit: Just have a single `nextBuildBlocker` variable which you overwrite, you aren't binding the closure to the initial value just referencing this variable directly so it will pick up changes.

build_test/lib/src/builder.dart
```diff
@@ -0,0 +1,82 @@
...
+      final newPath = _replaceSuffix(
+          buildStep.inputId.path, inputExtension, outputExtension);
+      final id = new AssetId(buildStep.inputId.package, newPath);
+      buildStep.writeAsString(
+          id, buildStep.readAsString(readFrom(buildStep.inputId)));
+    }
+  }
+}
+
+FutureOr Function(BuildStep, Map<String, List<String>>) copyFrom(
```
jakemac53: Lol wow I _really_ dislike this new syntax. Imo just use a typedef because it looks nicer and is `<1.x` compatible (even if we don't declare support for that I would prefer to be as compatible as possible).

build_test/lib/src/builder.dart
```diff
@@ -0,0 +1,82 @@
...
+Map<String, List<String>> replaceExtension(String from, String to) => {
+      from: [to]
+    };
+
+/// A [Builder] whose [build] method can be replaced with a closure.
+class TestBuilder implements Builder {
+  @override
+  final Map<String, List<String>> buildExtensions;
+
+  final FutureOr Function(BuildStep, Map<String, List<String>>) _build;
```
jakemac53: Ya, definitely typedef these it would be way more terse and they are copied around quite a bit. It will also be easier to update if needed.

build_test/lib/src/builder.dart
```diff
@@ -0,0 +1,82 @@
...
+    FutureOr Function(BuildStep, Map<String, List<String>>) build,
+    FutureOr Function(BuildStep, Map<String, List<String>>) extraWork,
+  })
+      : buildExtensions = buildExtensions ?? appendExtension('.copy'),
+        _build = build ?? _defaultBehavior,
+        _extraWork = extraWork;
+
+  @override
+  Future build(BuildStep buildStep) async {
+    if (buildStep.inputId.path.endsWith(r'$')) return;
```
jakemac53: nit: use `canRead` instead? If you can't then we should fix that.

build_test/pubspec.yaml
```diff
@@ -1,11 +1,11 @@
 name: build_test
 description: Utilities for writing unit tests of Builders.
-version: 0.9.4
+version: 0.9.5-dev
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build
 
 environment:
-  sdk: ">=1.22.1 <2.0.0"
+  sdk: ">=2.0.0-dev <2.0.0"
```
jakemac53: lets leave this back at 1.22.1 as well

build_runner/pubspec.yaml
```diff
@@ -39,3 +39,7 @@ dev_dependencies:
   package_resolver: ^1.0.2
   test: ^0.12.0
   test_descriptor: ^1.0.0
+
+dependency_overrides:
+  build_test:
+    path: ../build_test
```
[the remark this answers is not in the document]
natebosch: Done

build_runner/test/generate/serve_test.dart
natebosch: Done

build_test/lib/src/builder.dart
natebosch: Done

build_test/lib/src/builder.dart
natebosch: Done

build_test/lib/src/builder.dart
natebosch: Done

build_test/pubspec.yaml
natebosch: Done

PR #834, opened 2018-01-11
Disallow reading hidden file to generate source file

natebosch:
- Pass through `outputsHidden` down to `SingleStepAssetReader`.
- Disallow reading a hidden generated output if the step isn't hiding it's
  outputs.
- Replace the `ExistsBuilder` with a new behavior for the `TestBuilder`.
- Add `buildsCompleted` Stream to `TestBuilder`.

### DIFF
--- a/build_runner/lib/src/asset/reader.dart
+++ b/build_runner/lib/src/asset/reader.dart
@@ -73,6 +73,8 @@ class SingleStepReader implements DigestAssetReader {
   /// builder may behave differently on the next build.
   Iterable<Glob> get globsRan => _globsRan;
 
+  /// Checks whether [id] can be read by this step - attempting to build the
+  /// asset if necessary.
   Future<bool> _isReadable(AssetId id) {
     _assetsRead.add(id);
     var node = _assetGraph.get(id);
@@ -83,6 +85,8 @@ class SingleStepReader implements DigestAssetReader {
     return _isReadableNode(node);
   }
 
+  /// Checks whether [node] can be read by this step - attempting to build the
+  /// asset if necessary.
   Future<bool> _isReadableNode(AssetNode node) async {
     if (node.isGenerated) {
       final generatedNode = node as GeneratedAssetNode;
--- a/build_runner/lib/src/asset/reader.dart
+++ b/build_runner/lib/src/asset/reader.dart
@@ -34,6 +34,8 @@ abstract class Md5DigestReader implements DigestAssetReader {
   }
 }
 
+final _futureFalse = new Future<bool>.value(false);
+
 /// An [AssetReader] with a lifetime equivalent to that of a single step in a
 /// build.
 ///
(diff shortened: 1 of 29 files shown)


jakemac53:
That gets a bit complicated with package cycles as well though...

natebosch:
> Assets generated to source should be able to read assets generated to cache in other packages I think.

I disagree. I think anything you publish should only depend on other stuff that you or a dependency has published... I don't think it makes sense to generate something hidden in a dependency and then publish it.

If we do decide we want to allow that I don't know why we'd disallow reading hidden assets in the root package. It doesn't make sense for the root package to be _more_ restricted than dependencies.

jakemac53:
I guess my concern is this might encourage bad patterns such as always generating to source and publishing things that really shouldn't be published... but I am fine with this for now and we can always loosen restrictions if required. It is certainly easier to loosen them later than tighten them.

build_test/lib/src/builder.dart
```diff
@@ -29,16 +34,24 @@ void _copyToAll(BuildStep buildStep, Map<String, List<String>> buildExtensions,
...
 
 /// A build behavior which reads [assetId] and copies it's content into every
 /// output.
 BuildBehavior copyFrom(AssetId assetId) => (buildStep, buildExtensions) =>
-    _copyToAll(buildStep, buildExtensions, (_) => assetId);
+    _copyToAll(buildStep, buildExtensions, readFrom: (_) => assetId);
+
+/// A build behavior which writes either 'true' or 'false' depending on whether
+/// [assetId] can be read.
+BuildBehavior checkCanRead(AssetId assetId) =>
```
jakemac53: nit: `writeCanRead`?

build_runner/test/generate/build_test.dart
```diff
@@ -361,6 +363,23 @@ void main() {
             });
       });
 
+      test('disallows reading hidden outputs to create a non-hidden output',
```
jakemac53: It would be good to add a test for the behavior of not being able to read hidden assets in other packages as well, or merge that into this test.

build_test/lib/src/builder.dart
natebosch: Done

natebosch:
I made some other changes to align `listAssets` with the other methods. PTAL

build_runner/lib/src/asset/reader.dart
```diff
@@ -71,31 +73,36 @@ class SingleStepReader implements DigestAssetReader {
...
     var node = _assetGraph.get(id);
     if (node == null) {
       _assetGraph.add(new SyntheticSourceAssetNode(id));
-      return false;
+      return _futureFalse;
     }
+    return _isReadableNode(node);
+  }
+
+  Future<bool> _isReadableNode(AssetNode node) async {
```
jakemac53: lets at least leave a comment that this can do lazy builds

PR #837, opened 2018-01-11
Use an Isolate rather than a subprocess

natebosch:
We can avoid the `--assume-tty` hack since the Isolate sees the same TTY
as the VM. Leave that flag in for now as an option for higher levels of
wrapping scripts.

Only generate the String for the build script in
`build_script_generate.dart`. Handle writint it to a file in the main
method. This will open up the possibility of skipping the write to disk
and using a `data:` URI if we can solve the problem of build script
updates. If we can do that we can add a separate command to write the
build script to disk.

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -7,6 +7,7 @@ import 'dart:io';
 import 'dart:isolate';
 
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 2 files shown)


natebosch:
Hmm, signal handling is different now. If I hit `^c` during a build in watch mode it just dies, it doesn't go through the cleanup process.

We might need some work on this.

natebosch:
Ah nvm. I was on the wrong branch when I tested `^c`.

It has bad behavior on `master` and the isolate fixes it :)

PR #838, opened 2018-01-12
Add generate-build-script command

### LINKED ISSUE #778: Add pub run build_runner update command?
Something to just ensure that the build script is up to date, but does nothing else.

This will be useful for CLIs that wrap/delegate and perform a "doctor"-like command.

(Happy to add this, just want a 👍 first)

natebosch:
Closes #778

Wrapping entrypoints can use this to generate the build script and then
run it using `Isolate.spawnUri` and avoid workarounds like
`--assume-tty`.

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -50,7 +50,8 @@ const _generateCommand = 'generate-build-script';
 
 class _GenerateBuildScript extends Command {
   @override
-  final description = 'Generate a script to run builds and print the file path';
+  final description = 'Generate a script to run builds and print the file path '
+      'with no other logging. Useful for wrapping builds with other tools.';
 
   @override
   final name = _generateCommand;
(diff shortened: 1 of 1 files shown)


build_runner/bin/build_runner.dart
```diff
@@ -14,26 +15,43 @@ import 'package:build_runner/src/entrypoint/options.dart';
...
 
 Future<Null> main(List<String> args) async {
-  var logListener = Logger.root.onRecord.listen(stdIOLogListener);
-
   // Use the actual command runner to parse the args and immediately print the
   // usage information if there is no command provided or the help command was
   // explicitly invoked.
-  var commandRunner = new BuildCommandRunner([]);
+  var commandRunner = new BuildCommandRunner([])
+    ..addCommand(new _GenerateBuildScript());
```
jakemac53: Any reason we shouldn't add this to the command runner generally?

natebosch: The other place we use this CommandRunner doesn't have the capability of performing this command.

PR #839, opened 2018-01-12
Add explicit types to for-in loops

natebosch:
Work around for https://github.com/dart-lang/sdk/issues/31884 in the
dev.17.0 version of the SDK

### DIFF
--- a/build_runner/test/common/test_phases.dart
+++ b/build_runner/test/common/test_phases.dart
@@ -144,7 +144,7 @@ void checkBuild(BuildResult result,
 
   final unhiddenOutputs = <String, dynamic>{};
   final unhiddenAssets = new Set<AssetId>();
-  for (final id in outputs?.keys ?? const []) {
+  for (final String id in outputs?.keys ?? const []) {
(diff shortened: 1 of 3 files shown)


build_runner/test/common/test_phases.dart
```diff
@@ -144,7 +144,7 @@ void checkBuild(BuildResult result,
 
   final unhiddenOutputs = <String, dynamic>{};
   final unhiddenAssets = new Set<AssetId>();
-  for (final id in outputs?.keys ?? const []) {
+  for (final String id in outputs?.keys ?? const []) {
```
jakemac53: nit: we should try to get more consistent about final vs not in for loops... :)

not a blocker anything as its a preexisting issue

natebosch: Yeah I know that the larger community pushed back against "final everywhere" but I find that I occasionally slip back into that habit...

jakemac53: ya honestly I don't care, I can see it either way, but I do care that it is consistent :)

PR #840, opened 2018-01-12
Add digest method to AssetReader

jakemac53:
Closes https://github.com/dart-lang/build/issues/710


build_web_compilers/lib/src/summary_builder.dart
```diff
@@ -141,6 +142,16 @@ Future createLinkedSummary(Module module, BuildStep buildStep,
 
   // Add all the files to include in the linked summary bundle.
   request.arguments.addAll(_analyzerSourceArgsForModule(module, scratchSpace));
+
+  /// Add the [Input]s with `Digest`s.
```
natebosch: [nit] block comment, not doc comment

jakemac53:
Note that I am expecting analyzer failures here, they are all just hints about the deprecated CopyBuilder, which are a result of the path dependencies.

I can follow up with a pr to fix that after this lands.

jakemac53:
Looks like we have had a failure in build_web_compilers for a while, not actually triggered by this cl but the travis.yml changes fixed it to actually run that test. Will submit with broken travis later today and we can fix next week.

Issue #843, opened 2018-01-12
resolveSources: Assets not resolvable until next event loop


matanlurey:
I'm in the process of writing tests in `source_gen`, and hit the following:

```dart
// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
// for details. All rights reserved. Use of this source code is governed by a
// BSD-style license that can be found in the LICENSE file.

// Increase timeouts on this test which resolves source code and can be slow.
@Timeout.factor(2.0)
import 'dart:async';

import 'package:analyzer/dart/element/element.dart';
import 'package:build/build.dart';
import 'package:build_test/build_test.dart';
import 'package:source_gen/source_gen.dart';
import 'package:test/test.dart';

void main() {
  // Holds the analyzer open until all of the tests complete.
  final doneWithResolver = new Completer<Null>();

  LibraryReader b;

  setUpAll(() async {
    final resolver = await resolveSources(
      {
        'test_lib|lib/a.dart': sourceA,
        'test_lib|lib/b.dart': sourceB,
      },
      (r) => r,
      tearDown: doneWithResolver.future,
    );
    // Some bug in build_test causes the 'libraryFor' statement below to throw
    // saying it is not a dart library source unless we wait for the next event
    // loop. 
    await new Future.delayed(Duration.ZERO);
    b = new LibraryReader(
      await resolver.libraryFor(new AssetId('test_lib', 'lib/b.dart')),
    );
    expect(b.element, isNotNull);
  });

  tearDownAll(doneWithResolver.complete);

  group('top level fields', () {
    List<Element> topLevelFields;

    setUpAll(() {
      topLevelFields = b.element.definingCompilationUnit.topLevelVariables;
    });

    test('should read the prefix of a type', () {
      expect(b.prefixFor(topLevelFields[0]), 'a_prefixed');
    });

    test('should read null when the type is not prefixed', () {
      expect(b.prefixFor(topLevelFields[1]), isNull);
    });
  });
}

const sourceA = r'''
  class A {}
  class B {}
''';

const sourceB = r'''
  import 'a.dart' show B;
  import 'a.dart' as a_prefixed;

  a_prefixed.A topLevelFieldWithPrefix;   // [0]
  B topLevelFieldWithoutPrefix;           // [1]
''';
```

... where a `await new Future.delayed(...)` is required to go forwards. This is probably a bug.

matanlurey:
Ok, figured it out, it's not a timing issue, it's a bug in the implementation:

[...]

PR #846, opened 2018-01-12
Add resolverFor to resolveSources.

matanlurey:
Closes https://github.com/dart-lang/build/issues/843.

### DIFF
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -214,9 +214,7 @@ class _ResolveSourceBuilder<T> implements Builder {
   Future<Null> build(BuildStep buildStep) async {
     if (onDone.isCompleted) return;
     var result = await _action(buildStep.resolver);
-    if (!onDone.isCompleted && _resolverFor == buildStep.inputId) {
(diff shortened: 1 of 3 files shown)


build_test/lib/src/resolve_source.dart
```diff
@@ -192,15 +204,17 @@ typedef FutureOr<T> _ResolverAction<T>(Resolver resolver);
...
-  _ResolveSourceBuilder(this._action, this._tearDown);
+  _ResolveSourceBuilder(this._action, this._resolverFor, this._tearDown);
 
   @override
   Future<Null> build(BuildStep buildStep) async {
     if (onDone.isCompleted) return;
     var result = await _action(buildStep.resolver);
-    if (!onDone.isCompleted) {
+    if (!onDone.isCompleted && _resolverFor == buildStep.inputId) {
       // With resolveSources (plural), this function is called multiple times
```
natebosch: Is this comment still true? I think we will only see a given `inputId` once so we can remove the other check and the comment.

matanlurey: Good catch! Done.

PR #850, opened 2018-01-16
use async/await less in the hot paths

jakemac53:
This shaves off another ~18 seconds from angular_components_example build times


jakemac53:
Yes that was the issue with the failing test, updated now. Note that I didn't use a try/catch because that made the angular_components_example build time ~4 seconds slower :(.

jakemac53:
Merging, ran tests locally and travis looks like it is still having issues.

Issue #847, opened 2018-01-13
Add a README for build_web_compilers


matanlurey:
Since, y'know, it is going to be there for most web users :)

PR #853, opened 2018-01-16
add a README for build_web_compilers

jakemac53:
Fixes https://github.com/dart-lang/build/issues/847


build_web_compilers/README.md
```diff
@@ -0,0 +1,98 @@
...
+following to your `pubspec.yaml`:
+
+```yaml
+dev_dependencies:
+  build_web_compilers:
+```
+
+## Usage
+
+If you are using the autogenerated build script (going through `pub run` instead
```
natebosch: Maybe `pub run build_runner` so users don't get confused with other `pub run` things

build_web_compilers/README.md
```diff
@@ -0,0 +1,98 @@
...
+of handwriting a `build.dart` file), then all you need is the `dev_dependency`
+listed above.
+
+If you are using a custom build script, you will need to add the following
```
natebosch: I wonder if we should expose a `List<BuilderApplication>` for this use case?

natebosch: Call out that these must be last.

Also, should we consider moving this below configuration to an 'advanced usage' or 'manual usage' section?

build_web_compilers/README.md
jakemac53: done

build_web_compilers/README.md
jakemac53: we could consider doing that.... it would likely be more future proof

jakemac53: Moved it under a `Manual Usage` section

PR #855, opened 2018-01-16
Include the SDK summaries and ensure the root app summary is built.

alan-knight:


natebosch:
@jakemac53 - this looks sane to me, can you also take a look?

Do you know why we didn't need the `linkedSummaryId` part before? Is that something we only use at runtime?

jakemac53:
The linked summary id isn't required to build the root module, but it is required if you want to import the root module which the dev extension presumably does.

My only concern is this will regress performance to some degree for all users, we should possibly hide it behind an option?

natebosch:
Should we work on being able to lazily build assets when they are requested by the client?

jakemac53:
Personally, I don't like the idea of lazily building assets on request because it diverges `serve` from `watch` and `build` to much.

If we did do it, I think we would need to make it so you couldn't lazily request `optional` assets directly, you would have to lazily ask for something non-optional, which in turn caused the optional thing to be built. Otherwise you could end up in a situation where something was built in serve mode but not in build/watch mode which could cause a _lot_ of confusion.

jakemac53:
Ok @alan-knight we talked and decided to add this as an option, which defaults to off for now. Once the extension ships we can flip it to on by default if we decide that is the right thing to do (likely will be).

That means for demo purposes you will need a build.yaml which overrides the option to enable it but it will be a trivial one.

I will work on the change and push it to this PR.

Issue #858, opened 2018-01-17
Silent failures from build_runner serve


nshahan:
In the latest version at 71b79a63e49d98d4aa89c45c389cd6aa56d6e682 `build_runner serve` will fail silently. To repro make sure you are using that version and launch a simple web server on localhost:8080. Then try to run `build_runner serve` on a package. It should fail because the port is in use by another process but the failure doesn't log anything to the console.

jakemac53:
You must have an old server hanging around listening on that port, try this: `fuser -k 8080/tcp`.

jakemac53:
Sorry didn't real the comment fully :). Looks like it is an issue with the isolate approach.

jakemac53:
cc @natebosch

nshahan:
Yeah, reverting back to 76216ea2856b0886c44d7288d456a55a547222a7 (right before the isolate change) gets back to the state where a crash stack trace is printed in the console.

natebosch:
does it work as expected when you use `dart .dart_tool/build/entrypoint/build.dart serve` instead of `pub run`?

nshahan:
@natebosch Yes.

jakemac53:
See https://api.dartlang.org/stable/1.24.3/dart-isolate/Isolate/errors.html?

natebosch:
Fixed by #860

PR #860, opened 2018-01-17
Print errors coming from the Isolate

natebosch:


natebosch:
Fixes #858

natebosch:
@nshahan - you'll want the same thing in your wrapper script

PR #861, opened 2018-01-18
Add build_resolvers package

natebosch:
This will allow us to drop the dependency on `build_barback`, as well as
give us a place to put future enhancements for Resolvers.

- Copy relevant code from `code_transformers` and replace barback
imports with build imports
- Drop dependency on `build_barback` from `build_runner`
- Use the new resolver in `build_runner`
- Tweak some logging and remove usage of `package:source_span`


natebosch:
I don't see any significant performance improvement with this - we had hoped there might be some with less translation between `AssetId` representations etc.

This is a hacky copy - we can us it to drive future improvements and once we can start removing other places we use `build_barback` we'll start to see pruned dependencies as well.

natebosch:
I'm fine holding this until after we've published `build_runner` as it is now.

Will take a look at bringing over the tests

natebosch:
Ported over the tests and modified them to use the `resolveSource` stuff from `build_test`. Deleted the ones that had to do with handling file updates.

build_resolvers/lib/src/resolver.dart
```diff
@@ -195,8 +195,11 @@ class AnalyzerResolver implements ReleasableResolver {
 
   @override
   Future<LibraryElement> findLibraryByName(String libraryName) async =>
-      libraries.firstWhere((l) => l.name == libraryName,
-          defaultValue: () => null) as LibraryElement;
+      (await libraries.firstWhere((l) {
+        print('Name: ${l.name}');
```
jakemac53: probably want to remove this :)

build_resolvers/test/resolver_test.dart
```diff
@@ -0,0 +1,174 @@
...
+
+              main() {
+              } ''',
+      }, (resolver) async {
+        var lib = await resolver.libraryFor(entryPoint);
+        expect(lib.importedLibraries.length, 2);
+      }, resolvers: new AnalyzerResolvers());
+    });
+
+    //test('should list all libraries', () {
```
jakemac53: is this still failing or just accidentally commented out?

natebosch: Just a bad rebase after we got the tests working.

build_runner/lib/src/generate/fold_frames.dart
```diff
@@ -8,6 +8,7 @@ const _foldFramesFor = const [
   '_bazel_codegen',
   'build',
   'build_barback',
```
kevmoo: Remove this?

natebosch: if it showed up (you can still pass the `BarbackResolvers` through if you want) we'd want it folded. We should remove this if and when we delete the package.

PR #865, opened 2018-01-18
Allow `$default` as a target name

### LINKED ISSUE #864: Allow `$default` as package name so it's copy pasteable

natebosch:
Closes #864

This makes some simple cases of configuring builders for the entire
package able to get copied directly rather than rewriting the target nam
for each package.


kevmoo:
Associated readme update?

PR #870, opened 2018-01-18
update all packages for release supporting the latest package:build

jakemac53:


jakemac53:
My strategy is going to be to publish these before I merge (travis will fail). Then I will publish and kick off a rebuild, which should pass.

If it doesn't pass then I did something wrong :).

natebosch:
Publishing before merging would be OK but it's not usually how I approach it.

Typically what I do is merge with a failing travis, then start publishing. This guarantees that I can tag the exact state of the code when it's published.

If we instead publish then merge we just need to be careful that if there are any changes that need to be made in between for some reason that we have some commit to tag with the exact state of the package that was published. I do still find it really useful to have tags that match version numbers.

PR #872, opened 2018-01-19
edited docs/getting_started.md

kwalrath:
An attempt to make getting started a little easier.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -1,50 +1,88 @@
 # Getting started with `build_runner`
 
-`build_runner` can be used as a development server.
+To use `build_runner`, you need a 2.x (dev channel or fresher) version of
+the Dart SDK.
 
-1. Add a `dev_dependency` on `build_runner: ^0.7.0`
-2. Add a `dev_dependency` on `build_web_compilers: ^0.2.0`
-3. Run `pub get`
-4. To start the server run `pub run build_runner serve`
+* [Automated installers](https://www.dartlang.org/install#automated-installation-and-updates)
+* [Direct downloads](https://www.dartlang.org/install/archive#dev-channel)
 
-The first build will be the slowest. After that assets are cached on disk and
-incremental builds will be faster. While the `serve` command is running any
-changes you save to files will trigger rebuilds automatically.
+If you have issues using `build_runner`, see the
+[Troubleshooting section](#troubleshooting), below.
 
-## Available commands
 
-In addition to `serve` you can use:
+## Using `build_runner` as a development server
 
-- `build` - run a single build and exit. This is most useful if your build also
-  generates outputs to your source directory. With `--output <dirname>` this
+1. Edit your package's **pubspec.yaml** file,
+   removing all transformers and
+   adding dev dependencies on **build_runner** and **build_web_compilers**:
+
+   ```yaml
+   ...  
+   environment:
+     sdk: '>=2.0.0-dev <2.0.0'
+   ...  
+   dev_dependencies:
+     build_runner: ^0.7.0
+     build_web_compilers: ^0.2.0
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -1,58 +1,122 @@
...
 
-## Available commands
 
-In addition to `serve` you can use:
+## Using `build_runner` as a development server
 
-- `build` - run a single build and exit. This is most useful if your build also
-  generates outputs to your source directory. With `--output <dirname>` this
+1. Edit your package's **pubspec.yaml** file,
+   removing all transformers and
```
natebosch: removing transformers is optional - people can try it out without breaking their existing `pub serve` workflow.

docs/getting_started.md
```diff
@@ -1,58 +1,122 @@
...
-- `watch` - like `build` but reruns after file changes. With
+
+- **watch:** Like `build` but reruns after file changes. With
   `--output <dirname>` the merged output directory will be kept up to date with
   changes. This can be used to keep the outputs updated for use with another
   filed-based development server.
-- `test` - Creates an output directory and runs `pub run test` within it.
 
-## Switching to Dart2Js
+- **test:** Creates an output directory and runs `pub run test` within it.
```
natebosch: Should we call out that `build_test` needs to be a dev dependency for this to work? The latest version of the command will check for it and warn you - but that might be a surprise

docs/getting_started.md
```diff
@@ -1,58 +1,122 @@
...
 targets:
   my_package_name:
     builders:
-      dart_web_compilers|entrypiont:
+      dart_web_compilers|entrypoint:
         options:
           compiler: dart2js
 ```
 
+<!--PENDING: Why does this show a `builders` line? Is that necessary for
```
natebosch: Is it is necessary. The `dart_web_compiler|entrypoint` is a "builder" just like angular or any other builder - there isn't anything special about compilers in build_runner like there was with pub.

docs/getting_started.md
```diff
@@ -1,58 +1,122 @@
...
+with `pub run build_runner`, because it will be kept up to date with changes
+in the build packages. If your pubspec lists builders, switch to a version of
+the builder-containing package that has a `build.yaml` file.
+
+
+## Troubleshooting
+
+<!-- summarize here. -->
+
+<!-- add something about symptoms if you don't remove transformers -->
```
natebosch: There shouldn't be any symptoms for including or removing the config.

docs/getting_started.md
jakemac53: ya we should add that imo

docs/getting_started.md
```diff
@@ -1,58 +1,122 @@
...
 Older versions of `build_runner` were designed to run with manually written
 build scripts referencing the Builders available in the local package or in
 dependencies. This pattern can still be used when customization is needed
-outside of `build.yaml`, but we recommend most users use the generated build
-script with `pub run build_runner` since it will be kept up to date with changes
-in the build packages. Check with the packages that provide Builders you are
-using for versions published with a `build.yaml` file.
+outside of `build.yaml`, but we recommend using the generated build script
+with `pub run build_runner`, because it will be kept up to date with changes
+in the build packages. If your pubspec lists builders, switch to a version of
```
jakemac53: s/builders/transformers

although this doesn't seem relevant to this section (its duplicated from the above section where it is relevant?)

docs/getting_started.md
jakemac53: We could add a warning if we still see a transformers section or something, idk

docs/getting_started.md
kwalrath: Done.

PR #873, opened 2018-01-19
Serialize AssetGraph to/from List<int>

natebosch:
Helpful for #41

Centralizes all references to `JSON` within the serialization so that we
can update a single place to experiment with other serialization
formats.

### DIFF
--- a/build_runner/test/asset_graph/graph_test.dart
+++ b/build_runner/test/asset_graph/graph_test.dart
@@ -117,17 +117,9 @@ void main() {
           }
         }
 
-        var encoded = JSON.encode(graph.serialize());
-        var decoded = new AssetGraph.deserialize(JSON.decode(encoded) as Map);
+        var encoded = graph.serialize();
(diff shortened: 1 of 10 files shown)


build_runner/test/asset_graph/graph_test.dart
```diff
@@ -117,18 +117,10 @@ void main() {
...
         }
 
-        var encoded = JSON.encode(graph.serialize());
-        var decoded = new AssetGraph.deserialize(JSON.decode(encoded) as Map);
+        var encoded = graph.serialize();
+        var decoded = new AssetGraph.deserialize(encoded);
         expect(graph, equalsAssetGraph(decoded));
       });
-
-      test('Throws an AssetGraphVersionError if versions dont match up', () {
```
jakemac53: We should restore this test in some manner

natebosch: Done

Issue #874, opened 2018-01-19
`pub run build_runner test` returns exit code 0 when tests fail or aren't run


chalin:
Two issues are being reported here.

(1) When tests fail, I'd expect the shell exit code to be nonzero as opposed to (notice the `echo $?` at the end of the command):

```nocode
toh-0 > pub run build_runner test ...; echo $?
[INFO] ensureBuildScript: Generating build script completed, took 332ms
[INFO] BuildDefinition: Reading cached asset graph completed, took 303ms
[INFO] BuildDefinition: Checking for updates since last build completed, took 441ms
[INFO] Build: Running build completed, took 8.0s
[INFO] Build: Caching finalized dependency graph completed, took 103ms
[INFO] CreateOutputDir: Deleting existing output dir `build` completed, took 503ms
[INFO] CreateOutputDir: Creating merged output dir `build` completed, took 1.9s
[INFO] Build: Succeeded after 10.5s with 19 outputs

Running tests...

00:02 +0 -1: test/app_test.dart: Default greeting [E]                                                                                                                                                  
  Expected: 'Hello Angular2'
    Actual: 'Hello Angular'
     Which: is different. Both strings start the same, but the actual value is missing the following trailing characters: 2
  
  
00:02 +2 -1: Some tests failed.                                                                                                                                                                        
0 # <-- tests failed but the exit code is still 0
```

(2) Also, looking at [this Travis log](https://travis-ci.org/dart-lang/site-webdev/jobs/330814689), you'll notice that the test run fails because 

```nocode
UnexpectedExistingOutputsException: Either you opted not to delete existing files, or you are not running in interactive mode and did not specify `deleteFilesByDefault` as `true`.
```

and yet the exit code returned was still 0 (I also confirmed this behavior locally). If tests didn't run successfully, as in this case, then the exit code should be nonzero.

jakemac53:
Fwiw, the exception you saw indicates that you are generating some files to the source tree instead of to cache - which is fine generally - but you will need to add the `--delete-conflicting-outputs` flag for things to work right on travis today.

chalin:
Right, already I've updated the script to include that flag, but it is still misleading IMHO for the command to return a status code of 0 if the tests didn't successfully run (because they weren't run at all :).

jakemac53:
Published in build_runner 0.7.7

PR #878, opened 2018-01-19
Return exit codes from `run`, and send them to the parent isolate.

jakemac53:
Fixes https://github.com/dart-lang/build/issues/874

cc @chalin


natebosch:
cc @nshahan

PR #880, opened 2018-01-19
Don't add watchers to hosted dependencies

### LINKED ISSUE #868: Don't set up watchers for packages in the pub cache
This would be more optimal at the cost of being slightly unsafe if someone does modify a file.

We really don't want people to do this in general so it's probably fine to skip the watch.

natebosch:
Closes #868

We don't expect files in `pub`, `hosted`, or `github` dependencies to
ever change locally.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.7.8
+## 0.7.7+1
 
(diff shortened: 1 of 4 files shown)


jakemac53:
oh we should update the changelog/pubspec and probably do another release as well

PR #882, opened 2018-01-20
Add support for `--config` overriding build.yaml

### LINKED ISSUE #881: Allow specifying a different build.yaml
Related to https://github.com/dart-lang/build/issues/801.

`$ pub run build_runner:build --config build.prod.yaml`

> @matanlurey: Would be really nice to have a dev (DDC mode) with certain flags, and a prod (Dart2JS) with others.

/cc @jakemac53

natebosch:
Closes #881

Plump the flag `--config` through to where we find build config
overrides. If it's non-null look for the file `build.$configKey.yaml`
and use it instead of `build.yaml`.

Rename the `dart2js` build config for `e2e_example` to the new pattern
and use it rather than rewrite it as `build.yaml` during the test.


build_runner/lib/src/package_graph/build_config_overrides.dart
```diff
@@ -12,18 +12,25 @@ import 'package:path/path.dart' as p;
...
-    PackageGraph packageGraph) async {
+    PackageGraph packageGraph, String configKey) async {
   final configs = <String, BuildConfig>{};
   final configFiles = new Glob('*.build.yaml').list();
   await for (final file in configFiles) {
     if (file is File) {
       final packageName = p.basename(file.path).split('.').first;
       final packageNode = packageGraph.allPackages[packageName];
-      final yaml = await file.readAsString();
+      final yaml = file.readAsStringSync();
```
jakemac53: https://github.com/dart-lang/build/issues/884

e2e_example/test/common/utils.dart
```diff
@@ -177,23 +177,28 @@ Future<String> nextStdOutLine(String message) =>
     _stdOutLines.firstWhere((line) => line.contains(message)) as Future<String>;
 
 /// Runs tests using the manual build script.
-Future<ProcessResult> _runManualTests({bool usePrecompiled}) {
+Future<ProcessResult> _runManualTests(
+    {bool usePrecompiled, List<String> args}) {
```
jakemac53: nit: `buildArgs` for consistency/clarity (same below)

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -18,11 +18,13 @@ import 'builder_ordering.dart';
...
-Future<String> generateBuildScript() => logTimedAsync(
-    new Logger('Entrypoint'), 'Generating build script', _generateBuildScript);
+Future<String> generateBuildScript(String configKey) => logTimedAsync(
+    new Logger('Entrypoint'),
+    'Generating build script',
+    () => _generateBuildScript(configKey));
 
-Future<String> _generateBuildScript() async {
-  final builders = await _findBuilderApplications();
+Future<String> _generateBuildScript(String configKey) async {
```
jakemac53: nit: I would make this an optional arg to indicate that it can be null

build_runner/bin/build_runner.dart
```diff
@@ -31,7 +30,8 @@ Future<Null> main(List<String> args) async {
   if (commandName != _generateCommand) {
     logListener = Logger.root.onRecord.listen(stdIOLogListener);
   }
-  var buildScript = await generateBuildScript();
+  var configKey = parsedArgs.command['config'] as String;
```
jakemac53: I don't really like that this is assuming the `config` option exists on all commands (although I don't see why that would change either). Maybe just check that `config` is a valid option first for future compatibility? It would be pretty unlikely that we would catch it with a test if we did add another command that didn't support this option.

natebosch: I'm almost tempted to revert this part of it and not allow the `--config` flag to impact build script generation. Effectively it would mean that only `targets` can be changed, not `builders`. Thoughts? It could end up making sense if we decide to split targets and builders into different files. #885

jakemac53: Sounds good to me as long as we add in some error handling to validate you don't have a builders section.

natebosch: Updated #885 to mention that we should validate this case. I can't think of a simple way do to it in this PR without expanding the scope a lot (or needing a hack like this arg) so I'll go with ignoring it for now.

build_runner/lib/src/build_script_generate/build_script_generate.dart

e2e_example/test/common/utils.dart
natebosch: Done

PR #891, opened 2018-01-22
Export build_script_generate.dart

nshahan:
Allow wrapping scripts to directly call generateBuildScript() without having to run build_runner as a process.

### DIFF
--- a/build_runner/lib/build_runner.dart
+++ b/build_runner/lib/build_runner.dart
@@ -4,7 +4,8 @@
 export 'src/asset/file_based.dart';
 export 'src/asset/reader.dart' show RunnerAssetReader;
 export 'src/asset/writer.dart';
-export 'src/build_script_generate/build_script_generate.dart';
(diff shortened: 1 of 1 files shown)


build_runner/lib/build_runner.dart
```diff
@@ -4,6 +4,7 @@
 export 'src/asset/file_based.dart';
 export 'src/asset/reader.dart' show RunnerAssetReader;
 export 'src/asset/writer.dart';
+export 'src/build_script_generate/build_script_generate.dart';
```
natebosch: can you add a `show` with what we need explicitly? I'm worried we'll add some other public function that we don't want to leak.

natebosch:
We discussed this and I think it's sensible to do - but I want @jakemac53  to approve too.

build_runner/lib/build_runner.dart
nshahan: Done.

PR #895, opened 2018-01-23
Clear last known digest on file changes

### LINKED ISSUE #894: "web/index.html"'s ETag never invalidates
Simple reproduction case:

Files:

```
web/
  index.html
```

Pubspec:

```yaml
name: angular_hello_world

environment:
  sdk: '>=2.0.0-dev <2.0.0'

dependencies:
  angular: ^5.0.0-alpha+4

dev_dependencies:
  build_runner: ^0.7.8
  build_web_compilers: ^0.2.1

dependency_overrides:
  analyzer: ^0.31.0-alpha.1
```

Terminal:

```bash
pub run build_runner serve
```

-----

As viewed by @natebosch as well, a hard-refresh of the browser is required to see changes.

We suspect the ETag is not being invalidated. We might want to add testing around ETag in general.

natebosch:
Fixes #894

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -241,9 +241,7 @@ class AssetGraph {
     // Pre-emptively compute digests for the new and modified nodes we know have
     // outputs.
     await _setLastKnownDigests(
-        newAndModifiedNodes
-            .where((node) => node.isValidInput && node.outputs.isNotEmpty),
-        digestReader);
+        newAndModifiedNodes.where((node) => node.isValidInput), digestReader);
 
     // Collects the set of all transitive ids to be removed from the graph,
     // based on the removed `SourceAssetNode`s by following the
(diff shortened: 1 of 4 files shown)


kevmoo:
No test ❓

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -241,9 +241,7 @@ class AssetGraph {
     // Pre-emptively compute digests for the new and modified nodes we know have
     // outputs.
     await _setLastKnownDigests(
-        newAndModifiedNodes
-            .where((node) => node.isValidInput && node.outputs.isNotEmpty),
-        digestReader);
+        newAndModifiedNodes.where((node) => node.isValidInput), digestReader);
```
jakemac53: We should make sure this doesn't hurt performance significantly but I don't think it really will, unless somebody has a ton of static files (but even if they are ever served we would want to do this).

The other option would be to only set the digest if it previously had a digest set.

natebosch: I think in general we should only end up here if it did have a digest set, but we could also be more explicit about it.

PR #909, opened 2018-01-26
Add section header for `--output`

natebosch:
This should make the `--output` option easier to find and will give us a
section to link to.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -41,11 +41,12 @@ While the `serve` command runs, every change you save triggers a rebuild.
 The first build is the slowest. After that, assets are cached on disk and
 incremental builds are faster.
 
-## Building output served by another server
+## Creating an output directory
 
 Build with `--output <directory name>` to write files into a merged output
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -41,12 +41,18 @@ While the `serve` command runs, every change you save triggers a rebuild.
 The first build is the slowest. After that, assets are cached on disk and
 incremental builds are faster.
 
+## Building output served by another server
```
jakemac53: This title is confusing, sounds like you are building with inputs from some other server. The option is also generally useful for other things.

Maybe just `Creating an output directory`?

natebosch: I was hoping to find a header that matched the users intent rather than the solution.

How about "Building output to be served by another server"

jakemac53: That is just one use case for this flag though, and not necessarily the main one?

natebosch: How about the latest? I changed the heading to something more general and mentioned the particular use case in the body.

PR #923, opened 2018-01-29
Allow overriding Resolvers

natebosch:
This will let us use these utilities to test new Resolver
implementations. Also fix a bug where the `action` was being called once
per input because the complete future wasn't called until the first
action finished.

### DIFF
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -215,18 +215,13 @@ class _ResolveSourceBuilder<T> implements Builder {
 
   final onDone = new Completer<T>();
 
-  var hasStartedAction = false;
-
   _ResolveSourceBuilder(this._action, this._resolverFor, this._tearDown);
 
   @override
   Future<Null> build(BuildStep buildStep) async {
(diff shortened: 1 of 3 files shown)


natebosch:
This will be used to test #861

build_test/lib/src/resolve_source.dart
```diff
@@ -208,11 +215,14 @@ class _ResolveSourceBuilder<T> implements Builder {
...
   final onDone = new Completer<T>();
 
+  var hasStartedAction = false;
+
   _ResolveSourceBuilder(this._action, this._resolverFor, this._tearDown);
 
   @override
   Future<Null> build(BuildStep buildStep) async {
-    if (onDone.isCompleted) return;
+    if (hasStartedAction) return;
```
jakemac53: Not sure this is really the right solution.... its arbitrary (although probably deterministic?) which input would actually get resolved.

I think we should instead force the user to provide a single primary input if they supply more than one input?

natebosch: Ah, we already had `_resolveFor`. Checking against that now.

PR #933, opened 2018-01-30
Delete manual travis changes against mono_repo

natebosch:


jakemac53:
looks like chrome is still failing to launch :(

natebosch:
firefox won't launch either :(

keeping as is for now.

Issue #821, opened 2018-01-08
`test` command doesn't seem to use the compact reporter


jakemac53:
See https://github.com/dart-lang/test/issues/749 which looks like the underlying issue.

kevmoo:
ProcessStartMode.INHERIT_STDIO landed in 2.0.0-dev.20 - https://github.com/dart-lang/sdk/commit/6523896c6fb27a7ead2ced038da4d15164d894d4

jakemac53:
wooohooooooo!!!

PR #934, opened 2018-01-30
use ProcessStartMode.INHERIT_STDIO for running tests

jakemac53:
Fixes https://github.com/dart-lang/build/issues/821

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -391,8 +391,6 @@ class _TestCommand extends BuildRunnerCommand {
       if (options.outputDir == null && outputDir != null) {
         await new Directory(outputDir).delete(recursive: true);
       }
-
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -391,23 +391,22 @@ class _TestCommand extends BuildRunnerCommand {
...
+    var testProcess = await Process.start(
         _pubBinary,
         [
           'run',
           'test',
           '--precompiled',
           precompiledPath,
-        ]..addAll(extraTestArgs));
+        ]..addAll(extraTestArgs),
+        mode: ProcessStartMode.INHERIT_STDIO);
```
natebosch: need to update pubspec with the SDK constraint

jakemac53: ya I forgot to commit that change, updated

Issue #910, opened 2018-01-26
`pub run build_runner build` returns exit code 0 even when SEVERE errors reported


chalin:
If I seed a template error and run test over one of the toh-0 example app like this:

```console
pub run build_runner build --delete-conflicting-outputs --output=build; echo $?
```

I get:

---

[INFO] Entrypoint: Generating build script completed, took 334ms
[INFO] BuildDefinition: Building new asset graph completed, took 663ms
[INFO] BuildDefinition: Checking for unexpected pre-existing outputs. completed, took 1ms
[**SEVERE**] Instance of 'LibraryBuilder' on angular_tour_of_heroes|lib/app_component.dart: Error running TemplateGenerator for angular_tour_of_heroes|lib/app_component.dart.
Template parse errors:
... [some error]...

**Please fix all errors before compiling** (warnings are okay).
}
[WARNING] Instance of 'WebEntrypointBuilder' on angular_tour_of_heroes|test/app_test.dart: Unable to read angular_tour_of_heroes|test/app_test.ddc.js, check your console for compilation errors.
[WARNING] Instance of 'WebEntrypointBuilder' on angular_tour_of_heroes|test/app_test.dart.browser_test.dart: Unable to read angular_tour_of_heroes|test/app_test.ddc.js, check your console for compilation errors.
[INFO] Build: Running build completed, took 20.0s
[INFO] Build: Caching finalized dependency graph completed, took 100ms
[INFO] CreateOutputDir: Creating merged output dir `build` completed, took 1.4s
[INFO] **Build: Succeeded** after 21.6s with 2074 outputs
**0**

---

So, in summary:

- The template compiler reports SEVERE errors
- A message suggests "Please fix all errors before compiling", but yet 
- We get "Build: Succeeded"
- An exit code of 0

I'm using the latest build_* packages:
```console
toh-0 > grep build_ pubspec.yaml
  build_runner: ^0.7.8
  build_test: ^0.10.0
  build_web_compilers: ^0.2.1
```

cc @kwalrath

jakemac53:
There is a flag to fail builds, I can't recall exactly what it is but --help should show it. This is a bit of a grey area because at dev time it isn't clear we want to fail the build, there is value in letting you see what it was able to complete

alorenzen:
I'm not sure I agree that we don't want to fail the build at dev time. Internally, we fail the build if any of the builders report an error.

[...]

PR #937, opened 2018-01-31
Remember failures from previous builds

jakemac53:
Further work towards https://github.com/dart-lang/build/issues/910, we will now remember all failures (regardless of --fail-on-severe flag), and if you provide that flag we will fail the build.

The current message is somewhat lacking, looks roughly like this:
```
[SEVERE] Build: Failed after 6.9s
There were 1 actions with `severe` logs and --fail-on-severe was passed
```
I added another action item to https://github.com/dart-lang/build/issues/910 to add a `--rerun-failing-actions` flag to get more information. I don't think its worth trying to store all the information about previous failures as it could get quite large if one failure caused a large number of other cascading failures.

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -22,6 +22,12 @@ part 'serialization.dart';
 
 /// All the [AssetId]s involved in a build, and all of their outputs.
 class AssetGraph {
+  /// All the [AssetId]s that failed the last time they were ran.
+  ///
+  /// See [markFailed] and [markSucceeded] for more info.
+  final Set<AssetId> _failedAssets;
+  Iterable<AssetId> get failedAssets => _failedAssets;
+
   /// All the [AssetNode]s in the graph, indexed by package and then path.
   final _nodesByPackage = <String, Map<String, AssetNode>>{};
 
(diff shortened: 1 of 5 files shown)


jakemac53:
@chalin This should at least provide all the functionality you need in order to make sure builds fail always with --fail-on-severe, even if its not entirely clear exactly what caused the failure in the followup build case.

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -22,6 +22,13 @@ part 'serialization.dart';
 
 /// All the [AssetId]s involved in a build, and all of their outputs.
 class AssetGraph {
+  /// A map of phase number to primary inputs that failed during that phase.
+  ///
+  /// See [markActionFailed] and [markActionSucceeded] for more info.
+  final Map<int, Set<AssetId>> _failedActions;
+  UnmodifiableMapView<int, Iterable<AssetId>> get failedActions =>
```
natebosch: why do we need to expose this as `UnmodifiableMapView` instead of `Map`?

natebosch: I wasn't clear...

It makes sense to me to wrap it in `UnmodifiableMapView` for defensiveness.. I'm not sure why the getter's signature also needs to say that.

jakemac53: Oh it probably doesn't I guess, it makes it a bit more clear that its not modifiable though?

natebosch: The interface is the same, the methods you can't use will still autocomplete in your editor etc. It's only if you hover over it and look for the type that you'd notice.

jakemac53: > It's only if you hover over it and look for the type that you'd notice.

Ya but that is a fairly common use case right? At least that is usually how I figure out the type of a thing.

Why wouldn't we want to expose this type?

natebosch: 🤷‍♂️ 

I'm fine with it if you think it adds value.

PR #944, opened 2018-02-01
IGNORE: trying out build stages

jakemac53:


jakemac53:
Looks like it works as we hoped :)

PR #946, opened 2018-02-01
Playing around with build stages

jakemac53:


jakemac53:
@kevmoo what do you think about these changes? Essentially we are no longer relying on the matrix expansion at all, which in some ways makes things more clear but does potentially clutter up the generated yaml file.

If you think this format is acceptable I can work on a PR to mono_repo.

natebosch:
Before updating mono_repo I think we need to validate whether caches can carry over using this strategy.

kevmoo:
@jakemac53 – sooo 🔥

Agree with @natebosch , though – let's verify the caching.

Maybe try to roll this w/ angular first and then generalize?

tool/travis.sh
```diff
@@ -7,36 +7,44 @@ set -e
 if [ -z "$PKG" ]; then
   echo -e '\033[31mPKG environment variable must be set!\033[0m'
   exit 1
-elif [ -z "$TASK" ]; then
-  echo -e '\033[31mTASK environment variable must be set!\033[0m'
-  exit 1
+fi
+
+if [ "$#" == "0" ]; then
+	echo -e "\033[31mExpected at least one task!\033[0m"
```
kevmoo: whitespace! Not a biggy...

tool/travis.sh
```diff
@@ -7,36 +7,44 @@ set -e
...
-test_1) echo
-  echo -e '\033[1mTASK: test_1\033[22m'
-  echo -e 'pub run test --run-skipped'
-  pub run test --run-skipped
-  ;;
-*) echo -e "\033[31mNot expecting TASK '${TASK}'. Error!\033[0m"
-  exit 1
-  ;;
-esac
+while (( "$#" )); do
```
kevmoo: huh?

jakemac53: I added support for passing multiple tasks as arguments, this is "while there are more arguments" basically

jakemac53: basically just lets you combine the analyzer/dartfmt tasks into one to save time

tool/travis.sh
jakemac53: oh weird apparently vscode uses tabs by default in shell scripts lol

tool/travis.sh
kevmoo: Ah. So if we want to stack tasks together.

It's a trade off. Fewer things to startup is good.

Losing the clear signal that a failure was due to dartfmt or analyzer is a bit more annoying.

Idea: can we do fun bash tricks to run all tasks (even if one fails) so we get all of the results – and then do a `succeed if every subtask succeeds`.

I hate getting a dartfmt error. Fixing it. Only to find out there is an analyzer error, too...

kevmoo: Of course this also leads me to think we should ponder creating a `.monorepo` config file or something so we can more clearly express this kind of intent without bending over backwards to create a valid `.travis.yml` file

jakemac53: Ya I haven't thought to much about exactly how users would express grouping tasks like this - just playing around with it.

jakemac53: > Idea: can we do fun bash tricks to run all tasks (even if one fails) so we get all of the results – and then do a succeed if every subtask succeeds.

I like that idea for this if we do decide its worth supporting in general (I think it probably is though)

.travis.yml
```diff
@@ -9,9 +9,10 @@ dist: trusty
 jobs:
   include:
     - &analyze_and_format
-      stage: analyze/format
+      stage: analyze_and_format
       dart: dev
       script: ./tool/travis.sh dartanalyzer dartfmt
+      if: env(PKG) IS present
```
kevmoo: when is it not?

jakemac53: For this very first one - I think its weird that it tries to run this one at all but we could just set the first PKG here as well. It just bothered me so I did this instead.

tool/travis.sh
jakemac53: done https://travis-ci.org/dart-lang/build/jobs/336315880

jakemac53:
Fwiw [this](https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices) and [this](https://docs.travis-ci.com/user/build-stages/warm-cache/) indicate to me that if we have a single environment variable (the package) then the cache will be shared across stages for that package and do what we want. Still need to confirm though.

jakemac53:
Ok, @natebosch @kevmoo going to go ahead and submit this for now if you all approve, we can always revert back to the mono_repo version if we want later on.

I will also update angular, and assuming we like the new workflow we can update mono_repo in the near future.

Issue #948, opened 2018-02-02
`pub run build_runner build --help` crashes


kevmoo:
With build_runner 0.7.9

```
> pub run build_runner build --help
[INFO] Entrypoint: Generating build script completed, took 266ms
Performs a single build on the specified targets and then exits.

Usage: build_runner build [arguments]
-h, --help                          Print this usage information.
    --[no-]assume-tty               Enables colors and interactive input when the script does not appear to be running directly in a terminal, for instance when it is a subprocess
    --delete-conflicting-outputs    By default, the user will be prompted to delete any files which already exist but were not known to be generated by this specific build script.

                                    Enabling this option skips the prompt and deletes the files. This should typically be used in continues integration servers and tests, but not otherwise.

    --low-resources-mode            Reduce the amount of memory consumed by the build process. This will slow down builds but allow them to progress in resource constrained environments.
-c, --config                        Read `build.<name>.yaml` instead of the default `build.yaml`
    --[no-]fail-on-severe           Whether to consider the build a failure on an error logged.
-o, --output                        A directory to write the result of a build to.
-v, --verbose                       Enables verbose logging.
    --define                        Sets the global `options` config for a builder by key.

Run "build_runner help" to see global options.
Unhandled exception:
Bad state: Bad response from isolate, expected an exit code but got null
#0      main.<anonymous closure> (http://localhost:61905/build_runner.dart:55)
#1      _RootZone.runUnaryGuarded (dart:async/zone.dart:1316)
#2      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:330)
#3      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:257)
#4      _StreamController&&_SyncStreamControllerDispatch._sendData (dart:async/stream_controller.dart:763)
#5      _StreamController._add (dart:async/stream_controller.dart:639)
#6      _StreamController.add (dart:async/stream_controller.dart:585)
#7      _RawReceivePortImpl._handleMessage (dart:isolate-patch/dart:isolate/isolate_patch.dart:164)
```

PR #953, opened 2018-02-02
make sure the --help command doesn't crash

jakemac53:
Fixes https://github.com/dart-lang/build/issues/948

### DIFF
--- a/build_runner/lib/src/entrypoint/run.dart
+++ b/build_runner/lib/src/entrypoint/run.dart
@@ -10,7 +10,7 @@ import 'options.dart';
 
 /// A common entrypoint to parse command line arguments and build or serve with
 /// [builders].
-Future<int> run(List<String> args, List<BuilderApplication> builders) {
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/entrypoint/run.dart
```diff
@@ -10,7 +10,7 @@ import 'options.dart';
 
 /// A common entrypoint to parse command line arguments and build or serve with
 /// [builders].
-Future<int> run(List<String> args, List<BuilderApplication> builders) {
+Future<int> run(List<String> args, List<BuilderApplication> builders) async {
   var runner = new BuildCommandRunner(builders);
-  return runner.run(args);
+  return (await runner.run(args)) ?? 0;
```
kevmoo: Might could be nice to add a note here that `--help` is the only command we expect to return `null`.

natebosch: Anything that returns null and doesn't `throw` should be fine to use `0`.  I don't think we need to add a comment based on details that could change outside this code

Issue #908, opened 2018-01-26
Push a release for the stable Dart SDK that ...


matanlurey:
... notifies that using `pub run build_runner` requires Dart 2.0+A newer version.

We could link to a `.md` file on this repo.

The rationale here is folks from the conference not yet on a `dev` SDK will try to reproduce the commands and get frustrated when pub says there is no executable not understanding they need a new version of the SDK and the package.

Thoughts? @natebosch @jakemac53 @kevmoo

kevmoo:
Could work in theory. We'd have to do something weird...like release `0.6.9+fyi` or something...

matanlurey:
Yeah, that's reasonable, I think.

jakemac53:
closed by https://github.com/dart-lang/build/pull/954

PR #954, opened 2018-02-02
Add old binary

jakemac53:
Fixes https://github.com/dart-lang/build/issues/908


build_runner/bin/build_runner.dart
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch: [nit] date

PR #957, opened 2018-02-03
Remove `_resolver` field from sources

natebosch:
Drop some unused methods the field.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -114,7 +114,7 @@ class AnalyzerResolver implements ReleasableResolver {
       visiting.add(buildStep.readAsString(assetId).then((contents) {
         var source = sources[assetId];
         if (source == null) {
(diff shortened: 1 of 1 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -272,30 +269,6 @@ class AssetBasedSource extends Source {
 
   @override
   bool get isInSystemLibrary => false;
-
-  Source resolveRelative(Uri relativeUri) {
```
kevmoo: public API? breaking change?

natebosch: we don't export this class, so no it's not breaking

natebosch: `AnalyzerResolvers` is the only class that's public from this file
https://github.com/dart-lang/build/blob/c49f2a6330624fc86ea470d5b2ad141b03ed9cea/build_resolvers/lib/build_resolvers.dart#L5

PR #958, opened 2018-02-03
Log non-empty stderr on build_web_compilers test

natebosch:
Sometimes on Travis the build can be slow and our output format for the
time changes when it takes over 1 minute.

### DIFF
--- a/build_web_compilers/test/build_test.dart
+++ b/build_web_compilers/test/build_test.dart
@@ -33,8 +33,10 @@ void main() {
 
     // 2 - run build - should be no output, since nothing should change
     var result = _runProc('dart', ['--checked', 'tool/build.dart', 'build']);
-    expect(result,
-        contains(new RegExp(r'Build: Succeeded after \S+ with \d+ outputs')));
+    expect(
+        result,
+        contains(
+            new RegExp(r'Build: Succeeded after \S+( \S+)? with \d+ outputs')));
 
     // 3 - get a list of modified `.g.dart` files - should still be empty
     expect(_changedGeneratedFiles(), isEmpty);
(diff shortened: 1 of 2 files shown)


natebosch:
hmm, I thought it was just that the regex didn't match, but it looks like this doesn't even wait for the build to finish...

natebosch:
Ah, so we have a couple problems here.

1. Even when there is output on stderr and the build fails we're returning a 0 exit code
2. Removing the dependency override for analyzer caused the build to stop working.

jakemac53:
> Even when there is output on stderr and the build fails we're returning a 0 exit code

Ya, we should add `--fail-on-severe` I think.

> Removing the dependency override for analyzer caused the build to stop working.

Ouch, ya I think since we weren't running with --fail-on-severe this has been failing for a while?

natebosch:
> this has been failing for a while?

It has been passing until we removed the dependency override that was forcing the latest analyzer for these tests.
https://github.com/dart-lang/build/commit/2e2ec6e34fc59303382484df341129f2c201a9c8#diff-8c687f66dd166104d209dd9440bd6608L33

Started passing again when other packages were published allowing the `0.31` analyzer and this started picking it up without the override.

> Ya, we should add `--fail-on-severe` I think.

I suspect that might not be enough. The build wasn't finishing at all, it would die after the uncaught exception was printed but I guess the VM still exited 0... In any case this is using a custom build script so it might be less critical to figure this out.

jakemac53:
> I suspect that might not be enough. The build wasn't finishing at all, it would die after the uncaught exception was printed but I guess the VM still exited 0... In any case this is using a custom build script so it might be less critical to figure this out.

We probably just need to set `exitCode` to the result of the call to `run`

natebosch:
> We probably just need to set `exitCode` to the result of the call to `run`

Good call. Done.

PR #963, opened 2018-02-05
Update pubspecs and changelogs for release to support latest analyzer

jakemac53:
cc @matanlurey


jakemac53:
oh I will need to remove dependency overrides as well forgot about that

PR #966, opened 2018-02-06
Add override for build_resolvers

natebosch:
We want to run e2e_example with local overrides of everything


natebosch:
Leaving `build_barback` because we still pull that in through some dependency.

PR #969, opened 2018-02-06
Prepare build_web_compilers for release

jakemac53:
Depend on published build_modules.

Also added build_modules to the travis config since I realized its tests weren't actually being ran.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,9 +76,6 @@ jobs:
       dart: stable
     - <<: *unit_test
       env: PKG="build_modules"
-    - <<: *unit_test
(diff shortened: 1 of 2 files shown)


matanlurey:
```
Resolving dependencies... 
Package json_serializable has no versions that match >=0.3.0 <0.4.0 derived from:
- build_modules depends on version ^0.3.0
```

jakemac53:
Ya I don't get the travis failure, I can get a version solve locally with both pub get and pub upgrade....

jakemac53:
Oh nvm it was only with the stable sdk, removing the stable sdk unit tests

PR #970, opened 2018-02-07
Use latest monorepo

jakemac53:
Used mono_repo from https://github.com/dart-lang/mono_repo/pull/51

We lose the combining of dartfmt/dartanalyzer but I think for now its worth just being back on mono_repo to limit hand edits which are error prone.

I would like to add that as a feature to mono_repo anyways (combining of tasks)


natebosch:
> We lose the combining of dartfmt/dartanalyzer

This adds ~10 minutes of task time due to not sharing pub solves and other per-task work.

jakemac53:
> This adds ~10 minutes of task time due to not sharing pub solves and other per-task work.

Ya, we can improve this incrementally from here though. For instance we might want to add a `pub upgrade` stage? Possibly mono_repo would even add that by default instead of doing its hardcoded `pub upgrade`.

jakemac53:
Also note that as a part of this we are actually running a few things on stable that we weren't before, I restored exactly the same tasks that we had in the original travis files.

PR #973, opened 2018-02-07
split e2e tests into 4 shards

jakemac53:
Fixes https://github.com/dart-lang/build/issues/972

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -149,7 +149,19 @@ jobs:
       env: PKG="e2e_example"
       dart: dev
     - stage: e2e_test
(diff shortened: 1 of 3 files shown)


jakemac53:
Looks like this does a reasonable job of sharding :)

jakemac53:
> hmm, one of the shards is much smaller than the others but we cut the wall time in half and only end up using about 5 minutes more task time. Looks like a good change to me.

Ya it isn't perfect but its better than manually trying to split things up I think, we won't forget to add new tests to a shard.

Issue #915, opened 2018-01-27
Add detailed timing breakdown


alorenzen:
For debugging purposes, it would be nice if we could have some sort of per-task breakdown of the build timing.
This could help us see the critical path and prioritize what to optimize.

jakemac53:
This is unfortunately quite difficult because of the async (and lazy in some cases) nature of the builds. One approach is to allow builder authors to do their own tracking - as filed in https://github.com/dart-lang/build/issues/823.

This allows authors to track a bit more specifically the things they care about but it still doesn't solve the problem of accidentally tracking the time spent doing completely unrelated things during async gaps.

matanlurey:
I think #823 is a good start.

It should be possible to tell total-time from `build()` to `build()` complete, right?

jakemac53:
> It should be possible to tell total-time from build() to build() complete, right?

Yes - but unfortunately if you have any lazy inputs the number  you get back would not be meaningful in any way because its doing an arbitrary amount of unrelated work.

PR #988, opened 2018-02-09
Add a basic performance visualization a `/$perf`

jakemac53:
Fixes https://github.com/dart-lang/build/issues/915

For now incremental builds are a lot more interesting to look at, because they do a better job of capturing time spent in any individual actual builder (previous inputs will be built as a part of the `setup` phase).

cc @matanlurey @alorenzen 

[examples.zip](https://github.com/dart-lang/build/files/1712132/examples.zip)


jakemac53:
You can also just save the web page and attach that to github issues, so I did that with a full and incremental build for angular_components_example here to illustrate how its done (and so you can easily see what it looks like).

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -122,6 +122,8 @@ class BuildImpl {
   final bool _verbose;
   final BuildEnvironment _environment;
 
+  BuildPerformanceTracker _performanceTracker;
```
natebosch: non-final fields in this class make me a little nervous... I'm not sure I have a better suggestion though and at least its private 🤷‍♂️

jakemac53:
Oh I forgot to add even the most basic of tests :D, ill go ahead and at least add some unit tests really quick.

build_runner/lib/src/generate/performance_tracker.dart
```diff
@@ -19,45 +19,82 @@ abstract class Timings {
 
 /// The [Timings] of an entire build, including all its [actions].
 abstract class BuildPerformance implements Timings {
-  /// The [Timings] of each [BuildAction] ran in this build.
-  Iterable<BuildActionPerformance> get actions;
+  /// The [Timings] of each phase ran in this build.
+  Iterable<BuildPhasePerformance> get phases;
```
natebosch: were you planning on adding an `@experimental` annotation to this library?

build_runner/lib/src/server/server.dart
```diff
@@ -120,3 +142,85 @@ class AssetHandler {
...
+    for (var phase in action.phases) {
+      var start = phase.startTime.millisecondsSinceEpoch -
+          performance.startTime.millisecondsSinceEpoch;
+      var end = phase.stopTime.millisecondsSinceEpoch -
+          performance.startTime.millisecondsSinceEpoch;
+
+      String phaseName;
+      switch (phase.phase) {
+        case BuilderActionPhase.Setup:
+          phaseName = "setup";
```
natebosch: inconsistent quotes in this file

build_runner/lib/src/server/server.dart
```diff
@@ -120,3 +142,85 @@ class AssetHandler {
...
+        dataTable.addColumn({ type: 'string', id: 'Phase' });
+        dataTable.addColumn({ type: 'number', id: 'Start' });
+        dataTable.addColumn({ type: 'number', id: 'End' });
+        dataTable.addRows([
+$rows
+        ]);
+
+        var options = {
+          colors: ['#cbb69d', '#603913', '#c69c6e']
+          // timeline: {
```
natebosch: delete

build_runner/lib/src/server/server.dart
jakemac53: done

build_runner/lib/src/generate/build_impl.dart
jakemac53: Ya, it isn't ideal. We actually have a lot of these that have to get cleaned up after each build (_lazyPhases, _resourceManager).

The real solution is probably to create another class that handles running a single build, but we can do that separately.

build_runner/lib/src/generate/performance_tracker.dart
jakemac53: done

PR #989, opened 2018-02-09
fix dartjs worker bug

jakemac53:


jakemac53:
Merging with broken appveyor as that is related to the latest dev release not this PR

Issue #995, opened 2018-02-10
build/serve crash:  Unhandled exception: Could not import file builder.dart


afpatmin:
When trying to run command 'pub run build_runner serve' / 'pub run build_runner build',
the terminal outputs:
```
[INFO] Entrypoint: Generating build script completed, took 421ms
Unhandled exception:
IsolateSpawnException: Unable to spawn isolate: Unhandled exception:
Could not import "file:///X:/project/tool/src/builder.dart" 
from "file:///X:/project/.dart_tool/build/entrypoint/build.dart": 
FileSystemException: Cannot open file, path = 'X:\project\tool\src\builder.dart' 
(OS Error: Det går inte att hitta sökvägen., errno = 3)
null
```

The issue occured after upgrading build_runner from v 0.7.9+2 to v 0.7.10. 

The project doesn't contain a build.yaml or any build.dart scripts

build_runner: 0.7.10
OS: Windows 10
SDK: 2.0.0-dev.22.0
IDE: Webstorm 2017.3.4

afpatmin:
travis reported this too: [https://travis-ci.org/dart-lang/build/jobs/339725849](url)

luisvt:
The error is caused because of `built_value` dependency. It looks likt its configuration is asking to create a `builder.dart` file into `tool/src` folder. As work around I created `tool/src/builder.dart` containing next code:

```dart
import 'package:build/build.dart';

Builder builtValueBuilder(_) => null;```

jakemac53:
Ya this is caused by a local builder defined in code_builder which we end up importing even though it never applies to anything.

You can force code_builder version 3.0.1 in dependency_overrides to fix. We will come out with a fix soon as well.

matanlurey:
Fixed by https://github.com/dart-lang/code_builder/pull/198, and published as `3.0.3`.

If you run `pub upgrade` the crashes should go away.

jakemac53:
thanks @matanlurey, I am working on a fix in build_runner as well

PR #996, opened 2018-02-10
skip builders with relative imports that arent defined in the root package

jakemac53:
Fixes the underlying issue for https://github.com/dart-lang/build/issues/995.

We still need to do more at some point (not importing builders which will never be applied) but this should catch this really basic case which can never work.


jakemac53:
Looks like a bunch of other unrelated analyzer hints, probably a result of the latest sdk. Merging this and publishing for now and we can get master green on monday.

PR #1002, opened 2018-02-12
Refer to barback in the README instead of pub

nex3:
Barback is the build system; pub just embeds it.


build/README.md
```diff
@@ -19,18 +19,19 @@ and allows reading/writing files and resolving Dart source code.
 An interface into the dart [analyzer][pub:analyzer] to allow resolution of code
 that needs static analysis and/or code generation.
 
-## Differences between the `build` package and `pub`.
+## Differences between the `build` package and `barback`.
```
natebosch: [nit] keep a mention of pub in the header - this is how most of our users know it

nex3: Done.

PR #1003, opened 2018-02-12
Disable performance tracking by default, add --track-performance to enable

jakemac53:
Did this by creating no-op performance trackers as that seemed the most straightforward. It does mean a lot of abstract classes etc though.


build_runner/lib/src/generate/performance_tracker.dart
```diff
@@ -87,14 +111,45 @@ class BuildPerformanceTracker extends TimeTracker implements BuildPerformance {
...
+  @override
+  Iterable<BuilderActionTracker> get actions => throw new UnimplementedError();
+
+  @override
+  Iterable<BuildPhaseTracker> get phases => throw new UnimplementedError();
+
+  @override
+  BuilderActionTracker startBuilderAction(
+          AssetId primaryInput, Builder builder) =>
+      new BuilderActionTracker.noOp();
```
natebosch: can we avoid a `new` here?

build_runner/lib/src/generate/performance_tracker.dart
```diff
@@ -143,19 +218,60 @@ class BuilderActionTracker extends TimeTracker
...
+  factory TimeTracker.noOp() => _NoOpTimeTracker.sharedInstance;
+
+  void start();
+  void stop();
+}
+
+/// Implementation of a real [TimeTracker].
+///
+/// Use [TimeTracker] factory to get an instance.
+class _TimeTrackerImpl implements TimeTracker {
```
natebosch: IMO the default TimeTracker class should be this and the NoOp version should extend it.

build_runner/lib/src/generate/performance_tracker.dart
jakemac53: We could make it not a factory, but instead a static getter or something. In reality it isn't a new one though, it is returning a static instance (maybe thats what you don't like though haha).

Thoughts?

build_runner/lib/src/generate/performance_tracker.dart
jakemac53: My only problem with that is if we add new methods/fields, the NoOp version would end up inheriting the real implementation. With it set up this way we explicitly have to implement anything new with either a no-op or throw which is a bit safer long term.

build_runner/lib/src/generate/performance_tracker.dart
natebosch: ah, I didn't read far enough down when I wrote this comment. I'm fine with the `new` since the factory keeps the single instance for us.

Issue #1005, opened 2018-02-13
--help: document [target directory]


kevmoo:
Should we update usage to be `build_runner serve [arguments] [directory]` ?

Also for build?

```
Runs a development server that serves the specified targets and runs builds based on file system updates.

Usage: build_runner serve [arguments]
-h, --help                          Print this usage information.
    --[no-]assume-tty               Enables colors and interactive input when the script does not appear to be running directly in a terminal, for instance when it is a subprocess
    --delete-conflicting-outputs    By default, the user will be prompted to delete any files which already exist but were not known to be generated by this specific build script.

                                    Enabling this option skips the prompt and deletes the files. This should typically be used in continues integration servers and tests, but not otherwise.

    --low-resources-mode            Reduce the amount of memory consumed by the build process. This will slow down builds but allow them to progress in resource constrained environments.
-c, --config                        Read `build.<name>.yaml` instead of the default `build.yaml`
    --[no-]fail-on-severe           Whether to consider the build a failure on an error logged.
-o, --output                        A directory to write the result of a build to.
-v, --verbose                       Enables verbose logging.
    --define                        Sets the global `options` config for a builder by key.
    --hostname                      Specify the hostname to serve on
                                    (defaults to "localhost")

Run "build_runner help" to see global options.
```

PR #1014, opened 2018-02-14
update serve usage to mention directory/port trailing args

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1005

I didn't see a 100% clear standard on how to represent this arg (specifically, the optional `:port` part) in a usage statement, this seems to be fairly reasonable though.

I also fixed it so you can provide a series of dirs with no ports and won't get an error.

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -314,7 +314,7 @@ class _ServeCommand extends _WatchCommand {
   }
 
   @override
-  String get invocation => '${super.invocation} [directory[:port]] ...';
+  String get invocation => '${super.invocation} [directory[:port]]...';
 
   @override
   String get name => 'serve';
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -312,6 +313,9 @@ class _ServeCommand extends _WatchCommand {
           help: 'Specify the hostname to serve on', defaultsTo: 'localhost');
   }
 
+  @override
+  String get invocation => '${super.invocation} [directory[:port]] ...';
```
natebosch: [nit] no space between the closing brace and the `...`

natebosch:
> I didn't see a 100% clear standard on how to represent this arg (specifically, the optional `:port` part) in a usage statement, this seems to be fairly reasonable though.

Yup, quite reasonable 😄 

I think if we wanted to follow http://docopt.org/ it would be

`[<directory>[:<port>]]...`

Issue #1007, opened 2018-02-13
Only serve a directory if it exists


kevmoo:
I see

```
Serving `web` on port 8080
Serving `test` on port 8081
```

Even though I don't have `web` or `test`. Should only serve directories that exist.

Related to https://github.com/dart-lang/build/issues/1006

Issue #1006, opened 2018-02-13
Include `example` in default directory list


kevmoo:
right now we serve `web` and `test` – `example` should be included

natebosch:
I'm not convinced we should do this by default, but I suppose with #1007 the impact is minimal

PR #1015, opened 2018-02-14
Add more default dirs to serve and compile

jakemac53:
... but only actually serve them if they exist.

Fixes https://github.com/dart-lang/build/issues/1006
Fixes https://github.com/dart-lang/build/issues/1007

### DIFF
--- a/build_web_compilers/build.yaml
+++ b/build_web_compilers/build.yaml
@@ -12,12 +12,12 @@ builders:
     is_optional: True
     auto_apply: all_packages
     required_inputs:
-      - ".dart"
+      - .dart
   entrypoint:
     target: "build_web_compilers"
     import: "package:build_web_compilers/builders.dart"
     builder_factories:
-      - "webEntrypointBuilder"
+      - webEntrypointBuilder
     build_extensions:
       .dart:
         - .dart.bootstrap.js
--- a/build_web_compilers/build.yaml
+++ b/build_web_compilers/build.yaml
@@ -11,20 +11,31 @@ builders:
         - .ddc.js.map
     is_optional: True
     auto_apply: all_packages
-    required_inputs: [".dart"]
+    required_inputs:
+      - ".dart"
   entrypoint:
     target: "build_web_compilers"
     import: "package:build_web_compilers/builders.dart"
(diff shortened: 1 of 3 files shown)


build_web_compilers/build.yaml
```diff
@@ -11,20 +11,31 @@ builders:
         - .ddc.js.map
     is_optional: True
     auto_apply: all_packages
-    required_inputs: [".dart"]
+    required_inputs:
+      - ".dart"
```
natebosch: [nit] inconsistent quoting with line 27

Out of curiosity - is there a reason to prefer this style over the more compact version?

build_web_compilers/build.yaml
```diff
@@ -11,20 +11,31 @@ builders:
...
   entrypoint:
     target: "build_web_compilers"
     import: "package:build_web_compilers/builders.dart"
-    builder_factories: ["webEntrypointBuilder"]
+    builder_factories:
+      - "webEntrypointBuilder"
```
natebosch: [nit] inconsistent quoting

build_web_compilers/build.yaml
```diff
@@ -11,20 +11,31 @@ builders:
...
     build_to: cache
     auto_apply: root_package
     defaults:
       generate_for:
-        include: ["web/**", "test/**_test.dart"]
-        exclude: ["test/**.node_test.dart", "test/**.vm_test.dart"]
+        include:
+          - web/**
+          - test/**_test.dart
+          - example/**
```
natebosch: I'm a bit concerned about this change. It means we'll spam errors when there are non-web entrypoints in example/

build_web_compilers/build.yaml
jakemac53: Just for consistency, we have other lists that would end up wrapping. We could change everything to this style too, but it feels a bit weird in yaml to me.

build_web_compilers/build.yaml
jakemac53: Only if there are non-web examples in example, and you have a dependency on build_web_compilers. The idea is just to try and do the right most of the time - I think its fairly safe to assume web entrypoints everywhere if somebody has this dependency.

Issue #1021, opened 2018-02-15
Add test for stack trace mapping in the browser console


jakemac53:
This has broken multiple times, so we need to set up a web driver test or something along those lines to check the actual console output.

We should also check the package:test output in the terminal once that is actually working (won't require webdriver at least).

matanlurey:
I don't mind trying this today/tomorrow assuming nobody was going to start earlier.

matanlurey:
Some questions:

* Do we want to verify `pub run test` maps or just in the browser?
* Do we care about DDC and Dart2JS or just DDC?

jakemac53:
We do already have a test for the stack traces printed by the test command (but only for DDC).

Ideally we would check both dart2js and ddc though.

davidmorgan:
Closing as stale.

Issue #916, opened 2018-01-29
Generate source maps with build_web_compilers


eredo:
The `build_web_compilers` should also generate source map files. The `.ddc.js` output files contain the reference to a `.ddc.js.map` file but the files are not generated.

The verbose logging shows that `.js.map` files are generated. However it looks like they are not persisted to disk and chrome doesn't recognise them in regular stack logging.

jakemac53:
I see the sourcemaps for e2e_example, and they are working in chrome. Can you confirm you have js sourcemaps enabled in devtools (enable javascript sourcemaps under dev tools settings)?

matanlurey:
I've seen a few people murmur about this on Gitter. We need more information.

@eredo Could you provide:

1. What version of `build_runner` and `build_web_compilers` is being used?
2. What operating system and version?
3. What version of the Dart SDK? (`dart --version`)
4. An example terminal output that shows `.js.map` files being generated
5. An example of on-disk missing files
6. An example of Chrome failing to fetch the files

That will help greatly in debugging this.

eredo:
After doing some further digging, I realised that Intellij simply hides the source map. So the files do exist, sorry! But the chrome console output doesn't contain the proper source location in the exception. Is there maybe some extra work necessary? Source maps are activated in the chrome developer tools settings.
Setting up a break point in the `.dart` file actually works, so I assume loading of the source maps works as expected.

Example project: https://github.com/eredo/poc_angulardart_subapps
Example component:

```dart
import 'package:angular/angular.dart';

@Component(
  selector: 'stack-trace',
  template: '<button (click)="throwIt()">Throw exception</button>',
)
class StackTraceComponent {
  void throwIt() => throw new Exception('this is a test exception.');
}
```

I'm adding the details @matanlurey asked for, in case anybody else is having the originally assumed issue.
Dependencies:

```
build 0.12.0
build_barback 0.5.0+2
build_config 0.2.2+1
build_runner 0.7.8
build_test 0.10.0
build_web_compilers 0.2.1
```

OS: `MacOS 10.13.2`
Dart SDK: `Dart VM version: 2.0.0-dev.19.0 (Fri Jan 19 08:01:12 2018 +0100) on "macos_x64"`
Output directory in `.dart_tool/generated/...`:

```
-rw-r--r--   1 user  group   1794 Jan 29 19:30 stack_trace_component.ddc.js
[...]

PR #1022, opened 2018-02-15
Work towards fixing ddc source maps

jakemac53:
Towards https://github.com/dart-lang/build/issues/916 and https://github.com/dart-lang/angular/issues/871.

After this we are now getting an empty stack trace in the console, but stack traces in the browser work as expected. The console logging for package:test looks like a package:test issue where it is trying to re-map the already mapped stack trace, going to pass that part off to @grouma for further investigation. 

This is pretty difficult to write a test for, we will need to set up webdriver I think. I filed https://github.com/dart-lang/build/issues/1021 to track adding that test as a followup.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,24 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Dart command line",
+            "type": "dart-cli",
+            "request": "launch",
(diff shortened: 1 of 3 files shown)


jakemac53:
Actually I have some other changes coming which will make this better as well

grouma:
@jakemac53 and I investigated the issue and the additional mapping from dart2js is happening due to the top level empty source map. Removing this sourcemap resolves the issue. We have one final issue in that sdk stack frames are not properly folded. There are a couple of potential solutions but I'd like to discuss with @nex3 first. I believe there is some default magic that we are skipping over for some reason.

jakemac53:
Ok so I did end up needing a small change in the sdk, https://dart-review.googlesource.com/c/sdk/+/41921, but after that and this pr everything should be working.

Issue #1008, opened 2018-02-13
Changing `build.yaml` should cause rebuild w/ serve/watch


kevmoo:

jakemac53:
note that we should technically watch all path dependencies build.yaml files as well

natebosch:
[nit] all `build.yaml` files _that we read_.
With `--config` or with `packageName.build.yaml` overrides things aren't as straightforward

jakemac53:
Or also, if you create a new _build.yaml_ file that _would have been_ read :|

PR #1027, opened 2018-02-16
Exit on build.yaml changes

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1008

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -11,6 +11,7 @@ import 'package:build_runner/src/watcher/node_watcher.dart';
 import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
 import 'package:logging/logging.dart';
+import 'package:path/path.dart' as p;
 import 'package:stream_transform/stream_transform.dart';
 import 'package:watcher/watcher.dart';
 
@@ -73,6 +74,7 @@ Future<ServeHandler> watch(
       directoryWatcherFactory: directoryWatcherFactory,
       onLog: onLog);
   var options = new BuildOptions(environment,
+      configKey: configKey,
       deleteFilesByDefault: deleteFilesByDefault,
       failOnSevere: failOnSevere,
       packageGraph: packageGraph,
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -212,16 +216,26 @@ class WatchImpl implements BuildState {
         })
         .asyncMap<AssetChange>((change) {
           // Kill future builds if the root packages file changes.
```
natebosch: does this comment need to change?

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -212,16 +216,26 @@ class WatchImpl implements BuildState {
...
+                _terminateCompleter.complete();
+                _logger
+                    .severe('Terminating builds due to package graph update, '
+                        'please restart the build.');
+              }
+              return change;
+            });
+          } else if (p
+              .basename(id.path)
+              .contains(new RegExp('^build(\.$_configKey)?\.yaml\$'))) {
```
natebosch: I think we also need to terminate for all `build.yaml` file changes. That file impacts the build script.

natebosch:
Should we consider this fixed if we aren't handling `package_name.build.yaml` files?

Does this handle `build.yaml` file changes in path dependencies? - If so we should ignore the `$configKey` changes for path dependencies because only the root package honors that config.

jakemac53:
> Should we consider this fixed if we aren't handling package_name.build.yaml files?

Ah I forgot about those, easy enough to add to this.

> Does this handle build.yaml file changes in path dependencies?

Yes it does

> If so we should ignore the $configKey changes for path dependencies because only the root package honors that config.

Ah ok I didn't know that, I can change that behavior.

build_runner/lib/src/generate/watch_impl.dart
jakemac53: ill move it down

build_runner/lib/src/generate/watch_impl.dart
jakemac53: Yes, this catches both `build.yaml` and `build.<config>.yaml`

Issue #1033, opened 2018-02-20
--output shouldn't output old optional assets that were not needed by the last build


jakemac53:
Today, if you build with dartdevc, and then do a dart2js build, you actually end up with the dartdevc modules output to the merged directory. Everything still works, but there are a lot of turds in the output as a result.

I think to resolve this we can skip optional assets initially when iterating the graph, but then make sure to include all optional inputs of required outputs.

jakemac53:
Re-opening because the serve portion of this hasn't been implemented

kevmoo:
> Re-opening because the serve portion of this hasn't been implemented

Just hit this w/ `webdev` tests.

PR #1034, opened 2018-02-20
Remove extra non-required outputs from the merged dir

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1033

I also deleted the `create_merged_dir` executable because it can't work as it was implemented any more and has been replaced by the `--output` option anyways.


jakemac53:
> I wonder if we want any of these utility methods somewhere more general?
> 
> For example, do we want to serve these files through the serve handlers or should they be skipped there as well?

Hmm, that is a good point. We would still serve these files today which makes things a bit inconsistent.

To do this efficiently enough for serve mode I think would require adding a field to `GeneratedAssetNode` though... and the logic to update that could be a bit challenging.

jakemac53:
Heh, it turns out the travis failures here are caused by exactly this issue - the `.errors` files aren't inputs to any required action so they aren't being written to the output dir, and we end up with test timeouts instead of explicit failures as a result.

natebosch:
So we depend on having _all_ of the outputs from any action which had _any_ output read later?

jakemac53:
> So we depend on having all of the outputs from any action which had any output read later?

In this specific use case we were relying on the fact that the error files for ddc were output as part of the same action, yes.

I have updated it to specifically ask for those files for any missing js module.

In general, this actually seems like a _good_ change, as it means you have to be more explicit. We should try and figure out a good way of enforcing this at serve time as well though.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -95,6 +95,7 @@ Future<List<Module>> _ensureTransitiveModules(
   // Check that each module is readable, and warn otherwise.
   await Future.wait(jsModules.map((jsId) async {
     if (await reader.canRead(jsId)) return;
+    await reader.canRead(jsId.addExtension('.errors'));
     log.warning(
         'Unable to read $jsId, check your console for compilation errors.');
```
natebosch: Should we mention where to find the `.errors` files in case this was run without `--rerun-failing-actions` and so the message might not be on the console?

jakemac53: Sure ya, I like that idea

jakemac53:
Going to sit on this pr for a little bit and see if I can come up with a model that works in both serve and build mode efficiently. If not I think this is still worth landing, but it does introduce a bit of a weird corner case.

PR #1035, opened 2018-02-21
Remove locking from AnalyzerResolver

natebosch:
- Remove the BuildStep and `lastPhaseComplete` fields.
- Rename `performResolve` as `resolve` since no additional Future
  handling needs to happen.
- Add `PerActionResolver` which tracks the entrypoints for a given
  fetched resolver to crawl `libraries` and `findLibraryByName`.

We may want to cache the `libraries` with something like a
`CachingIterable` rather than recrawling imports each time if we find
that the current implementation is expensive.


natebosch:
On its own this doesn't give any noticeable performance improvement, but it opens up the possibility for things like making the Angular builder optional.

build_resolvers/lib/src/resolver.dart
```diff
@@ -18,26 +19,59 @@ import 'package:path/path.dart' as native_path;
...
+      if (!await _delegate.isLibrary(entryPoint)) continue;
+      final library = await _delegate.libraryFor(entryPoint);
+      toVisit.add(library);
+      seen.add(library);
+    }
+    while (toVisit.isNotEmpty) {
+      final current = toVisit.removeFirst();
+      yield current;
+      final toCrawl = current.importedLibraries
+          .followedBy(current.exportedLibraries)
```
jakemac53: nice, I didn't realize this had landed

build_resolvers/lib/src/resolver.dart
```diff
@@ -18,26 +19,59 @@ import 'package:path/path.dart' as native_path;
...
+/// Implements [Resolver.libraries] and [Resolver.findLibraryByName] by crawling
+/// down from entrypoints.
+class PerActionResolver implements ReleasableResolver {
+  final ReleasableResolver _delegate;
+  final Iterable<AssetId> _entryPoints;
+
+  PerActionResolver(this._delegate, this._entryPoints);
+
+  @override
+  Stream<LibraryElement> get libraries async* {
```
jakemac53: should we add the `canRead` check here?

build_resolvers/lib/src/resolver.dart
```diff
@@ -138,40 +143,18 @@ class AnalyzerResolver implements ReleasableResolver {
...
-    }
-    return new Stream.fromIterable(_libraries);
-  }
-
-  void _collectLibraries(LibraryElement lib) {
-    if (lib == null || _libraries.contains(lib)) return;
-    _libraries.add(lib);
-    lib.importedLibraries.forEach(_collectLibraries);
-    lib.exportedLibraries.forEach(_collectLibraries);
+    throw new UnimplementedError();
```
jakemac53: Probably worth adding a comment about why this always throws (ditto below). Or maybe a single comment on the class saying why a couple of its methods throw.

jakemac53:
I believe that this currently attempts to update all sources any time you call `Future<ReleasableResolver> get(BuildStep buildStep)` (because that calls `_performResolve`), which going to be broken in certain scenarios.

Specifically, lazy actions which don't have access to certain sources can end up removing those sources from view of the later actions that do have access to them (if they call `_performResolve` after the later action does, but while the later action is still processing).

I think that `_performResolve` needs to be smarter in this case. Specifically, I think the resolver needs to be a `Resource`, so it can track when a build starts/stops, and it should never remove a source during a build. We could also probably be smarter about checking for source updates but that could be an optimization.

build_resolvers/lib/src/resolver.dart
natebosch: I was hesitant to use it and bump the SDK constraint, but it was too tempting to pass up :)

build_resolvers/lib/src/resolver.dart
natebosch: Added a TODO. Do you think we need it right away?

build_resolvers/lib/src/resolver.dart
natebosch: Done

natebosch:
Added some checking to avoid removing assets which were readable during the current build as well as a `reset` called in between builds.

This is done manually in build_runner for now. We can investigate making this a Resource later.

PTAL

build_resolvers/lib/src/resolver.dart
jakemac53: I don't think its strictly necessary no

natebosch:
Added the test we talked about. Confirmed that it does find the bug with not adding to `_seenAssets` and  passes after the fix.

PR #1036, opened 2018-02-21
Asset graph handler

jakemac53:
Add a basic handler for the `$graph` path to the dev server which lets you search for nodes and view their inputs/outputs and some basic data about the node.

This could be improved a lot but we can do that as one-offs as needed or as we feel like it.


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -297,9 +299,9 @@ class WatchImpl implements BuildState {
 
   /// Checks if we should skip a watch event for this [change].
   bool _shouldProcess(AssetChange change) {
-    assert(_assetGraph != null);
-    if (_isCacheFile(change) && !_assetGraph.contains(change.id)) return false;
-    var node = _assetGraph.get(change.id);
+    assert(assetGraph != null);
```
natebosch: are these changes necessary? should we still refer to `_assetGraph` from within this class?

build_runner/lib/src/server/graph_viz.js
```diff
@@ -0,0 +1,90 @@
+window.$build = {}
```
natebosch: copyright notice on this and other new files

build_runner/lib/src/server/graph_viz_main.dart
```diff
@@ -0,0 +1,49 @@
+import 'dart:convert';
```
natebosch: why does the file name mention graph viz?

build_runner/lib/src/generate/watch_impl.dart
jakemac53: done

build_runner/lib/src/server/graph_viz_main.dart
jakemac53: I just wanted it to be unique, its the entrypoint for the asset graph visualizer

build_runner/lib/src/server/graph_viz.js
jakemac53: done

Issue #622, opened 2017-11-16
Consider changing `needsUpdate` to a `state` field which is an enum of possible states


jakemac53:
This should allow us to avoid issues like the one discovered [here](https://github.com/dart-lang/build/pull/621)

jakemac53:
We would also want to guard against updating any of the fields unless the node is in a certain state, etc.

natebosch:
So would the states `wasOutput` and `wasNotOutput` imply it no longer needs an update?

jakemac53:
I would leave the `wasOutput` field as is - other than we could potentially throw if you try and read it when the node is not in a fully updated state.

jakemac53:
We might not really need an enum - I am not sure if we need more than 2 states. The real win would be enforcing that you can't modify any of the fields unless `needsUpdate == false`.

jakemac53:
For clarity though it might make sense to have 3 states:

- Up to date
- Needs update
- Currently building

natebosch:
Yeah I'm not sure if we'd be able to do anything with currentlyBuilding.

Disallowing updates would be interesting...

jakemac53:
I think its worth doing this just for the glob case, we can have a state enum as follows:

```
enum NodeState {
  upToDate,
  mayNeedUpdate,
  definitelyNeedsUpdate,
}
```

jakemac53:
cc @natebosch how do you feel about those names?

natebosch:
synced offline. The other option we could at some point investigate is a synthetic asset node representing the result of a glob - we could then invalidate it's digest and have better cohesiveness with how we handle other changes. That has other costs that the enum don't have so I think we should go with the enum you've proposed.

PR #1037, opened 2018-02-21
Change `needsUpdate` to a `state` enum and use that for better glob invalidation

jakemac53:
Fixes https://github.com/dart-lang/build/issues/622


build_runner/lib/src/asset_graph/node.dart
```diff
@@ -105,8 +116,8 @@ class GeneratedAssetNode extends AssetNode {
   /// The primary input which generated this node.
   final AssetId primaryInput;
 
-  /// Whether or not this asset needs to be updated.
-  bool needsUpdate;
+  /// The current [GeneratedNodeState] for this node.
```
natebosch: drop the comment?

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -469,20 +469,23 @@ class _SingleBuild {
...
     // We only check the first output, because all outputs share the same inputs
     // and invalidation state.
-    if (!node.needsUpdate) return false;
+    if (node.state == GeneratedNodeState.upToDate) return false;
+    // Early bail out condition, this is a forced update.
+    if (node.state == GeneratedNodeState.definitelyNeedsUpdate) return true;
     // TODO: Don't assume the worst for globs
     // https://github.com/dart-lang/build/issues/624
-    if (node.previousInputsDigest == null || node.globs.isNotEmpty) {
+    if (node.previousInputsDigest == null) {
```
natebosch: ah, I guess we could have also considered clearing out the previous digest to indicate a glob change...

I think I like the enum better.

build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.7.12-prerelease
```
natebosch: is "prerelease" better than "dev" in some way?

jakemac53: I think so yes, dev released in theory could be published, this more clearly indicates that it's not imo.

build_runner/lib/src/asset_graph/node.dart

build_runner/lib/src/generate/build_impl.dart
jakemac53: ack

jakemac53:
Merged on red because the hanging tests look related to a package dependency or sdk update (they fail on master as well)

PR #1040, opened 2018-02-22
Ignore cast failures by default

jakemac53:
cc @vsmenon @matanlurey

### DIFF
--- a/build_web_compilers/lib/src/web_entrypoint_builder.dart
+++ b/build_web_compilers/lib/src/web_entrypoint_builder.dart
@@ -23,10 +23,16 @@ enum WebCompiler {
 
 /// The top level keys supported for the `options` config for the
 /// [WebEntrypointBuilder].
-const _supportedOptions = const [_compiler, _dart2jsArgs, _buildRootAppSummary];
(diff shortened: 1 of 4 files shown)


build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -23,10 +23,16 @@ enum WebCompiler {
...
+const _supportedOptions = const [
+  _compiler,
+  _dart2jsArgs,
+  _buildRootAppSummary,
+  _ignoreCastFailures
+];
 const _buildRootAppSummary = 'build_root_app_summary';
 const _compiler = 'compiler';
 const _dart2jsArgs = 'dart2js_args';
+const _ignoreCastFailures = 'ignore_cast_failures';
```
kevmoo: Would be nice to have the behavior documented *somewhere*.

Re: DDC-only, likely to change default in the future, etc

jakemac53: It's in the changelog - but ya we should get in the habit of updating real docs with new features.

Issue #1044, opened 2018-02-23
If no source directories match then serve mode never starts a server.


jakemac53:
This is pretty confusing, and includes if you are generating files under `web` or some folder that would match. At a minimum we should give a warning that no directories matched.

nshahan:
I think this issue should probably stay open. The message is helpful, but it would be nice if serve checked for source or generated directories.

jakemac53:
Re-opened, but assigned low priority since this is a pretty narrow use case

natebosch:
I think if you manually specify the directory it works, is that good enough for your use case?

`pub run build_runner serve web`

jakemac53:
That is the workaround but it shouldn't be required technically. The issue is if you have a project where the entire web directory is generated.

natebosch:
Does this come up often enough that it's worth making smooth?

nshahan:
Probably not very common, just very confusing when it doesn't work. For my use case the new message and the ability to specify the directory manually works fine. 

Hypothetically it could be more confusing if you have multiple web entry points, some that exist as source code and some that are completely generated.

natebosch:
I think this is working now.

PR #1045, opened 2018-02-23
Add warning if nothing will be served in `serve` mode

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1044

cc @nshahan


jakemac53:
merging on red as this looks like the same chrome issue angular was seeing, and its not related to this pr

PR #1056, opened 2018-02-26
[WIP] Remove extra non-required outputs from the merged dir

jakemac53:
Alternative to https://github.com/dart-lang/build/pull/1034

Fixes https://github.com/dart-lang/build/issues/1033

I also deleted the `create_merged_dir` executable because it can't work as it was implemented any more and has been replaced by the `--output` option anyways.

@natebosch if you like this approach more I can implement it in the serve handler as well.


jakemac53:
Basically the approach here is to track the number of required outputs which depend on a given input, and then only output optional outputs which have a number > 0.

Note that I still need to update the serialization/deserialization logic as well.

jakemac53:
hmmm I did just realize that technically we would need to go increment these counters for all transitive inputs as well...

jakemac53:
ya  closing this as its not going to help over the other approach

Issue #1065, opened 2018-02-27
Add `runs_before` *and* `applies_builders` to builder config


natebosch:
Indicates that a different builder should be run on this target (package) after this builder, including it's outputs. Can be used to effect builder ordering when that builder would have run anyways, and can cause builders to run that would not have run otherwise.

Necessary for a "CombineOutputs" builder that can be opted in for source_gen like scenarios. #975 

cc @nshahan

jakemac53:
`requires_builder` imo because it can cause a builder to run that was not applied at all previously.

natebosch:
synced offline, these probably end up needing to be separate options.

Going with `runs_before` and `applies_builders` for now.

PR #1066, opened 2018-02-28
Add runs_before Builder config and use it for ordering

natebosch:
Towards #1065

- Add `runs_before` parsing and include it in the test.
- Don't split builders into groups based on whether they have
  `requires_input`. This could slightly change ordering of some builders
  but not in a way that has any negative impact.
- Use `runs_before` to set dependencies for builders when ordering them.

### DIFF
--- a/build_runner/test/build_script_generate/builder_ordering_test.dart
+++ b/build_runner/test/build_script_generate/builder_ordering_test.dart
@@ -1,0 +1,67 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'package:build_runner/src/build_script_generate/builder_ordering.dart';
+
+import '../common/build_configs.dart';
+
+void main() {
+  group('Builder ordering', () {
+    test('orders builders with `runs_before`', () async {
+      final buildConfigs = parseBuildConfigs({
+        'a': {
+          'builders': {
+            'runs_second': {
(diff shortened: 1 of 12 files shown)


natebosch:
cc @nshahan

nshahan:
You beat me to the PR. :smiley:

I am still having a confusing issue with a file not being ready for the angular or ddc builders at the right time in the gallery project I am working on. Same thing with your version and mine. Maybe you can look at it with me tomorrow.

build_runner/test/build_script_generate/builder_ordering_test.dart
```diff
@@ -0,0 +1,67 @@
...
+
+import 'package:test/test.dart';
+
+import 'package:build_runner/src/build_script_generate/builder_ordering.dart';
+
+import '../common/build_configs.dart';
+
+void main() {
+  group('Builder ordering', () {
+    test('orders builders with `runs_before`', () async {
```
jakemac53: nit: add some error validations tests for a cycle created by `runs_before` and `required_inputs`?

nshahan:
Deleting the .dart_tool directory is what finally got the missing file to generate for me. I'm not sure if that was overkill but now I can see the order of the builders get changed as I would expect it to with this change.

build_runner/test/build_script_generate/builder_ordering_test.dart
nshahan: nit: add a warning to the documentation that describes a the danger of creating a cycle.

natebosch: Added a separate section in the build_config README about builder ordering. Added a test that when there is a cycle it throws.

Issue #1046, opened 2018-02-24
Check for `foo.dart` when `foo.module` doesn't exist


natebosch:
We lost a chunk of time trying to investigate why a `.module` file wasn't getting created when the issue was that we had an import to a non-existent dart file. The real dart file was in a different directory.

We should check for the dart file and adjust the error message since this points to a problem with the code, not a problem with the build order etc.

cc @nshahan

matanlurey:
Not sure the best way to solve it, but I investigated a bit:

https://github.com/dart-lang/build/blob/20f533b243d6f8b0c0b0f219ee531a9b504082da/build_modules/lib/src/modules.dart#L125-L127

... is confusing. You might want to consider making it a `.fine`.

https://github.com/dart-lang/build/blob/20f533b243d6f8b0c0b0f219ee531a9b504082da/build_web_compilers/lib/src/dev_compiler_builder.dart#L45-L51

... prints a blank error.

jakemac53:
I am going to just make `computeTransitiveDependencies` throw if it fails to find a module. I don't think it ever makes sense to continue in that case.

PR #1068, opened 2018-02-28
Add nicer missing module reporting

jakemac53:
In normal workflows, with this change, you will now only see errors for invalid imports reported by the entrypoint builder, and they will look something like this:

```
[SEVERE] Instance of 'WebEntrypointBuilder' on e2e_example|web/main.dart:  

Unable to find modules for some sources, check the following imports:

`import 'missing.dart';` from e2e_example|web/sub/message.dart at 3:1
```

Fixes https://github.com/dart-lang/build/issues/1046

### DIFF
--- a/build_modules/lib/src/errors.dart
+++ b/build_modules/lib/src/errors.dart
@@ -91,9 +91,11 @@ Future<String> _missingImportMessage(
   var parsed = parseDirectives(contents, suppressErrors: true);
   var import =
       parsed.directives.whereType<UriBasedDirective>().firstWhere((directive) {
-    var id = new AssetId.resolve(directive.uri.stringValue, from: sourceId);
+    var uriString = directive.uri.stringValue;
+    if (uriString.startsWith('dart:')) return false;
+    var id = new AssetId.resolve(uriString, from: sourceId);
     return id == missingId;
-  });
+  }, orElse: () => null);
   var lineInfo = parsed.lineInfo.getLocation(import.offset);
   return '`$import` from $sourceId at $lineInfo';
 }
--- a/build_modules/lib/src/errors.dart
+++ b/build_modules/lib/src/errors.dart
@@ -58,7 +58,7 @@ class MissingModulesException implements Exception {
       List<Module> transitiveModules,
       AssetReader reader) async {
     var buffer = new StringBuffer(
-        ' \n\nUnable to find modules for some sources, check the following '
+        ' Unable to find modules for some sources, check the following '
         'imports:\n\n');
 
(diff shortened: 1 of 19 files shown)


build_modules/lib/src/errors.dart
```diff
@@ -37,3 +42,58 @@ class KernelSummaryException extends _WorkerException {
...
+
+  MissingModulesException._(this.message);
+
+  static Future<MissingModulesException> create(
+      Module module,
+      Set<AssetId> missingSources,
+      List<Module> transitiveModules,
+      AssetReader reader) async {
+    var buffer = new StringBuffer(
+        ' \n\nUnable to find modules for some sources, check the following '
```
natebosch: why do we need the preceding newlines here? That seems like it should be the responsibility of whatever is using this string.

build_modules/test/summary_builder_test.dart
```diff
@@ -13,60 +14,89 @@ import 'util.dart';
...
-      () async {
-    var expectedOutputs = <String, Matcher>{
-      'b|lib/b.unlinked.sum': new HasUnlinkedUris(['package:b/b.dart']),
-      'a|lib/a.unlinked.sum': new HasUnlinkedUris(['package:a/a.dart']),
-      'a|web/index.unlinked.sum':
-          new HasUnlinkedUris([endsWith('web/index.dart')]),
-    };
-    await testBuilder(new UnlinkedSummaryBuilder(), assets,
-        outputs: expectedOutputs);
+    test("can output linked analyzer summaries for modules under lib and web",
```
natebosch: [nit] mixed quote types

same in other test files

build_modules/lib/src/errors.dart
jakemac53: It looks nicer and these are directly logged in 3 places :man_shrugging: 

We could make this always happen for severe logs?

build_runner/lib/src/logging/build_for_input_logger.dart
```diff
@@ -60,6 +60,15 @@ class ErrorRecordingLogger implements Logger {
     if (logLevel >= Level.SEVERE) {
       _errorWasSeen = true;
     }
+    if (logLevel >= Level.WARNING) {
```
natebosch: I just realized this method should never have worked... including the `_errorWasSeen` tracking. Are we missing tests?

Since we have a `_delegate` rather than `extends` the calls to `_delegate.sever()` never hit this method..

natebosch: Ah, because we also set `_errorWasSeen` directly in the call to `severe` :)

PR #1080, opened 2018-03-02
add e2e test that runs on latest raw release in cron jobs only

jakemac53:
cc @kevmoo @matanlurey for fyi, I think this should work... :D

### DIFF
--- a/build_resolvers/pubspec.yaml
+++ b/build_resolvers/pubspec.yaml
@@ -5,7 +5,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_resolvers
 
 environment:
-  sdk: '>=2.0.0-dev.23 <2.0.0'
+  sdk: '>=2.0.0-dev.32 <2.0.0'
 
 dependencies:
   analyzer: ">=0.27.1 <0.32.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,6 +136,10 @@ jobs:
       script: ./tool/travis.sh test_5
       env: PKG="e2e_example"
       dart: dev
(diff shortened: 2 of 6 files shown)


kevmoo:
Once this lands I'll bump the cron run frequency to daily – and we'll see.

Maybe add "api" to the list, too? Then we can hit it manually?

build_resolvers/pubspec.yaml
```diff
@@ -5,7 +5,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_resolvers
 
 environment:
-  sdk: '>=2.0.0-dev.23 <2.0.0'
+  sdk: '>=2.0.0-dev.32 <2.0.0'
```
natebosch: why so high? I thought `orElse` was added a few releases ago?

jakemac53: @kevmoo said it just got released, I took his word for it because digging for it is more effort than its worth.

matanlurey: I'm supportive of this, less variance in having to figure out what dev SDK works and doesn't (especially given most dev users are using brew/automatically updating anyway).

jakemac53:
merging on red as we are red for other issues (stack traces getting broken)

kevmoo:
It worked!

https://travis-ci.org/dart-lang/build/builds/348382639

Used the trigger build option – scroll to the bottom to see the cron item

kevmoo:
Just updated the cron job to run daily...

This is where we could wire up the Chat webhook 😄

jakemac53:
nice, will fix the redness shortly so we can see it actually run but ya looks like its working!

PR #1081, opened 2018-03-02
try adding an after_failure script that pings the build chat

jakemac53:
This should just ping us for non-pull request related build failures, can tweak later if its not working right or gets spammy.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,24 @@
 # Created with https://github.com/dart-lang/mono_repo
 language: dart
 
-# Hand-edited
+####################
+# Begin hand edits #
+####################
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -169,3 +169,6 @@ before_install:
  - export DISPLAY=:99.0
  - sh -e /etc/init.d/xvfb start
  - "t=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done"
+
+after_failure:
```
natebosch: Can you organize the file so that any hand edited stuff is contiguous?

Issue #981, opened 2018-02-08
Add a BuildFailure class, and catch in the runners


matanlurey:
We should not display the stack trace (or do it as `FINE`) for these type of errors.

/cc @natebosch 

* [ ] Add the class to `build`
* [ ] Add support in `build_runner`
* [ ] Add support in `bazel_codegen` (https://github.com/dart-lang/bazel/issues/212)

natebosch:
Some other options for completeness:

- Stop printing stack traces by default and only print them when some option is passed
- Check the type, if its a subclass of `Error` don't print the stack trace, and if it's a subclass of `Exception` print it.

matanlurey:
I personally like `BuildError`, because it could also do other nice things:

* Capture the current `primaryInput` (?)
* Be extended by clients, i.e. `TemplateBuildError`, `InjectionBuildError` for us

matanlurey:
@jakemac53 any thoughts here?

jakemac53:
+1 I am in favor, we should also automagically wrap any normal exceptions from builders in one of these

leonsenft:
I'm considering trying this before #2624, but it's unclear to me where such errors should be handled in `build_runner` and `bazel_codegen`.

`build_runner` (and friends) in particular has so much indirection I can't quite figure out where an individual builder is even invoked. Do you have any pointers on where to get started? Thanks!

jakemac53:
https://github.com/dart-lang/build/blob/master/build_runner_core/lib/src/generate/build_impl.dart#L486 is roughly where the builder actually gets ran - that is calling this api from `build` https://github.com/dart-lang/build/blob/master/build/lib/src/generate/run_builder.dart.

natebosch:
Builders are invoked here:

https://github.com/dart-lang/build/blob/04ba2196bca63a6999365d788b1cd66062f423fc/build_runner_core/lib/src/generate/build_impl.dart#L486

I would expect that special handling for this exception might involve the logging handler:
https://github.com/dart-lang/build/blob/04ba2196bca63a6999365d788b1cd66062f423fc/build/lib/src/builder/logging.dart#L28

leonsenft:
Great, thanks for the help! One last quick question before I get started. The proposal suggests an `Error` (based on the name), but this would be better suited as an `Exception` because it's intended to be caught, right?

natebosch:
Yes. This should be an exception. I would implement that interface but not include it in the name `BuildFailure` is my first choice.

[...]

PR #1100, opened 2018-03-06
Use a 'builder label' for log messages

### LINKED ISSUE #1069: Log lines from builders are really long
Just the description of the logger can be really long today, for instance `[SEVERE] Instance of 'DevCompilerBuilder' on e2e_example|web/sub/main.module: `.

This is partly because of the default `toString` for objects and partly because we include the full primary input id in the description.

We could leave it as is and always add newlines after the description, or we could try to shorten the descriptions in some way, but I think we should do something to clean it up.

natebosch:
Closes #1069

Replaces the builder `toString()` with a more relevant label, and
changes AssetIds to something more reconizable.

- Keep a `buildLabel` on `BuildAction`. Where possible this is a
  `builderKey`, but when that is missing or empty we fall back on the
  `toString` of the Builder.
- Normalize AssetIds to `package:` URIs for non-root packages, and file
  path for root package.
- Don't prefix newlines as messages are logged, these weren't working as
  implemented and would get swallowed by the StdIoLogger anyway.
- Hardcode known `build_runner` Logger names and don't print their name
  with messages.
- Always prefix messages from Builders (loggers with names we don't
  recognize) with a newline. This works well in the WARNING and SEVERE
  case but might not work as well for other messages - we can revist if
  necessary.
- Don't print stack traces unless verbose logging is enabled.

### DIFF
--- a/build_runner/lib/src/logging/build_for_input_logger.dart
+++ b/build_runner/lib/src/logging/build_for_input_logger.dart
@@ -60,15 +60,6 @@ class BuildForInputLogger implements Logger {
     if (logLevel >= Level.SEVERE) {
       _errorWasSeen = true;
     }
-    if (logLevel >= Level.WARNING) {
-      if (_delegate.isLoggable(logLevel)) {
-        var original = message;
-        if (message is Function) message = () => '\n${original()}';
-        if (message is String) {
-          message = '\n$message';
-        }
-      }
-    }
     _delegate.log(logLevel, message, error, stackTrace, zone);
   }
 
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -58,7 +58,8 @@ void stdIOLogListener(LogRecord record, {bool verbose}) {
   }
 }
 
-/// Filter out the Logger names known to come from `build_runner`.
+/// Filter out the Logger names known to come from `build_runner` and splits the
+/// header for levels >= WARNING.
 String _loggerName(LogRecord record, bool verbose) {
   var knownNames = const [
     'Entrypoint',
@@ -66,7 +67,8 @@ String _loggerName(LogRecord record, bool verbose) {
     'BuildDefinition',
     'Heartbeat'
   ];
+  var maybeSplit = record.level >= Level.WARNING ? '\n' : '';
   return verbose || !knownNames.contains(record.loggerName)
-      ? '${record.loggerName}:\n'
+      ? '${record.loggerName}:$maybeSplit'
       : '';
 }
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -27,7 +27,7 @@ void stdIOLogListener(LogRecord record, {bool verbose}) {
         headerMessage.split('\n').takeWhile((line) => line.isEmpty).length;
     headerMessage = headerMessage.substring(blankLineCount);
   }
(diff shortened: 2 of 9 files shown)


natebosch:
The hardcoded logger names make me a littler nervous - some of the existing newline magic is also somewhat troubling, but I think it's still worth it to get the behavior we have.

This might solve #981 since it hides stack traces by default, but we can still discuss a `BuildError` class.

build_runner/lib/src/logging/build_for_input_logger.dart
```diff
@@ -60,15 +60,6 @@ class BuildForInputLogger implements Logger {
     if (logLevel >= Level.SEVERE) {
       _errorWasSeen = true;
     }
-    if (logLevel >= Level.WARNING) {
```
jakemac53: I think we still want this - unless we stopped removing the asset that its running on from the description as well (and maybe even then).

build_runner/test/generate/build_integration_test.dart
```diff
@@ -60,10 +60,8 @@ main(List<String> args) async {
         var result = await runDart('a', 'tool/build.dart',
             args: ['build', '--delete-conflicting-outputs']);
         expect(result.exitCode, 0, reason: result.stderr as String);
-        expect(
-            result.stdout,
-            contains('BuildDefinition: Invalidating asset graph due to '
```
jakemac53: are these names no longer present or you just wanted to simplify the tests? I think they do provide some value personally (the names being there, I am ambivalent about the tests).

jakemac53:
Should have read the full description first :).

The main thing I don't think we want is always prefixing messages with a newline - I believe that will break the existing logic that overwrites messages under WARNING level?

build_runner/lib/src/logging/build_for_input_logger.dart
natebosch: Header splitting happens where we print now. This code path isn't realistically getting hit now, but if we fixed this we'd end up with a black line before the header, not splitting the header like we intended.

build_runner/test/generate/build_integration_test.dart
natebosch: They are no longer present. I felt like they mainly added value for us. The end user probably didn't care about the particulars ways we split up functionality in this package or the labels we gave them.

build_runner/lib/src/logging/build_for_input_logger.dart
jakemac53: ack

build_runner/test/generate/build_integration_test.dart
jakemac53: It gave you some indication of where the log was coming from but its probably true that for most users that didn't mean much.... I am ok with removing them.

build_runner/lib/src/logging/std_io_logging.dart
```diff
@@ -57,3 +57,16 @@ void stdIOLogListener(LogRecord record, {bool verbose}) {
...
+
+/// Filter out the Logger names known to come from `build_runner`.
+String _loggerName(LogRecord record, bool verbose) {
+  var knownNames = const [
+    'Entrypoint',
+    'Build',
+    'BuildDefinition',
+    'Heartbeat'
+  ];
+  return verbose || !knownNames.contains(record.loggerName)
```
jakemac53: looks like this is where we would only add the newline for >= WARNING?

natebosch: Yeah, I can add a level check here.

natebosch:
@jakemac53 - now only splitting on WARNING and above - any other concerns?

Issue #1105, opened 2018-03-06
contents of build.yaml not overridden by --define values


kevmoo:
For these versions

```yaml
  build_runner:
    dependency: "direct dev"
    description:
      name: build_runner
      url: "https://pub.dartlang.org"
    source: hosted
    version: "0.7.12"
  build_web_compilers:
    dependency: "direct dev"
    description:
      name: build_web_compilers
      url: "https://pub.dartlang.org"
    source: hosted
    version: "0.3.2"
```

jakemac53:
Looks like we do the right thing here, will investigate further into the problem (the specific issue here was providing multiple defines for the same builder and the 2nd one not seeming to be applied).

jakemac53:
Ok, it looks like the issue here was actually an extra wrapping of quotes in the args (using process.run but adding in explicit quotes when dart:io already does its own escaping/wrapping). 

@natebosch can you look into giving an error message if we don't recognize a builder key (assuming that is what happened here?)

jakemac53:
Ya it looks like we don't do any validation that builder keys exist - either in build.yaml or with --define

natebosch:
Can you give an example of an invocation that failed because of this?

PR #1106, opened 2018-03-06
add test that --define overrides build.yaml

jakemac53:
Added as a result of https://github.com/dart-lang/build/issues/1105, but it looks like we are already doing the right thing - cc @kevmoo.

### DIFF
--- a/build_runner/test/generate/build_integration_test.dart
+++ b/build_runner/test/generate/build_integration_test.dart
@@ -357,6 +357,90 @@ main(List<String> args) async {
         await expectBuildOutput(expectCopyInOutput: false);
       });
     });
+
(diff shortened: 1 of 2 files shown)


build_runner/test/generate/build_integration_test.dart
```diff
@@ -357,6 +357,90 @@ main(List<String> args) async {
         await expectBuildOutput(expectCopyInOutput: false);
       });
     });
+
+    group('--define overrides build.yaml', () {
+      String buildContent = '''
```
natebosch: [nit] maybe `final buildContent` or `const buildContent` so that it's clear this doesn't change?

jakemac53: done

Issue #1075, opened 2018-03-01
Can't use $default in other parts of build_config


matanlurey:
Source: https://github.com/dart-lang/build/issues/1073#issuecomment-369627741

That might be WAI, but we should fail nicer then.

```
[INFO] Entrypoint: Generating build script completed, took 267ms
NoSuchMethodError: The getter 'target' was called on null.
Receiver: null
Tried calling: target
#0      Object.noSuchMethod (dart:core-patch/dart:core/object_patch.dart:46)
#1      createBuildActions.<anonymous closure> (package:build_runner/src/package_graph/apply_builders.dart:114:22)
#2      stronglyConnectedComponents.strongConnect (package:graphs/src/strongly_connected_components.dart:44:24)
#3      stronglyConnectedComponents (package:graphs/src/strongly_connected_components.dart:66:55)
#4      createBuildActions (package:build_runner/src/package_graph/apply_builders.dart:112:16)
<asynchronous suspension>
#5      build (package:build_runner/src/generate/build_impl.dart:93:13)
```

matanlurey:
If this is a legit use-case I'd love to help.

natebosch:
yeah I think this is worth supporting.

Fixing this will involve something like updating [`normalizeTargetUsage`](https://github.com/dart-lang/build/blob/91d76a5e7e6b280827b5f4165d2932cf5619f5a3/build_config/lib/src/key_normalization.dart#L47) to understand `_defaultTargetNamePlaceholder`

matanlurey:
Ack! Unless you need it on a schedule I can take this one.

matanlurey:
As-per @natebosch on latest (closed PR #1113) we just need to release build_config.

PR #1108, opened 2018-03-06
Add support for normalizing $default:$default

matanlurey:
This is the first part of work needed for https://github.com/dart-lang/build/issues/1075.

After this, `apply_builders.dart` will just need to normalize the key.

(Also ran dartfmt, which looks like it changed again?)

### DIFF
--- a/build_config/lib/build_config.dart
+++ b/build_config/lib/build_config.dart
@@ -13,7 +13,4 @@ export 'src/build_config.dart'
         TargetBuilderConfig,
         TargetBuilderConfigDefaults;
 export 'src/key_normalization.dart'
-    show
-        normalizeBuilderKeyUsage,
-        normalizeTargetKeyUsage,
-        normalizeTargetKeyDefinition;
+    show normalizeBuilderKeyUsage, normalizeTargetKeyUsage;
--- a/build_config/lib/build_config.dart
+++ b/build_config/lib/build_config.dart
@@ -13,4 +13,7 @@ export 'src/build_config.dart'
         TargetBuilderConfig,
         TargetBuilderConfigDefaults;
 export 'src/key_normalization.dart'
-    show normalizeBuilderKeyUsage, normalizeTargetKeyUsage;
+    show
+        normalizeBuilderKeyUsage,
+        normalizeTargetKeyUsage,
+        normalizeTargetKeyDefinition;
--- a/build_config/lib/src/key_normalization.dart
+++ b/build_config/lib/src/key_normalization.dart
@@ -3,6 +3,8 @@
 // BSD-style license that can be found in the LICENSE file.
 
 const _defaultTargetNamePlaceholder = r'$default';
+const _defaultTargetNamePlaceholderKey =
+    '$_defaultTargetNamePlaceholder:$_defaultTargetNamePlaceholder';
 
 /// Returns the normalized [builderKey] definition when used from [packageName].
 ///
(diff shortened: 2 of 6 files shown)


build_config/lib/build_config.dart
```diff
@@ -13,4 +13,7 @@ export 'src/build_config.dart'
         TargetBuilderConfig,
         TargetBuilderConfigDefaults;
 export 'src/key_normalization.dart'
-    show normalizeBuilderKeyUsage, normalizeTargetKeyUsage;
+    show
+        normalizeBuilderKeyUsage,
+        normalizeTargetKeyUsage,
+        normalizeTargetKeyDefinition;
```
jakemac53: I think we really just want `normalizeTargetKeyUsage` to understand this as opposed to exposing `normalizeTargetKeyDefinition`? 

The things we actually want to add support for are all usages I believe.

matanlurey: Ah my bad, let me do that instead :)

matanlurey: PTAL

build_config/lib/src/key_normalization.dart
```diff
@@ -41,11 +41,17 @@ String normalizeTargetKeyDefinition(String targetKey, String packageName) =>
...
+///   - "$default" => "$packageName:$packageName"
+///   - "$default:$default": => "$packageName:$packageName"
 ///   - "some_package" => "some_package:some_package"
 ///   - ":some_target" => "$packageName:some_target"
 ///   - "some_package:some_target" => "some_package:some_target"
 String normalizeTargetKeyUsage(String targetKey, String packageName) =>
-    _normalizeUsage(targetKey, packageName, ':');
+    targetKey == _defaultTargetNamePlaceholder ||
+            targetKey ==
+                '$_defaultTargetNamePlaceholder:$_defaultTargetNamePlaceholder'
```
natebosch: might not be necessary, but do we want to accept `:$default`? We do accept `:my_package` I think

build_config/lib/src/key_normalization.dart
```diff
@@ -41,11 +41,17 @@ String normalizeTargetKeyDefinition(String targetKey, String packageName) =>
 ///
 /// Example normalizations:
 ///
+///   - "$default" => "$packageName:$packageName"
+///   - "$default:$default": => "$packageName:$packageName"
```
natebosch: [nit] One extra `:` snuck in before the `=>`

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -157,31 +157,31 @@ Iterable<BuildAction> _createBuildActionsWithinCycle(
...
-              ..addAll(builderConfigOverrides));
-        return new BuildAction(b(options), node.package.name,
-            builderOptions: options,
-            targetSources: node.target.sources,
-            generateFor: generateFor,
-            isOptional: builderApplication.isOptional,
-            hideOutput: builderApplication.hideOutput);
-      }));
+      .where((targetNode) =>
+      _shouldApply(builderApplication, targetNode, applyWith))
```
natebosch: did dartfmt go haywire here? Some of this looks off to me.

build_config/lib/build_config.dart
natebosch: I can't find the usage of `normalize*` from apply_builders... did something change?

build_config/lib/src/key_normalization.dart
matanlurey: PTAL again, thanks @natebosch

build_config/lib/src/key_normalization.dart
jakemac53: I am not sure if we need to actually support this variant, but I guess I don't see a reason why not either?

PR #1110, opened 2018-03-06
Post process builder

jakemac53:
Next iteration - this:

- renames `builderFactories` => `builderFactory` (thought I did this before but i guess not).
- Adds `PostProcessBuilder` class to build_runner.
- ~~Adds new `GeneratedForPhaseAssetNode` which represents all existing generated nodes.~~
- Update all related code to the point where tests are all passing etc.

~~There is a bit of weirdness in that the post process nodes are actually just `GeneratedAssetNode`s right now... I don't think we need extra information for them. I could make them be a different type just to make the type checks in the code more clear though...~~


build_runner/lib/src/asset_graph/node.dart
```diff
@@ -106,13 +108,16 @@ enum GeneratedNodeState {
 }
 
 /// A generated node in the asset graph.
+///
+/// Most generated nodes should actually be [GeneratedForPhaseAssetNode]s which
```
natebosch: [nit] drop `actually`

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -169,6 +166,41 @@ class GeneratedAssetNode extends AssetNode {
       'GeneratedAssetNode: $id generated from input $primaryInput.';
 }
 
+/// A typical generated node which was generated in a specific phase, and used a
```
natebosch: [nit] and may have used a set of...

build_runner/lib/src/asset_graph/node.dart
```diff
@@ -169,6 +166,41 @@ class GeneratedAssetNode extends AssetNode {
...
+/// set of [globs] during its generation.
+class GeneratedForPhaseAssetNode extends GeneratedAssetNode {
```
natebosch: My Java days have trained me to be scared of extending a non-abstract class. Should we consider a separate sub-type for post-build outputs?

natebosch: What are your thoughts on `OutputNode` and `PhaseOutputNode`?

build_runner/lib/src/builder/post_process_builder.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+
+/// A [PostProcessBuilder] runs in a special phase at the end of the build.
```
natebosch: [nit] noun-phrase.

"A builder which runes in a special phase at the end of the build."

build_runner/lib/src/builder/post_process_builder.dart
```diff
@@ -0,0 +1,32 @@
...
+///
+/// They are different from a normal [Builder] in several ways:
+///
+/// - They don't have to declare output extensions, and can output any file as
+///   long as it doesn't conflict with an existing one.
+/// - They can only read their primary input id.
```
natebosch: "They can only read their primary input". It isn't the ID that they are reading.

build_runner/lib/src/asset_graph/node.dart

build_runner/lib/src/asset_graph/node.dart

build_runner/lib/src/builder/post_process_builder.dart
jakemac53: done

build_runner/lib/src/builder/post_process_builder.dart
jakemac53: done

build_runner/lib/src/asset_graph/node.dart
jakemac53: So actually I may be able to revert this part - I am testing out just making `BuildAction` not have a `builder` field, and instead there would be a `BuilderBuildAction` and `PostBuildAction`.

That would allow me to actually give a `phaseNumber` to the post build actions and get some other stuff for free as well.

It turns out we don't reference the `builder` field from `BuildAction` all that much so adding the type checks where necessary shouldn't be difficult, and it would allow me to provide an `Iterable<PostProcessBuilder> builders` field for the `PostBuildAction`.

jakemac53: Depending on how the `BuildAction` stuff pans out ya I like these names.

jakemac53:
Ok, definitely reverting the GeneratedAssetNode changes, just pushed that commit

build_runner/lib/src/builder/post_process_builder.dart
```diff
@@ -0,0 +1,32 @@
...
+///   that were built as a part of the normal build.
+/// - They all run in a single phase, and thus can not see the outputs of any
+///   other [PostProcessBuilder]s.
+/// - Because they run in a separate phase, after other builders, none of thier
+///   outputs can be consumed by [Builder]s.
+///
+/// Because of these restrictions, these builders should never be used to output
+/// Dart files, or any other file which would should be processed by normal
+/// [Builder]s.
+abstract class PostProcessBuilder {
```
natebosch: do we need to `export` this from some file under `lib/`?

jakemac53: Yes, we will :).

I will do that in a followup as its not super important right now and I am tired of waiting on travis lol.

PR #1113, opened 2018-03-07
Use 'normalizeTargetKeyUsage' in apply_builders.

matanlurey:
Next part of https://github.com/dart-lang/build/issues/1075.

I can't add a test (or can I?) until we release the latest `build_config`, but this should be backwards compatible (you just can't use the new feature unless build_config is released).


matanlurey:
Oh I see, I didn't realize that! I guess my prior Q around releasing build_config stands, but I'll close this PR and be happy :)

PR #1118, opened 2018-03-07
Add PostProcessAction

jakemac53:
- Added `applyPostProccess` method which takes `PostProcessBuilderFactory`s instead of `BuilderFactory`s.
- Split `BuildAction` into `BuilderBuildAction` and `PostProcessBuildAction`. The base class no longer has a `builder` field.
- `BuilderApplication` now has a `builderActionFactories` getter instead of a `builderFactories` getter, and it controls which type of build action will get created.
- The default constructor for `BuilderApplication` has been replaced with `BuilderApplication.forBuilder` and `BuilderApplication.forPostProcessBuilder`.


natebosch:
synced offline - `package`, `inputs`, `generate_for` and a single `builderOptions` all don't make sense for a post process builder action. I'm concerned that modeling both types as a `BuildAction` is going to get hairy, but we can keep trying it and see if there is a version that will work.

jakemac53:
@natebosch updated PTAL

build_runner/CHANGELOG.md
```diff
@@ -1,4 +1,21 @@
...
+## 0.8.0-prerelease
+
+### New Features
+
+- Added `applyPostProccess` method which takes `PostProcessBuilderFactory`s
+  instead of `BuilderFactory`s.
+
+### Breaking Changes
+
+- Split `BuildAction` into `BuilderBuildAction` and `PostProcessBuildAction`.
```
natebosch: Is this class public? If it isn't this isn't a breaking change. If it is we should probably figure out why

build_runner/lib/src/generate/phase.dart
```diff
@@ -53,7 +70,7 @@ class BuildAction {
   ///
   /// [hideOutput] specifies that the generated asses should be placed in the
   /// build cache rather than the source tree.
-  factory BuildAction(
+  factory BuilderBuildAction(
```
natebosch: Maybe `BuilderAction` to match `PostProcessBuildAction` vs `PostProcessBuilderAction`?

Or maybe `InBuildAction` and then `PostBuildPhase` and `PostBuildAction`?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -81,9 +87,27 @@ BuilderApplication apply(String builderKey,
       appliesBuilders: appliesBuilders,
     );
 
-/// A description of which packages need a given [Builder] applied.
+/// Same as [apply] except it takes [PostProcessBuilderFactory]s and does not
+/// provide options for `isOptional` or `hideOutput` because they aren't
+/// configurable for these types of builders. They are never optional and always
```
natebosch: Need a 1 sentence summary then a blank line

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -81,9 +87,27 @@ BuilderApplication apply(String builderKey,
...
+/// hidden.
+BuilderApplication applyPostProcess(String builderKey,
+        List<PostProcessBuilderFactory> builderFactories, PackageFilter filter,
```
natebosch: Should this be a single factory?

natebosch: `PostProcessBuilder` in `build_config` doesn't have an `auto_apply` - does it need `filter` here? Or can we start with it automatically being `toNoneByDefault()` and add an optional arg later if we need it?

It also doesn't have `applies_builders`, does it need it here?

https://github.com/dart-lang/build/blob/a98f4b12e0344ee3b8e8fb079d073e4689221036/build_config/lib/src/parse.dart#L43

PR #1119, opened 2018-03-07
Add support for IPv4 loopback

### LINKED ISSUE #1103: Use pkg:http_multi_server for serve – listen on v4 and v6
https://github.com/dart-lang/build/blob/20f533b243d6f8b0c0b0f219ee531a9b504082da/build_runner/lib/src/entrypoint/options.dart#L355

Support both IPv4 & v6 endpoints – helpful for systems that have one but not the other

natebosch:
Closes #1103

- Add dependnecy on `http_multi_server` and use it when hostname is
  'localhost'.
- Listen on any connection when hostname is 'any'.

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:args/args.dart';
 import 'package:args/command_runner.dart';
 import 'package:build_config/build_config.dart';
+import 'package:http_multi_server/http_multi_server.dart';
 import 'package:io/io.dart';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -376,6 +375,26 @@ class _ServeCommand extends _WatchCommand {
...
+Future<HttpServer> _startServer(
+    _ServeOptions options, _ServeTarget target, ServeHandler handler) async {
+  var server = await _bindServer(options, target);
+  serveRequests(
+      server, handler.handlerFor(target.dir, logRequests: options.logRequests));
+  return server;
+}
+
+Future<HttpServer> _bindServer(_ServeOptions options, _ServeTarget target) {
+  switch (options.hostName) {
```
kevmoo: @nex3 – we now do this 3-ish places, right?

pub, test, build_runner – could this be something we create as a util in `http_multi_server`?

nex3: I'm not entirely sure what you mean by "do this"...

PR #1122, opened 2018-03-07
Add validation to builder keys

### LINKED ISSUE #1105: contents of build.yaml not overridden by --define values

natebosch:
Closes #1105

Add a warning if invalid builder keys attempt configuration with either
`--define` arguments or in the root package `build.yaml`.

This only checks the root package since there are valid use cases for
configuration builders which don't end up a dependents build.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -30,7 +30,6 @@ import '../package_graph/package_graph.dart';
 import '../package_graph/target_graph.dart';
 import '../performance_tracking/performance_tracking_resolvers.dart';
 import '../util/constants.dart';
-import '../validation/config_validation.dart';
 import 'build_definition.dart';
 import 'build_result.dart';
 import 'create_merged_dir.dart';
@@ -90,9 +89,6 @@ Future<BuildResult> build(
 
   final buildActions =
       await createBuildActions(targetGraph, builders, builderConfigOverrides);
-
-  validateBuilderConfig(
-      builders, targetGraph.rootPackageConfig, builderConfigOverrides, _logger);
 
   var result = await singleBuild(environment, options, buildActions);
 
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -30,6 +30,7 @@ import '../package_graph/package_graph.dart';
 import '../package_graph/target_graph.dart';
 import '../performance_tracking/performance_tracking_resolvers.dart';
 import '../util/constants.dart';
+import '../validation/config_validation.dart';
 import 'build_definition.dart';
 import 'build_result.dart';
 import 'create_merged_dir.dart';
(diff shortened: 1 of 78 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -90,6 +91,9 @@ Future<BuildResult> build(
   final buildActions =
       await createBuildActions(targetGraph, builders, builderConfigOverrides);
 
+  validateBuilderConfig(
```
jakemac53: Should we just do this as a part of `createBuildActions` so we don't need to explicitly do it (less likely to forget if we do `createBuildActions` somewhere else later on also)

build_runner/lib/src/validation/config_validation.dart
```diff
@@ -0,0 +1,31 @@
...
+/// Checks that all configuration is for valid builder keys.
+void validateBuilderConfig(
+    Iterable<BuilderApplication> builders,
+    BuildConfig rootPackageConfig,
+    Map<String, Map<String, dynamic>> builderConfigOverrides,
+    Logger logger) {
+  final builderKeys = builders.map((b) => b.builderKey).toSet();
+  for (final key in builderConfigOverrides.keys) {
+    if (!builderKeys.contains(key)) {
+      logger.warning('Overriding configuration for  `$key` but this is not a '
```
jakemac53: nit: extra space before `$key`

e2e_example/test/build_integration_test.dart
```diff
@@ -31,4 +31,21 @@ void main() {
       expect(result.stdout, contains('Skipping tests due to build failure'));
     });
   });
+
+  group('config validation', () {
```
jakemac53: nit: It might be better to move this to `build_runner/test/generate/build_integration_test.dart` because you can have a lot fewer dependencies and the test will run faster (and also not run ddc or anything else).

e2e_example/test/build_integration_test.dart
```diff
@@ -39,7 +39,7 @@ void main() {
 
     test('warns when using an invalid override', () async {
       var result =
-          await runAutoBuild(trailingArgs: ['--define', 'bad|key=foo=bar']);
+          await runAutoBuild(trailingArgs: ['--define=badkey=foo=bar']);
```
jakemac53: Pipe should work

build_runner/lib/src/generate/build_impl.dart
natebosch: Done

e2e_example/test/build_integration_test.dart
natebosch: Done

PR #1136, opened 2018-03-09
Log a warning when pbr serve is used w/o build_web_compilers.

### LINKED ISSUE #1124: `serve` command should warn if there is no `build_web_compilers` dependency

matanlurey:
Closes #1124.

A strictly "better" version of this would see if there is at least a single Dart file being potentially served (i.e. this isn't an entirely static site or plain JavaScript), but I imagine that's quite rare - and it's just a warning.

... sorry for the wacky test, I just couldn't find the matcher that works.

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -342,6 +342,7 @@ class _ServeCommand extends _WatchCommand {
   @override
   Future<int> run() async {
     var options = _readOptions();
+    var logger = new Logger('Serve');
     var handler = await watch(builderApplications,
         deleteFilesByDefault: options.deleteFilesByDefault,
         enableLowResourcesMode: options.enableLowResourcesMode,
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -476,11 +477,27 @@ class _TestCommand extends BuildRunnerCommand {
...
+    Missing dev dependency on package:build_web_compilers, which is required to serve Dart compiled to JavaScript.
+
+    Please update your dev_dependencies section of your pubspec.yaml:
+
+    dev_dependencies:
+      build_runner: any
+      build_test: any
+      build_web_compilers: any''');
+  } else {
+    log.warning('YAY!');
```
jakemac53: maybe get rid of this one ;)

build_runner/test/generate/build_integration_test.dart
```diff
@@ -507,5 +508,33 @@ main() async {
...
+          'build_runner',
+        ]),
+        d.dir('web', [
+          d.file('a.dart', 'void main() {}'),
+        ]),
+      ]).create();
+
+      await pubGet('a');
+      var result = await startPub('a', 'run', args: ['build_runner', 'serve']);
+      var error = 'Missing dev dependency on package:build_web_compilers';
```
jakemac53: I think you can do an `addTearDown(() => result.kill())` which would make it get killed even if the test times out or w/e.

matanlurey: Nice idea, done.

build_runner/lib/src/entrypoint/options.dart
matanlurey: Done.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -524,7 +539,7 @@ Map<String, Map<String, dynamic>> _parseBuilderConfigOverrides(
 
 class BuildTestDependencyError extends StateError {
   BuildTestDependencyError() : super('''
-Missing dev dependecy on package:build_test, which is required to run tests.
+Missing dev dependency on package:build_test, which is required to run tests.
```
jakemac53: Looks like you need to update the test for this one, but then lgtm!

PR #1137, opened 2018-03-09
update serialization logic to have separate field enums for node types

jakemac53:
No functional change - but makes it easier to add new node types.


jakemac53:
> Some of the numbers will change so we need to increment _version

I don't think they will?

natebosch:
> I don't think they will?

You're right. Ship it!

Issue #1134, opened 2018-03-08
Watcher crashed for emacs temp file


alorenzen:
Edited a file in emacs, which created a temporary file, while running `build_runner serve`, and got the crash listed below:
```
FormatException: Invalid character (at character 37)
examples.registration_form/address/.#address_component.html
                                    ^

#0      _Uri._fail (dart:core/uri.dart:1597)
#1      _Uri._normalize (dart:core/uri.dart:2292)
#2      _Uri._normalizeOrSubstring (dart:core/uri.dart:2248)
#3      _Uri._makePath (dart:core/uri.dart:2096)
#4      new _Uri (dart:core/uri.dart:1497)
#5      AssetId.uri (package:build/src/asset/id.dart:102:13)
#6      PackageGraphWatcher._watch.<anonymous closure>.<anonymous closure> (package:build_runner/src/watcher/graph_watcher.dart:84:54)
#7      _RootZone.runUnaryGuarded (dart:async/zone.dart:1316)
#8      _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:330)
#9      _BufferingStreamSubscription._add (dart:async/stream_impl.dart:257)
#10     _ForwardingStreamSubscription._add (dart:async/stream_pipe.dart:132)
#11     _MapStream._handleData (dart:async/stream_pipe.dart:232)
#12     _ForwardingStreamSubscription._handleData (dart:async/stream_pipe.dart:164)
#13     _RootZone.runUnaryGuarded (dart:async/zone.dart:1316)
#14     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:330)
#15     _BufferingStreamSubscription._add (dart:async/stream_impl.dart:257)
#16     _SyncBroadcastStreamController._sendData (dart:async/broadcast_stream_controller.dart:379)
#17     _BroadcastStreamController.add (dart:async/broadcast_stream_controller.dart:254)
#18     _RootZone.runUnaryGuarded (dart:async/zone.dart:1316)
#19     _BufferingStreamSubscription._sendData (dart:async/stream_impl.dart:330)
#20     _DelayedData.perform (dart:async/stream_impl.dart:578)
#21     _StreamImplEvents.handleNext (dart:async/stream_impl.dart:694)
#22     _PendingEvents.schedule.<anonymous closure> (dart:async/stream_impl.dart:654)
#23     _microtaskLoop (dart:async/schedule_microtask.dart:41)
[...]

PR #1139, opened 2018-03-09
Print terse stack traces for uncaught errors

### LINKED ISSUE #1135: Use Trace.terse (pkg/stack_trace) for crashes
... we used to do this when `Chain.capture` was used, but now it is lost.

For example:
https://github.com/dart-lang/build/issues/1134

... has a bunch of junk in it unrelated to user code :)

natebosch:
Closes #1135

Exceptions from Builders should be caught by a `runZoned` call. The only
exceptions we expect to escape the Isolate are bugs within build_runner.

- Add a message saying that an issue should be file.
- Parse the trace and print the Trace.terse version

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -9,6 +9,7 @@ import 'dart:isolate';
 import 'package:args/command_runner.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
+import 'package:stack_trace/stack_trace.dart';
 
 import 'package:build_runner/src/build_script_generate/build_script_generate.dart';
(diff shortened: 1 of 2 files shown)


natebosch:
The error from #1134 now looks like:

```


You have hit a bug in build_runner
Please file an issue with reproduction steps at https://github.com/dart-lang/build/issues


FormatException: Invalid character (at character 13)
e2e_example/#strange_file_name
            ^

dart:core                                                  new _Uri
package:build/src/asset/id.dart 102:13                     AssetId.uri
package:build_runner/src/watcher/graph_watcher.dart 84:54  PackageGraphWatcher._watch.<fn>.<fn>
```

PR #1142, opened 2018-03-10
Get the e2e working for post process builders

jakemac53:
So, I thought I was being smart and merged master into this branch so I wouldn't have to do the merge twice (once on the dart2js-deferred-loading branch and once on this branch).

However, that means this pull request looks much larger than it really is and there's a bunch of upstream changes mixed in... my bad :(. At this point though I think it would be quite difficult to resolve it in a sane manner.

At a high level, this pr does the following:

- Adds a `BuildAction` class which `InBuildPhase` and `PostBuildAction` implement
- Adds a `PostProcessAnchorNode` synthetic node type which is used to track previous outputs of these post build actions for an input and check when builds need to be redone etc.
- Adds a `PostProcessBuildStep` class and `runPostProcessBuilder` method.
- Implements the actual building in `BuildImpl`.
- Adds the `applyPostProcess` applications in the generated build script.

This does add some e2e tests, but no unit tests. I will follow up with some unit tests separately.


jakemac53:
Closing in favor of clean merge to master https://github.com/dart-lang/build/pull/1143

PR #1143, opened 2018-03-10
Add PostProcessBuilders

jakemac53:
This PR would be in favor of https://github.com/dart-lang/build/pull/1142, which got pretty ugly due to merges from master.

This merges the entire feature into the master branch, but it does work e2e now so I think at this time that would be fine. The main thing missing is unit tests, but there is at least an e2e test. I will follow up with unit tests.


build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,31 @@
...
+    that were built as a part of the normal build.
+  - They can not be optional themselves, and can only output to cache.
+  - Because they all run in a single phase, after other builders, none of their
+    outputs can be used as inputs to any actions.
+- Added `applyPostProccess` method which takes `PostProcessBuilderFactory`s
+  instead of `BuilderFactory`s.
+
+### Breaking Changes
+
+- Split `BuildAction` into `BuilderBuildAction` and `PostProcessBuildAction`.
```
natebosch: Names don't match latest changes. But I think we could also delete since `BuildAction` was not public?

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -52,18 +52,28 @@ Future<Iterable<Expression>> _findBuilderApplications(String configKey) async {
         package.name, package.dependencies.map((n) => n.name), package.path);
   }
 
-  final builderDefinitions =
-      (await Future.wait(orderedPackages.map(_packageBuildConfig)))
-          .expand((c) => c.builderDefinitions.values)
-          .where((definition) {
+  bool _isValidDefinition(dynamic definition) {
```
natebosch: as a followup we should see whether there is some Interface that would make sense for this. Maybe something like `PackageProvidedCode` that has a `package` and `import` field. Then make `BuilderDefinition implement PackageProvidedCode`.

build_runner/lib/src/builder/post_process_build_step.dart
```diff
@@ -0,0 +1,90 @@
...
+        (String c) => _writer.writeAsString(id, c, encoding: encoding));
+    _writeResults.add(Result.capture(done));
+    return done;
+  }
+
+  /// Waits for work to finish and cleans up resources.
+  ///
+  /// This method should be called after a build has completed. After the
+  /// returned [Future] completes then all outputs have been written and the
+  /// [Resolver] for this build step - if any - has been released.
```
natebosch: There will never be a `Resolver`.

build_runner/CHANGELOG.md
jakemac53: removed

build_runner/lib/src/build_script_generate/build_script_generate.dart
jakemac53: acknowledged

build_runner/lib/src/builder/post_process_build_step.dart
jakemac53: done

Issue #814, opened 2018-01-05
Dart2js deferred loading support


jakemac53:
Eventually we need to figure out how to support dart2js with deferred loading. If at all possible this should not involve anything hardcoded the `build` or `build_runner` packages, and could be completely isolated to the `build_compilers` package.

Some ideas and previous solutions include:

- Output archive files and add logic for serving from archive files with specific uri patterns. We would also need specific logic for expanding these archives when creating merged output directories.
- Require users to hardcode the number of deferred outputs they expect. We would need to give good errors when this number needs to be updated. We could allow users to over-declare this number though.
  - This would still be blocked on user config being able to override the output extensions of builders, ideally on a per-target basis.
- Hardcode a certain number of potential output files, and just omit most of them. This doesn't scale though, and also involves some amount of cpu/memory overhead.
- Add some way of a builder declaring output patterns as opposed to specific outputs. This would mean the asset graph would have to become a lot more flexible (today it is completely fixed before a build starts). This also makes checking for overlapping outputs more expensive and complicated.

natebosch:
I like archive files.

I think we've used both archives and hardcoding number of deferred parts in the config with bazel right? IIRC archives are our current preferred approach in bazel and so if we can support that here we'll drive alignment.

Archives are also a more generally useful solution if there are other builders that want this flexibility in outputs (which can't be read during the rest of the build)

jakemac53:
The main issue with archive files is figuring out how and what to serve from them.

We would need to add some sort of special case logic, maybe a special type of asset node, which contains information about what files can actually be served by the tar file. We could potentially create synthetic asset nodes as well which just point at a tar file for their contents, and can't be used as inputs to other actions.

natebosch:
Would it work to force a certain layout for the archive and serve any file in the archive which isn't available as a normal asset?

Request comes in for a path, translate to assetId and check if it's in the graph. If it isn't translate to an archive path and walk the archives we have 1 by 1 and serve the file from the first archive that has it. Archives can be walked in buildaction order.

jakemac53:
We really don't want to be searching all potential archives _on each request_ like that though. We need to know ahead of time what archive to serve from.

For instance you would potentially have one of these archives _per test_, which could be very slow.

natebosch:
We can read the list of files ahead of time and cache that. We'd only read the list of files once per build, not each request

natebosch:
[...]

PR #1147, opened 2018-03-12
Add dart2js deferred loading support

jakemac53:
With this change dart2js now puts all the extra outputs (dump_info, deferred parts) in an archive file, which is extracted using one of the new `PostProcessBuilder` (`Dart2JsArchiveExtractor`) which is applied by the build_web_compilers|entrypoint builder.

To write unit tests we will need to create a series of helper utils for testing post process builders etc - that should happen but the provided e2e test should be sufficient for now.

Closes https://github.com/dart-lang/build/issues/814


jakemac53:
> Do we want a test confirming behavior with --output?

Done

PR #1150, opened 2018-03-13
release build_runner

jakemac53:


natebosch:
Is there anything else we should be looking at rolling with this breaking release? Maybe doing the default of fail-on-severe?

jakemac53:
> Is there anything else we should be looking at rolling with this breaking release? Maybe doing the default of fail-on-severe?

I am fine with flipping that if you want - but won't have time to do it before I go on vacation so you would need to do the release while I am out (unless you have time today to do it).

natebosch:
Ah, I realized that the serve case is going to be a bit more effort to fix than I had anticipated. Let's go ahead and publish as is and we'll use `0.9.0` for other changes.

PR #1162, opened 2018-03-13
Drop -dev

natebosch:


natebosch:
One the one hand you could make the argument that I should be paying closer attention.

On the other hand I *really* wish I could push to master for cases like this :(

PR #1179, opened 2018-03-16
Add many more lints, update dependencies in _bazel_codegen

kevmoo:


analysis_options.yaml
```diff
@@ -10,45 +10,65 @@ analyzer:
...
+    - comment_references
+    #- constant_identifier_names
+    - control_flow_in_finally
     - directives_ordering
     - empty_catches
     - empty_constructor_bodies
+    - empty_statements
+    - hash_and_equals
+    #- implementation_imports
+    - invariant_booleans
```
natebosch: I thought this one slowed things way down

analysis_options.yaml
```diff
@@ -10,45 +10,65 @@ analyzer:
...
     - library_names
     - library_prefixes
+    - list_remove_unrelated_type
+    - no_adjacent_strings_in_list
+    - no_duplicate_case_values
     - non_constant_identifier_names
+    - omit_local_variable_types
+    #- only_throw_errors
+    - overridden_fields
+    - package_api_docs
```
natebosch: Drop this. It's encourages worthless comments.

analysis_options.yaml
kevmoo: I'm not sure...

@pq @bwilkerson – ideas on `invariant_booleans` lint – RE speed?

analysis_options.yaml
kevmoo: Fixed

analysis_options.yaml
bwilkerson: I don't know what the performance characteristics are, at this point. I seem to recall that @pq has a script that will measure the performance, so we might be able to get that information.

pq: Yes.  We now have a "benchmark" bot.  This doesn't give us absolute metrics but does get us relative timings.  Here's one run for example:

https://travis-ci.org/dart-lang/linter/jobs/354257315

which confirms that `invariant_booleans` is the slowest of the bunch (by a long shot).

image

Whether it's worth it or not is another question altogether!

PR #1190, opened 2018-03-19
Exit with an error if a nonexistent command is attempted

kevmoo:


build_runner/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 - Improved the layout of `/$perf`, especially after browser window resize.
 
+- `pub run build_runner` exits with a error when invoked with an unsupported
```
natebosch: Remove the extra blank line above

build_runner/bin/build_runner.dart
```diff
@@ -23,6 +25,16 @@ Future<Null> main(List<String> args) async {
     ..addCommand(new _GenerateBuildScript());
   var parsedArgs = commandRunner.parse(args);
   var commandName = parsedArgs.command?.name;
+
+  if (parsedArgs.rest.isNotEmpty) {
```
natebosch: we detect this today with `commandName == null` - should we continue to do that and remove that condition below?

kevmoo: `null` is no command. I think we still want to print help and exit code 0 there.

build_runner/CHANGELOG.md
kevmoo: ack

PR #1195, opened 2018-03-22
Update --output if source with no outputs changes

### LINKED ISSUE #1095: Document and/or fix how non-source files work with --output
From Gitter for context:

```
Matan Lurey @matanlurey 12:07
(Work as intended)
We should update https://github.com/dart-lang/build/blob/master/docs/getting_started.md#creating-an-output-directory once we find out 

Jacob MacDonald @jakemac53 12:18
Oh.... ya interesting if the node is considered "not interesting" we don't trigger a rebuild but its true that with --output we still need to update that directory.

Nate Bosch @natebosch 12:21
we might need to cause source file digests to get read during the --output step. That's effectively how we make things work with serve

Jacob MacDonald @jakemac53 12:34
ah true if we just computed the digest at that point it would end up doing what we want
```

natebosch:
Fixes #1095

We use the presence of a digest as an indication that a source file has
an impact on the build.

- Ensure the digest is tracked for any asset which is copied to the
  merged output directory.
- Add a regession test which fails due to a timeout before the change
  and passes after.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -4,7 +4,9 @@
 - `pub run build_runner` exits with a error when invoked with an unsupported
   command.
 - Bug Fix: Update outputs in merged directory for sources which are not used
-  during the build.
+  during the build. For example if `web/index.html` is not read to produce any
+  generated outputs changes to this file will now get picked up during `pub run
+  build_runner watch --output build`.
 
 ## 0.8.0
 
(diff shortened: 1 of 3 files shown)


build_runner/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - Improved the layout of `/$perf`, especially after browser window resize.
 - `pub run build_runner` exits with a error when invoked with an unsupported
   command.
+- Bug Fix: Update outputs in merged directory for sources which are not used
```
matanlurey: It might be nice to give an example, i.e.:

"during the build - a common example is changing web/index.html"

natebosch: Done

PR #1198, opened 2018-03-23
Ignore deprecated barback imports

kevmoo:
This package is now deprecated


kevmoo:
@natebosch See https://travis-ci.org/dart-lang/build/jobs/357563395#L540

PR #1204, opened 2018-03-26
A couple more lints - new as of -dev.41

kevmoo:


build_resolvers/lib/src/resolver.dart
```diff
@@ -240,8 +240,8 @@ class AssetBasedSource extends Source {
   bool exists() => _contents != null;
 
   @override
-  bool operator ==(Object other) =>
-      other is AssetBasedSource && assetId == other.assetId;
+  bool operator ==(Object object) =>
```
natebosch: why the flip flop with the other changes?

kevmoo: The thing we're inheriting from breaks the rule – so to be consistent w/ it, we have to change.

...honestly, I should have put ignore here.

How the heck did you come back to this now?

natebosch: oh, the lint tells us to use the same name as the super type?

I was prepping to publish and looking at the diff since the last publish and wasn't sure why this was different from the other cases

kevmoo:
ack

Issue #1018, opened 2018-02-15
Build failures cause cascading errors that are confusing


matanlurey:
For example, this error reported by a user on Gitter:

> @matanlurey Ok still trying to make sure I have everything setup correctly.
>
> I'm even getting errors in some the async libs:

```
[error] Invalid override. The type of 'TypeSafeStream.lastWhere' ('((T) → bool, {defaultValue: () → Object}) → Future<dynamic>') isn't a subtype of 'Stream<T>.lastWhere' ('((T) → bool, {defaultValue: () → dynamic, orElse: () → T}) → Future<T>'). (package:async/src/typed/stream.dart, line 59, col 3)
[error] Invalid override. The type of 'TypeSafeStream.singleWhere' ('((T) → bool) → Future<T>') isn't a subtype of 'Stream<T>.singleWhere' ('((T) → bool, {orElse: () → T}) → Future<T>'). (package:async/src/typed/stream.dart, line 62, col 3)

Please fix all errors before compiling (warnings are okay).
[WARNING] Instance of 'WebEntrypointBuilder' on one_veenee|web/main.dart: Unable to read async|lib/src/typed/stream.ddc.js, check your console for compilation errors.
[SEVERE] Instance of 'DevCompilerBuilder' on async|lib/src/typed_stream_transformer.module: ```
```

In this case they did nothing wrong, the analyzer step failed because a package is out of date.

It would be nice to make these analysis errors (i.e. compilation errors) stand out better, but I've filed [TODO] about that. In the mean time ... maybe we should flip `--fail-on-severe` to `true` by default, because clearly its not possible to move forward in these cases.

matanlurey:
Another example? https://github.com/dart-lang/build/issues/1020

natebosch:
Yeah I think one thing we need to do is mark an output as failed and do something smarter when a future step tries to read it. The behavior today of it just looking missing is weird.

natebosch:
We're now throwing when a build step attempts to read a file that failed, though they can still check `canRead` and it will get treated as if the file doesn't exist.

Taking a look at how to handle serving successful assets when the overall build failed and hitting some trouble - we currently rely on serving `.ddc.errors` files to communicate build failures for DDC, but if we refuse to serve assets produced by failing build steps we'll hide these. We might need a new scheme for communicating failures to the browser.

jakemac53:
The `.ddc.errors` files are not a general solution anyways so that would probably be best if we can come up with something general.

[...]

PR #1208, opened 2018-03-28
Catch exceptions and mark builds as failures

### LINKED ISSUE #1207: Catch exceptions and treat them as build step failures rather than killing the entire build
Today we treat "log.severe" very differently from throwing an exception, even with `--fail-on-severe`.

With a severe log we record a failing action and let the build continue. With a thrown exception the build dies and we fail to write the asset graph.

I think we shouldn't rely on builder authors to only `log.severe` to indicate a failed build - we should allow them to throw and treat them the same.

natebosch:
Fixes #1207

Catch exceptions and use them to mark individual actions as failures
rather than allowing the exception to bubble up and stop the entire
build.

Move the handling for `--fail-on-severe` to whether or not we mark an
action as failing rather than whether or not we treat failed actions as
a build failure.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -7,6 +7,9 @@
   during the build. For example if `web/index.html` is not read to produce any
   generated outputs changes to this file will now get picked up during `pub run
   build_runner watch --output build`.
+- Don't allow a thrown exception from a Builder to take down the entire build
+  process - instead record it as a failed action. The overall build will still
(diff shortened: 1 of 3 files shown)


grouma:
At the gym. Will review shortly.

natebosch:
Ideally there would be a test, but I'd end up changing it pretty soon anyway. I expect this behavior to get covered when I fix https://github.com/dart-lang/build/issues/1018 and add tests for that.

In the short term I don't think it's worth adding one.

PR #1210, opened 2018-03-29
Drop .js.deps file

natebosch:
Given that this is referring to files that won't exist anywhere this is
run (other than Jake's desktop) I don't think it's a useful file.


natebosch:
I don't think we need to update it, since AFAICT the update process is pretty manual and we won't end up with this file sitting around very often.

Depending on #1211 we might end up with it in our gitignore.

PR #1219, opened 2018-03-29
Display the full URL when using `build_runner serve`

### LINKED ISSUE #1217: [Feature request] `build_runner serve` should display a URL instead of just the port
Using `build_runner` v0.8.0 with SDK v2.0.0-dev.43 on Ubuntu 17.10.
This is NOT an issue: just a little feature that I want.

When using `pub serve`, the terminal displays : `Serving appname web on http://localhost:8080`

When using `pub run build_runner serve`, the terminal displays: `Serving 'web' on port 8080`

Having the full URL displayed is convenient: it let me "CTRL+click" on the URL and have my browser opened with that URL (I'm using Gnome Terminal). Having only the port displayed forces me to manually open my browser and to enter the full URL before testing my application.

cedx:
Fixes #1217.

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -374,7 +374,7 @@ class _ServeCommand extends _WatchCommand {
           'args in <dir>[:<port>] format.');
     } else {
       for (var target in options.serveTargets) {
-        stdout.writeln('Serving `${target.dir}` on port ${target.port}');
+        stdout.writeln('Serving `${target.dir}` on http://${options.hostName}:${target.port}');
       }
     }
     await handler.buildResults.drain();


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -374,7 +374,7 @@ class _ServeCommand extends _WatchCommand {
           'args in <dir>[:<port>] format.');
     } else {
       for (var target in options.serveTargets) {
-        stdout.writeln('Serving `${target.dir}` on port ${target.port}');
+        stdout.writeln('Serving `${target.dir}` on http://${options.hostName}:${target.port}');
```
natebosch: This will be slightly odd for the magic hostName `any`... but I think that's probably OK.

PR #1229, opened 2018-04-02
Track failures per output rather than per phase

natebosch:
Towards #1018

This will allow us to be smarter about serving partial results for
builds which had some failure. This also starts refusing to allow later
steps to read outputs from builders that failed.

- Track failures per node rather than per phase, add a boolean on
  `GeneratedAssetNode` and update serialization.
- Set `isFailure` when a build step failed, and cascade the failure down
  the line of primary outputs.
- Don't use failed asset as primary inputs even if they were written.
- Throw an exception if a later step attempts to read an output which is
  failed.


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -344,6 +322,7 @@ class AssetGraph {
           //
           // The regular input hash checks won't pick up glob changes.
           node.state = GeneratedNodeState.definitelyNeedsUpdate;
+          node.isFailure = false;
```
grouma: Discussed offline. It might make things cleaner if failed is considered another state.

natebosch: I was wrong about needing to reset isFailure. We explicitly need to track this separately from the state, and we can't reset it when the state becomes "needs update", because in the case where the inputs were identical and we decide not to rerun we have to remember that the result is a failure rather than success.

The `definitelyNeedsUpdate` case would have been fine since we'd always rebuild it anyway, but the `mayNeedUpdate` case was getting clobbered by https://github.com/dart-lang/build/blob/c460c4a47baf16f29dedb4382f9260e51d30cf80/build_runner/lib/src/generate/build_impl.dart#L548-L552 so subsequent builds with identical digest were looking like they were successful. I added a test for this case.

Issue #1211, opened 2018-03-29
Asset graph handler compiled JS can get stale


natebosch:
The file `graph_viz_main.dart.js` needs to be recompiled any time we change the format for AssetGraph serialization.

We should probably have a test that this file is up to date, and we should document how to rebuild it.

natebosch:
It's not clear to me what the process should be - the process I found involved some manual renames. If there is a better way it might also involve wanting to ignore the .js.deps file so we should update .gitignore if that's the case.

See #1210

jakemac53:
Hmm ya, we could in theory add a build.yaml and build this file as a part of users builds but that would slow down initial builds :(.

Probably just some sort of test would be best.

natebosch:
We never actually documented how to rebuild this...

jakemac53:
and I just updated the version too so we definitely need to rebuild it :grimacing:

natebosch:
The compiled code no longer depends on the asset graph serialization which is why this was originally closed.

PR #1230, opened 2018-04-02
Recompile graph_viz_main.dart

natebosch:


natebosch:
I change the serialization format in https://github.com/dart-lang/build/pull/1229 and forgot to recompile then.

I have https://github.com/dart-lang/build/issues/1211 to add a test to prevent this issue.

Issue #1223, opened 2018-03-29
Allow access to pkg:build top-level `log` within `new BuilderApplication.forBuilder`


kevmoo:
It'd be nice to use this when options are being parsed.

PR #1231, opened 2018-04-02
Factory logger

grouma:
Closes https://github.com/dart-lang/build/issues/1223

### DIFF
--- a/build_runner/lib/src/package_graph/apply_builders.dart
+++ b/build_runner/lib/src/package_graph/apply_builders.dart
@@ -5,6 +5,7 @@
 import 'dart:async';
 
 import 'package:build/build.dart';
+import 'package:build/src/builder/logging.dart';
 import 'package:build_config/build_config.dart';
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -289,3 +291,19 @@ Map<String, List<BuilderApplication>> _applyWith(
   }
   return applyWith;
 }
+
+/// Runs [fn] in an error handling [Zone].
+///
+/// Any calls to [print] will be logged with `log.info`, and any errors will be
+/// logged with `log.severe`.
```
natebosch: I think we need to find a way to bubble up the error and cancel the overall build. Maybe it will be enough to do a null check later on?

grouma: Checking for null was enough. Added a corresponding test.

PR #1233, opened 2018-04-03
Load $graph view by node

natebosch:
Will allow us to expose build errors to this view for #1018

This should be functionally equivalent to what we had before, with a new
ability to request the node information for a single asset.
Responsiveness for refocusing the graph is a bit worse than before.

- Add AssetGraphHandler class to pull out individual nodes from the
  graph and serialize to a json representation of a node and it's edges.
- Cut out most behavior from `graph_viz.js` and move it into the server
  side or the Dart side.
- Collaps the `$perf`, `$graph`, and asset handlers into a single step
  in the Cascade since we want to allow different messages on `404`
  rather than falling through to the message for missing assets.


natebosch:
This will close #1211 since the compiled JS no longer depends on the asset graph serialization, only on other details from `asset_graph_handler.dart` and we're very unlikely to commit a change without recompiling.

build_runner/lib/src/server/asset_graph_handler.dart
```diff
@@ -0,0 +1,97 @@
...
+  /// For the path `$graph` returns the HTML page to render the graph.
+  ///
+  /// For any other path under `$graph/` will look for the assetNode referenced.
+  /// Path patterns can be:
+  /// - `$graph/packages/<package>/<path_under_lib>`
+  /// - `$graph/<path_under_root_package>`
+  /// - `$graph/<path_under_rootDir>`
+  ///
+  /// There may be some ambiguity between paths which are under the top-level of
+  /// the root package, and those which are under the rootDir.
```
jakemac53: Lets note here how those ambiguities are resolved

Issue #1097, opened 2018-03-05
Change default for --fail-on-severe


natebosch:
Maybe even remove the flag?

matanlurey:
Would be part of the way to https://github.com/dart-lang/build/issues/913.

Though FWIW, I find this really confusing, including internally. In AngularDart I've started throwing `BuildError` when you hit a "severe" state; it avoids getting into weird error cases (i.e. step 1 fails, so step 2 hits an NPE):

https://github.com/dart-lang/angular/blob/1d79577139085d927d6ee1b4a8cb9e41256d1336/angular_compiler/lib/src/cli/logging.dart#L112-L129

jakemac53:
I don't know that we actually want to do this until we have a way of doing partial builds. Or at least we shouldn't kill the build early as a result - I would be fine with treating the overall build as failed as long as we will still serve apps.

The use case I am primarily worried about is refactoring - I should be able to hack around during refactoring and serve any apps that still work. I shouldn't have to update my tests to match until after I am done with the refactor - or maybe I update them first if I am into TDD, either way would put you in a bad place if we bailed early on the builds.

matanlurey:
@jakemac53:

> I don't know that we actually want to do this until we have a way of doing partial builds. Or at least we shouldn't kill the build early as a result - I would be fine with treating the overall build as failed as long as we will still serve apps.

I guess I'm confused. Isn't a `severe` usually something that will prevent a successful serve/build anyway? If it's not, maybe we should encourage downgrading to a warning?

> The use case I am primarily worried about is refactoring - I should be able to hack around during refactoring and serve any apps that still work.

Ack, but again, I see this as `severe` == `your_app_will_not_work`. For example:

```dart
void main() {
  String x = 5;
}
```

... won't compile. There is no value in continuing*, right?

_*I realize there might be value on continuing other unrelated builds._

jakemac53:
@matanlurey the problem is we don't have a way right now of not building your entire package - so if any of your tests don't compile it would break the entire build and you wouldn't be able to serve your app etc.

Once we do have that, I am fully in favor of flipping this default.

natebosch:
> I would be fine with treating the overall build as failed as long as we will still serve apps.

Yeah I think what we should do is let builds continue, and serve any assets that were written even if that asset is marked failed, but track that things are failed. We can separately consider a flag which does/doesn't serve or copy to `--output` the assets which are marked failed.

matanlurey:
[...]

PR #1237, opened 2018-04-03
Allow serving assets for failed builds

natebosch:
Towards #1097

No longer refuse to server any assets if the overall build fails. Check
the asset node for individual requests and send an internal server error
if the build step producing that asset failed.

Add `canReadReason` to `FinalizedReader` to give more distinction than a boolean. We use this to determine whether to send a `500` or `404` for missing assets.

Take a `FinalizedReader` rather than generic `AssetReader` in `AssetHandler`.

### DIFF
--- a/build_runner/lib/src/asset/finalized_reader.dart
+++ b/build_runner/lib/src/asset/finalized_reader.dart
@@ -4,8 +4,10 @@
 
 import 'dart:async';
 import 'dart:convert';
+
 import 'package:build/build.dart';
 import 'package:build_runner/src/asset_graph/graph.dart';
+import 'package:build_runner/src/asset_graph/node.dart';
 import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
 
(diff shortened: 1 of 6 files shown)


natebosch:
This will likely change after https://github.com/dart-lang/build/pull/1238 goes in since I'll implement the logic for "hiding" failed assets in the reader.

Hold off on review for now.

natebosch:
Merged in and updated to use `FinalizedReader` PTAL.

build_runner/lib/src/asset/finalized_reader.dart
```diff
@@ -19,6 +21,20 @@ class FinalizedReader implements AssetReader {
     this._assetGraph,
   );
 
+  /// Returns a reason why [id] is not readable, or null if it is readable.
+  Future<UnreadableAssetReason> canReadReason(AssetId id) async {
```
jakemac53: This name implies the opposite of what it returns to me :D, maybe `unreadableReason`?

build_runner/lib/src/asset/finalized_reader.dart
```diff
@@ -19,6 +21,20 @@ class FinalizedReader implements AssetReader {
...
+    if (!_assetGraph.contains(id)) return UnreadableAssetReason.notFound;
+    var node = _assetGraph.get(id);
+    if (node.isDeleted) return UnreadableAssetReason.deleted;
+    if (!node.isReadable) return UnreadableAssetReason.assetType;
+    if (node is GeneratedAssetNode) {
+      if (node.isFailure) return UnreadableAssetReason.failed;
+      if (!node.wasOutput) return UnreadableAssetReason.notOutput;
+    }
+    if (await _delegate.canRead(id)) return null;
```
jakemac53: maybe throw?

natebosch: This is a strange api, `null` is actually the typical case.

Maybe I should try it more like a `Maybe` or `Optional` or something, where I add a read API that either gives you the `List<int>` or gives you an `UnreadableAssetReason`...

natebosch: Hmm, even that wouldn't accomplish what we need since we only want to get the bytes after checking the digest and deciding we have a cache miss on the browser side.

I suppose I could reintroduce a call to `canRead` and only check the reason if `canRead` returned false. I had hoped to not do all the checking twice but that might be a premature optimization.

build_runner/lib/src/asset/finalized_reader.dart
grouma: +1

build_runner/lib/src/asset/finalized_reader.dart
jakemac53: Ya I think just leaving in the `canRead` call and then having a separate call that expects to only be given unreadable asset ids would be a bit cleaner.

PR #1243, opened 2018-04-04
Add serialization for isDeleted

grouma:
Also add missing copyright.

### DIFF
--- a/build_runner/lib/src/asset/finalized_reader.dart
+++ b/build_runner/lib/src/asset/finalized_reader.dart
@@ -1,3 +1,6 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
 import 'dart:async';
 import 'dart:convert';
 import 'package:build/build.dart';
--- a/build_runner/lib/src/asset/finalized_reader.dart
+++ b/build_runner/lib/src/asset/finalized_reader.dart
@@ -1,6 +1,7 @@
 // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/asset/finalized_reader.dart
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
```
natebosch: [nit] extra blank line below the copyright to separate it from imports

build_runner/test/common/matchers.dart
```diff
@@ -30,6 +30,10 @@ class _AssetGraphMatcher extends Matcher {
     if (graph.allNodes.length != _expected.allNodes.length) matches = false;
     for (var node in graph.allNodes) {
       var expectedNode = _expected.get(node.id);
+      if (node.isDeleted != node.isDeleted) {
```
jakemac53: one of these should be `expectedNode.isDeleted` ;)

build_runner/test/common/matchers.dart
```diff
@@ -30,6 +30,10 @@ class _AssetGraphMatcher extends Matcher {
...
+        matchState['IsDeleted'] = [node.isDeleted, expectedNode.isDeleted];
```
jakemac53: the string here is the description and should include the asset id, something like `IsDeleted of ${node.id}`

build_runner/lib/src/asset/finalized_reader.dart
grouma: Done here and elsewhere.

build_runner/test/common/matchers.dart
grouma: Done.

build_runner/test/common/matchers.dart
grouma: You don't like tautologies?

PR #1251, opened 2018-04-06
Make example in build/README.md valid code

kevmoo:


build/README.md
```diff
@@ -52,27 +52,27 @@ with the same name, but an additional extension:
...
     var copy = inputId.addExtension(extension);
-    var contents = await buildStep.readAsString(input);
+    var contents = await buildStep.readAsString(inputId);
 
     /// Write out the new asset.
     ///
     /// There is no need to `await` here, the system handles waiting on these
     /// files as necessary before advancing to the next phase.
-    buildStep.writeAsString(copy, contents);
+    await buildStep.writeAsString(copy, contents);
```
natebosch: See 2 lines above - "There is no need to `await` here"

The reason this returns a `Future` is so it can implement `AssetReader`, but in this usage it isn't necessary to wait for the write to finish since it doesn't have an effect that impacts you.

If we had a way we'd tell the linter to ignore this.

kevmoo: Fixed here

See https://github.com/dart-lang/build/issues/1254

kevmoo:
@natebosch – PTAL

PR #1253, opened 2018-04-06
Implement build modes

natebosch:


build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -17,7 +17,7 @@ import 'package_graph.dart';
 import 'target_graph.dart';
 
 typedef BuildPhase BuildPhaseFactory(String package, BuilderOptions options,
-    InputSet targetSources, InputSet generateFor);
+    InputSet targetSources, InputSet generateFor, bool isReleaseBuild);
```
natebosch: I think this is technically breaking.  `BuildPhaseFactory` isn't public itself, but there is a public field on `BuilderApplication` which exposes it...

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -79,6 +91,9 @@ BuilderApplication apply(String builderKey,
         {bool isOptional,
         bool hideOutput,
         InputSet defaultGenerateFor,
+        BuilderOptions defaultOptions,
```
natebosch: After adding these arguments it might be better to take a `TargetBuilderConfigDefaults` rather than all of it's fields separately... but that would also make it a breaking change.

PR #1256, opened 2018-04-06
Add simple examples

kevmoo:
Things we can reference in docs/readmes

### DIFF
--- a/example/build.yaml
+++ b/example/build.yaml
@@ -2,7 +2,6 @@
 builders:
 
   copyBuilder:
-    target: ":example"
     import: "package:example/builder.dart"
     builder_factories: ["copyBuilder"]
     build_extensions: {".txt": [".txt.copy"]}
@@ -10,7 +9,6 @@ builders:
     auto_apply: root_package
 
   resolvingBuilder:
-    target: ":example"
     import: "package:example/builder.dart"
     builder_factories: ["resolvingBuilder"]
     build_extensions: {".dart": [".dart.info.json"]}
@@ -23,7 +21,6 @@ builders:
         - web/**
 
   cssBuilder:
-    target: ":example"
     import: "package:example/builder.dart"
     builder_factories: ["cssBuilder"]
     build_extensions: {"$web$": ["generated.css"]}
--- a/example/build.yaml
+++ b/example/build.yaml
@@ -26,3 +26,10 @@ builders:
     build_extensions: {".dart": [".dart.info.json"]}
     auto_apply: dependents
     build_to: source
+
+  cssBuilder:
+    target: ":example"
+    import: "package:example/builder.dart"
+    builder_factories: ["cssBuilder"]
+    build_extensions: {"$web$": ["generated.css"]}
+    auto_apply: root_package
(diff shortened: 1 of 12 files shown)


kevmoo:
Issue to rename `e2e_example` - https://github.com/dart-lang/build/issues/1257

example/build.yaml
```diff
@@ -0,0 +1,30 @@
+# Read about `build.yaml` at https://pub.dartlang.org/packages/build_config
+builders:
+
+  copyBuilder:
+    target: ":example"
```
jakemac53: Should we add comments for these now? Also I think you can actually omit `target`, its only needed for bazel and is just confusing right now imo.

kevmoo: https://github.com/dart-lang/build/tree/master/build_config Not commented as required/optional here...

kevmoo: See https://github.com/dart-lang/build/issues/1258

example/lib/example.dart
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: annotate_overrides
```
natebosch: would it be too noisy to put them in?

example/lib/example.dart
```diff
@@ -0,0 +1,98 @@
...
+          'name': entryLib.name,
+          'member count': entryLib.unit.declarations.length,
+          'visible libraries': visibleLibraries
+        }));
+  }
+
+  /// Configure output extensions. All possible inputs match the empty input
+  /// extension. For each input 1 output is created with `extension` appended to
+  /// the path.
+  Map<String, List<String>> get buildExtensions => {
```
natebosch: [nit] what do you think of

```dart
final buildExtensions = const {
  '.dart': const ['.dart.info.json']
};
```

I like the idea of demonstrating const extensions in our examples since I'd like most cases to use them. Hopefully the places where builders need to fall back on a `get` and use dynamic information on the instance should be rare.

example/lib/example.dart
kevmoo: Eh, if we want to copy-paste these into the README.

example/lib/example.dart
kevmoo: ack

example/lib/example.dart
natebosch: I'd be fine with copying them to the README. Except in cases where doing the "right"/idiomatic thing is very verbose I like for examples to follow good conventions.

kevmoo:
Agreed!

example/build.yaml
jakemac53: I meant all the keys - to document what each one means

kevmoo:
@natebosch PTAL – config is much more readable now!

example/lib/example.dart
```diff
@@ -0,0 +1,82 @@
...
+
+import 'package:build/build.dart';
+
+/// Copy contents of a `txt` files into `name.txt.copy`.
+///
+/// A header row is added pointing to the input file name.
+class CopyBuilder implements Builder {
+  @override
+  Future build(BuildStep buildStep) async {
+    /// Each [buildStep] has a single input.
```
natebosch: [nit] double slash here and below

kevmoo: ack

PR #1262, opened 2018-04-06
Add build mode support to build_runner

natebosch:
Towards #1109

Add a `--release` flag and a `isReleasMode` boolean to signal the mode.
Honor builder definition default options and target config options for
the current mode.

- Serialize and pass through  `options`, `devOptions`, and
  `releaseOptions` from builder definition defaults when generating the
  build script.
- Apply overrides to options starting with the builder definition
  defaults, then builder definitions defaults for the current mode, then
  target level options, then target level options for the current mode.
- Add arg parsting an plumb through `isReleaseBuild` through the `build`
  and `watch` functions.


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -272,11 +280,16 @@ abstract class BuildRunnerCommand extends Command<int> {
...
-      ..addFlag('verbose',
+      ..addFlag(_verbose,
           abbr: 'v',
           defaultsTo: false,
           negatable: false,
           help: 'Enables verbose logging.')
+      ..addFlag(_release,
+          abbr: 'r',
+          defaultsTo: false,
+          negatable: false,
```
grouma: Make negatable true

PR #1263, opened 2018-04-06
Require build_extensions again

natebosch:
Partially rolls back 265432a7c39c5b93d2bab8c2e945b1fd9dc8767e

I forgot that these _are_ used by `build_runner` since they can impact
the order we run builders in because of `required_inputs`.


kevmoo:
Oops!

PR #1268, opened 2018-04-07
Update getting started to use `--release`

natebosch:


natebosch:
cc @kwalrath - still need to publish, and users will need to be on the latest for `build_runner` and `build_web_compilers`, but we should start updating docs to refer to `--release` instead of `--config release` if there are other places we talk about this.

We expect most user shouldn't need  to use `--config` anymore.

Issue #1269, opened 2018-04-08
Feature request : Add more separation between messages printed by build tasks


dineshbhosale:
When `build_runner watch` is watching for file changes and re-running builds, it will be useful to clear old output messages or include extra separation between messages, sometimes it gets difficult to distinguish between old error messages and new error messages.

When build is successful it appears as if it has failed because old error messages are still there.
Some command line flag for clearing old output will be useful or adding extra separation will be useful.

In my use I was triggering `build_runner watch` again and again to clear out old messages because they were hard to distinguish. 
<img src="https://i.imgur.com/j32zgHy.png"></img>

PR #1271, opened 2018-04-08
Update build_runner/generate/build_impl.dart

dineshbhosale:
An attempt to fix :

https://github.com/dart-lang/build/issues/1269

Before adding : 
scrrrrr

After adding :
scr2

Now it prints a separator whenever a build task is finished.

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -221,6 +221,7 @@ class _SingleBuild {
       _logger.severe('Failed after ${humanReadable(watch.elapsed)}',
           result.exception, result.stackTrace);
     }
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -221,6 +221,7 @@ class _SingleBuild {
       _logger.severe('Failed after ${humanReadable(watch.elapsed)}',
           result.exception, result.stackTrace);
     }
+    print('-----------------------------------------------------------------');
```
natebosch: We'll need to go through the `_logger` rather than `print`. We also only want this for the watch/serve case, not for every build.

Will add a similar line with https://github.com/dart-lang/build/pull/1276

PR #1277, opened 2018-04-10
move the .packages file to the root directory always

jakemac53:
This is required for the node/vm --precompiled support.


jakemac53:
I don't know of any scenario in which it was being used before, it wasn't even valid really to have it under there.

jakemac53:
Submitting with appveyor flake (got it 3x in a row... but its definitely the flake).

PR #1278, opened 2018-04-10
stop creating missing html test files

jakemac53:
Can't be published until after the new package:test update is released (and I won't merge until that happens).

Unfortunately, we can't require the latest version of package:test since this is in build_runner... any ideas? 

As long as we release package:test first, when people `pub upgrade` everything should work fine and they shouldn't actually get broken (as long as they actually get the latest version of package:test).


jakemac53:
https://github.com/dart-lang/test/pull/809 will fix the travis issues

natebosch:
We should hold off on this for now, angular is explicitly upgrading only build_runner and not test right now since they are broken by `--precompiled` working in VM tests.

https://github.com/dart-lang/build/issues/1285

PR #1288, opened 2018-04-12
Avoid `packagePath` use in `PackageReader`

### LINKED ISSUE #1285: PackageAssetReader does not work with `build_runner --output` style .packages files
The documentation for `PackageResolver.packagePath` calls out that it works only for `.packages` style package map if it points to a real packages' `lib/` directory. In build runner's output we generate a `packages/` directory style layout which _doesn't_ include `lib/` and point the package map at it.

When we get the `packagePath` we're getting a value that doesn't make sense, and then attempting to append a `lib/something` path to it and not finding a real file.
https://github.com/dart-lang/build/blob/1b8cb5d04c82c0517c36961a3d5a3d748c7710f0/build_test/lib/src/package_reader.dart#L77-L81

For that case we can do something like `new File(_packageResolver.resolveUri(id.uri).path)` which seems to work. The `listAssets` case is tricker because we'd have to do some deeper analysis to figure out if we're seeing a package lib directory _not_ with the name `lib/` then somehow adjust the glob...

https://github.com/dart-lang/build/blob/1b8cb5d04c82c0517c36961a3d5a3d748c7710f0/build_test/lib/src/package_reader.dart#L92-L96

And I haven't figured out how this works when we try to resolve something in the "root" package.

natebosch:
Closes #1285

This method does not work when the `.packages` file does not have the
same semantics as the one created by pub - that is when it does use
point to directories named `lib/` under real packages. The package map
we create with `build_runner --output` does not have the same semantics,
it points to directoreis that are not naemd `lib/`.

- Implement `_resolve` on top of the package resolver directly resolving
  by URI for `package:` uris. This should always work.
- Implement `asset:` URIs by assuming the cwd is the root package, which
  is the only place we should get asset: URIs. This may be brittle but
  is the best we can do for now.
- Remove support for `findAssets` since the implementation is broken in
  these cases and it's not clear how we'd fix it.


build_test/CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
-## 0.10.2-dev
+## 0.10.2
 
 - Added a `DebugIndexBuilder`, which by default generates a `test/index.html`
   with links to debug tests in your `test/**_test.dart` folder, linking to the
   generated `*_test.debug.html` files. **NOTE**: This only works for web-based
   tests.
+- Fix `PackageAssetReader` when running with a package map generated by
```
jakemac53: nit: `when running a package map linking to "packages" folder structure`?

build_test/lib/src/package_reader.dart
```diff
@@ -74,26 +74,21 @@ class PackageAssetReader extends AssetReader
   }
 
   File _resolve(AssetId id) {
-    final packagePath = _packageResolver.packagePath(id.package);
-    if (packagePath == null) {
-      return null;
+    final uri = id.uri;
+    if (uri.isScheme('package')) {
+      return new File(_packageResolver.resolveUri(id.uri).path);
```
denniskaselow: Using the `path` from an uri fails under windows.

Doing something like 

```dart
      final file = new File(new File('.').absolute.uri.path);
      file.readAsString();
```

produces a `(OS Error: The filename, directory name, or volume label syntax is incorrect.
, errno = 123)` because the path start with a leading `/`.

denniskaselow: Can you use 
```dart
return new File.fromUri(_packageResolver.resolveUri(id.uri));
``` 
instead?

natebosch: Thanks!

Can you try with https://github.com/dart-lang/build/pull/1294 and see if it's working for you?

denniskaselow: Tested and with this change it's working again.

PR #1291, opened 2018-04-12
Reimplement findAssets in PackageAssetReader

natebosch:
This was removed because I (incorrectly) thought there were no usages
and it was hard to support. Add it back in with a hack of listing all
the files in the searched package, reconsituting a "lib/" structure, and
then running the glob against it.

This is potentially much less performant for large packages and narrow
globs since we always list all files instead of just listing the glob.

### DIFF
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -110,7 +110,7 @@ class PackageAssetReader extends AssetReader
           .list(recursive: true)
           .where((e) => e is File)
           .map((f) => p.relative(f.path, from: Directory.current.path))
-          .where((p) => !p.startsWith('packages/'))));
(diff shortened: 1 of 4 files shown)


build_test/lib/src/package_reader.dart
```diff
@@ -87,8 +88,31 @@ class PackageAssetReader extends AssetReader
...
+    }
+
+    var packageFiles = new Directory(packageLibDir.path)
+        .list(recursive: true)
+        .where((e) => e is File)
+        .map(
+            (f) => p.join('lib', p.relative(f.path, from: packageLibDir.path)));
+    if (package == _rootPackage) {
+      // TODO this assumes the cwd is the root package
+      packageFiles = packageFiles.transform(merge(Directory.current
```
jakemac53: I think this will duplicate files under `lib`

PR #1294, opened 2018-04-13
Avoid using `Uri.path`

natebosch:
This may not work on windows, instead use `p.fromUri` or `File.fromUri`
as needed. Remaining paths are from `File`, `Directory`, or `AssetId`
instances which should be safe.

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.10.2+2
+
+- Avoid looking for files from `Uri.path` paths.
+
 ## 0.10.2+1
 
 - Add back an implementation of `findAssets` in `PackageAssetReader`.
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -77,7 +77,7 @@ class PackageAssetReader extends AssetReader
   File _resolve(AssetId id) {
     final uri = id.uri;
     if (uri.isScheme('package')) {
-      return new File(_packageResolver.resolveUri(id.uri).path);
+      return new File.fromUri(_packageResolver.resolveUri(id.uri));
     }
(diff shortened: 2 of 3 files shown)


build_test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.2+2
+
+- Avoid looking for files from `Uri.path` paths.
```
kevmoo: Why? Just curious...

natebosch: https://github.com/dart-lang/build/pull/1288#pullrequestreview-111985549

build_test/lib/src/package_reader.dart
```diff
@@ -77,7 +77,7 @@ class PackageAssetReader extends AssetReader
   File _resolve(AssetId id) {
     final uri = id.uri;
     if (uri.isScheme('package')) {
-      return new File(_packageResolver.resolveUri(id.uri).path);
+      return new File.fromUri(_packageResolver.resolveUri(id.uri));
```
jakemac53: I think you can also do `.filePath`, which is how I fixed this before? Either way probably works though?

jakemac53:
If this was failing on windows it means we don't have test coverage for it - please add some so we can validate this change/prevent regressions.

jakemac53:
Oh wait - its because we only run e2e tests on windows...... bleh

natebosch:
Yeah I'm not sure if we want to run build_test tests on windows...

build_test/lib/src/package_reader.dart
natebosch: yeah I don't know if there is much reason to prefer one over the other.

I have a slight preference for `File.fromUri` if only because the type errors are nicer than "you gave `path` but should have given `toFilePath()` and they're both strings"

PR #1296, opened 2018-04-13
Clean build config

jakemac53:
Duplicate of https://github.com/dart-lang/build/pull/1295/files that just up front converts everything away from Yaml* to normal Map/List/etc.

Opening for posterity sake since I had already written it, not sure which is better really. I kind of like just making a total complete break from the yaml types though.


jakemac53:
Also only ran build_config tests ;)

jakemac53:
Just going to close this for now - as there are likely some other issues

PR #1298, opened 2018-04-13
Avoid deprecated onError in int.parse

natebosch:


natebosch:
Yeah I like `tryParse(something) ?? -1` a lot, I'm much less sure about the other case where I needed to add a method to set a top level variable...

PR #1311, opened 2018-04-18
crawl transitive modules in parallel

natebosch:
Previous implementation is serialized across file IO because we aren't
doing any work during `await reader.canRead` or `await
reader.readAsString`. `crawlAsync` implements a more aggressive
algorithm which attempts resolves nodes in parallel when possible.

### DIFF
--- a/build_modules/lib/src/modules.dart
+++ b/build_modules/lib/src/modules.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 
 import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
+import 'package:graphs/graphs.dart';
 import 'package:json_annotation/json_annotation.dart';
 
 import 'errors.dart';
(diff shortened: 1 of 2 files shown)


natebosch:
I'm not sure if we expect https://github.com/dart-lang/build/blob/39c2c0ecf65939b83170983bd7ac86a7100aa60e/build_modules/lib/src/modules.dart#L130 to trigger a build of a module which wasn't already output...

If so this change might not be desirable - it could cause more task churn computing the module output.

If not, or possibly even if so, this change should improve our ability to do work in parallel with disk IO for incremental builds (or with --low-resource-mode)

jakemac53:
I think it might cause builds, yes. I don't think it is necessarily worth optimizing since we already are scheduling so much async work. It is possible we could see benefit in low resource mode though.

natebosch:
Yeah I'm not seeing this make much of a difference in either direction.

PR #1312, opened 2018-04-19
Build modules from the bottom up with a resource pool

### LINKED ISSUE #1272: DDC builders memory usage can be reduced
I have a project with DDC builders setted up as in readme:
[README](https://github.com/dart-lang/build/blob/master/build_web_compilers/README.md#manual-usage)

It seems, that without lowResourceMode project can't be built on my machine at all.
With lowResourceMode it's possible, but take too much time, so I've investigated, what happens without lowResourceMode. My best guess is following:

Every builder (module, summaries, compiler) is marked as optional.
When entry point builder starts working, all dependent assets added to queue for building.
When any builder starts its build, it continues until first `await`. 
Then next builder (from queue) starts working, also till first `await`.
Process continues until builders for every asset is added to queue, but none of them can be completed, until every builder finishes all work before last `await` in its code.
So, none of the local variables (e.g. `transitiveDeps` from linked summary builder), can be cleaned up by GC.
It seems, that my machine can't finish normal build, because it tries to compile every asset in parallel, without posiibility to clean local variables, needed by any builder.

What fix I can suggest:
There can be some pool of resources for builders: do not start compiling new asset, until compilation of any previous N assets is finished.
Or may be there are some flags of dartVM, which can manage behaviour of async-await, but I couldn't find any.

Used versions:
Dart VM version: 2.0.0-dev.46.0 on "macos_x64"
build_runner: 0.8.1
build_modules: 0.2.0+2
build_web_compilers: 0.3.5

natebosch:
Fixes #1272

Each time an output is read for each module in transitive modules use a Pool and use strongly connected components to order them.

The assumption is that if we aren't running in dependency order we'll try to start builds for _all_ modules before any of them complete. By changing up the order we should only start builds that can complete without triggering further lazy builds.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -8,6 +8,7 @@ analyzer:
     override_on_non_overriding_method: error
   exclude:
     - "test/goldens/generated_build_script.dart"
+    - "e2e_example/build"
 linter:
   rules:
     - annotate_overrides
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.2.1
+
+- Give a guaranteed post-order traversal for
+  `Module.computeTransitiveDependencies`
+
 # 0.2.0+2
 
 - Fix use of `whereType` in `MissingModulesException`,
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 0.2.1
 
-- Give a guaranteed post-order traversal for
+- Give a guaranteed reverse dependency order for
   `Module.computeTransitiveDependencies`
 
 # 0.2.0+2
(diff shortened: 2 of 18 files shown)


natebosch:
Tested by running a build on my macbook against a package with 1000 angular components. In both cases it runs out of memory, but it gets further with this change.

Before, runs out of memory somewhere in the main process:

```
[INFO] 56.4s elapsed, 6378/6379 actions completed.Exhausted heap space, trying to allocate 56 bytes.
```

After, runs out of memory somewhere in the main process, and also gets severe errors from `build_web_compilers|entrypoint`

```

jakemac53:
Seems like a good idea in general. I think it would be ideal if we could do it in a more general way, but if that is impractical then something like this is fine for now and doesn't need to be blocked. I do think its worth filing a general issue describing the pattern/problem so we can track eventually providing a more general solution though, this is an easy trap to fall into.

natebosch:
I stopped using `stronglyConnectedComponents` and added some changelos. PTAL

natebosch:
I went ahead and split out https://github.com/dart-lang/build/pull/1316 since it'll have a green travis

Issue #1104, opened 2018-03-06
Add "clean" command


matanlurey:
... especially now that more tools will use `.dart_tool`, deleting that entire directory is sort of a smell. Ideally it would also mention that this shouldn't be needed, and we'd love a bug, i.e.:

```bash
$ pub run build_runner clean
$
$ > Deleting cache. This shouldn't be necessary for most applications, unless you
$ > have made intentional edits to generated files (i.e. for testing). Consider
$ > filing a bug at ...
```

matanlurey:
Thoughts @jakemac53 @natebosch?

kevmoo:
SGTM

natebosch:
We think today it'll delete `.dart_tool/build`? I think that makes sense for now.

jakemac53:
SGTM

PR #1314, opened 2018-04-19
Add a clean command

jakemac53:
This deletes generated to source outputs as well as the `.dart_tool/build` directory.

Fixes https://github.com/dart-lang/build/issues/1104

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -17,6 +17,13 @@ import 'package:shelf/shelf_io.dart';
 
 import 'package:build_runner/build_runner.dart';
 
+import '../asset/file_based.dart';
+import '../asset_graph/graph.dart';
+import '../asset_graph/node.dart';
+import '../logging/logging.dart';
+import '../logging/std_io_logging.dart';
+import '../util/constants.dart';
+
 const _assumeTty = 'assume-tty';
 const _define = 'define';
 const _deleteFilesByDefault = 'delete-conflicting-outputs';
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -539,6 +548,67 @@ class _TestCommand extends BuildRunnerCommand {
   }
 }
 
+class _CleanCommand extends Command<int> {
```
natebosch: at some point we need to consider splitting up this file... Doesn't need to happen now though.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -539,6 +548,67 @@ class _TestCommand extends BuildRunnerCommand {
...
+  @override
+  Future<int> run() async {
+    var logSubscription = Logger.root.onRecord.listen(stdIOLogListener);
+
+    logger.warning('Deleting cache and generated source files.\n'
+        'This shouldn\'t be necessary for most applications, unless you have '
+        'made intentional edits to generated files (i.e. for testing). '
+        'Consider filing a bug at '
+        'https://github.com/dart-lang/build/issues/new if you are using this '
+        'to work around an apparent bug.');
```
natebosch: maybe "apparent reproducible bug"

build_runner/lib/src/entrypoint/options.dart
jakemac53: Ya, we would probably want to use part files though to keep these commands private (at least for now?), and I would personally rather deal with a large file than part files most of the time :man_shrugging: .

build_runner/lib/src/entrypoint/options.dart
jakemac53: done

Issue #1186, opened 2018-03-19
Add a course-grained modules for non-path dependencies.


natebosch:
When files are not likely to change often we can group them into bigger DDC modules which means fewer (but bigger) summaries used for later actions.
Our intuition is that 100 large summaries will perform better than 1000 small ones.

jakemac53:
Note that this is also far fewer total actions and nodes in the graph, as well as fewer dependencies for the nodes that exist.

natebosch:
High level summary for our current and proposed module grouping algorithms:

Currently `build_runner` is a fine-grained as possible. Any Dart library which _can_ be it's own module _will_ be its own module. We only group multiple Dart libraries when the is a library cycle. We directly use the `libraryCycle` from analyzer: https://github.com/dart-lang/build/blob/1b8cb5d04c82c0517c36961a3d5a3d748c7710f0/build_modules/lib/src/modules.dart#L93

In `pub` things are grouped into bigger modules. @jakemac53 wrote it so would know better than me, but I think at a high level its:
- Every library under `lib/` get's its won module, or a grouped module if it's in a cycle like above
- Libraries under `lib/src` get grouped into as large of groups as possible such that each of the modules from `lib/` can satisfy it's imports without pulling in extra compiled code.

This is all implemented in `module_computer.dart`: https://github.com/dart-lang/pub/blob/711f515553a6b87e522715944391cced4dc96d13/lib/src/dartdevc/module_computer.dart

jakemac53:
> Every library under lib/ get's its won module, or a grouped module if it's in a cycle like above

It also rolls into its module anything under `lib/src` that it imports that isn't imported by any other public file under `lib`.

jakemac53:
Also libraries under `lib/src` that aren't imported by any public library I believe get their own module.

PR #1315, opened 2018-04-19
Allow for coarse module strategy

grouma:
Introduce the idea of "coarsely" built modules. These modules contain as many files as possible with the guarantee that an asset will be added to no more than one module.

High level changes:
- Introduce a MetaModule builder which creates a .meta_module file
  - This file contains a list of coarsely built modules for a given library
- Update Module builder to make use of .meta_module file
  - The .meta_module file is used to "clean" the module dependencies such that they refer to primary sources when necessary
- Enable configuration of Module builder to make use of new "coarse" strategy.
  - Note this is now the default strategy.
- Add tons of tests

Note that much of the MetaModule logic was borrowed from the pub build module computation:
https://github.com/dart-lang/pub/commit/5ca86eb6e7be45ade4e44cbb7402d296cdf8b32f#diff-dd41b3ddc504f3950e11d7ee4cf31780

Results:
With the previous strategy the angular components gallery took roughly 1 minute and 38 seconds on my machine to build. With the new strategy the build time is reduced to 1 minute and 6 seconds. Furthermore, initial page load time is reduced by roughly 2 seconds. 

Work towards https://github.com/dart-lang/build/issues/1186


build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
...
+import 'package:json_annotation/json_annotation.dart';
+
+import 'modules.dart';
+
+part 'meta_module.g.dart';
+
+/// Returns the top level directory in [uri].
+///
+/// Throws an [ArgumentError] if [uri] is just a filename with no directory.
+String _topLevelDir(String uri) {
```
jakemac53: nit: rename argument to `path` since its not a `Uri`. I think we already have this implemented somewhere as well.

build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
...
+/// by an `AssetId`.
+///
+/// Throws an [ArgumentError] if an [AssetId] can otherwise not be created. This
+/// might happen if:
+///
+/// * [importUri] is absolute but has a scheme other than `dart:` or `package:`.
+/// * [importUri] is a relative path that reaches outside of the current top
+///   level directory of a package (relative import from `web` to `lib` for
+///   instance).
+AssetId _importUriToAssetId(AssetId source, String importUri) {
```
jakemac53: You can use `AssetId.parse` instead of this, https://github.com/dart-lang/build/blob/master/build/lib/src/asset/id.dart#L54

build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
...
+    }
+    return new AssetId(source.package, targetPath);
+  }
+}
+
+/// An [AssetId] and all of its internal/external deps based on it's
+/// Directives.
+///
+/// Used to compute strongly connected components in the import graph for all
+/// "internal" deps. Any "external" deps are ignored during that computation
```
jakemac53: So this is an interesting aspect of this algorithm that I hadn't thought about.

We should figure out a way to support modules which have a cycle containing files in more than one package. I think this is doable - with one requirement - which is that a a coarse-grained package can not depend on a fine-grained package.

We should talk offline about the strategy here.

build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
...
+      }
+    }
+    var module = new Module(primaryNode.id, allAssetIds, allDepIds);
+    modules[module.primarySource] = module;
+  }
+  return modules;
+}
+
+/// Computes the strongly connected components reachable from entrypoints.
+List<Set<_AssetNode>> _stronglyConnectedComponents(
```
jakemac53: Note that we have a canonical strongly connected components implementation now - we should migrate this to use that. I think we still want this class.

build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
...
+    nodesById[asset] = node;
+  }
+  var connectedComponents = _stronglyConnectedComponents(nodesById);
+  var modulesById = _createModulesFromComponents(connectedComponents);
+  var modules = _mergeModules(modulesById, entryIds);
+  return modules;
+}
+
+@JsonSerializable()
+class MetaModule extends Object with _$MetaModuleSerializerMixin {
```
jakemac53: nit: consider separating this into its own file.

build_modules/lib/src/meta_module_builder.dart
```diff
@@ -0,0 +1,32 @@
...
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+
+import 'meta_module.dart';
+
+/// The extension for serialized meta module assets.
+const metaModuleExtension = '.meta_module';
+
+/// Creates `.meta_module` file for any Dart library.
```
jakemac53: nit: describe this a bit more `This module contains information about the full computed module structure for the package.`

build_modules/lib/src/module_builder.dart
```diff
@@ -7,15 +7,95 @@ import 'dart:convert';
...
+          .canRead(dep.changeExtension(moduleExtension))) {
+        cleanedDeps.add(dep);
+        continue;
+      }
+    }
+    var primaryDep = assetToPrimary[dep];
+    if (primaryDep == null) {
+      // We don't know the primary source for the dep so look for it
+      // using the meta module file.
+      var depMeta = new MetaModule.fromJson(json.decode(
```
jakemac53: Related to my comment above, but I believe this will fail if there is a module cycle between two packages. We can discuss later :D.

build_modules/lib/src/module_builder.dart
```diff
@@ -7,15 +7,95 @@ import 'dart:convert';
...
+    MetaModule meta, Map<AssetId, AssetId> assetToPrimary) {
+  for (var module in meta.modules) {
+    for (var source in module.sources) {
+      assetToPrimary[source] = module.primarySource;
+    }
+  }
+}
+
+/// Removes dependencies from the provided [Module] so that only those with
+/// corresponding module files remain.
```
jakemac53: nit: Note that this doesn't just remove dependencies - it modifies them to point at the primary source

build_modules/lib/src/module_builder.dart
```diff
@@ -7,15 +7,95 @@ import 'dart:convert';
...
+
 /// Creates `.module` files for any `.dart` file that is the primary dart
 /// source of a [Module].
 class ModuleBuilder implements Builder {
-  const ModuleBuilder();
+  final bool _isCoarse;
+  const ModuleBuilder({bool isCourse: false}) : _isCoarse = isCourse;
+
+  factory ModuleBuilder.forOptions(BuilderOptions options) {
+    return new ModuleBuilder(isCourse: options.config['strategy'] == 'coarse');
```
jakemac53: Lets add some validation around the acceptable values for `strategy`, otherwise typos can be frustrating.

Also, lets validate that we don't see any unrecognized top level keys.

build_modules/test/matchers.dart
```diff
@@ -36,6 +44,65 @@ class HasUnlinkedUris extends CustomMatcher {
...
+
+  EncodedMetaModuleMatcher(this.expected);
+
+  @override
+  bool matches(actual, description) {
+    if (actual is List<int>) {
+      actual = utf8.decode(actual as List<int>);
+    }
+    if (actual is! String) return false;
+    var jSon = json.decode(actual as String) as Map<String, dynamic>;
```
jakemac53: lol :man_facepalming: what have we done

build_modules/test/summary_builder_test.dart
```diff
@@ -18,6 +18,10 @@ main() {
     setUp(() async {
       assets = {
         'build_modules|lib/src/analysis_options.default.yaml': '',
+        'a|lib/.meta_module':
+            '{"m":[{"p":["a","lib/a.dart"],"s":[["a","lib/a.dart"]],"d":[]}]}',
```
jakemac53: nit: Create a MetaModule programatically and serialize it so this test is less brittle and easier to read

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.2.0+3
```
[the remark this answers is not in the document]
grouma: Done.

build_modules/lib/src/meta_module.dart
```diff
@@ -0,0 +1,433 @@
+import 'dart:async';
```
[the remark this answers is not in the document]
grouma: Done.

build_modules/lib/src/meta_module.dart
grouma: Done. Couldn't find any other implementation after some basic searches.

build_modules/lib/src/meta_module.dart
grouma: I can see moving the computerModule logic over to another file but going to leave it as is for now.

build_modules/lib/src/meta_module_builder.dart
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
grouma: Done.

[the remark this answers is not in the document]
grouma: Done.

build_modules/lib/src/meta_module_builder.dart
grouma: Done.

build_modules/lib/src/module_builder.dart
grouma: Should be more clear now.

build_modules/test/summary_builder_test.dart
grouma: Done.

build_modules/lib/src/meta_module.dart
grouma: Done.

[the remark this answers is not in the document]
grouma: Done.

grouma:
Will hold off on releasing this until the cycle issue is resolved which should be done by early next week. All other issues should be addressed. PTAL.

build_modules/test/module_builder_test.dart
```diff
@@ -36,6 +36,24 @@ main() {
       'a|lib/d.module': encodedMatchesModule(moduleD),
     });
   });
+  test('can output modules if there exists a cycle', () async {
```
jakemac53: Not sure if we need to keep this test really?

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.2.2
+
+- Add new `strategy` option for `ModuleBuilder`. If `coarse` is provided a
```
natebosch: mention what the default is

build_modules/build.yaml
```diff
@@ -1,6 +1,15 @@
 builders:
+  meta_module:
```
natebosch: Now that I'm looking at it, I wonder if we should make this a single builder definition from the start...

I'd merge this with the `modules` builder

build_modules/CHANGELOG.md
grouma: Done.

build_modules/build.yaml
grouma: Leaving for a follow up PR.

Issue #1248, opened 2018-04-05
Logging newlines on windows is broken


jakemac53:
The console overwrites all lines instead of only info lines, and even with --verbose some newlines are missing. We might be missing a carraige return somewhere or something?

PR #1317, opened 2018-04-19
fix newlines on windows

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1248

### DIFF
--- a/build_runner/lib/src/logging/logging.dart
+++ b/build_runner/lib/src/logging/logging.dart
@@ -3,7 +3,6 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:io';
 
 import 'package:logging/logging.dart';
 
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/logging/logging.dart
```diff
@@ -9,8 +9,8 @@ import 'package:logging/logging.dart';
 
 import 'human_readable_duration.dart';
 
-// Ensures on windows this message does not get overwritten by later logs.
-final _logSuffix = '${Platform.isWindows ? '' : '\n'}';
+// Ensures this message does not get overwritten by later logs.
```
jakemac53: The irony here is pretty amazing lol.

natebosch:
I was super curious about this so I dug.

This was added because on windows we used to not ever overwrite the line but on other platforms we did. So this was preventing an _extra_ newline on windows where we wouldn't be deleting lines. At some point we started doing the overwrite on windows too and didn't clean this up.

https://github.com/dart-lang/build/pull/83#discussion-diff-55567597

Issue #1283, opened 2018-04-11
Better message and exit code when merged directory cannot be written


kevmoo:
with Dart SDK 2.0.0-dev.47.0

- build_runner 0.8.2+2
- build_web_compilers 0.3.6

Get exit code `1` and output

```
    [INFO] Initializing inputs
    [INFO] Reading cached asset graph...
    [INFO] Reading cached asset graph completed, took 304ms

    [INFO] Checking for updates since last build...
    [INFO] Checking for updates since last build completed, took 667ms

    [INFO] Running build...
    [INFO] Running build completed, took 17ms

    [INFO] Caching finalized dependency graph...
    [INFO] Caching finalized dependency graph completed, took 118ms

    [WARNING] Skipped creation of the merged output directory.
[e] [SEVERE] Failed after 148ms
[e] Failed to create merged output directories.
```

Not helpful.

Suggestion:

- [Exit code 73 - "cannot create output file"](https://pub.dartlang.org/documentation/io/0.3.2+1/io/ExitCode/cantCreate-constant.html)
- Something about "target directory `XYZ` already exists. Choose a different directory or delete the contents of that directory"

PR #1319, opened 2018-04-20
Proper error values

grouma:
Allow configurable exit codes.

Towards https://github.com/dart-lang/build/issues/1283

We still need to provide better messaging for merge errors.

### DIFF
--- a/build_runner/lib/src/generate/build_result.dart
+++ b/build_runner/lib/src/generate/build_result.dart
@@ -65,13 +65,11 @@ enum BuildStatus {
 }
 
 /// The type of failure
-///
-/// Values correspond to proposed exitCode.
 class FailureType {
-  static const general = const FailureType(1);
-  static const cantCreate = const FailureType(73);
-  final int value;
-  const FailureType(this.value);
+  static const general = const FailureType._(1);
+  static const cantCreate = const FailureType._(73);
+  final int exitCode;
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/generate/build_result.dart
```diff
@@ -56,6 +64,16 @@ enum BuildStatus {
   failure,
 }
 
+/// The type of failure
+///
+/// Values correspond to proposed exitCode.
+class FailureType {
+  static const general = const FailureType(1);
+  static const cantCreate = const FailureType(73);
+  final int value;
```
jakemac53: nit: rename exitCode?

natebosch: +1

build_runner/lib/src/generate/build_result.dart
```diff
@@ -56,6 +64,16 @@ enum BuildStatus {
...
+  const FailureType(this.value);
```
kevmoo: Make this private? Keep the set of exit codes restricted to the ones provided here?

build_runner/lib/src/generate/build_result.dart
grouma: package:io doesn't expose those values but I have updated the name to exitCode.

PR #1320, opened 2018-04-20
Fix recording of input digests on the asset graph

natebosch:
Depending on the order that inputs were read we could compute a
different `previousInputsDigest` than we would with identical inputs on
the next build. This would cause more actions to be run than necessary -
though it only impacts the case where all inputs were identical which is
relatively rare in practice, however upcoming changes to the module
strategy will make it much more likely.

- Store read assets in a sorted set which matches the way we store
inputs on the asset graph.
- Don't always include the primary input as if it was read - if the
build step doesn't read the primary input we can assume it does not
depend on rerunning when the content changes.
- Pull the digest computation out of the loop since we can always run it
against  the first output's node.

### DIFF
--- a/build_runner/lib/src/asset/reader.dart
+++ b/build_runner/lib/src/asset/reader.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 
 import 'package:build/build.dart';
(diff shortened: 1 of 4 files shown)


natebosch:
I'll check on whether I can easily write a regression test for this, but I don't think we should wait to publish.

build_runner/lib/src/asset/reader.dart
```diff
@@ -29,7 +30,7 @@ abstract class RunnerAssetReader implements MultiPackageAssetReader {}
 /// tracking.
 class SingleStepReader implements AssetReader {
   final AssetGraph _assetGraph;
-  final _assetsRead = new Set<AssetId>();
+  final _assetsRead = new SplayTreeSet<AssetId>();
```
jakemac53: nit: we should add a comment here about ordering so this doesn't get reverted

jakemac53: Yes, this is the more important part if anything imo.

natebosch: https://github.com/dart-lang/build/pull/1327

PR #1322, opened 2018-04-21
Remove from _pendingWrites in whenComplete

natebosch:
Previously if readAsString threw it could get the build stuck because
we'd never remove the pendingWrite Future and could get stuck waiting
forever.

Closes https://github.com/dart-lang/angular/issues/1259.

### DIFF
--- a/scratch_space/lib/src/scratch_space.dart
+++ b/scratch_space/lib/src/scratch_space.dart
@@ -27,7 +27,7 @@ class ScratchSpace {
   final Directory tempDir;
 
   // Assets which have a file created but are still being written to.
-  final _pendingWrites = <AssetId, Future<Null>>{};
+  final _pendingWrites = <AssetId, Future<void>>{};
 
   ScratchSpace._(this.tempDir)
       : packagesDir = new Directory(p.join(tempDir.path, 'packages'));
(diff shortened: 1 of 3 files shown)


scratch_space/lib/src/scratch_space.dart
```diff
@@ -93,14 +93,13 @@ class ScratchSpace {
...
-            // ignore: unawaited_futures
-            _pendingWrites.remove(id);
-          });
-        }();
+        var done = reader
+            .readAsBytes(id)
+            .then((bytes) =>
+                _descriptorPool.withResource(() => file.writeAsBytes(bytes)))
+            .whenComplete(() {
+          _pendingWrites.remove(id);
```
leonsenft: nit: use .whenComplete(() => _pendingWrites.remove(id)) to be consistent with previous line 😀

natebosch: would need to be `..whenComplete` in that case. The `remove` would return `done` which turns this into an infinitely recursing Future

leonsenft: Actually looking at the documentation, `whenComplete` doesn't return the inner future. It still returns the original future: https://api.dartlang.org/stable/1.24.3/dart-async/Future/whenComplete.html

leonsenft: Oh weird indeed! When I try it that way it hangs _without_ an error just on the first compile. Perhaps `whenComplete` is not implemented as described.

leonsenft: Oh of course! I totally overlooked that despite returning the result of the original future, it will only await any returned future first before doing so. Your code is correct; I was mistaken!

natebosch: ah yeah that makes sense. It ignores the result but still waits for it

PR #1324, opened 2018-04-21
Handle module cycles

grouma:
- Introduce new MetaModuleCleanBuilder which outputs a .meta_module_clean
- This new output contains a list of cleaned modules with cyclic modules merged into one

I'm seeing even further performance benefits with this approach. Curious how it will impact the Wrike folks.

Towards https://github.com/dart-lang/build/issues/1186

### DIFF
--- a/build_modules/build.yaml
+++ b/build_modules/build.yaml
@@ -1,21 +1,15 @@
 builders:
-  meta_module:
-    import: "package:build_modules/builders.dart"
-    builder_factories:
-      - metaModuleBuilder
-    build_extensions:
-      $lib$:
-        - .meta_module
-    is_optional: True
-    auto_apply: all_packages
-    required_inputs: [".dart"]
   modules:
     import: "package:build_modules/builders.dart"
     builder_factories:
+      - metaModuleBuilder
+      - metaModuleCleanBuilder
       - moduleBuilder
       - unlinkedSummaryBuilder
       - linkedSummaryBuilder
     build_extensions:
+      $lib$:
+        - .meta_module
       .dart:
         - .module
         - .linked.sum
@@ -25,4 +19,4 @@ builders:
     defaults:
       options:
         strategy: 'coarse'
-    required_inputs: [".dart", ".meta_module"]
+    required_inputs: [".dart"]
--- a/build_modules/lib/src/meta_module_clean_builder.dart
+++ b/build_modules/lib/src/meta_module_clean_builder.dart
@@ -58,7 +58,11 @@ Future<Set<Module>> _transitiveModules(BuildStep buildStep, AssetId metaAsset,
   return dependentModules;
 }
 
-Module _cleanModule(
+/// Merges the modules in a strongly connected component.
+///
+/// Note this will clean the module dependencies as the merge happens.
+/// The result will be that all dependencies are primary sources.
+Module _mergeComponent(
     List<Module> connectedComponent, Map<AssetId, Module> assetToModule) {
   var sources = new Set<AssetId>();
   var deps = new Set<AssetId>();
@@ -106,7 +110,7 @@ class MetaModuleCleanBuilder implements Builder {
     var cleanModules = new SplayTreeSet<Module>(
         (a, b) => a.primarySource.compareTo(b.primarySource));
     for (var connectedComponent in connectedComponents) {
-      var cleanModule = _cleanModule(connectedComponent, assetToModule);
+      var cleanModule = _mergeComponent(connectedComponent, assetToModule);
       for (var source in cleanModule.sources) {
         if (source.package == buildStep.inputId.package) {
           cleanModules.add(cleanModule);
--- a/build_modules/lib/src/meta_module_clean_builder.dart
+++ b/build_modules/lib/src/meta_module_clean_builder.dart
@@ -1,0 +1,121 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:collection';
+import 'dart:convert';
+
+import 'package:graphs/graphs.dart';
+
+import 'package:build/build.dart';
+import 'meta_module.dart';
+import 'meta_module_builder.dart';
+import 'modules.dart';
+
+/// The extension for serialized clean meta module assets.
+const metaModuleCleanExtension = '.meta_module_clean';
+
+/// Map of [AssetId] to corresponding primary [AssetId] within the same
+/// [Module].
+final _assetToModule =
+    new Resource<Map<AssetId, Module>>(() => {}, dispose: (map) => map.clear());
+
+/// Returns a set of all modules transitvely reachable from the provided meta
+/// module asset.
+Future<Set<Module>> _transitiveModules(BuildStep buildStep, AssetId metaAsset,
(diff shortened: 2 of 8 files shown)


build_modules/build.yaml
```diff
@@ -1,21 +1,15 @@
...
     import: "package:build_modules/builders.dart"
     builder_factories:
+      - metaModuleBuilder
+      - metaModuleCleanBuilder
       - moduleBuilder
       - unlinkedSummaryBuilder
       - linkedSummaryBuilder
     build_extensions:
+      $lib$:
+        - .meta_module
```
natebosch:
```yaml
- .meta_module.raw
- .meta_module.clean
```

?

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+import 'dart:convert';
+
+import 'package:graphs/graphs.dart';
+
+import 'package:build/build.dart';
+import 'meta_module.dart';
+import 'meta_module_builder.dart';
+import 'modules.dart';
+
+/// The extension for serialized clean meta module assets.
```
natebosch: I'm unsure about the name "clean"...

Maybe `.package_modules` and `.cross_package_modules` instead?

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+  nextModules.addAll(meta.modules);
+  while (nextModules.isNotEmpty) {
+    var module = nextModules.removeLast();
+    dependentModules.add(module);
+    for (var source in module.sources) {
+      assetToModule[source] = module;
+    }
+    for (var dep in module.directDependencies) {
+      var depMetaAsset = new AssetId(dep.package, 'lib/$metaModuleExtension');
+      if (!seenMetas.contains(depMetaAsset)) {
```
natebosch: reduce nesting by one by reversing this condition with a `continue`

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+const metaModuleCleanExtension = '.meta_module_clean';
+
+/// Map of [AssetId] to corresponding primary [AssetId] within the same
+/// [Module].
+final _assetToModule =
+    new Resource<Map<AssetId, Module>>(() => {}, dispose: (map) => map.clear());
+
+/// Returns a set of all modules transitvely reachable from the provided meta
+/// module asset.
+Future<Set<Module>> _transitiveModules(BuildStep buildStep, AssetId metaAsset,
```
natebosch: I wonder how well this fits with `crawlAsync`?

https://pub.dartlang.org/documentation/graphs/0.1.1/graphs/crawlAsync.html

The benefit would be that we'd separate out the business logic of how we resolve modules and dependencies from the graph algorithm

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+  Future build(BuildStep buildStep) async {
+    var assetToModule = await buildStep.fetchResource(_assetToModule);
+    var modules =
+        await _transitiveModules(buildStep, buildStep.inputId, assetToModule);
+    var connectedComponents = stronglyConnectedComponents<AssetId, Module>(
+        modules,
+        (m) => m.primarySource,
+        (m) => m.directDependencies
+            .map((d) => assetToModule[d])
+            .where((d) => d != null));
```
natebosch: when would we not have a module in `assetToModule`?

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+    // Ensure deterministic output by sorting the modules.
+    var cleanModules = new SplayTreeSet<Module>(
+        (a, b) => a.primarySource.compareTo(b.primarySource));
+    for (var connectedComponent in connectedComponents) {
```
natebosch: could write this in a way that more directly expresses that we're interested the collection rather than side effects.

```dart
  Module merge(List<Module> c) => _mergeComponent(c, assetToModule);
  bool hasSourceInPackage(Module m) =>
      m.sources.any((s) => s.package == buildStep.inputId.package);
  new SplayTreeSet<Module>()
    ..addAll(connectedComponents.map(merge).where(hasSourceInPackage));
```

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+
+/// The extension for serialized clean meta module assets.
+const metaModuleCleanExtension = '.meta_module_clean';
+
+/// Map of [AssetId] to corresponding primary [AssetId] within the same
+/// [Module].
+final _assetToModule =
+    new Resource<Map<AssetId, Module>>(() => {}, dispose: (map) => map.clear());
+
+/// Returns a set of all modules transitvely reachable from the provided meta
```
jakemac53: transitively*

build_modules/lib/src/meta_module_clean_builder.dart
jakemac53: You could do a `crawlAsync` for `MetaModule`s (children would be all the merged MetaModule dependencies of all the contained modules), and then you could expand all the modules afterwords?

jakemac53: nit: I prefer to have private implementations below the public apis in the file. Also I think its a bit more natural to put private methods implementations below their actual usage in the file, roughly in the order in which they are invoked by the public methods.

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -0,0 +1,125 @@
...
+  var deps = new Set<AssetId>();
+  // Sort the modules to deterministicly select the primary source.
+  var components = new SplayTreeSet<Module>(
+      (a, b) => a.primarySource.compareTo(b.primarySource))
+    ..addAll(connectedComponent);
+  var primarySource = components.first.primarySource;
+  for (var module in connectedComponent) {
+    sources.addAll(module.sources);
+    for (var dep in module.directDependencies) {
+      var depModule = assetToModule[dep];
```
jakemac53: It is a bit hard to follow, but will this map contain un-merged modules? We need to make sure it ends up pointing at the merged modules returned by this function, right?

jakemac53:
I would really like to see some full E2E tests to confirm all the cases we care about are covered, and all the interactions work as we expect. At least these two cases:

- A root package cycle should make all packages in the cycle revert to the fine strategy.
- An example of a cycle in package dependencies.

build_modules/lib/src/meta_module_clean_builder.dart
grouma: I can't use the crawlAsync method since it doesn't propagate errors as expected.

build_modules/lib/src/meta_module_clean_builder.dart
natebosch: Can the doc comment explain what "clean" means?

build_modules/lib/src/meta_module_clean_builder.dart
natebosch: https://github.com/dart-lang/graphs/pull/18

build_modules/lib/src/meta_module_clean_builder.dart
grouma: Done.

PR #1326, opened 2018-04-21
Plumb through a skip-build-script-check argument

natebosch:
Hidden from help and no changelog because most end users should not need
this.


kevmoo:
Ya'll need to use https://pub.dartlang.org/packages/build_cli

Issue #1087, opened 2018-03-05
Devcompiler dynamic loading doesnt work with chrome extensions


Meai:
For content scripts, the baseUrl is going to be whatever page you are on and so it will never find all these extra scripts that it's trying to load: 

`
https://reddit.com/packages/$sdk/dev_compiler/amd/require.js net::ERR_ABORTED
https://reddit.com/packages/$sdk/dev_compiler/web/dart_stack_trace_mapper.js net::ERR_ABORTED
`

jakemac53:
Maybe we just need a `baseUrl` configuration option which would be set for these scripts as well as the requireJs global config?

dineshbhosale:
When a chrome extensions inject a content script into browser that references other dart files then nothing works because DDC is referencing incorrect URLs, so developer is forced to use `dart2js`.

But, developing chrome extensions with `dart2js` is slow as compared to DDC.
While using dart2js When a global dart file is edited then almost entire source code have to be recompiled.

Setting a custom `baseUrl` for applications will ease the development process.

jakemac53:
Just to confirm, the `<base>` tag support doesn't solve this issue still right?

davidmorgan:
Closing as stale.

PR #1330, opened 2018-04-23

dineshbhosale:
Partial fix for https://github.com/dart-lang/build/issues/1087

This will attempt to detect `baseUrl` using `<base href="/">` html element.

Most angular dart apps specify base href using `<base href="/">`, so this will work for them without any issues.

Works with both `routerProviders` , `routerProvidersHash`

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -314,7 +314,6 @@ var baseUrl = (function() {
   if(baseTag && baseTag[0] && baseTag[0].href){
     return baseTag[0].href;
   }
-
   var pathParts = location.pathname.split("/");
   if (pathParts[0] == "") {
     pathParts.shift();
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -310,6 +310,12 @@ final _baseUrlScript = '''
 // Attempt to detect --precompiled mode for tests, and set the base url
 // appropriately, otherwise set it to "/".
 var baseUrl = (function() {
+  if(document.getElementsByTagName("base")){
+  if(document.getElementsByTagName("base")[0]){
+  if(document.getElementsByTagName("base")[0].href){
+    return document.getElementsByTagName("base")[0].href;
+  }}}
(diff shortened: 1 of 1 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -310,6 +310,12 @@ final _baseUrlScript = '''
 // Attempt to detect --precompiled mode for tests, and set the base url
 // appropriately, otherwise set it to "/".
 var baseUrl = (function() {
+  if(document.getElementsByTagName("base")){
+  if(document.getElementsByTagName("base")[0]){
```
jakemac53: nit: only do one query on the document and assign it to a local variable

dineshbhosale: If we don't do it this way then it will cause an error if the element is not found. I don't know how else we can implement this without causing an error.

jakemac53:
I am a little bit concerned that this could break some existing users - but we could release it as a breaking change even.

I don't know how common it is to set this to some other domain, which would obviously break things. Ideally we would check that the original script url was relative before doing this.

@natebosch what do you think?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
natebosch: not sure why it would cause an error...

```js
var baseTag = document.getElementsByTagName("base");
if(baseTag && baseTag[0] && baseTag[0].href){
  return baseTag[0].href;
}
```

kevmoo:
> an you try this out with a web worker and see what the impact is?

Tried patching this into the hacker news app in the angular repo – it breaks the service worker. 😕

dineshbhosale:
Yes we shouldn't merge this, We need a better solution to set `baseUrl` variable.

For now I can not find any other way to set baseUrl other than this method. 

For example, when root URL of an Angular Dart 5 application served using nginx is `https://example.com/toolkit_for_facebook/`

and `index.html` includes following tag
`<base href="/toolkit_for_facebook/">`

Then script generated by `build_web_compilers` should be able to properly set values for `baseUrl` variable but it does not and gives errors as shown in following image.

baseurlerr


So, adding above lines will properly detect value for baseUrl variable.

Issue #896, opened 2018-01-23
Deployed Angular 5 app results in 404 for require.js


chalin:
TL;DR `<base href>` is ignored by `main.dart.js` generated for Angular 5 app.

---

The deployed version of the 5.0.0-alpha+4 version of the [quickstart][] example app fails to load with the following **error appearing in the console**:

```nocode
GET http://angular-examples.github.io/packages/$sdk/dev_compiler/web/dart_stack_trace_mapper.js net::ERR_ABORTED
(anonymous) @ main.dart.js:50

GET http://angular-examples.github.io/packages/$sdk/dev_compiler/amd/require.js net::ERR_ABORTED
(anonymous) @ main.dart.js:56
```

To reproduce, visit: http://angular-examples.github.io/quickstart/5

This problem did not appear using the old build system under Angular 4.

What seems to be happening is that the `<base href="/quickstart/5/">` contained in `index.html` is being ignored by by [main.dart.js](https://github.com/angular-examples/quickstart/blob/gh-pages/5/main.dart.js#L32-L54):

```js
  var baseUrl = ...
  var el;
  el = document.createElement("script");
  ...
  el.src =
    baseUrl + "packages/$sdk/dev_compiler/web/dart_stack_trace_mapper.js";
 ...
```

The setting of the baseUrl doesn't seem to take the `<base href>`.

cc @kwalrath @kevmoo @matanlurey 

[quickstart]: https://webdev-dartlang-org-dev.firebaseapp.com/angular/guide/setup#create-a-starter-project

matanlurey:
The example you are showing is with `dartdevc`, did you used to do that with v4?

chalin:
No, under v4 we used dart2js.

matanlurey:
So that's not comparing apples and oranges then :)

Any reason you'd want to _deploy_ `dartdevc`? That usually isn't desired.

chalin:
Sure, the goal wasn't to compare but to point out the issue.

> Any reason you'd want to deploy dartdevc? That usually isn't desired.

Jan 17, 18:13, on Gitter [dart-lang/source_gen](https://gitter.im/dart-lang/source_gen), we were discussing the building and testing of the Angular example apps. In the end you wrote:

> There probably isn’t huge value in running in dart2js
> We don’t internally

To which Jacob agreed. Until there is an easier way to switch between web compilers (#801), I was planning on building and deploying using dartdevc.

[...]

PR #1332, opened 2018-04-23

dineshbhosale:
Partial fix for #1087 
Works with service workers, `routerProviders`, `routerProvidersHash`.


jakemac53:
@chalin would this resolve the issues you were having with deployed ddc apps?

natebosch:
Previous context: https://github.com/dart-lang/build/pull/1330

jakemac53:
Also cc @matanlurey @alorenzen does this seem reasonable to you guys from an angular perspective?

matanlurey:
I don't know enough about how this works, unfortunately.

For AngularDart, there are two ways to set the base URL:
* In the HTML, above
* As a provider

I assume that the former only didn't work, and this makes it work?

dineshbhosale:
Yes,
This will allow to develop using DDC while using different root directories, `http://example.com/abc/` while using `routerProviders` at the same time, not breaking redirect links (for example password reset link) links sent by backend.

dineshbhosale:
Also it is very useful for chrome extensions, because for now there is no other way to set `baseUrl` for chrome extensions.

chalin:
> @chalin would this resolve the issues you were having with deployed ddc apps?

Thanks for the ping @jakemac53. The issue was: _Deployed Angular 5 app results in 404 for require.js_ #896. Looking at this PR now.

jakemac53:
> Thanks for the ping @jakemac53. The issue was: Deployed Angular 5 app results in 404 for require.js #896. Looking at this PR now.

Ok yes - so it looks like you are setting the `<base href="...">` so i think this would resolve the issue potentially?

chalin:
> looks like you are setting the <base href="..."> so i think this would resolve the issue potentially?

Yes, though if we're going to fix the code, I'd clean it up too (e.g., we should be using `===`, etc). I'd also want to ensure that we only accept href's that start with '/'.

I'd suggest something like this:

```js
var baseUrl = (function () {
  var href = '/';
    var el = document.getElementsByTagName('base')[0];
    if (el && el.href && el.href.startsWith('/')) href = el.href;
  } else {
    var pathParts = location.pathname.split('/');
    if (pathParts[0] === '') pathParts.shift();
    if (pathParts[1] === 'test') href = '/' + pathParts.slice(0, 2).join('/') + '/';
  }
  return href;
}());
```

Btw, is the "Attempt to detect --precompiled mode" still relevant?

cc @kwalrath

dineshbhosale:
After trying different variations, this one works without any errors

```js
var baseUrl = (function () {
  // base href should start with "/"
    if (el && el[0] && el[0].href && el[0].href.startsWith('/')){
    	return el[0].href;
    }
  }
  if (pathParts[0] == "") {
    pathParts.shift();
  }
  var baseUrl;
  }
  return "/";
}());
```

jakemac53:
Ok, I think we are going to go with this change, thanks @dineshbhosale.

Please push a CHANGELOG.md update as well, something along the lines of `dartdevc now respects the <base> tag".

jakemac53:
I went ahead and just pushed the changelog/pubspec updates, will merge on green. Thanks @dineshbhosale!

jakemac53:
Released as 0.3.7+3 of build_web_compilers

Issue #1334, opened 2018-04-23
Add `isRoot` boolean to `BuilderOptions`.


jakemac53:
This allows builders to have defaults which are different for the root package than other packages.

PR #1336, opened 2018-04-23
add isRoot and support it in build_runner

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1334

cc @grouma

### DIFF
--- a/build/lib/src/builder/builder.dart
+++ b/build/lib/src/builder/builder.dart
@@ -43,8 +43,8 @@ class BuilderOptions {
   /// this instance.
   ///
   /// Config values are overridden at a per-key granularity. There is no value
-  /// level merging. [other] may be null or empty, in which case this instance
-  /// is returned directly.
+  /// level merging. [other] may be null, in which case this instance is
+  /// returned directly.
   ///
   /// The `isRoot` value will also be overridden to value from [other].
   BuilderOptions overrideWith(BuilderOptions other) {
--- a/build/lib/src/builder/builder.dart
+++ b/build/lib/src/builder/builder.dart
@@ -28,6 +28,9 @@ class BuilderOptions {
   /// A configuration with no options set.
   static const empty = const BuilderOptions(const {});
 
+  /// A configuration with [isRoot] set to `true`, and no options set.
+  static const forRoot = const BuilderOptions(const {}, isRoot: true);
+
   /// The configuration to apply to a given usage of a [Builder].
   ///
   /// A `Map` parsed from json or yaml. The value types will be `String`, `num`,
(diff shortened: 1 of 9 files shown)


build/lib/src/builder/builder.dart
```diff
@@ -34,18 +34,24 @@ class BuilderOptions {
...
   ///
-  /// Values are overridden at a per-key granularity. There is no value level
-  /// merging. [other] may be null or empty, in which case this instance is
-  /// returned directly.
-  BuilderOptions overrideWith(BuilderOptions other) =>
-      other == null || other.config.isEmpty
-          ? this
-          : new BuilderOptions({}..addAll(config)..addAll(other.config));
+  /// Config values are overridden at a per-key granularity. There is no value
+  /// level merging. [other] may be null or empty, in which case this instance
```
natebosch: no longer returned directly if other is empty. (Also the meaning of "empty" might not make sense anymore)

do we want to do any other checking for the case where `other.config` is empty and `other.isRoot` is already the same as `this.isRoot`?

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -189,16 +197,20 @@ class BuilderApplication {
...
+    var phaseFactory = (PackageNode package, BuilderOptions options,
         InputSet targetSources, InputSet generateFor, bool isReleaseBuild) {
       generateFor ??= defaultGenerateFor;
 
       final optionsWithDefaults = (defaultOptions ?? BuilderOptions.empty)
           .overrideWith(
               isReleaseBuild ? defaultReleaseOptions : defaultDevOptions)
           .overrideWith(options);
+      if (package.isRoot) {
+        optionsWithDefaults.overrideWith(new BuilderOptions({}, isRoot: true));
```
natebosch: I think we need `optionsWithDefaults = optionsWithDefaults.overrideWith`...

jakemac53: yep.... good catch

build/lib/src/builder/builder.dart
jakemac53: I don't think it is worth optimizing really, these objects are cheap

build/test/builder/builder_options_test.dart
```diff
@@ -13,11 +13,26 @@ void main() {
...
       expect(overridden.config,
           {'foo': 'bar', 'baz': 'different', 'more': 'added'});
+      expect(overridden.isRoot, isFalse);
     });
 
-    test('changes nothing when overriding with empty options', () {
+    test('overrides isRoot', () {
+      var defaults = const BuilderOptions(const {}, isRoot: false);
+      var overridden =
+          defaults.overrideWith(const BuilderOptions(const {}, isRoot: true));
```
natebosch: what do you think of making this const instance a static member of the class?

`defaults.overrideWith(BuilderOptions.forRoot);`

jakemac53: sure, why not

PR #1337, opened 2018-04-24
Move PostProcessBuilder to package:build

### LINKED ISSUE #1284: Move PostProcessBuilder interfaces to package:build
We don't want angular depending on build_runner (for a non-dev dependency) but they want to implement a builder to clean up the placeholder files.

natebosch:
Closes #1284

The constructor change is technicallly breaking for
`package:build_runner` but we didn't expect anyone outside of
build_runner to construct it. The constructor is now hidden.

- Move the relevant interfaces and export them from build_runner.
- Move the relevate tests and test utilities.


build/lib/src/builder/post_process_build_step.dart
```diff
@@ -9,25 +9,30 @@ import 'package:async/async.dart';
 import 'package:build/build.dart';
 import 'package:crypto/src/digest.dart';
 
-import '../asset_graph/graph.dart';
-import '../asset_graph/node.dart';
+// This is not exported to hack around a package-private constructor.
```
grouma: Not familiar with this limitation. Can you explain further?

build/test/common/builders.dart
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+
+class CopyingPostProcessBuilder implements PostProcessBuilder {
```
grouma: [nit] I know this is a test file but doc comment?

build_runner/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 0.8.4
 
 - Log the number of completed actions on successful builds.
+- Deprecated `PostProcessBuilder` and `PosProcessBuildStep`. These should be
```
grouma: I wouldn't say deprecated. They were completely removed...

build/lib/src/builder/post_process_build_step.dart
```diff
@@ -70,26 +72,6 @@ class PostProcessBuildStep {
...
-        phaseNumber: _phaseNum,
-        wasOutput: true,
-        isFailure: false,
-        state: GeneratedNodeState.upToDate);
-    _assetGraph.add(node);
-    _anchorNode.outputs.add(id);
-  }
-
-  /// Checks if [id] is a valid output id.
-  void _checkOutput(AssetId id) {
```
grouma: I see this was moved to the build_runner. Can you explain the move?

build/lib/src/builder/post_process_build_step.dart
```diff
@@ -9,25 +9,30 @@ import 'package:async/async.dart';
...
 import 'package:crypto/src/digest.dart';
 
-import '../asset_graph/graph.dart';
-import '../asset_graph/node.dart';
+// This is not exported to hack around a package-private constructor.
+PostProcessBuildStep postProcessBuildStep(
+        AssetId inputId,
+        AssetReader reader,
+        AssetWriter writer,
+        void Function(AssetId) addAsset,
```
jakemac53: nit: please add some comments for when/how these should be used

build/lib/src/builder/post_process_build_step.dart
jakemac53: package:build doesn't know anything about asset graphs or nodes - it only has the general interfaces which can work with any build system. These implementations are build_runner specific.

build/lib/src/builder/post_process_build_step.dart
jakemac53: What is the advantage of this versus just not exporting the constructor and making it public?

natebosch: We export the type because users need to use it to implement a `PostProcessBuilder` but we don't want users outside the package to be able to instantiate one.

build/lib/src/builder/post_process_build_step.dart
natebosch: Will add them at `runPostProcessBuilder` since that is where these are exposed.

build_runner/CHANGELOG.md
natebosch: They are still `export`ed so that it's not a breaking change, but using them through a build_runner imports is what is deprecated.

PR #1341, opened 2018-04-24
default strategy for root

grouma:
- Default the strategy for root and non-root packages
- Only allow overriding the strategy for the root package
- Fix tons of typos
- Prep for release

This should close https://github.com/dart-lang/build/issues/1186

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -7,6 +7,16 @@
   connected component. You can override this behavior by providing `coarse`
   to the `strategy` option.
 
+  Example configuration:
+  ```
+  targets:
+  $default:
+    builders:
+      build_modules|modules:
+        options:
+          strategy: coarse
+  ```
+
 # 0.2.1
 
 - Give a guaranteed reverse dependency order for
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,9 +1,11 @@
 # 0.2.2
 
-- Add new `strategy` option for `ModuleBuilder`. If `coarse` is provided a
-  minimum number of modules will be created. Otherwise, a module will be
-  created for each strongly connected component. Note that `coarse` is
-  the default strategy.
+- Add new `ModuleBuilder` strategies. By default the `coarse` strategy is used
+  for all non-root packages and will create a minimum number of modules. This
+  strategy can not be overridden. However, for the root package, the `fine`
+  strategy will be used which creates a module for each strongly
+  connected component. You can override this behavior by providing `coarse`
+  to the `strategy` option.
 
 # 0.2.1
 
--- a/build_modules/build.yaml
+++ b/build_modules/build.yaml
@@ -17,7 +17,4 @@ builders:
         - .unlinked.sum
     is_optional: True
     auto_apply: all_packages
-    defaults:
-      options:
-        strategy: 'coarse'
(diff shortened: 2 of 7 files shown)


build_modules/lib/src/module_builder.dart
```diff
@@ -82,30 +82,28 @@ Future<Null> _processMeta(
...
-      case 'coarse':
-        return Strategy.course;
-      case 'fine':
-        return Strategy.fine;
-        break;
-      default:
-        throw 'Unexpected ModuleBuilder strategy: ${options.config['strategy']}';
+    var config = options.config['strategy'] as String;
+    if (config != null && config != 'coarse' && config != 'fine') {
+      throw 'Unexpected ModuleBuilder strategy: $config';
```
natebosch: can we delay this check until after the `isRoot` check? That way you can't break consumers with a bad published config.

build_modules/lib/src/module_builder.dart
```diff
@@ -82,30 +82,28 @@ Future<Null> _processMeta(
...
+    } else if (!options.isRoot || (options.isRoot && config == 'coarse')) {
```
natebosch: no need to repeat the `isRoot` check. `(!options.isRoot || config == 'coarse')`

build_modules/CHANGELOG.md
```diff
@@ -1,9 +1,11 @@
...
-- Add new `strategy` option for `ModuleBuilder`. If `coarse` is provided a
-  minimum number of modules will be created. Otherwise, a module will be
-  created for each strongly connected component. Note that `coarse` is
-  the default strategy.
+- Add new `ModuleBuilder` strategies. By default the `coarse` strategy is used
+  for all non-root packages and will create a minimum number of modules. This
+  strategy can not be overridden. However, for the root package, the `fine`
+  strategy will be used which creates a module for each strongly
+  connected component. You can override this behavior by providing `coarse`
+  to the `strategy` option.
```
jakemac53: nit: add an example of overriding this:

```yaml
targets:
  $default:
    builders:
      build_modules|modules:
        options:
          strategy: coarse
```

Note that we should actually rename the builder in the build.yaml in this package to `build_modules`, so that you don't have to do `build_modules|modules` above.

build_modules/build.yaml
```diff
@@ -17,7 +17,4 @@ builders:
         - .unlinked.sum
     is_optional: True
     auto_apply: all_packages
-    defaults:
```
jakemac53: nit: consider a release default to use the `coarse` strategy? If you do release with dartdevc you would want fewer modules.

build_modules/lib/src/module_builder.dart
```diff
@@ -82,30 +82,28 @@ Future<Null> _processMeta(
...
-    var config = (options.config['strategy'] ?? 'coarse') as String;
-    switch (config) {
-      case 'coarse':
-        return Strategy.course;
-      case 'fine':
-        return Strategy.fine;
-        break;
-      default:
-        throw 'Unexpected ModuleBuilder strategy: ${options.config['strategy']}';
+    var config = options.config['strategy'] as String;
```
jakemac53: nit: Add an error if you try to override to `fine`

build_modules/CHANGELOG.md
grouma: Done.

Issue #1308, opened 2018-04-18
META: Clean up the build directory in release mode


jakemac53:
All packages should clean up their outputs that aren't required for deployment, at least in --release mode. 

This issue will track the status of each:

* [x] `.module` and `.sum` files should be deleted by build_modules
* [x] `.dart` files should be deleted by build_web_compilers
* [x] `.ng_placeholder` files should be deleted by angular
* [x] `.js[.map]` files in the sdk should be deleted by ?
* [x] `.scss` and `.sass` files should be deleted by sass_builder
* [x] Extra dart2js outputs (`.js.deps, .js.tar.gz`, etc) should be deleted by build_web_compilers. This is a little different as it has to do more with just not extracting some of the files from the archive.
* [x] https://github.com/dart-lang/build/issues/1348

jakemac53:
cc @natebosch @kevmoo @matanlurey 

This could be considered blocked on https://github.com/dart-lang/build/issues/1284

natebosch:
@matanlurey - you'd leave the template HTML files in the output directory for release mode? We plan on removing `.dart` files from the output in release mode...

natebosch:
Waiting on a release of Angular and Sass builder but the changes are in.

Only remaining issue is the `.js` files from the SDK

jakemac53:
Random hacky but "safe" (for now) idea for the SDK js files:

The build_web_compilers|entrypoint builder could have a top level boolean field `dartdevcIsUsed` which defaults to false. It would get toggled to true inside the constructor of the builder itself, if the `BuilderOptoins` have dartdevc configured as the compiler.

The cleanup builder which deletes these files would read that top level boolean to determine if it can or cannot clean up the SDK files.

The obvious problem is if we moved to a more parallelized model for builds where we spawn builders in separate isolates or processes, then this would fall apart.

natebosch:
I'd hope to avoid communicating with global state like that. It's an anti-pattern that we want builder authors to avoid so we should model better practices.

What happens when we toggle back to a non-release build? I don't know if we'd correctly invalidate the builder because none of it's inputs will have changed.

jakemac53:
>I'd hope to avoid communicating with global state like that. It's an anti-pattern that we want builder authors to avoid so we should model better practices.

Definitely agree I just don't have any other solutions I can think of that aren't equally bad.

> What happens when we toggle back to a non-release build? I don't know if we'd correctly invalidate the builder because none of it's inputs will have changed.

[...]

PR #1350, opened 2018-04-25
Add dart_source_cleanup builder

natebosch:
Towards #1308

Enabled by default in release mode. Removes .dart source files from the
output.

### DIFF
--- a/build_web_compilers/build.yaml
+++ b/build_web_compilers/build.yaml
@@ -12,6 +12,11 @@ builders:
     auto_apply: all_packages
     required_inputs:
       - .dart
+    applies_builders:
+      # We want this to apply for dart2js as well - luckily this builder doesn't
+      # need to be disabled since it's lazy so we can use is as a handle to get
+      # the cleanup builder applied.
+      - build_web_compilers|dart_source_cleanup
   entrypoint:
     import: "package:build_web_compilers/builders.dart"
     builder_factories:
@@ -49,3 +54,9 @@ post_process_builders:
   dart2js_archive_extractor:
     import: "package:build_web_compilers/builders.dart"
     builder_factory: dart2JsArchiveExtractor
+  dart_source_cleanup:
+    import: "package:build_web_compilers/builders.dart"
+    builder_factory: dartSourceCleanup
+    defaults:
+      release_options:
+        enabled: true
--- a/build_web_compilers/lib/src/archive_extractor.dart
+++ b/build_web_compilers/lib/src/archive_extractor.dart
@@ -6,8 +6,6 @@ import 'dart:async';
 
 import 'package:archive/archive.dart';
 import 'package:build/build.dart';
-import 'package:build_runner/build_runner.dart'
-    show PostProcessBuildStep, PostProcessBuilder;
 import 'package:path/path.dart' as p;
 
 import 'web_entrypoint_builder.dart';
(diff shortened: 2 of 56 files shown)


natebosch:
Some reasons we might want to hold off on submitting this:

- We don't have a nice way to set global configuration so if someone _doesn't_ want this in release mode they'd need to use command line args to avoid it.
- Should we do this from `build_module` instead? It also has a builder applied to all packages that we could attach this to, and it would allow it to be used for free with `build_node_compilers`.

build_web_compilers/build.yaml
```diff
@@ -12,6 +12,11 @@ builders:
     auto_apply: all_packages
     required_inputs:
       - .dart
+    applies_builders:
+      # We want this to apply for dart2js as well - luckily this builder doesn't
+      # need to be disabled since it's lazy so we can use is as a handle to get
+      # the cleanup builder applied.
```
jakemac53: This could instead be set up in its own builder with `auto_appy: all_packages`? It doesn't need to be applied by another builder like this.

build_web_compilers/lib/src/archive_extractor.dart
```diff
@@ -6,8 +6,6 @@ import 'dart:async';
 
 import 'package:archive/archive.dart';
 import 'package:build/build.dart';
-import 'package:build_runner/build_runner.dart'
```
jakemac53: weird how was this not an analysis error?

build_web_compilers/lib/src/dart_source_cleanup.dart
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+
```
jakemac53: nit: add a small comment here `deletes all dart files it finds if its enabled`.

It might actually be worth creating a `FileDeleterBuilder` which takes the input extensions in the constructor, and then we can easily just add a factory method for each one we want to add? This would probably live in the `build` package in that case but its such a common thing to want to do I think that would make sense.

build_web_compilers/build.yaml
natebosch: We didn't add `auto_apply` to `PostProcessBuilders` yet because for our known use cases we had other builders we could attach them too.

I'd be fine adding support though, want me to start working on that?

build_web_compilers/lib/src/archive_extractor.dart
natebosch: It's fine to have the same class imported through two files and the `build_runner` one still exports these.

I'll probably move this to a separate PR though as cleanup.

build_web_compilers/lib/src/dart_source_cleanup.dart
natebosch: Agree on making a `FileDeletingBuilder`. Will take a look at that.

build_web_compilers/build.yaml
jakemac53: Oh, I forgot about that. I think this is fine for now then, since the same reasoning still applies?

Issue #1193, opened 2018-03-20
Indicate what actions are pending when they get 'stuck'


natebosch:
We log the number of actions completed, but once things stop moving it's not clear what actions we're waiting on.

Since this will likely be multi-line we should output output it when actions have been running longer than some threshold - maybe 60 seconds for a single action?

Dart2JS will regularly take that long for bigger projects.

jakemac53:
bumping this to P1, we really need it imo

jakemac53:
I have an alternate proposal for how it could work though - we should just have a global timer where if _no_ actions complete after some much shorter threshold (5s?) then we start logging warnings with a slow (not exponential) backoff.

This would be just as useful imo as tracking the actions individually, but should be simpler and more efficient.

PR #1352, opened 2018-04-25
Add heartbeat for stuck actions

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1193

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.8.6-dev
+## 0.8.7
 
 - If a build appears to be not making progress (no actions completed for x
   amount of time), then a warning will now be logged with the pending actions.
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.8.7
 
-- If a build appears to be not making progress (no actions completed for x
-  amount of time), then a warning will now be logged with the pending actions.
+- If a build appears to be not making progress (no actions completed for 15
+  seconds), then a warning will now be logged with the pending actions.
 
 ## 0.8.6
 
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.8.5
+
+- If a build appears to be not making progress (no actions completed for x
+  amount of time), then a warning will now be logged with the pending actions.
+
 ## 0.8.4
 
 - Log the number of completed actions on successful builds.
@@ -17,10 +22,12 @@
   different.
 
 ## 0.8.2+2
+
 - The `.packages` file is now always created in the root of the output directory
   instead of under each top level directory.
 
 ## 0.8.2+1
+
 - Bug Fix: Correctly parse Window's paths with new `--output` semantics.
 
 ## 0.8.2
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -174,8 +174,8 @@ class _SingleBuild {
   final bool _verbose;
   final RunnerAssetWriter _writer;
 
-  int numActionsCompleted = 0;
-  int numActionsStarted = 0;
+  int actionsCompletedCount = 0;
+  int actionsStartedCount = 0;
 
   final pendingActions = new SplayTreeMap<int, Set<String>>();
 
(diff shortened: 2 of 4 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -192,7 +197,25 @@ class _SingleBuild {
         _resolvers = buildImpl._resolvers,
         _resourceManager = buildImpl._resourceManager,
         _verbose = buildImpl._verbose,
-        _writer = buildImpl._writer;
+        _writer = buildImpl._writer {
+    hungActionsHeartbeat = new HungActionsHeartbeat(() {
+      final message = new StringBuffer();
+      const numActionsToLog = 5;
```
natebosch: [nit] Effective Dart calls out `numFoo` as a naming pattern to avoid.
https://www.dartlang.org/guides/language/effective-dart/design#prefer-putting-the-most-descriptive-noun-last

I think `maxActionsToLog` would work here, then `additionalActionCount` below

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -192,7 +197,25 @@ class _SingleBuild {
         _resolvers = buildImpl._resolvers,
         _resourceManager = buildImpl._resourceManager,
         _verbose = buildImpl._verbose,
-        _writer = buildImpl._writer;
+        _writer = buildImpl._writer {
+    hungActionsHeartbeat = new HungActionsHeartbeat(() {
```
natebosch: [optional] I wonder if we can move the details to another file? I see its using variables from this file so maybe it would be too much overhaul for this PR but we should consider whether we can do more to separate out the behavior of the build from the UX of the logging...

build_runner/lib/src/generate/heartbeat.dart
```diff
@@ -10,75 +10,127 @@ import '../logging/human_readable_duration.dart';
...
     if (transformLog != null) {
       message = transformLog(message);
     }
     _logger.info(message);
-    _resetHeartbeat();
+  }
+}
+
+class HungActionsHeartbeat extends Heartbeat {
+  /// An arbitraty function that will return a description of pending actions.
```
natebosch: How about `/// Returns a description of pending actions`

build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.8.7
+
+- If a build appears to be not making progress (no actions completed for x
```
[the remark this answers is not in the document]
jakemac53: updated, I was trying to leave it ambiguous as to the amount of time so we can change it more easily but it does read better and we could always update it in subsequent changelogs.

build_runner/lib/src/generate/build_impl.dart
jakemac53: I get the sentiment but I don't think its really worth doing in this case - this is just significantly simpler than the other options.

build_runner/lib/src/generate/build_impl.dart
jakemac53: That guideline isn't about the word `num` specifially, I guess what it really wants me to do is something like `actionsToLogMax` (the fact that its a maximum is the important part and should come last?).

Updated here and also refactored the `numActionsCompleted` and `numActionsToLog`.

In general this seems like an oddly specific thing to have in effective dart but im happy to conform to it :man_shrugging:.

build_runner/lib/src/generate/heartbeat.dart
jakemac53: done

build_runner/lib/src/generate/build_impl.dart
natebosch: I think long term we might want to consider pulling out all of the UX focused code from this file.

build_runner/lib/src/generate/build_impl.dart
natebosch: `num` also goes against https://www.dartlang.org/guides/language/effective-dart/design#avoid-abbreviations

😉

jakemac53: lol oh ya, the other guideline I disagree with in there lmao :grinning:

jakemac53: I would also agree that `num` is more common than `numberOf`, making it an exception to the rule ("Unless the abbreviation is more common than the unabbreviated term,"), even though it is explicitly called out in the do not section haha.

I think whoever wrote this doc just really doesn't like `num*` variables for some reason.

PR #1353, opened 2018-04-25

dineshbhosale:
`el[0].href.startsWith("/")` should be replaced with `el[0].getAttribute("href").startsWith("/")` because `el[0].href` returns an absolute path.

For example if base href is `<base href="/abc/">` then 
`el[0].href` returns `http://example.com/abc/` , so we can't perform a successful `startsWith("/")` check on `el[0].href`, whereas `el[0].getAttribute("href")` will return `/abc/` and `el[0].getAttribute("href").startsWith("/")` will return `true`.


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -317,8 +317,8 @@ var baseUrl = (function () {
   // base href should start with "/"
   if (typeof document !== 'undefined') {
     var el = document.getElementsByTagName('base');
-    if (el && el[0] && el[0].href && el[0].href.startsWith('/')){
-    	return el[0].href;
+    if (el && el[0] && el[0].href && el[0].getAttribute("href") && el[0].getAttribute("href").startsWith("/")){
```
natebosch: I'm confused why we need both `el[0].href` and `el[0].getAttribute("href")`

Should the first one be removed?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -317,8 +317,8 @@ var baseUrl = (function () {
   // base href should start with "/"
   if (typeof document !== 'undefined') {
     var el = document.getElementsByTagName('base');
-    if (el && el[0] && el[0].href && el[0].href.startsWith('/')){
-    	return el[0].href;
+    if (el && el[0] && el[0].getAttribute("href") && el[0].getAttribute("href").startsWith("/")){
+      return el[0].href;
```
natebosch: should this also be `getAttribute("href")`?

dineshbhosale: I removed unnecessary check for `el[0].href`.

dineshbhosale:
I added an entry in change log as well.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: maybe just create a local variable containing the href since its accessed 3 times?

natebosch: I'm referring to how it's inconsistent. We have two place uses `.getAttribute("href")` and one using `.href`. Why are we picking one or the other in each place?

If the `.getAttribute("href")` pattern is for when we aren't sure if it exists, and `.href` pattern is for when we know it exists, then we should use `.href.startsWith("/")` above since in that location we know it exists...

If there is some other condition making us choose one pattern or the other - what is that condition?

dineshbhosale: @natebosch 
`el[0].href` returns `http://example.com/abc/` , so we can't perform a successful `startsWith("/")` check on `el[0].href`, whereas `el[0].getAttribute("href")` will return `/abc/` and `el[0].getAttribute("href").startsWith("/")` will return `true`.

dineshbhosale: @jakemac53 
`el[0].getAttribute("href")` is used twice while checking condition
`el[0].href` is used only once while returning value.

Here I'm trying to make sure it works and doesn't fail.

dineshbhosale:
```js
var baseUrl = (function () {
  // base href should start with "/"
    if (el && el[0] && el[0].getAttribute("href") && el[0].getAttribute("href").startsWith("/")){
      return el[0].href;
    }
  }
  if (pathParts[0] == "") {
    pathParts.shift();
  }
  var baseUrl;
  }
  return "/";
}());
```

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
natebosch: Thanks for the explanation!

dineshbhosale:
Difference between `el[0].href.startsWith('/')` and `el[0].getAttribute("href").startsWith("/")`

screenshot_2018-04-27_19-17-19

PR #1357, opened 2018-04-25
Run with headless

grouma:
This will speed up our E2E tests slightly since https://github.com/dart-lang/test/issues/772 is now closed.

### DIFF
--- a/e2e_example/dart_test.yaml
+++ b/e2e_example/dart_test.yaml
@@ -6,4 +6,4 @@ concurrency: 1
 override_platforms:
   chrome:
     settings:
-      headless: false
+      headless: true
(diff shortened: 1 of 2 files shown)


natebosch:
Can we get rid of the xvfb stuff?

https://github.com/dart-lang/build/blob/0fdde9abb6bacbd8daf02806217278f27dd5b44d/.travis.yml#L11-L15

grouma:
I believe we can remove the xvfb.

e2e_example/dart_test.yaml
```diff
@@ -6,4 +6,4 @@ concurrency: 1
 override_platforms:
```
natebosch: I think we can remove the section altogether since headless is the default?

PR #1360, opened 2018-04-25
Add PostProcessBuilder utilities

natebosch:
- Add FileDeletingBuilder since we're likely to have multiple
  implementations which vary only in their input extensions.
- Add NoOp builder as a shorcut for PostProcessBuilders which can be
  disabled by options.
- Add the typedef PostProcessBuilderFactory for consistency.


natebosch:
I'm less sure about having a utility for checking an enabled field in options but mostly because I wasn't sure what I'd want the usage to look like.

What do you think about:

```dart
PostProcessBuilder someOptionalBuilder(BuilderOptions options) =>
    PostProcessBuilder
        .whenEnabled((options) => new SomeOptionalBuilder(options));
```

natebosch:
I'm also unsure about how often the "enabled" key is going to be the deciding factor.

For instance I was just considering make the `FileDeletingBuilder` automatically get applied to all packages with an empty input extensions (which make the factory want to switch to the no op builder to avoid of worthless empt instances). Then you could configure it like this:

```yaml
targets:
  $defaults:
    builders:
      build|file_cleanup:
        extensions: [".yaml"]
```

jakemac53:
> PostProcessBuilder someOptionalBuilder(BuilderOptions options) =>
>     PostProcessBuilder
>         .whenEnabled((options) => new SomeOptionalBuilder(options));

I think I like that, but it would have to take file extensions too right?

> For instance I was just considering make the FileDeletingBuilder automatically get applied to all packages with an empty input extensions (which make the factory want to switch to the no op builder to avoid of worthless empt instances). Then you could configure it like this:

I think we decided we don't want to require users to provide this explicitly, and it would have to be applied globally which we can't do right now.

natebosch:
> and it would have to be applied globally which we can't do right now.

true, though I think this might be the use case that makes us want to add it. The motivating use case for me is being able to clean up [this file](https://github.com/dart-lang/build/blob/master/build_modules/lib/src/analysis_options.default.yaml) by default for people since even when we remove Dart sources we end up dragging this into the packages directory when it's a build time only file.

natebosch:
hmm, I also need to look at whether the `noOp` thing will invalidate the asset graph - how do we handle that case, where the `inputExtensions` change between builds?

natebosch:
We use use the runtime type of the builder so it would invalidate things to switch to a no-op builder based on options...

https://github.com/dart-lang/build/blob/0fdde9abb6bacbd8daf02806217278f27dd5b44d/build_runner/lib/src/generate/phase.dart#L193

natebosch:
Yeah I think we're definitely not ready for the input extensions to change based on options...

jakemac53:
> Yeah I think we're definitely not ready for the input extensions to change based on options...

Ya so instead of using the NoOp builder just make the FileDeletingBuilder know how to skip files?

natebosch:
>  just make the FileDeletingBuilder know how to skip files

That's what I'll do for now.

Issue #1200, opened 2018-03-23
Warn or fail if any command is run with 0 builders possible


natebosch:
If I add a dependency on `build_runner` today and nothing else that exposes builders and run a build it will "successfully" do a build with 0 outputs. We should either warn or fail in this case.

cc @kevmoo

kevmoo:
What about `serve` and/or build/watch with -o ?

natebosch:
Yeah `serve` and `-o` are the two cases where there is _some_ possibility of a real use case.

I'm somewhat inclined to say we don't support those and see if we get any bugs about it since it's just easier to not have special cases in our check.

The other thing that makes this slightly lower value than I had originally thought is we can expect the vast majority of packages to have a dep on package:test and it won't be too long before that ships a builder...

kevmoo:
ack. Easier to get less restrictive later...

PR #1362, opened 2018-04-25
Fail on empty build

grouma:
Closes https://github.com/dart-lang/build/issues/1200

### DIFF
--- a/build_runner/lib/src/generate/build_impl.dart
+++ b/build_runner/lib/src/generate/build_impl.dart
@@ -92,7 +92,13 @@ Future<BuildResult> build(
   final buildPhases = await createBuildPhases(
       targetGraph, builders, builderConfigOverrides, isReleaseBuild ?? false);
 
-  var result = await singleBuild(environment, options, buildPhases);
+  BuildResult result;
+  if (buildPhases.isEmpty) {
+    _logger.severe('Nothing was built, yet a build was requested.');
+    result = new BuildResult(BuildStatus.failure, []);
+  } else {
+    result = await singleBuild(environment, options, buildPhases);
+  }
 
   await terminator.cancel();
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -92,7 +92,13 @@ Future<BuildResult> build(
   final buildPhases = await createBuildPhases(
       targetGraph, builders, builderConfigOverrides, isReleaseBuild ?? false);
 
-  var result = await singleBuild(environment, options, buildPhases);
+  BuildResult result;
+  if (buildPhases.isEmpty) {
+    _logger.severe('Nothing was built, yet a build was requested.');
```
natebosch: maybe "Nothing can be built, yet a build was requested."

We also want a similar check in watch mode around here: https://github.com/dart-lang/build/blob/0fdde9abb6bacbd8daf02806217278f27dd5b44d/build_runner/lib/src/generate/watch_impl.dart#L92

grouma: Added the check. Should we exit as well? My initial thought is no.

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -92,6 +92,10 @@ Future<ServeHandler> watch(
   final buildPhases = await createBuildPhases(
       targetGraph, builders, builderConfigOverrides, isReleaseBuild ?? false);
 
+  if (buildPhases.isEmpty) {
+    _logger.severe('Nothing can be built, yet a build was requested.');
```
natebosch: can we bail out too?

grouma:
Had to fix some tests. PTAL.

Issue #1358, opened 2018-04-25
Adding new entrypoints crashes the build


jakemac53:
This looks related to the recent module changes, you can repro by running the `test/test_integration_test.dart` in the e2e_example package.

PR #1363, opened 2018-04-25
Make sure we don't output meta module files unless in coarse mode

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1358

### DIFF
--- a/build_modules/lib/src/module_builder.dart
+++ b/build_modules/lib/src/module_builder.dart
@@ -7,6 +7,7 @@ import 'dart:convert';
 
 import 'package:build/build.dart';
 
+import 'common.dart';
 import 'meta_module.dart';
 import 'meta_module_clean_builder.dart';
 import 'modules.dart';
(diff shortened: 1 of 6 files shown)


build_modules/lib/src/module_builder.dart
```diff
@@ -122,14 +105,14 @@ class ModuleBuilder implements Builder {
...
-    var asset =
+    var cleanMetaAsset =
         new AssetId(buildStep.inputId.package, 'lib/$metaModuleCleanExtension');
     // If we can't read the clean meta module it is likely that this package
     // is in a module cycle so fall back to the fine strategy.
-    if (await buildStep.canRead(asset) && _isCoarse) {
-      if (!readMetas.contains(asset)) {
-        await _processMeta(
-            buildStep, asset, primaryToClean, assetToPrimary, readMetas);
+    if (await buildStep.canRead(cleanMetaAsset) && _isCoarse) {
```
grouma: Discussed offline. The issue that the file was getting created is because of this conditional. I will follow up with a PR that reverses the check.

PR #1364, opened 2018-04-25
Avoid using Builder.runtimeType

natebosch:
It should be legal for a BuilderFactory to pick a concrete subclass to
return based on the BuilderOptions.

- Change logging/toString to always use the builder label since that is
  what we care about anyway.
- Change the identity to hash the build extensions since that is what
  would actually invalidate the AssetGraph and we don't otherwise have
  anything preventing the same class of Builder from changing it's
  extensions based on options.

### DIFF
--- a/build_runner/lib/src/generate/phase.dart
+++ b/build_runner/lib/src/generate/phase.dart
@@ -117,14 +117,14 @@ class InBuildPhase extends BuildPhase implements BuildAction {
     final settings = <String>[];
     if (isOptional) settings.add('optional');
     if (hideOutput) settings.add('hidden');
-    var result = '${builder.runtimeType} on $targetSources in $package';
+    var result = '$builderLabel on $targetSources in $package';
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/generate/phase.dart
```diff
@@ -117,14 +117,14 @@ class InBuildPhase extends BuildPhase implements BuildAction {
...
-    var result = '${builder.runtimeType} on $targetSources in $package';
+    var result = '$builderLabel on $targetSources in $package';
     if (settings.isNotEmpty) result += ' $settings';
     return result;
   }
 
   @override
   int get identity => _deepEquals.hash([
-        '${builder.runtimeType}',
+        builder.buildExtensions,
```
jakemac53: We should probably also include the `builderLabel`?

natebosch: we may as well. Done

Issue #1372, opened 2018-04-26
GraphViz: handle bad queries better


kevmoo:
https://github.com/dart-lang/build/blob/c4c9d10769cb783873bc4b199e7d0b6893260d31/build_runner/lib/src/server/graph_viz_main.dart#L26

At the moment, if there is no `|` in the string, it just crashes.

PR #1373, opened 2018-04-26
Log some semi-helpful messages if the graph page fails a query request

kevmoo:
Fixes https://github.com/dart-lang/build/issues/1372


build_runner/lib/src/server/graph_viz_main.dart
```diff
@@ -15,21 +16,42 @@ main() async {
   searchForm.onSubmit.listen((e) async {
     e.preventDefault();
     var query = searchBox.value;
-    _focus(query);
+    await _focus(query);
```
natebosch: could also make the method not async instead

build_runner/lib/src/server/graph_viz_main.dart
```diff
@@ -15,21 +16,42 @@ main() async {
...
+  String requestPath;
+  try {
+    var parts = query.split('|');
+    var package = parts[0];
+    var path = parts[1];
+    requestPath = path.startsWith('lib/')
+        ? 'packages/$package/${path.substring(4)}'
+        : '${path.split('/').skip(1).join('/')}';
+  } catch (e, stack) {
+    _error('The query you provided ($query) could not be parsed.', e, stack);
```
grouma: What do the queries look like here? Should we wrap it in back tics?

build_runner/lib/src/server/graph_viz_main.dart
kevmoo: true!

build_runner/lib/src/server/graph_viz_main.dart
kevmoo: It's just an asset ID – I think. The parens should be fine.

Issue #1375, opened 2018-04-26
WARNING: No actions completed, waiting on build_web_compilers|ddc


chalin:
Attempting to build most of the example apps results in the following (e.g., see this [Travis log](https://travis-ci.org/dart-lang/site-webdev/jobs/371725070)):

```nocode
architecture > pub run build_runner build --fail-on-severe --delete-conflicting-outputs --output=build
[INFO] Generating build script completed, took 309ms
[INFO] Building new asset graph completed, took 677ms
[INFO] Checking for unexpected pre-existing outputs. completed, took 1ms
[WARNING] No actions completed for 15.1s, waiting on:
  - build_web_compilers|ddc on package:analyzer/analyzer.module
  - build_web_compilers|entrypoint on web/main.dart

[INFO] Running build completed, took 41.9s
[INFO] Caching finalized dependency graph completed, took 118ms
[INFO] Creating merged output dir `build` completed, took 1.5s
[INFO] Writing asset manifest completed, took 1ms
[INFO] Succeeded after 43.7s with 1289 outputs (1724 actions)
chalin@CC5K16 Thu Apr 26 16:26:14 ~/git/wd/examples/ng/doc/architecture (chalin-2-dev-50-sdk-0425)
```

Environment:
- **Dart VM version: 2.0.0-dev.50.0** on "macos_x64"

It was working fine under Dart 2-dev.49.

cc @kwalrath

natebosch:
I think this is more likely to be caused by a change in `build_modules` thank in the SDK. @matanlurey was also seeing some problems with trying to run DDC on the analyzer package which I think we don't normally expect...

I wonder if the course modules are triggering some condition in angular which is grouping compiler code in with runtime code or something.

cc @grouma

grouma:
What exactly are you trying to build? I have tried to build a handful of examples and I can't reproduce the problem with Dart VM version: 2.0.0-dev.50.0 on "macos_x64"

grouma:
Scratch that. I was able to reproduce the problem building hacker_news_pwa on
Dart VM version: 2.0.0-dev.50.0 on "macos_x64"

but not:
Dart VM version: 2.0.0-dev.50.0 (Unknown timestamp) on "linux_x64"

chalin:
[...]

PR #1378, opened 2018-04-27
Correctly create shared modules

grouma:
Closes https://github.com/dart-lang/build/issues/1375

We weren't creating a shared module for resources that were used from multiple entry points. Now create a module for those resources.


kevmoo:
Just tried this on hacker news. Dropped 264 items from build.manefest

clean build (with DDC) went from 58 to 19 seconds!

PR #1379, opened 2018-04-27
Sort contents of the .build.manifest and bootstrap files

kevmoo:


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -58,7 +59,8 @@ Future<Null> bootstrapDdc(BuildStep buildStep,
   appModuleScope = appModuleScope.replaceAll('.', '\$46');
 
   // Map from module name to module path for custom modules.
-  var modulePaths = {'dart_sdk': 'packages/\$sdk/dev_compiler/amd/dart_sdk'};
+  var modulePaths = new SplayTreeMap<String, String>.from(
```
natebosch: [nit] I think you could drop the `<String, String>` if you change `from` to `of`

kevmoo: indeed!

PR #1380, opened 2018-04-27
Meta module tweaks

natebosch:
- Avoid `(toList()..sort()).first`, use `reduce(_min)` instead.
- Add a `merge` method to `Module` so it can be used from multiple
  places.
- Don't make copies of collections that are unreferenced later.
- Use `Iterable<Module>` where possible instead of `Set<AssetId>` or
  `Map<AssetId, Module>`
- Keep separate collections for the different varieties of modules and
  concatenate them after merging.
- Change some for loops to `map` when we care more about the collection
  than the side effects.

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -99,48 +99,32 @@ class _AssetNode {
   }
 }
 
-/// Creates simple modules based strictly off of [connectedComponents].
+/// Creates a module based strictly off of a strongly connected component of
+/// asset nodes.
 ///
 /// This creates more modules than we want, but we collapse them later on.
-Map<AssetId, Module> _createModulesFromComponents(
-    Iterable<List<_AssetNode>> connectedComponents) {
-  var modules = <AssetId, Module>{};
-  for (var componentNodes in connectedComponents) {
-    // Name components based on first alphabetically sorted node, preferring
-    // public srcs (not under lib/src).
-    var sortedNodes = componentNodes.toList()
-      ..sort((a, b) => a.id.path.compareTo(b.id.path));
-    var primaryNode = sortedNodes.firstWhere(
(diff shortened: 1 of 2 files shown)


build_modules/lib/src/meta_module.dart
```diff
@@ -167,68 +151,86 @@ Set<AssetId> _localTransitiveDeps(
...
 ///
-///   * If it is an entrypoint module, skip it.
-///   * Else merge it into a module whose name is a the entrypoint
-///   that import it (create that module if it doesn't exist).
-List<Module> _mergeModules(
-    Map<AssetId, Module> originalModulesById, Set<AssetId> entrypoints) {
-  var modulesById = new Map<AssetId, Module>.from(originalModulesById);
+///   * If it is an entrypoint module do not merge it.
+///   * If it is not depended on my any entrypoint do not merge it.
+///   * If it is depended on by on entrypoint merge it into the entrypoint
```
grouma: typo

build_modules/lib/src/meta_module.dart
```diff
@@ -167,68 +151,86 @@ Set<AssetId> _localTransitiveDeps(
...
+///     modules
+///   * Else merge it into with others that are depended on by the same set of
+///     entrypoints
```
grouma: Dart highlighters act funny if you indent past the bullet point. Keep this under the bullet.

build_modules/lib/src/meta_module.dart
natebosch: Done

build_modules/lib/src/meta_module.dart
natebosch: Done

natebosch:
Pushing on red because the tests are flaky but this commit is not introducing the flakes and this makes the code easier to understand while we are debugging.

PR #1384, opened 2018-04-27
Build in verbose mode for integration tests

natebosch:


natebosch:
Pushing on red because it's not introducing new failures and it makes the flakes easier to diagnose.

PR #1385, opened 2018-04-27
Cache only complete modules

natebosch:
Previous there was a race condition between updating `readMetas` and
populating `assetToPrimary` so we could attempt to pull out data that
was not ready yet and miss out on modules that should exist.

This changes the usage of `Resource` to include only data that will
never change, and includes access checks to make sure we only read data
that is allowed and tracked by the build system.

### DIFF
--- a/build_modules/lib/src/module_builder.dart
+++ b/build_modules/lib/src/module_builder.dart
@@ -46,6 +46,7 @@ class _CleanMetaModuleCache {
 Future<Module> _cleanModuleDeps(
     BuildStep buildStep, Module module, _CleanMetaModuleCache cache) async {
   var cleanedDeps = new Set<AssetId>();
+  var depAssetToModules = <AssetId, Module>{};
   for (var dep in module.directDependencies) {
     // Since we are not using the course strategy we can safely add
     // all dependencies in the same package as they will have a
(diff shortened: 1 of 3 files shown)


build_modules/lib/src/module_builder.dart
```diff
@@ -55,34 +61,19 @@ Future<Module> _cleanModuleDeps(
     }
     // The dep must have come from a coarse module. Look for the corresponding
     // primary source.
-    var depAsset = new AssetId(dep.package, 'lib/$metaModuleCleanExtension');
-    if (await buildStep.canRead(depAsset) && !readMetas.contains(depAsset)) {
-      await _processMeta(
-          buildStep, depAsset, primaryToClean, assetToPrimary, readMetas);
+    var metaModule = await cache.find(dep.package, buildStep);
+    if (metaModule == null) {
+      // TODO amek a new exception class for this
```
grouma: typo

natebosch:
I added back a fallback to recompute library cycles for dependencies if they haven't already output individual modules. This should be similar to what we used to do with fine grained modules and package cycles.

build_modules/lib/src/module_builder.dart
```diff
@@ -53,36 +60,28 @@ Future<Module> _cleanModuleDeps(
...
-    // primary source.
-    var depAsset = new AssetId(dep.package, 'lib/$metaModuleCleanExtension');
-    if (await buildStep.canRead(depAsset) && !readMetas.contains(depAsset)) {
-      await _processMeta(
-          buildStep, depAsset, primaryToClean, assetToPrimary, readMetas);
+    var metaModule = await cache.find(dep.package, buildStep);
+    if (metaModule == null) {
+      // The dep is also fine but it's individual modules will come in a later
+      // phase. Need to recompute them.
+      if (!depAssetToModules.containsKey(dep)) {
```
grouma: Any good way to test this?

natebosch: Hopefully...

Will follow up.

PR #1386, opened 2018-04-27
Add excludes to FileDeletingBuilder

natebosch:
Towards #1308

Angular will want to use this builder, but will need the ability to
exclude certain files from being removed.

See discussion at https://github.com/dart-lang/angular/issues/1282


natebosch:
hmm, maybe this could be accomplished with just `generate_for`... should I walk this back?

natebosch:
Ah right. Makes sense.

PR #1391, opened 2018-04-28
Cleanup /$graph

kevmoo:
Make the query syntax a bit more clear, support paths and asset IDs


build_runner/test/server/serve_integration_test.dart
```diff
@@ -108,6 +109,74 @@ example:file://fake/pkg/path
...
+        expect(json, containsPair('primary', containsPair('id', expectedId)));
+      });
+    }
+
+    testSuccess('valid path', '?q=web/initial.txt', 'example|web/initial.txt');
+    testSuccess('valid path, leading slash', '?q=/web/initial.txt',
+        'example|web/initial.txt');
+    testSuccess('valid path, assuming root', '?q=initial.txt',
+        'example|web/initial.txt');
+    testSuccess('valid path, assuming root, leading slash', '?q=/initial.txt',
```
jakemac53: Should we treat a leading slash as forced to the root of the package?

kevmoo: 🤷‍♂️  – I'm thinking of a user copy-pasting from their browser.

Since we actually print the AssetId that is understood, I'm not sure it's needed...

PR #1393, opened 2018-04-29
Add AssetId.pathSegments

kevmoo:


kevmoo:
Not very consistent about using a non-Windows split. Good to generalize this...

```
build_runner/lib/src/generate/create_merged_dir.dart:    var parts = p.url.split(node.id.path);
build_runner/lib/src/generate/create_merged_dir.dart:        p.join('packages', id.package, p.joinAll(p.url.split(id.path).skip(1)));
build_test/lib/src/debug_test_builder.dart:      final path = p.joinAll(p.split(_debugHtmlId(test).path)..removeAt(0));
build_web_compilers/lib/src/dev_compiler_bootstrap.dart:            : _context.joinAll(_context.split(jsId.path).skip(1)));
build_web_compilers/lib/src/dev_compiler_builder.dart:    var libraryRoot = '/${p.split(p.dirname(jsId.path)).first}';
scratch_space/lib/src/scratch_space.dart:        p.url.join(id.package, p.url.joinAll(p.url.split(id.path).skip(1)));
scratch_space/test/scratch_space_test.dart:    var idNum = int.parse(p.split(id.path).last);
```

build/lib/src/asset/id.dart
```diff
@@ -19,6 +19,9 @@ class AssetId implements Comparable<AssetId> {
   /// the host platform.
   final String path;
 
+  /// Splits [path] into its components.
+  List<String> get pathSegments => p.posix.split(path);
```
jakemac53: Generally we use `p.url`, I don't know where that is different but it would be good to be consistent.

PR #1396, opened 2018-04-30
Invalidate nodes after graph structure changes

### LINKED ISSUE #1354: Failed actions might be getting stuck in the asset graph
https://gitter.im/dart-lang/angular?at=5ae0b11e2d0e228d7baff65b

If the build fails such that a generated asset node gets marked as a failure, then the build structure changes such that we won't attempt to run that action anymore (but it's still in the graph), we might not be cleaning things up properly.

natebosch:
Fixes #1354

### DIFF
--- a/build_runner/lib/src/asset_graph/graph.dart
+++ b/build_runner/lib/src/asset_graph/graph.dart
@@ -234,7 +234,6 @@ class AssetGraph {
       String rootPackage,
       Future delete(AssetId id),
       AssetReader digestReader) async {
-    var invalidatedIds = new Set<AssetId>();
 
     var newIds = new Set<AssetId>();
     var modifyIds = new Set<AssetId>();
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -191,7 +191,9 @@ class AssetGraph {
       var builderOptionsNode = get(node.builderOptionsId);
       builderOptionsNode.outputs.remove(id);
     }
-    _nodesByPackage[id.package].remove(id.path);
+    if (node is! SyntheticSourceAssetNode) {
```
grouma: Add a comment here for why we do this.

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -309,6 +292,24 @@ class AssetGraph {
         _addOutputsForSources(buildPhases, newIds, rootPackage)
           ..addAll(transitiveRemovedIds);
 
+    // Transitively invalidates all assets.
```
grouma: Optionally add a comment here indicating that this must be done after updates to the graph are completed.

build_runner/lib/src/asset_graph/graph.dart
```diff
@@ -191,7 +191,10 @@ class AssetGraph {
       var builderOptionsNode = get(node.builderOptionsId);
       builderOptionsNode.outputs.remove(id);
     }
-    _nodesByPackage[id.package].remove(id.path);
+    // Synthetic nodes need to be kept to retain dependency tracking.
+    if (node is! SyntheticSourceAssetNode) {
+      _nodesByPackage[id.package].remove(id.path);
```
jakemac53: Note that we are still adding thise node to `removedIds`... is that correct?

natebosch: Good question...

We apparently also explicitly intended for SyntheticSourceAssetNodes to be removed:

https://github.com/dart-lang/build/blob/b7bea8daa7f8fecc866e2b98b20edeb513a6c548/build_runner/lib/src/asset_graph/graph.dart#L449-L451

We count on these removed nodes being tracked so we can remove them from the set of assets consider inputs through a series of phases

https://github.com/dart-lang/build/blob/b7bea8daa7f8fecc866e2b98b20edeb513a6c548/build_runner/lib/src/asset_graph/graph.dart#L419

`allInputs` there starts as the set of sources in the environment, not in the graph, so the synthetic nodes wouldn't be in that set. I _think_ this is safe as is.

PR #1398, opened 2018-05-01
Track required outputs by following output chains

### LINKED ISSUE #1397: Failed optional outputs that aren't required in subsequent builds make the overall build appear failed
This is a different issue than the one fixed in https://github.com/dart-lang/build/issues/1354

This is a close match to the issue in https://github.com/dart-lang/build/pull/1034

natebosch:
Fixes #1397

Add `OptionalOutputTracker` to cache the status of which outputs from
optional phases are required. Use the tracker to check which failed
outputs were required during the current build. Also pass the tracker
`createMergedOutputDirectories` and replace the logic which crawled the
other direction.


natebosch:
I took a look at putting this logic in `FinalizedAssetReader` and using that from `createMergedOutputDirectories` but it was snowballing a bit so I held off for now.

If we do that though this will close out #1033 since it would be reflected by the server as well.

build_runner/lib/src/asset_graph/optional_output_tracker.dart
```diff
@@ -0,0 +1,38 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:build/build.dart';
+
+import '../generate/phase.dart';
+import 'graph.dart';
+import 'node.dart';
+
+class OptionalOutputTracker {
```
grouma: Doc comment here and below.

build_runner/lib/src/asset_graph/optional_output_tracker.dart
```diff
@@ -0,0 +1,38 @@
...
+    if (node is! GeneratedAssetNode) return true;
+    final generatedNode = node as GeneratedAssetNode;
+    final phase = _buildPhases[generatedNode.phaseNumber];
+    if (!phase.isOptional) return true;
+    return _checkedOutputs.putIfAbsent(
+        output,
+        () =>
+            generatedNode.outputs
+                .any((o) => isRequired(o, currentlyChecking)) ||
+            _assetGraph
```
jakemac53: I don't quite get this logic - why do we care about other outputs in the phase?

jakemac53: Oh - I think this is trying to handle the additional outputs of the same action?

Right now I think it is checking all outputs of the same phase which is a lot more than we actually want I believe. We could add an additional `where` after `outputsForPhase` that checks if the primary input is the same?

build_runner/lib/src/asset_graph/optional_output_tracker.dart
natebosch: Done

build_runner/lib/src/asset_graph/optional_output_tracker.dart
natebosch: https://github.com/dart-lang/build/pull/1034/commits/6dda418fa6879656c73cbcf7cc67da6414037edc

https://github.com/dart-lang/build/pull/1034#issuecomment-367117339

natebosch:
Added a regression test and doc comments. PTAL

build_runner/lib/src/asset_graph/optional_output_tracker.dart
natebosch: Ah I see, yeah that makes sense. Will switch to only those with the same primary input

jakemac53: We still need to either do the `expectedOutputs` trick I did there or add a `where` after the `outputsToPhase` though to retain the old logic right?

natebosch: yeah added a where clause to check the primary inputs

jakemac53: Lol I really wish github updated comments without having to refresh lol

jakemac53:
This should make finishing https://github.com/dart-lang/build/issues/1033 easier as well yes?

natebosch:
Yes I believe that gets a lot easier to fix. Just need to decide how to structure it. There isn't a clear way to pass the `OptionalOutputTracker` through to the `FinalizedAssetReader` (or to the server handler directly) so we'll need to consider exactly whether we're OK with multiple instances, or if we want to add a field to `BuildResult` or something.

natebosch:
> or we could expose a `wasRequired` method on `BuildResult` which takes an `AssetId`

I was thinking I'd like to make a `FinalizedAssetReader` field on `BuildResult`, use `OptionalOutputTracker` in the reader (or just move the logic there), and make `createMergedOutputDirectories` use a `FinalizedAssetReader`.

I have been playing around a bit but I'm not really happy with it yet. I didn't want that work to block getting this fixed so I figured we could submit as is and refactor soon.

e2e_example/test/build_integration_test.dart
```diff
@@ -75,4 +76,30 @@ void main() {
...
+        () async {
+      // Run a build with DDC that should fail
+      final path = p.join('test', 'common', 'message.dart');
+      await replaceAllInFile(path, "'Hello World!'", 'can not parse this');
+      final result = await runAutoBuild(trailingArgs: ['--fail-on-severe']);
+      expect(result.exitCode, isNot(0));
+      expect(result.stderr, contains('Failed'));
+
+      // Run a build with dart2js that should also fail. Makes the failing DDC
+      // outptus no longer necessary for the build but does not invalidate them
```
jakemac53: s/outptus/outputs

e2e_example/test/build_integration_test.dart
```diff
@@ -75,4 +76,30 @@ void main() {
...
+    test('Failing optional outputs which are required during the next build',
+        () async {
+      // Run a build with DDC that should fail
+      final path = p.join('test', 'common', 'message.dart');
+      await replaceAllInFile(path, "'Hello World!'", 'can not parse this');
+      final result = await runAutoBuild(trailingArgs: ['--fail-on-severe']);
+      expect(result.exitCode, isNot(0));
+      expect(result.stderr, contains('Failed'));
+
+      // Run a build with dart2js that should also fail. Makes the failing DDC
```
jakemac53: Not super stoked on relying on the internals here... also note that the ddc output is going to get marked as needing an update as well which might change the logic or hit different codepaths (not sure how we treat failures that didn't rerun).

I would instead create a new file with an error in it (but not imported anywhere) and expect as successful build, then add an import to it and expect a failure, then remove the import again and expect it to pass.

natebosch: As a bonus it's also much faster :)

Skipping the check for the successful build since I don't think it changes the meaning of the test.

PR #1404, opened 2018-05-02
Undelete files that were previously deleted

### LINKED ISSUE #1402: PostProcessBuilder deletes are not invalidated
If I run a build with a PostProcessBuilder deleting an asset, then change the builder options so that it won't delete, and run another build then the output is still not present.

Somewhere around here where we are cleaning up the previous effects of the builder we'll need to "undelete": https://github.com/dart-lang/build/blob/36cbd8a5108654e7862cf3a1caba50425ac35c26/build_runner/lib/src/generate/build_impl.dart#L537

This gets tricky though because an asset could be deleted by more than one post process builder.

I think what we need to do is:

- Change `isDeleted` bool to a Set of builders which requested it to be deleted. Add a getter for `isDeleted` that returns whether this set is empty
- Add a `deleted` set to `PostProcessBuilderAnchorNode` so that we can trace what needs to be cleaned up. Remove the right value from the set during cleanup.

We can simplify step 2 with the assumption that _only_ the primary input can be deleted but it might be a source for bugs if we ever change that limitation.

natebosch:
Fixes #1402

### DIFF
--- a/_test/test/build_integration_test.dart
+++ b/_test/test/build_integration_test.dart
@@ -96,5 +96,17 @@ void main() {
       final nextBuild = await runBuild(trailingArgs: ['--fail-on-severe']);
       expect(nextBuild.exitCode, 0);
     });
+
+    test(
+        'Restores previously deleted outputs if they are not deleted in subsequent builds',
+        () async {
+      final dartSource =
+          new File(p.join('build', 'web', 'packages', '_test', 'app.dart'));
+      await runBuild(trailingArgs: ['--release', '--output', 'build']);
+      expect(dartSource.existsSync(), false);
+
+      await runBuild(trailingArgs: ['--output', 'build']);
+      expect(dartSource.existsSync(), true);
+    });
   });
 }
--- a/_test/test/build_integration_test.dart
+++ b/_test/test/build_integration_test.dart
@@ -102,7 +102,11 @@ void main() {
         () async {
       final dartSource =
           new File(p.join('build', 'web', 'packages', '_test', 'app.dart'));
-      await runBuild(trailingArgs: ['--release', '--output', 'build']);
+      await runBuild(trailingArgs: [
+        '--define=build_web_compilers|dart_source_cleanup=enabled=true',
+        '--output',
+        'build'
+      ]);
       expect(dartSource.existsSync(), false);
 
       await runBuild(trailingArgs: ['--output', 'build']);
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/generate/build_impl.dart
```diff
@@ -568,7 +569,11 @@ class _SingleBuild {
       if (!_assetGraph.contains(assetId)) {
         throw new AssetNotFoundException(assetId);
       }
-      _assetGraph.get(assetId).isDeleted = true;
+      if (assetId != input) {
```
natebosch: I had written the interface for `deleteAsset` in what I had hoped was a future proof way. `PostProcessBuildStep` is safe - it only lets you delete the primary input. The interface we pass through is more general since I thought we could avoid future breaking changes if we did want to allow deleting other assets.

If we want we can clean this up with a point release of `package:build` and change `deleteAsset` to `deletePrimaryInput` with no arguments. Otherwise we can leave it as is and if we want to allow deleting other assets through the build step we can change the way we track deletes here.

_test/test/build_integration_test.dart
```diff
@@ -96,5 +96,17 @@ void main() {
       final nextBuild = await runBuild(trailingArgs: ['--fail-on-severe']);
       expect(nextBuild.exitCode, 0);
     });
+
+    test(
+        'Restores previously deleted outputs if they are not deleted in subsequent builds',
```
natebosch: I went ahead and wrote a regression test but my guess is @jakemac53  won't like it 😝 

I'm also not covering the case where multiple post process builders want to delete the same asset with any test.

We might need to bite the bullet and start writing tests with test_descriptor so we can have isolated little sample packages...

grouma: Note sure what's wrong with the test. Looks fine to me :/

natebosch: It has some of the same downsides as https://github.com/dart-lang/build/pull/1398#discussion_r185284625 though maybe not quite as drastic

The main thing I don't like about it is having to build with Dart2Js again which slows down the tests.

The test would be faster, and would be a more direct expression of what we're covering, if we make some builders that explicitly do or don't delete their outputs and change only that. The Dart source cleanup builder does that - but it's also tied to other details.

build_runner/lib/src/generate/build_impl.dart
jakemac53: I think this is fine, we may very well want to allow deleting other assets in the future. I have also come up with at least one  reasonable use cases for reading other assets as well.

jakemac53:
I think we could do this without adding additional information to the asset graph.

Instead of keeping track of all the things that deleted a file, we could simply clear out the `isDeleted` field for all assets that were updated prior to running the post process builders. Those builders will need to rerun anyways and would reset it to `isDeleted`.

PR #1405, opened 2018-05-02
Drop dependency overrides

natebosch:


natebosch:
Couldn't pub publish with these still here.

kevmoo:
...have the lower constraints been updated as needed?

natebosch:
Yes, they were updated when adding the overrides:

https://github.com/dart-lang/build/pull/1350/files#diff-8c687f66dd166104d209dd9440bd6608
https://github.com/dart-lang/build/pull/1355/files#diff-8c687f66dd166104d209dd9440bd6608

PR #1408, opened 2018-05-02
Delete empty directories in output

### LINKED ISSUE #1348: Don't create a package directory if there are no files under the package
If we have a PostProcessBuilder that remove dart source files a package in the output could become empty. We should probably just not write it at all.

natebosch:
Closes #1348


natebosch:
This is pretty naive, performance doesn't seem _terrible_ but I'm worried about what it might look like on windows.

Should we take the time to collect up a Set of directories that may have become empty and avoid the repeated checks for whether they are empty after every file?

natebosch:
Ok, collapsing everything to a Set of directories to delete was pretty easy and brings the runtime right back to where it was on linux in the `_test` package.

We're still potentially repeating some work here, but I think it should be minimal.

PR #1409, opened 2018-05-03
Add a warning when serving with no files

### LINKED ISSUE #1091: Print a warning when attempting to serve a directory with no assets
If I `pub run build_runner server non_standard_directory` we will start up a server but unless this directory was manually added to a `target` all requests will fail since it's outside the default whitelist. We won't know until the first build is complete, but we can check the asset graph and if there are no assets that match print a warning.

natebosch:
Closes #1091

The error message is somewhat vague because it isn't trivial to tell
_why_ a directory has no assets. We're mostly trying to make it easier
to tell that there is something wrong before getting confused that every
single request is 404ing.

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -98,10 +98,9 @@ class ServeHandler implements BuildState {
   }
 
   void _warnForEmptyDirectory(String rootDir) {
-    if (_state.assetGraph
+    if (!_state.assetGraph
         .packageNodes(_rootPackage)
-        .where((n) => n.id.path.startsWith('$rootDir/'))
-        .isEmpty) {
+        .any((n) => n.id.path.startsWith('$rootDir/'))) {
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/server/server.dart
```diff
@@ -95,6 +96,17 @@ class ServeHandler implements BuildState {
         _renderPerformance(_lastBuildResult.performance, hideSkipped),
         headers: {HttpHeaders.CONTENT_TYPE: 'text/html'});
   }
+
+  void _warnForEmptyDirectory(String rootDir) {
+    if (_state.assetGraph
+        .packageNodes(_rootPackage)
+        .where((n) => n.id.path.startsWith('$rootDir/'))
```
jakemac53: nit: `.any`?

natebosch: Good call. Done

PR #1410, opened 2018-05-03
changed the default for `enable_sync_async` to `true`

jakemac53:
I also fixed the original changelog where we added the option because it was cut off halfway through and referenced the wrong name....lol

cc @vsmenon


jakemac53:
cc @chalin to update the webdev site since we are treating this as a breaking change for build_web_compilers

PR #1412, opened 2018-05-03
Don't serve optional assets unused in the build

### LINKED ISSUE #1033: --output shouldn't output old optional assets that were not needed by the last build

natebosch:
Closes #1033

Use the `OptionalOutputTracker` in `FinalizedAssetReader` so the asset
handler sees these files as if they were not output.

### DIFF
--- a/build_runner/lib/src/asset/finalized_reader.dart
+++ b/build_runner/lib/src/asset/finalized_reader.dart
@@ -6,20 +6,21 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:build/build.dart';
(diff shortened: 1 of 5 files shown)


jakemac53:
lets add a test to confirm the behavior but lgtm

natebosch:
And of course adding a test revealed a bug 😄 

I need to reset things in between each build.

PR #1415, opened 2018-05-04
Add tools to generate test packages with builders

natebosch:
This aims to have a nicer workflow than the places where we have an
entire Dart file as a String in a test. It primarily relies on a trick
where it reades lines out of the _calling_ script and inserts them into
a test file.

A package can be generated with a `lib/builders.dart` and `build.yaml`,
or a package gan be generated with a `tool/build.dart`. In either case
the builders are wrapped automatically with factories and the remaining
configuration is plumbed through as appropriate.

There are limitations around what code can be imported and needed to
used only top-level variables rather than factories that look at
options. These limitations can be relaxed if the use case comes up.

Adds a `BuildTool` and `BuildServer` class to abstract the details of
interacting with the build system. Relies on the fact that the manual
script still uses `run` to have consistent interaction.

### DIFF
--- a/build_runner/test/integration_tests/utils/build_descriptor.dart
+++ b/build_runner/test/integration_tests/utils/build_descriptor.dart
@@ -1,0 +1,305 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:build/build.dart';
+import 'package:package_resolver/package_resolver.dart';
+import 'package:path/path.dart' as p;
+import 'package:stack_trace/stack_trace.dart';
+import 'package:test/test.dart' show expect;
+import 'package:test_descriptor/test_descriptor.dart' as d;
+import 'package:test_process/test_process.dart';
+
+class TestBuilderDefinition {
+  final String key;
+  final bool isOptional;
+  Builder builder;
+  final List<String> requiredInputs;
+
+  TestBuilderDefinition(
+      this.key, this.isOptional, this.builder, this.requiredInputs);
+}
+
+/// Define a builder with key [key] that can be assembled into a `build.dart` or
+/// `build.yaml`.
+///
+/// [key] must match the top level variable name of [builder]. It must exist in
+/// the invoking script between the last `import` and `main()`. The builder
+/// should only use references to `package:build_test/build_test.dart`.
+///
+/// [requiredInputs] only has effect when using this builder with
+/// [packageWithBuilders]. In the [packageWithBuildScript] use case the ordering
+/// of the `builders` argument determines builder ordering.
+TestBuilderDefinition builder(String key, Builder builder,
+        {bool isOptional: false, List<String> requiredInputs: const []}) =>
+    new TestBuilderDefinition(key, isOptional, builder, requiredInputs);
+
+/// Create a package in [d.sandbox] with a `buid.yaml` file exporting [builders]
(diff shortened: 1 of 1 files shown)


build_runner/test/integration_tests/optional_outputs_test.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  group('serve', () {
+    test('only serves assets that were actually required', () async {
+      var server = await buildTool.serve();
+      await server.started;
+
+      await server.expect404('a.txt.copy');
+      await server.expect404('a.txt.extra');
+
+      await updateContent('true');
```
jakemac53: nit: this feels a bit magic (update what content?). It isn't super bad since its defined in this file at least... maybe just `updateTxtContent`?

build_runner/test/integration_tests/utils/build_descriptor.dart
```diff
@@ -0,0 +1,305 @@
...
+///
+/// Files other than the pubspec should be set up with [packageContents].
+Future<BuildTool> package(Iterable<d.Descriptor> otherPackages,
+    {Iterable<d.Descriptor> packageContents}) async {
+  await d
+      .dir(
+          'a',
+          <d.Descriptor>[
+            await _pubspecWithDeps('a',
+                currentIsolateDependencies: ['build_runner', 'build'],
```
jakemac53: should we add more dependencies here? if/when we override other dependencies by `path` this might break in the future (same below).

Specifically, build_config, build_resolvers, etc (the ones you usually get transitively)

build_runner/test/integration_tests/optional_outputs_test.dart
```diff
@@ -0,0 +1,113 @@
...
+import 'dart:async';
+
+import 'package:build_test/build_test.dart';
+import 'package:meta/meta.dart';
+import 'package:test/test.dart';
+import 'package:test_descriptor/test_descriptor.dart' as d;
+
+import 'utils/build_descriptor.dart';
+
+// test-package-start
```
jakemac53: Not a huge fan of these special comments as opposed to just a multiline string. Especially because they don't really stand out much.

natebosch: I originally had it grab everything between the last import and `main()` but that was even more magic. I wanted to make this change because as opposed to a multi line string this approach gives us analysis, refactoring, formatting, etc.

I found myself editing things in this file then cutting and pasting into the string and I thought it would be nice to automate that step and get the benefits of analysis errors when there are breaking changes.

Let's find some time to chat about this today.

jakemac53: > I found myself editing things in this file then cutting and pasting into the string and I thought it would be nice to automate that step and get the benefits of analysis errors when there are breaking changes.

Ya agreed there are benefits to not having it be completely in a string - I just don't think these existing comments are obvious enough. If you don't have pre-existing knowledge about what they are doing or how this test works then this would be pretty confusing I think. Happy to chat later today :D.

natebosch: Discussed offline. We decided that there are enough benefits with this approach to make the weirdness of snipping out a chunk of code from the file worth it. It doesn't make any difference to the code which does the snipping, but I added a bunch of `####` to the comments to make them stand out more and make it clear that the chunk isn't normal code. We can continue to do that by convention.

build_runner/test/integration_tests/optional_outputs_test.dart
natebosch: Done

build_runner/test/integration_tests/utils/build_descriptor.dart
natebosch: Done

Issue #539, opened 2017-10-24
Add skippable build actions


jakemac53:
When serving/editing your main app, you probably don't want to constantly be rebuilding your tests as well, but today you would. If you have a lot of tests that could potentially slow down your edit/refresh significantly.

In order to support something like `dbuild serve web` and `dbuild serve test`, I propose that we add a notion of skippable build actions. This means our AssetGraph can retain the structure of the build, but simply skip certain portions of it.

One possible option could be combining this with the `optional` build actions concept, so that skipped actions would still be ran if something non-optional requested them. More concretely, I suggest adding a named arg such as `List<Glob> forceBuildIfMatches` to `BuildAction`, which is only allowed if `isOptional` is also `true`. Instead of skipping all actions that have `isOptional == true`, we would actually run builds for anything matching any of the provided globs. It would be safe to change those globs for different build invocations, based on command line args or whatever.

This would enable you to build specific apps based on their entry point file etc (`dbuild web/main.dart` etc).

cc @natebosch wdyt?

natebosch:
I think this makes sense. I'd maybe call the parameter `alwaysBuild`

jakemac53:
@natebosch @grouma @kevmoo I am going to start working on this, but we should clarify what the api should look like.

I actually no longer think we need to add any special flag that says an option is skippable, we can just treat all actions as skippable instead. We can accept a glob of paths that should be built - and only build the non-optional things that match those globs (and any other outputs from those same actions).

Then there is a question of what the command line interface should look like, especially because we already support trailing arguments for the `serve` and `test` commands.

Some options:

- Add a flag `--if-matches` or `--only` or something like that. This seems verbose though.
- Extend the `serve` trailing args to support globs patterns like `web/main.dart.js:8080`. I think we would always serve the top level folder in that case - but then only build `web/main.dart.js`.
  - This allows us to support it via normal trailing args in the `build` command as well.
  - It is unclear exactly how to handle the `test` command here - we could change it so we forward these args on to `pub run test`, so to run a single test would look like `pub run build_runner test --build-args test/foo_test.dart -- --test-args`. This would probably get a bit nasty in the arg parser logic, but should be doable.

Any other ideas?

jakemac53:
Oh one other option is to support building `targets` by name, which might be nice in some ways but would require some configuration as well, and it still doesn't solve building individual tests unless you make a target for each one.

natebosch:
[...]

PR #1416, opened 2018-05-07
Allow building only specified directories

jakemac53:
Work towards https://github.com/dart-lang/build/issues/539.

Sending this out before adding tests to get general agreement on the strategy first.

Specifically, this has interesting interactions with the `-o` argument, and also somewhat with the `serve` positional args. I chose to resolve these using these rules:

- If `-o` is provided with a specific output directory, add that to the list of directories to build. So for example `pbr build web -o test:build` would build both `web` and `test`, but the output directory would only contain the hoisted files from `test`.
  - The possible weird edge case here is if you do this without a specified directory for -o `pbr build web -o build` that will still only build the `web` directory, but it won't hoist it in the output directory. I think this roughly correlates to the desired behavior in that case, but it might be a little surprising for some people.
- For `serve` it is more straightforward, I simply treat the serve targets as build directories. The only real difference is in serve mode you are allowed to add a port modifier. The `-o` argument is treated the same as above, so `pbr serve web:8080 -o test:build` would build both `web` and `test`, serve `web` on 8080, and create an output directory with the hoisted `test` directory.

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -161,9 +161,12 @@ class _SharedOptions {
   }
 
   factory _SharedOptions.fromParsedArgs(
-      ArgResults argResults, String rootPackage, Command command) {
+      ArgResults argResults, String rootPackage, Command command,
+      {Iterable<String> defaultBuildDirs}) {
+    defaultBuildDirs ??= [];
     var outputMap = _parseOutputMap(argResults);
-    var buildDirs = _buildDirsFromOutputMap(outputMap);
+    var buildDirs = _buildDirsFromOutputMap(outputMap)
+      ..addAll(defaultBuildDirs);
     for (var arg in argResults.rest) {
       var parts = p.split(arg);
       if (parts.length > 1) {
@@ -230,7 +233,9 @@ class _ServeOptions extends _SharedOptions {
         );
 
   factory _ServeOptions.fromParsedArgs(
-      ArgResults argResults, String rootPackage, Command command) {
+      ArgResults argResults, String rootPackage, Command command,
+      {Iterable<String> defaultBuildDirs}) {
+    defaultBuildDirs ??= [];
     var serveTargets = <_ServeTarget>[];
     var nextDefaultPort = 8080;
     for (var arg in argResults.rest) {
@@ -259,7 +264,8 @@ class _ServeOptions extends _SharedOptions {
 
     var outputMap = _parseOutputMap(argResults);
     var buildDirs = _buildDirsFromOutputMap(outputMap)
-      ..addAll(serveTargets.map((t) => t.dir));
+      ..addAll(serveTargets.map((t) => t.dir))
+      ..addAll(defaultBuildDirs);
 
     return new _ServeOptions._(
       hostName: argResults[_hostname] as String,
@@ -365,9 +371,10 @@ abstract class _BuildRunnerCommand extends Command<int> {
   ///
   /// You may override this to return more specific options if desired, but they
   /// must extend [_SharedOptions].
-  _SharedOptions _readOptions() {
+  _SharedOptions _readOptions({Iterable<String> defaultBuildDirs}) {
     return new _SharedOptions.fromParsedArgs(
-        argResults, packageGraph.root.name, this);
+        argResults, packageGraph.root.name, this,
+        defaultBuildDirs: defaultBuildDirs);
   }
 }
 
@@ -467,8 +474,9 @@ class _ServeCommand extends _WatchCommand {
       'builds based on file system updates.';
 
   @override
-  _ServeOptions _readOptions() => new _ServeOptions.fromParsedArgs(
-      argResults, packageGraph.root.name, this);
+  _ServeOptions _readOptions({Iterable<String> defaultBuildDirs}) =>
+      new _ServeOptions.fromParsedArgs(argResults, packageGraph.root.name, this,
+          defaultBuildDirs: defaultBuildDirs);
 
   @override
   Future<int> run() async {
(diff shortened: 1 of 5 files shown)


jakemac53:
Note that I do need to figure out what to do with the test arg parser still as well, currently there are some weird interactions there.

natebosch:
@kevmoo - can you take a look at the comment above regarding how args are handled?

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -521,6 +557,22 @@ class _TestCommand extends _BuildRunnerCommand {
       'Performs a single build on the specified targets and then runs tests '
       'using the compiled assets.';
 
+  @override
+  _SharedOptions _readOptions({Iterable<String> defaultBuildDirs}) {
+    // This command doesn't allow specifying directories to build, instead it
+    // always builds the `test` directory.
+    if (argResults.rest.contains('--')) {
```
natebosch: Why does including a directory to build mean that `'--'` shows up in `argResults.rest`?

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -649,6 +702,15 @@ class _CleanCommand extends Command<int> {
   }
 }
 
+Set<String> _buildDirsFromOutputMap(Map<String, String> outputMap) {
+  var dirs = new Set<String>();
+  outputMap.forEach((k, v) {
```
natebosch: we aren't using `k` here so we could use a for loop around `outputMap.values`. Even better we could express this as what it is rather than how we get it.

```dart
Set<String> _buildDirsFromOutputMap(Map<String, String> outputMap) =>
    outputMap.values.where((v) => v != null).toSet();
```

build_runner/lib/src/generate/build_impl.dart
```diff
@@ -335,6 +341,12 @@ class _SingleBuild {
       if (phase.isOptional) continue;
       await _performanceTracker.trackBuildPhase(phase, () async {
         if (phase is InBuildPhase) {
+          // Bail early if building a non-root package and we are building
```
natebosch: do we need to do something to convert a non-optional phase action into something that can be run lazily in this case?

jakemac53: this is done btw, going to add a test for and check the logic for the optional outputs tracker as well.

jakemac53:
Note the behavior demonstrated by the test https://github.com/dart-lang/build/pull/1416/files#diff-9f6a3d2b7986c5450eac408bb5fbf963R210.

Since the OptionalOutputsTracker always allows source files you still get multiple directories - but they won't be built, which is a little bit weird.

Also note that you don't get the generated files under the `packages` dir that weren't inputs to an output in the output dir. We could decide to still always generate all files under `lib` directories but I don't think that is actually what you want (for ddc, or anything consuming summaries/kernel files, it certainly isn't).

build_runner/lib/src/generate/build_impl.dart
jakemac53: this is done as well, PTAL

jakemac53:
Overview of the differences in builds `_test` package when building in release/debug and web/test/both. Note that each of these has an overhead of `31 items, totalling 58.4 MB` for the `packages/$sdk` directory.

- `- o build`:
  - 14.7s with 399 outputs (2099 actions)
  - 2,063 items (and 2 hidden), totalling 78.5 MB
- `- o build --release`:
  - 47.7s with 230 outputs (1863 actions)
  - 134 items (and 2 hidden), totalling 63.5 MB
- `-o web:build`:
  - 8.0s with 31 outputs (1523 actions)
  - 1,852 items (and 2 hidden), totalling 76.1 MB
- `-o web:build --release`:
  - 11.3s with 19 outputs (1505 actions)
  - 102 items (and 2 hidden), totalling 60.3 MB
- `-o test:build`:
  - 14.2s with 378 outputs (2069 actions)
  - 2,032 items (and 2 hidden), totalling 78.4 MB
- `-o test:build --release`:
  - 45.8s with 214 outputs (1841 actions)
  - 120 items (and 2 hidden), totalling 63.2 MB

build_runner/test/entrypoint/run_test.dart
```diff
@@ -16,41 +22,113 @@ main() {
...
+          process.kill();
+          await process.exitCode;
+          return ExitCode.success.code;
+        } else if (nextLine.contains('failed after')) {
+          process.kill();
+          await process.exitCode;
+          return 1;
+        }
+      }
+      // Shouldn't ever reach here.
```
natebosch: maybe throw?

build_runner/lib/src/asset_graph/optional_output_tracker.dart
```diff
@@ -23,36 +23,44 @@ import 'node.dart';
 class OptionalOutputTracker {
   final _checkedOutputs = <AssetId, bool>{};
   final AssetGraph _assetGraph;
+  final List<String> _buildDirs;
   final List<BuildPhase> _buildPhases;
+  final String _rootPackage;
 
-  OptionalOutputTracker(this._assetGraph, this._buildPhases);
+  OptionalOutputTracker(
```
natebosch: Maybe expand or rephrase the doc comment on line 18?

jakemac53: done

jakemac53:
@natebosch See the change I had to make to this test https://github.com/dart-lang/build/pull/1416/commits/9a473a0f657606e1a7ae0541cc84058869f6b937

This actually is making me rethink the logic a little bit, perhaps we should always build anything that is "required" that is in any `lib` directory? For instance somebody using Sass without angular wouldn't end up getting the compiled css from other packages (or the current package) except for the sass files under the top level dir you are building. That doesn't seem right.

jakemac53:
I went ahead and changed the logic here to always run required phases for `lib/` inputs, @natebosch PTAL

build_runner/lib/src/util/util.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:build/build.dart';
```
natebosch: I'm not a fan of `util.dart` (I'm fine with a directory named `util` though). Maybe `build_dirs.dart`?

kevmoo:
You did it!

Issue #1368, opened 2018-04-26
Hide "generate-build-script" from help?


matanlurey:
Given that this is a rather internal command not useful to 99% of users.

If we agree, happy to send a PR :)

PR #1420, opened 2018-05-09
Command cleanup

jakemac53:
Refactored the `entrypoint` directory to have a file per command.

Also fixes https://github.com/dart-lang/build/issues/1368


build_runner/lib/src/entrypoint/test.dart
```diff
@@ -0,0 +1,154 @@
...
+      await new Directory(tempPath).delete(recursive: true);
+    }
+  }
+
+  /// Runs tests using [precompiledPath] as the precompiled test directory.
+  Future<int> _runTests(String precompiledPath) async {
+    stdout.writeln('Running tests...\n');
+    var extraTestArgs = argResults.rest;
+    var testProcess = await Process.start(
+        pubBinary,
```
kevmoo: FYI: you don't want to assume `pub` is in the path. Use tricks here to find the `pub` associated w/ the SDK invoking the parent.

build_runner/lib/src/util/constants.dart
```diff
@@ -27,3 +27,6 @@ const String cacheDir = '.dart_tool/build';
 
 /// Returns a hash for a given path.
 String _scriptHashFor(String path) => md5.convert(path.codeUnits).toString();
+
+/// The name of the pub binary on the current platform.
+final pubBinary = Platform.isWindows ? 'pub.bat' : 'pub';
```
kevmoo: If this is only used in test, just put it there?

build_runner/lib/src/entrypoint/test.dart
jakemac53: Sure but im not changing any code in this pr just reorganizing :D

build_runner/test/common/sdk.dart
```diff
@@ -5,13 +5,13 @@
...
-import 'package:cli_util/cli_util.dart';
 import 'package:path/path.dart' as p;
 import 'package:test/test.dart';
 import 'package:test_descriptor/test_descriptor.dart' as d;
 
-String _dartBinary = p.join(getSdkPath(), 'bin', 'dart');
-String _pubBinary = p.join(getSdkPath(), 'bin', 'pub');
+import 'package:build_runner/src/util/constants.dart';
+
+String _dartBinary = p.join(sdkBin, 'dart');
```
kevmoo:
```dart
final String dartPath =
    p.join(_sdkDir, 'bin', Platform.isWindows ? 'dart.exe' : 'dart');
```

kevmoo: Don't for get the `.exe` for windows...

kevmoo: We really should just put this in `pkg:io`

jakemac53: Agreed that these things would be useful in a shared package. I don't think the `.exe` is required though.

kevmoo:
Double checked. It's not needed!

Issue #1426, opened 2018-05-11
build_runner bug: The getter 'target' was called on null


chalin:
I'm attempting to upgrade the Angular example apps to use [pageloader 3.0.0-dev (master@HEAD)](https://github.com/google/pageloader).

I've updated the toh-1 example app, but when I try to build, I get the message:

- **You have hit a bug in build_runner**

Here is part of the log:

```
> pub run build_runner build
[INFO] Generating build script completed, took 273ms


You have hit a bug in build_runner
Please file an issue with reproduction steps at https://github.com/dart-lang/build/issues


NoSuchMethodError: The getter 'target' was called on null.
Receiver: null
Tried calling: target
dart:core                                                          Object.noSuchMethod
package:build_runner/src/package_graph/apply_builders.dart 250:22  createBuildPhases.<fn>
package:graphs/src/strongly_connected_components.dart 44:24        stronglyConnectedComponents.strongConnect
package:graphs/src/strongly_connected_components.dart 66:55        stronglyConnectedComponents
package:build_runner/src/package_graph/apply_builders.dart 248:18  createBuildPhases
package:build_runner/src/generate/build_impl.dart 93:29            build
package:build_runner/src/generate/build.dart 70:5                  build
package:build_runner/src/entrypoint/options.dart 359:24            _BuildCommand.run
package:args/command_runner.dart 194:27                            CommandRunner.runCommand
package:args/command_runner.dart 109:29                            CommandRunner.run.<fn>
dart:async                                                         new Future.sync
package:args/command_runner.dart 109:11                            CommandRunner.run
package:build_runner/src/entrypoint/run.dart 22:31                 run
.dart_tool/build/entrypoint/build.dart 81:22                       main
```

**Steps to reproduce:**

- Get a copy of the site-webdev branch associated with PR
  https://github.com/dart-lang/site-webdev/pull/1585
- `cd examples/ng/doc/toh-1`
- `pub get`
- `pub run build_runner build`

**Environment:**

- Dart VM version: 2.0.0-dev.53.0 (Thu May 3 12:35:47 2018 +0200) on "macos_x64"
- For package versions, see the PR's `pubspec.lock` file, but essentially:
  - `build_runner` 0.8.8
  - `pageloader` @HEAD of master.

cc @kevmoo @kwalrath

natebosch:
[...]

PR #1429, opened 2018-05-11
Improve error when a dependency is missing

natebosch:
Towards #1426

Instead of a stack trace we will now see:

[SEVERE] angular_tour_of_heroes:angular_tour_of_heroes declares a dependency on pageloader:pageloader but it does not exist
[SEVERE] Nothing can be built, yet a build was requested.

### DIFF
--- a/build_runner/lib/src/package_graph/apply_builders.dart
+++ b/build_runner/lib/src/package_graph/apply_builders.dart
@@ -10,6 +10,7 @@ import 'package:build_config/build_config.dart';
 import 'package:graphs/graphs.dart';
 import 'package:logging/logging.dart';
 
+import '../generate/exceptions.dart';
 import '../generate/phase.dart';
 import '../validation/config_validation.dart';
 import 'package_graph.dart';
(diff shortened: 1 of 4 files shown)


natebosch:
Added a more specific exception and a test. PTAL

build_runner/lib/src/package_graph/apply_builders.dart
```diff
@@ -245,11 +249,23 @@ Future<List<BuildPhase>> createBuildPhases(
...
+  try {
+    cycles = stronglyConnectedComponents<String, TargetNode>(
+        targetGraph.allModules.values,
+        (node) => node.target.key,
+        (node) => node.target.dependencies?.map((key) {
+              if (!targetGraph.allModules.containsKey(key)) {
+                _logger
+                    .severe('${node.target.key} declares a dependency on $key '
+                        'but it does not exist');
+                throw const _MissingDependency();
```
jakemac53: In some ways I think it would be cleaner to simply throw here, and add more information to the exception so that it can have a nice toString().

Then, we can catch the error at a higher level and not have to return an empty list which doesn't really seem like the correct behavior (it happens to end up fine but you get the weird extra log about there being nothing to build)

natebosch: yeah I already added the `CannotBuildException` in another recent PR which is handled well.
I'm not sure why I didn't think to use it here 😝

PR #1432, opened 2018-05-12
Restore error for missing default target

natebosch:


natebosch:
See #1426

build_config/lib/src/parse.dart
```diff
@@ -124,6 +124,10 @@ BuildConfig parseFromMap(String packageName,
       sources: const InputSet(),
     );
   }
+  if (!buildTargets.containsKey(defaultTarget)) {
```
kevmoo: No test?

build_config/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 0.2.6+2
```
jakemac53: did one of these already go out?

natebosch: Yes

build_config/lib/src/parse.dart

natebosch:
Added test. PTAL

Issue #1438, opened 2018-05-14
Support pkg:analyzer latest


kevmoo:
- [x] build_resolves

- [x] build

kevmoo:
Done!

PR #1443, opened 2018-05-15
build_runner: support the latest version of pkg:build

kevmoo:
Towards https://github.com/dart-lang/build/issues/1438


kevmoo:
Why the tight constraint on pkg:build here?

jakemac53:
We changed a bit they way we do versions so that we can easily track api additions/modifications using the minor version (we implement those interfaces).

Issue #1431, opened 2018-05-12
Add an FAQ under docs


natebosch:
We probably want to shorten our getting started doc to just the happy case and direct other stuff to an FAQ.

jakemac53:
This exists now, and we have been adding things as they come up.

PR #1453, opened 2018-05-16
Start an FAQ for builder authors

natebosch:
Towards #1431

We can add answers to this doc as they come up in issues. Start with an
answer for https://github.com/google/pageloader/issues/132

### DIFF
--- a/docs/builder_author_faq.md
+++ b/docs/builder_author_faq.md
@@ -1,0 +1,18 @@
+## When should a Builder build to `cache` vs `source`?
+
+Use `build_to: source` when:
+- The generated code _does not_ depend on details outside of the package with
+  the generated code that might change with a different version of it's
+  dependencies. Any details from the _generating_ package, or any other relied
+  upon by the generated code must not change without a major version bump.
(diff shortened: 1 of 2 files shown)


natebosch:
> Should we just make this a general `faq.md`?

I think builder authors and end users are different enough audiences to have 2 docs.

docs/builder_author_faq.md
```diff
@@ -0,0 +1,18 @@
+## When should a Builder build to `cache` vs `source`?
+
+Use `build_to: source` when:
+- The generated code _does not_ depend on details outside of the package with
```
grouma: I think I understand this first sentence but it is very hard for me to parse. Can you reword it? The multiple references to generated code specifically cause the confusion. Maybe use "input sources" for the second reference?

natebosch:
> please add a link prominently on the main README.md for the repo

Done. PTAL

docs/builder_author_faq.md
natebosch: Rephrased. Is it more clean now?

PR #1456, opened 2018-05-16
Create a build runner abstraction

grouma:
- Create new `BuildRunner` abstraction which will be exposed from `build_runner_core`
  - Note this abstraction is a simple wrapper around `BuildImpl` which exposes less details e.g. `AssetGraph`
- Change `BuildImpl` to not automatically do a first build
- Update `WatchImpl` to now manually initiate the first build
- Update `generate/build.dart` to now use `BuildRunner` instead of `BuildImpl`
- Update `test_phases` to now use `BuildRunner` API

### DIFF
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -5,6 +5,10 @@
 import 'dart:async';
 import 'dart:io';
 
+import 'package:build_runner/src/environment/io_environment.dart';
+import 'package:build_runner/src/environment/overridable_environment.dart';
+import 'package:build_runner/src/generate/options.dart';
+import 'package:build_runner/src/generate/terminator.dart';
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart';
 
@@ -13,8 +17,8 @@ import '../asset/writer.dart';
 import '../package_graph/apply_builders.dart';
 import '../package_graph/package_graph.dart';
 import '../server/server.dart';
-import 'build_impl.dart' as build_impl;
 import 'build_result.dart';
+import 'build_runner.dart';
 import 'directory_watcher_factory.dart';
 import 'watch_impl.dart' as watch_impl;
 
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/build.dart
```diff
@@ -50,45 +54,58 @@ import 'watch_impl.dart' as watch_impl;
...
       trackPerformance: trackPerformance,
-      skipBuildScriptCheck: skipBuildScriptCheck,
       verbose: verbose,
-      builderConfigOverrides: builderConfigOverrides,
-      isReleaseBuild: isReleaseBuild,
-      buildDirs: buildDirs,
-    );
+      buildDirs: buildDirs);
+  var terminator = new Terminator(terminateEventStream);
+  var build = await BuildRunner.create(
```
natebosch: is returning null on failure the right behavior here? What types of reasons might this fail? Would throwing be more appropriate?

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -302,21 +302,19 @@ class WatchImpl implements BuildState {
...
-      _reader = _build?.finalizedReader;
-      _readyCompleter.complete(null);
-
+      BuildResult firstBuild;
       if (_build == null) {
-        var result = new BuildResult(BuildStatus.failure, []);
-        controller.add(result);
-        firstBuildCompleter.complete(result);
-        return;
+        firstBuild = new BuildResult(BuildStatus.failure, []);
```
natebosch: [nit] would read easier with a ternary conditional IMO

build_runner/lib/src/generate/build.dart
grouma: It now throws.

build_runner/lib/src/generate/watch_impl.dart
grouma: N/A

build_runner/lib/src/generate/build.dart
```diff
@@ -92,19 +93,22 @@ Future<BuildResult> build(List<BuilderApplication> builders,
...
+      builders,
+      builderConfigOverrides,
+      isReleaseBuild: isReleaseBuild ?? false,
+    );
+    var result = await build.run({});
+    await build.beforeExit();
+    await terminator.cancel();
+    await options.logListener.cancel();
+    return result;
+  } on CannotBuildException {
```
natebosch: I don't think we need to catch this here. We can let it bubble up to `run`

build_runner/test/common/test_phases.dart
```diff
@@ -124,21 +125,23 @@ Future<BuildResult> testBuilders(
...
+      options,
+      environment,
+      builders,
+      builderConfigOverrides,
+      isReleaseBuild: false,
+    );
     result = await build.run({});
     await build.beforeExit();
     await options.logListener.cancel();
+  } on CannotBuildException {
```
natebosch: do we need to catch this here? Do any of our tests hit this behavior? If so, are they worthwhile tests?

build_runner/test/generate/build_test.dart
```diff
@@ -399,16 +399,18 @@ void main() {
...
+      expect(
+          () => testBuilders(
+              [
+                apply('', [(_) => new TestBuilder()], toPackage('b'),
+                    hideOutput: false)
+              ],
+              {'b|lib/b.txt': 'b'},
+              packageGraph: packageGraph,
+              outputs: {},
+              // Nothing to build since everything is skipped
```
natebosch: can remove this and the `BuildStatus.failur` now.

build_runner/lib/src/generate/build.dart
```diff
@@ -50,45 +54,60 @@ import 'watch_impl.dart' as watch_impl;
...
+  try {
+    var build = await BuildRunner.create(
+      options,
+      environment,
+      builders,
+      builderConfigOverrides,
+      isReleaseBuild: isReleaseBuild ?? false,
     );
+    var result = await build.run({});
+    await build?.beforeExit();
```
jakemac53: nit: should we rename this shutdown?

PR #1459, opened 2018-05-17
Better missing module message

jakemac53:
Closes https://github.com/dart-lang/build/issues/1458

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.2.2+5
+
+- Updated the missing modules message to highlight that the error is likely due
+  to a missing dependency.
+
 ## 0.2.2+4
 
 - Support `package:analyzer` `0.32.0`.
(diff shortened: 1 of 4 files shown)


jakemac53:
> Could this also be triggered by a build failure where a file is not output?

Yes, that could also cause this error, but I think the missing dependency case is more common (and confusing), which is why I am highlighting it here.

In the case of a build error you should see that error in your console as well (not necessarily true today actually if its from a previous build but we should be able to fix that).

jakemac53:
Another option could be exposing something like `unreadableReason` on the main asset reader api and then use that to be smarter here?

natebosch:
I'd be in favor of adding a field to the exception if we think we can find a meaningful way to implement it with bazel...

jakemac53:
> In bazel there is still the case of multiple builder factories in a single dart_codegen_binary, though maybe we bail early if there is a failure?

We could choose to do something nice here, but I also think its fine if there is just an `Unknown` reason or something which we fall back on at least initially. I don't think we have many use cases in google3 where this would actually happen, its only if you had builders which read the outputs of an earlier builder as a secondary input that you would get the nicer error. 
 
> So would there just never be a case of trying to read a failed output in bazel?

Outside of the multiple builder factory case I don't think so.

jakemac53:
Filed https://github.com/dart-lang/build/issues/1460 to track future work to improve the ability to provide nice error messages in these cases.

Issue #1427, opened 2018-05-11
Create `package:build_runner_core`


grouma:
This package should have a minimal set of features to reduce the number of dependencies. It likely should only expose the `build` command. This work is related to https://github.com/dart-lang/test/issues/797 and will likely be used by `flutter_test`.

jakemac53:
this is released

PR #1465, opened 2018-05-17
Refactor override configs

grouma:
More towards #1427

`findBuildConfigOverrides` has a dependency on `watcher`. Refactor it out of `options`.

### DIFF
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -9,6 +9,7 @@ import 'package:build_runner/src/environment/io_environment.dart';
 import 'package:build_runner/src/environment/overridable_environment.dart';
 import 'package:build_runner/src/generate/options.dart';
 import 'package:build_runner/src/generate/terminator.dart';
+import 'package:build_runner/src/package_graph/build_config_overrides.dart';
 import 'package:logging/logging.dart';
 import 'package:shelf/shelf.dart';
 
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -69,8 +70,9 @@ Future<ServeHandler> watch(
       reader: reader,
       writer: writer,
       onLog: onLog);
+  overrideBuildConfig ??
```
natebosch: did you intend to use `??=`?

build_runner/lib/src/generate/build.dart
```diff
@@ -80,12 +81,13 @@ Future<BuildResult> build(List<BuilderApplication> builders,
...
       onLog: onLog);
   var options = await BuildOptions.create(environment,
-      configKey: configKey,
       deleteFilesByDefault: deleteFilesByDefault,
       failOnSevere: failOnSevere,
       packageGraph: packageGraph,
       logLevel: logLevel,
       skipBuildScriptCheck: skipBuildScriptCheck,
+      overrideBuildConfig:
+          await findBuildConfigOverrides(packageGraph, configKey),
```
natebosch: does this mean we need to expose `findBuildConfigOverrides` from `build_runner_core`? Could `BuildOptions.create` call it instead?

build_runner/lib/src/generate/watch_impl.dart
grouma: Yes. Fixed.

build_runner/lib/src/generate/build.dart
grouma: No. The intent behind this PR is to allow `findBuildConfigOverrides` to remain in `build_runner`.

natebosch: ah, this is the `package_name.build.yaml` handling? I guess Flutter won't need to support that.

jakemac53: This is actually, `build.<config>.yaml` handling. The main issue is actually that we are trying to do as little yaml parsing as possible so the main api in core can only deal with `BuildConfig` objects.

The `--config` flag and semantics of it are something build_runner specific imo, and doesn't belong in core.

PR #1467, opened 2018-05-17
Include builderLabels in duplicate asset exception

### LINKED ISSUE #1273: Report builder keys when there is a conflicting output

natebosch:
Closes #1273
Start towards #1431

- Pass through the `BuildPhase`s and `rootPackage` so that the builder
  labels and a friendlier string for the AssetId can be used.
- Catch and log the duplicate asset exception when it occurs while
  loading the build definition. Throw a `CannotBuildException` to cancel
  the overall build without printing a stack trace.
- Add an FAQ section for this problem.
- Catch all exceptions from updating the asset graph when they occur
  during the build.

### DIFF
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -104,8 +104,8 @@ class _Loader {
         try {
           assetGraph = await AssetGraph.build(_buildPhases, inputSources,
               internalSources, _options.packageGraph, _environment.reader);
-        } on DuplicateAssetNodeException catch (e) {
-          _logger.severe('$e');
+        } on DuplicateAssetNodeException catch (e, st) {
+          _logger.severe('Conflicting outputs', e, st);
           throw new CannotBuildException();
         }
         buildScriptUpdates = await BuildScriptUpdates.create(
--- a/build_runner/lib/src/generate/build_definition.dart
+++ b/build_runner/lib/src/generate/build_definition.dart
@@ -101,8 +101,13 @@ class _Loader {
       Set<AssetId> conflictingOutputs;
 
       await logTimedAsync(_logger, 'Building new asset graph', () async {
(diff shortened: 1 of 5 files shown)


natebosch:
New output looks like:

```
[INFO] Generating build script completed, took 421ms
[SEVERE] Both pageloader and built_value_generator|built_value may output lib/hero.g.dart. Potential outputs must be unique across all builders. See https://github.com/dart-lang/build/blob/master/docs/faq.md#why-do-builders-need-unique-outputs
```

build_runner/lib/src/generate/build_definition.dart
```diff
@@ -101,8 +101,13 @@ class _Loader {
       Set<AssetId> conflictingOutputs;
 
       await logTimedAsync(_logger, 'Building new asset graph', () async {
-        assetGraph = await AssetGraph.build(_buildPhases, inputSources,
-            internalSources, _options.packageGraph, _environment.reader);
+        try {
+          assetGraph = await AssetGraph.build(_buildPhases, inputSources,
+              internalSources, _options.packageGraph, _environment.reader);
+        } on DuplicateAssetNodeException catch (e) {
+          _logger.severe('$e');
```
jakemac53: lets attach the stack trace as well, so we get it in verbose mode.

docs/faq.md
```diff
@@ -0,0 +1,15 @@
+## Why do Builders need unique outputs?
+
+`build_runner` relies on determining a static build graph before starting a
+build - it needs to know every file that may be written and which Builder would
```
grouma: [nit] Why capitalize Builder here but not below?

natebosch: No idea. going with all capitals for now.

build_runner/lib/src/generate/build_definition.dart
natebosch: Done

Issue #1470, opened 2018-05-18
findAssets can return sources that don't match any targets


jakemac53:
today the `FileBasedAssetReader` just globs the package, and only has access to a package graph which does not have the information required to filter out things that are not actually sources.

This means you can get a file back from a glob that you can't actually read, and it probably won't be tracked right either.

cc @kevmoo

jakemac53:
I am going to fix this for now by filtering results in the single step reader I think - since that is the only thing with the required information (I can use the asset graph to filter out nodes that don't exist).

jakemac53:
Actually, it looks like this is caused by us using the package graph and not the target graph when constructing the asset graph originally... I think changing this might end up being a breaking change.

Basically it looks like right now we are exposing all files in the root package as readable, we just won't let them be used as primary inputs if they aren't in a target. That means we probably are also picking up files from build directories and other hidden directories :(.

PR #1471, opened 2018-05-18
Use target graph for globs

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1470

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -7,6 +7,8 @@
   was built.
 - Improve error message when a dependency has a bad `build.yaml` with a missing
   dependency.
+- Sources that are not a part of a `target` will no longer appear in the asset
+  graph, so they will not be readable or globbable.
 
 ## 0.8.9
 
--- a/build_runner/lib/src/generate/input_matcher.dart
+++ b/build_runner/lib/src/generate/input_matcher.dart
@@ -8,51 +8,52 @@ import 'package:collection/collection.dart';
 import 'package:glob/glob.dart';
 
 /// A filter on files to run through a Builder.
-abstract class InputMatcher {
-  /// Whether [input] is included in this set of assets.
-  bool matches(AssetId input);
-
-  factory InputMatcher(InputSet inputSet) = _GlobInputMatcher;
-}
-
-class _GlobInputMatcher implements InputMatcher {
+class InputMatcher {
   /// The files to include
   ///
   /// Null or empty means include everything.
-  final List<Glob> include;
+  final List<Glob> includeGlobs;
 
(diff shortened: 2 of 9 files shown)


build_runner/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   was built.
 - Improve error message when a dependency has a bad `build.yaml` with a missing
   dependency.
+- Sources that are not a part of a `target` will no longer appear in the asset
+  graph, so they will not be readable or globbable.
```
grouma: Is globbable a word haha?

build_runner/lib/src/generate/input_matcher.dart
```diff
@@ -8,65 +8,68 @@ import 'package:collection/collection.dart';
...
   ///
   /// Null or empty means exclude nothing.
-  final List<Glob> exclude;
+  final List<Glob> excludeGlobs;
 
-  _GlobInputMatcher(InputSet inputSet)
-      : this.include = inputSet.include?.map((p) => new Glob(p))?.toList(),
-        this.exclude = inputSet.exclude?.map((p) => new Glob(p))?.toList();
+  InputMatcher(InputSet inputSet, {List<String> defaultInclude})
+      : this.includeGlobs = (inputSet.include ?? defaultInclude)
```
natebosch: We talked offline about possibly adding a new API for handling `matchesPreFiltered` or something so that that we aren't spending a bunch of time globbing against defaults when we know they'll always match (since they're in the asset graph already).

build_runner/CHANGELOG.md
jakemac53: it is now!

build_runner/lib/src/generate/input_matcher.dart
jakemac53: So I looked into this a bit this morning - I don't see anywhere this api would end up getting used.

We are only using this api when doing the initial glob for sources, as far as I can tell.

For the `targetSources` on `BuildPhase`, we are using `node.target.sources` which is the original sources.

jakemac53:
Ahah! I had accidentally removed the special whitelist for the SDK, adding that back in brings times back to normal.

build_runner/lib/src/generate/options.dart
```diff
@@ -89,6 +89,8 @@ class BuildOptions {
 
     Logger.root.level = logLevel;
 
+    var logListener = Logger.root.onRecord.listen(environment.onLog);
```
jakemac53: note that I moved this up so that I could log inside of `TargetGraph.forPackageGraph`, figured I might as well submit that change too though.

PR #1479, opened 2018-05-22
Follow symlinks when listing files

### LINKED ISSUE #1445: symlinked .dart files not found when builidng
After installing 2.0.0-dev.54.0-1 (linux) and switching from "pub" to "webdev" 
for building and serving, the compilers complain about not being able to find symlinked files
in the web/ directory 
Replacing the symlinked files with the real version fixes the problem.
IIRC the problem occurred with both  webdev serve and webdev build

----

For general questions consider using Stack Overflow instead:
https://stackoverflow.com/questions/tagged/dart

Also consider our Gitter channel for light-weight/quick discussions:
https://gitter.im/dart-lang/build

-----

natebosch:
Fixes #1445


natebosch:
I can't figure out why the decision was made to not follow symlinks, it's been this way since the first commit in this repo.

Unless we can think of some problem this is likely to cause I'm inclined to allow symlinks into the build.

natebosch:
Watching doesn't work so I don't think we can reasonably support this.

https://github.com/dart-lang/watcher/issues/18

natebosch:
We _could_ support it without watch mode working.

I'm not sure which behavior is more potentially confusing - or if there is a smooth way to detect and warn for either one.

natebosch:
> What are the file paths that we get back here?

We get the path of the link itself. For the most part this looks to use as if the file contents where directly in place.

> Also it would be good to verify the edge case of editing a file which is symlinked to.

Changes across invocations work fine because we check the content digest. Changes while watching don't work because we don't get an update event.

jakemac53:
I don't think we should support this at all then - but maybe we should detect it and report an error when we see symlinks that match source globs?

jakemac53:
what is the plan here?

natebosch:
I have not been able to find a reasonable design to detect symlinks. We want to either:

- Warn anytime we see symlinks (and don't attempt to support them)
- Warn when you try to try to _watch_ with symlinks (and best-effort support them for builds)

natebosch:
After a recent sync we decided it's worth moving forward with this even if we can't get a nice warning for watch mode.

jakemac53:
cc @alan-knight

PR #1481, opened 2018-05-23
build_runner: Dart2 fixes

kevmoo:
The fix to server.dart is a work-around for dart-lang/sdk#33212

### DIFF
--- a/build_runner/lib/src/asset_graph/serialization.dart
+++ b/build_runner/lib/src/asset_graph/serialization.dart
@@ -30,12 +30,12 @@ class _AssetGraphDeserializer {
         _deserializeDigest(_serializedGraph['buildActionsDigest'] as String),
         _serializedGraph['dart_version'] as String);
 
-    var packageNames = _serializedGraph['packages'] as List<String>;
+    var packageNames = _serializedGraph['packages'] as List;
 
     // Read in the id => AssetId map from the graph first.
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/asset_graph/serialization.dart
```diff
@@ -30,12 +30,12 @@ class _AssetGraphDeserializer {
         _deserializeDigest(_serializedGraph['buildActionsDigest'] as String),
         _serializedGraph['dart_version'] as String);
 
-    var packageNames = _serializedGraph['packages'] as List<String>;
+    var packageNames = _serializedGraph['packages'] as List;
```
jakemac53: should we `.cast<String>()` here and remove the `as String` below?

kevmoo: Same-same, right? The cast call feels "heavier" – @natebosch ?

jakemac53: I am curious what effective dart says here now - its effectively the same though I think since its a lazy cast

kevmoo: @munificent ?

jakemac53: Ah, "Cast the elements on access. If you immediately iterate over the collection, cast each element inside the iteration.", so LGTM

Issue #1484, opened 2018-05-24
Logging lost colors etc, `assumeTty` probably not passed through correctly?


jakemac53:
cc @grouma

jakemac53:
note that things work in `build` mode, but not `watch`/`serve` modes

PR #1485, opened 2018-05-24
fix default logging for watch/serve

jakemac53:
fixes https://github.com/dart-lang/build/issues/1484


grouma:
Anyway we can add a test for this?

Issue #1293, opened 2018-04-12
Typecast error in build_config when using dart2


alorenzen:
Attempting to run the angular compiler using --preview-dart-2 and getting a runtime cast failure in build_config:

```
Unhandled exception:
type 'YamlMap' is not a subtype of type 'Map<String, dynamic>' in type cast where
  YamlMap is from package:yaml/src/yaml_node.dart
  Map is from dart:core
  String is from dart:core

#0      Object._as (dart:core/runtime/libobject_patch.dart:67:25)
#1      parseFromYaml (package:build_config/src/parse.dart:78:30)
#2      new BuildConfig.parse (package:build_config/src/build_config.dart:113:7)
#3      BuildConfig.fromBuildConfigDir (package:build_config/src/build_config.dart:72:18)
<asynchronous suspension>
#4      _packageBuildConfig (package:build_runner/src/package_graph/target_graph.dart:55:22)
<asynchronous suspension>
#5      TargetGraph.forPackageGraph (package:build_runner/src/package_graph/target_graph.dart:29:17)
<asynchronous suspension>
#6      build (package:build_runner/src/generate/build_impl.dart:73:41)
<asynchronous suspension>
#7      build (package:build_runner/src/generate/build.dart:69:16)
#8      _BuildCommand.run (package:build_runner/src/entrypoint/options.dart:319:24)
<asynchronous suspension>
#9      CommandRunner.runCommand (package:args/command_runner.dart:194:27)
<asynchronous suspension>
#10     CommandRunner.run.<anonymous closure> (package:args/command_runner.dart:109:29)
#11     new Future.sync (dart:async/future.dart:224:31)
#12     CommandRunner.run (package:args/command_runner.dart:109:11)
#13     run (package:build_runner/src/entrypoint/run.dart:22:31)
<asynchronous suspension>
#14     main (file:///Users/alorenzen/angular/examples/registration_form/.dart_tool/build/entrypoint/build.dart:63:26)
<asynchronous suspension>
#15     _startIsolate.<anonymous closure> (dart:isolate/runtime/libisolate_patch.dart:275:32)
#16     _RawReceivePortImpl._handleMessage (dart:isolate/runtime/libisolate_patch.dart:165:12)
```

jakemac53:
We can/will fix this for sure, but just fyi you are probably going to get a lot of errors here from other packages (analyzer, etc).

alorenzen:
[...]

PR #1486, opened 2018-05-24
Use JsonSerializable for build_config parsing

jakemac53:
There is a little bit of weirdness here which we can hopefully resolve eventually with fixes from JsonSerializable, but this gets the vm tests passing with --preview-dart-2, and gives us pretty nice mostly-automatic error checking and reporting.

As an example, this build.yaml:

```yaml
targets:
  $default:
    sources:
      include: foo
```

Gives this error:

```
[SEVERE] Failed to parse `build.yaml` for provides_builder.
Invalid argument(s): line 4, column 7: Could not create `InputSet`.
Unsupported value for `include`: type 'String' is not a subtype of type 'List' in type cast

      include: foo
      ^^^^^^^
```

And we have custom errors that look nice as well, for a missing `$default` target for instance:

```yaml
targets:
  floop:
    sources:
      - lib/foo.dart
```

Gives this custom error:

```
[SEVERE] Failed to parse `build.yaml` for <package-here>.
Invalid argument(s): line 1, column 1: Could not create `BuildConfig`.
Unsupported value for `targets`: Must specify a target with the name `provides_builder` or `$default`.

targets:
^^^^^^^
```

The primary thing this doesn't support right now is error reporting for unrecognized keys. I was thinking I would try and add that as a followup, but JsonSerializable should be able to add it by default as well so that might be wasted effort, cc @kevmoo.

Also note that I ended up using a combination of both zones and expandos to get around the issues with the `package` and `key` fields which don't exist in the actual yaml values.


natebosch:
> ```
> Unsupported value for `include`: type 'String' is not a subtype of type 'List' in type cast
> ```

I thought we supported automatically turning this into a `List<String>` for `include`?

jakemac53:
> I thought we supported automatically turning this into a List<String> for include?

Oh, I didn't think so but I could support that in this system still.

natebosch:
Hmm, double the checking the code I don't see that handling... perhaps I'm misremembered.

jakemac53:
> Hmm, double the checking the code I don't see that handling... perhaps I'm misremembered.

I tried to retain all the custom logic we originally had in parse.dart, I think I got it all.

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1293

Issue #1487, opened 2018-05-24
build_web_compilers issue w/ --preview-dart-2


kevmoo:
https://travis-ci.org/dart-lang/build/jobs/383404300#L563

```
[SEVERE] build_web_compilers|entrypoint:
Invalid argument (dart2js_args): Expected a list of strings, but got a List<dynamic>:: Instance(length:1) of '_GrowableList'
You have hit a bug in build_runner
Please file an issue with reproduction steps at https://github.com/dart-lang/build/issues
builderFactory did not return a builder.
package:build_runner/src/package_graph/apply_builders.dart 180:30  new BuilderApplication.forBuilder.<fn>.<fn>
package:build_runner/src/package_graph/apply_builders.dart 320:27  _createBuildPhasesForBuilderInCycle.<fn>.<fn>
dart:core                                                          List.addAll
package:build_runner/src/package_graph/apply_builders.dart 269:7   createBuildPhases
package:build_runner/src/generate/build_impl.dart 95:29            BuildImpl.create
package:build_runner/src/generate/build_runner.dart 31:41          BuildRunner.create
package:build_runner/src/generate/build.dart 99:35                 build
package:build_runner/src/entrypoint/test.dart 82:26                TestCommand.run
package:args/command_runner.dart 194:27                            CommandRunner.runCommand
package:args/command_runner.dart 109:29                            CommandRunner.run.<fn>
dart:async                                                         new Future.sync
package:args/command_runner.dart 109:11                            CommandRunner.run
package:build_runner/src/entrypoint/run.dart 23:31                 run
.dart_tool/build/entrypoint/build.dart 80:26                       main
```

PR #1489, opened 2018-05-24
fix dart2js args cast error in dart 2

jakemac53:
After this the _test package can now build/serve web apps with --preview-dart-2!

Fixes https://github.com/dart-lang/build/issues/1487

### DIFF
--- a/build_web_compilers/lib/src/web_entrypoint_builder.dart
+++ b/build_web_compilers/lib/src/web_entrypoint_builder.dart
@@ -78,7 +78,7 @@ class WebEntrypointBuilder implements Builder {
     }
 
     var dart2JsArgs =
-        options.config[_dart2jsArgs].cast<String>() ?? const <String>[];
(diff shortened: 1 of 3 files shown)


build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -77,7 +77,8 @@ class WebEntrypointBuilder implements Builder {
             'Only `dartdevc` and `dart2js` are supported.');
     }
 
-    var dart2JsArgs = options.config[_dart2jsArgs] ?? const <String>[];
+    var dart2JsArgs =
+        options.config[_dart2jsArgs].cast<String>() ?? const <String>[];
```
natebosch: needs to be `?.cast<String>()`. Otherwise the `??` isn't doing anything.

jakemac53: ah yep good catch

PR #1492, opened 2018-05-24
Fix sync-async issues in serve handler test

natebosch:


build_runner/test/server/serve_handler_test.dart
```diff
@@ -199,17 +199,14 @@ class MockWatchImpl implements WatchImpl {
   }
 
   MockWatchImpl(this.reader, this.packageGraph, this.assetGraph) {
+    var firstBuild = new Completer<BuildResult>();
```
kevmoo: Couldn't you use Stream.forEach or something here?

I get nervous when I see Completer and fields holding onto Future

kevmoo:
Will let smarter people review then :-)

PR #1493, opened 2018-05-24
Use latest build_config in build_runner

natebosch:
- Add a dependency override to use it before it's published.
- Fix some errors in the config in test code. Drop unnecessary targets
  config.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -84,20 +84,20 @@ class BuildConfig {
   /// Create a [BuildConfig] by parsing [configYaml].
   factory BuildConfig.parse(
       String packageName, Iterable<String> dependencies, String configYaml) {
-    try {
-      var parsed = loadYaml(configYaml) as Map;
-      return new BuildConfig.fromMap(
-          packageName, dependencies, parsed ?? const {});
-    } on CheckedFromJsonException catch (e) {
-      throw new ArgumentError(_prettyPrintCheckedFromJsonException(e));
-    }
+    var parsed = loadYaml(configYaml) as Map;
+    return new BuildConfig.fromMap(
+        packageName, dependencies, parsed ?? const {});
   }
 
   /// Create a [BuildConfig] read a map which was already parsed.
   factory BuildConfig.fromMap(
       String packageName, Iterable<String> dependencies, Map config) {
-    return runInBuildConfigZone(() => new BuildConfig._fromJson(config),
-        packageName, dependencies.toList());
+    try {
+      return runInBuildConfigZone(() => new BuildConfig._fromJson(config),
+          packageName, dependencies.toList());
+    } on CheckedFromJsonException catch (e) {
+      throw new ArgumentError(_prettyPrintCheckedFromJsonException(e));
+    }
   }
 
   BuildConfig({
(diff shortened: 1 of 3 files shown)


build_config/lib/src/build_config.dart
```diff
@@ -145,24 +145,27 @@ Map<String, T> _normalizeBuilderDefinitions<T>(
...
-  var yamlKey = yamlMap.nodes.keys.singleWhere(
-      (k) => (k as YamlScalar).value == err.key,
-      orElse: () => null) as YamlScalar;
-
   var message = 'Could not create `${err.className}`.';
-  if (yamlKey == null) {
-    assert(err.key == null);
-    message = '${yamlMap.span.message(message)} ${err.innerError}';
+  if (err.key == null) {
+    if (err.map is YamlMap) {
```
kevmoo: nit: if you pull out the `map` as a local var and do the type check there, you shouldn't have to cast it below.

natebosch:
> Don't see new test coverage for the new/better pretty things

It's an exception path which is little used. We don't generally test those.

> I don't understand how the changes to the test files map to the changes in the code.

These are changes required because the latest `build_config` is more strict than it used to be. The invalid config fails at parse time now rather than later at runtime.

build_config/lib/src/build_config.dart

build_config/lib/src/build_config.dart
```diff
@@ -145,24 +145,27 @@ Map<String, T> _normalizeBuilderDefinitions<T>(
...
       message += '${err.message}\n';
     } else {
       message += '${err.innerError}\n';
     }
-    message = yamlKey.span.message(message);
+    if (map is YamlMap) {
+      var yamlKey = map.nodes.keys.singleWhere(
+          (k) => (k as YamlScalar).value == err.key,
+          orElse: () => null) as YamlScalar;
+      message += yamlKey.span.message(message);
```
jakemac53: are you sure you want the += here? I think the yaml `.span.message` call already incorporates the message you pass in.

build_config/lib/src/build_config.dart
```diff
@@ -145,24 +145,27 @@ Map<String, T> _normalizeBuilderDefinitions<T>(
...
     if (err.message != null) {
       message += '${err.message}\n';
     } else {
       message += '${err.innerError}\n';
     }
-    message = yamlKey.span.message(message);
+    if (map is YamlMap) {
+      var yamlKey = map.nodes.keys.singleWhere(
+          (k) => (k as YamlScalar).value == err.key,
+          orElse: () => null) as YamlScalar;
```
jakemac53: nit: if adding an `orElse` which returns `null`, then we should handle the `null` case.

natebosch:
Holding off on the changes to the exception handling. The important thing here is that the test works with the stricter parsing and I had only changed the exception handling to get actionable errors from the test code.

natebosch:
Hmm, analyzer warnings point to a potential usage that we've broken with the latest build_config.

@jakemac53 can you take a look? Technically this isn't a usage pattern we expected people (outside of our tests) to use, but we did change the public surface area when we changed the signature of some of the constructors.

https://travis-ci.org/dart-lang/build/jobs/383788477

jakemac53:
https://github.com/dart-lang/build/pull/1500 now does this plus fixing the other tests and some other minor tweaks that were needed for that

PR #1497, opened 2018-05-29
create test common

grouma:
More towards #1427

- Create fake `test_common` package which will be used by `build_runner` and `build_runner_core` for tests.


build_runner/pubspec.yaml
```diff
@@ -40,3 +40,5 @@ dev_dependencies:
   test: ^0.12.0
   test_descriptor: ^1.0.0
   test_process: ^1.0.0
+  test_common:
+    path: ../test_common
```
natebosch: Can we make this `_test_common` to make it clear that we don't intend for it to be published?

test_common/pubspec.yaml
```diff
@@ -0,0 +1,18 @@
...
+version: 0.1.0
+description: Test infra for writing build tests. Is not published.
+author: Dart Team <misc@dartlang.org>
+
+environment:
+  sdk: '>=2.0.0-dev.54 <2.0.0'
+
+dependencies:
+  build: ">=0.12.4 <0.12.7"
+  build_config: ^0.2.6
```
natebosch: I think these should be path dependencies

build_runner/pubspec.yaml
grouma: Done.

PR #1500, opened 2018-05-30
Use the latest build_config from build_runner

jakemac53:
- Updated test_common to use dependency overrides
- Updated the pubspec/changelog for build_config
- Updated `BuildConfig.useDefault` factory constructor to auto-construct the zone
- Throw a better error if we detect that we are not running in a build config zone
- Fix all the tests 
- Export `runInBuildConfigZone`


jakemac53:
> Any other changes we need to sneak in with the breaking version bump?

Not that I am aware of

PR #1504, opened 2018-05-31
Build runner core

grouma:
Mostly finishes #1427

- Create a new `build_runner_core` package and update `build_runner` to now depend on it.

I will of course need to publish the package. Further clean up can be done following this PR to minimize the amount of imports from `src`.


jakemac53:
:tada:

PR #1506, opened 2018-06-01
Add back support for erroring on unrecognized keys in build.yaml

jakemac53:
- use `disallowUnrecognizedKeys: true` everywhere
- added special handling for the `UnrecognizedKeysException` to log all the unrecognized keys.
- fixed the error reporting for bad values to use the value span not the key span


_test/pubspec.yaml
```diff
@@ -13,8 +13,6 @@ dev_dependencies:
 dependency_overrides:
   build:
     path: ../build
-  build_barback:
```
natebosch: Maybe submit this separately?

PR #1510, opened 2018-06-04
Build config docs

jakemac53:
I think eventually it would be cool to auto-gen these docs using json serializable, and this was initially an effort to see what that format might eventually look like - but I think its worth submitting.

The only real downside is we might forget to update it if we change the format but its easy to fix.

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -3,6 +3,10 @@
 Customizing the build behavior of a package is done  by creating a `build.yaml`
 file, which describes your configuration.
 
+The full format is described in the
+[docs/build_yaml_format.md](../docs/build_yaml_format.md) file, while this
+documentation is more focused on specific usage scenarios of the file.
+
 ## Dividing a package into Build targets
 
 When a `Builder` should be applied to a subset of files in a package the package
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -4,8 +4,9 @@ Customizing the build behavior of a package is done  by creating a `build.yaml`
 file, which describes your configuration.
 
 The full format is described in the
-[docs/build_yaml_format.md](../docs/build_yaml_format.md) file, while this
-documentation is more focused on specific usage scenarios of the file.
+[docs/build_yaml_format.md](https://github.com/dart-lang/build/blob/master/docs/build_yaml_format.md)
+file, while this documentation is more focused on specific usage scenarios of
+the file.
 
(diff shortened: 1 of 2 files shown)


build_config/README.md
```diff
@@ -3,6 +3,10 @@
 Customizing the build behavior of a package is done  by creating a `build.yaml`
 file, which describes your configuration.
 
+The full format is described in the
+[docs/build_yaml_format.md](../docs/build_yaml_format.md) file, while this
```
natebosch: [nit] this link won't work from the README on the pub site. Maybe add a full URL?

docs/build_yaml_format.md
```diff
@@ -0,0 +1,125 @@
...
+include | List<String> | `**`
+exclude | List<String> | none
+
+## TargetBuilderConfig
+
+key | value | default
+--- | --- | ---
+enabled | bool | true
+generate_for | [InputSet](#inputset) | `**`
+options | [BuilderOptions](#builderoptions) | none
```
natebosch: I don't think this link goes anywhere?

docs/build_yaml_format.md
```diff
@@ -0,0 +1,125 @@
...
+`name`.
+
+To construct a key, you join the package and name with a `:`, so for instance
+the `bar` target in the `foo` package would be referenced like this `foo:bar`.
+
+When referring to targets in the current package, you can omit the package and
+start with a `:` instead, so the above example could be shorted to `:bar`.
+
+Additionally, for any target you can omit the target name, and it will be
+defaulted to the package name, so `foo` would become `foo:foo`.
```
natebosch: We could consider removing some of the magic... this description is not telling the whole story. We perform different normalization when a target is _defined_ from when it is _referenced_. This description is of the reference. In a definition `foo` would become `<current package>:foo`.


Same thing with the builder below. As a definition `foo` becomes `<current package>|foo`, but as a reference it becomes `foo|foo`.

build_config/README.md
jakemac53: gah! ok :(

docs/build_yaml_format.md
jakemac53: done

docs/build_yaml_format.md
jakemac53: hmm true, I will just remove these paragraphs for now.

Issue #1514, opened 2018-06-05
Release build_config w/ new parser


kevmoo:
Cannot run --preview-dart-2 w/ Angular yet because of this - https://travis-ci.org/dart-lang/angular/jobs/388055311#L633

jakemac53:
I can work on publishing today - this should all be ready now.

PR #1515, opened 2018-06-05
release build_config 0.3.0

jakemac53:
Note that travis is going to fail for the unit tests in this package because it won't be able to get build_runner/build_runner_core dependencies without the overrides. I have ran the tests locally with the overrides in place though.

I will merge/publish this, then send out a PR to publish build_runner_core, and then build_runner after that, at which point master will go green again.

Fixes https://github.com/dart-lang/build/issues/1514

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,6 +141,10 @@ jobs:
       env: PKG="scratch_space"
       dart: dev
 
+matrix:
+  allow_failures:
+    - env: PKG="build_config"
+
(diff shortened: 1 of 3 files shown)


jakemac53:
@natebosch just fyi I am testing out simply allowing failures for build_config so other jobs can at least run

jakemac53:
other crashes are related to the latest sdk, next release should fix those though

PR #1522, opened 2018-06-06
meta module handle conditional imports

nickclmb:
Hi everyone,

At the moment if you have conditional imports in dart file, meta module will not have them in module dependencies:
```dart
import 'package:RAL/html_i.dart'
if (dart.library.html) 'dart:html'
if (dart.library.io) 'package:RAL/io/html_io.dart';
```
As a result the module will have dependency only on RAL/html_i.dart, missing RAL/io/html_io.dart
This pull request fixes that.

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -89,10 +89,31 @@ class _AssetNode {
               'same package');
         }
         parts.add(linkedId);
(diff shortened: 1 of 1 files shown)


nickclmb:
Signed the cla

nickclmb:
Ok, fixed profile, check again pls

Issue #1041, opened 2018-02-22
Add a way of dumping the build performance information


jakemac53:
Today we only support viewing it in `serve` mode, but that doesn't work for custom build scripts that use `build` or `watch` directly.

PR #1533, opened 2018-06-07
Make performance information serializable

jakemac53:
Work towards https://github.com/dart-lang/build/issues/1041

Note that the approach here is a little bit weird because we have two sets of classes - the base classes which only expose performance related information (and used to be entirely abstract) - and the versions that actually do the tracking itself.

Only the "view" classes are actually serializable (they are the only public ones also), and they are no longer abstract. When deserializing you can just use the `fromJson` constructors on these directly, and both implementations mix in the _same_ `toJson` method which is generated from the base classes.

### DIFF
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_runner
-version: 0.8.10
+version: 0.9.0-dev
 description: Tools to write binaries that run builders.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_runner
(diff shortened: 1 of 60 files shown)


jakemac53:
cc @kevmoo just for fyi

build_runner/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   build: ">=0.12.4 <0.12.7"
   build_config: ^0.3.0
   build_resolvers: ^0.2.0
-  build_runner_core: ^0.1.0
+  build_runner_core: ">=0.2.0 <0.2.1"
```
kevmoo: Always good to put a comment here why this is so constrained – for other maintainers, etc

jakemac53: done

PR #1535, opened 2018-06-07
Add `logPerformanceDir` to `BuildOptions`

jakemac53:
Next step towards https://github.com/dart-lang/build/issues/1041

The next step will be adding the flag to build_runner.


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -267,6 +273,18 @@ class _SingleBuild {
...
       });
 
+      // Log performance information if requested
+      if (_logPerformanceDir != null) {
+        assert(result.performance != null);
+        var logPath =
+            p.join(_logPerformanceDir, new DateTime.now().toIso8601String());
+        await logTimedAsync(_logger, 'Writing performance log to $logPath', () {
+          var performanceLogId = new AssetId(_packageGraph.root.name, logPath);
+          var serialized = jsonEncode(result.performance.toJson());
```
natebosch: [optional] the call to `toJson` isn't required here, just that `toJson` exists.

PR #1536, opened 2018-06-07
add --log-performance <dir> argument

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1041

### DIFF
--- a/build_runner/lib/src/entrypoint/base_command.dart
+++ b/build_runner/lib/src/entrypoint/base_command.dart
@@ -55,6 +55,9 @@ abstract class BuildRunnerCommand extends Command<int> {
           help: r'Enables performance tracking and the /$perf page.',
           negatable: true,
           defaultsTo: false)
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/entrypoint/base_command.dart
```diff
@@ -77,7 +77,10 @@ abstract class BuildRunnerCommand extends Command<int> {
           help: 'Build with release mode defaults for builders.')
       ..addMultiOption(defineOption,
           splitCommas: false,
-          help: 'Sets the global `options` config for a builder by key.');
+          help: 'Sets the global `options` config for a builder by key.')
+      ..addOption(logPerformanceOption,
```
natebosch: [nit] maybe position next to the `trackPerformanceOption` instead of at the end?

Also make a not either here or with the other option that this implies the other.

jakemac53: doen

PR #1540, opened 2018-06-08
skip the asset graph version test which is really flaky on windows

jakemac53:


jakemac53:
We have a general windows flakyness bug filed, and its pretty self commenting imo (the Skip instance has the reason and it's reported on the command line). I don't think its worth actually resolving at this time.

PR #1543, opened 2018-06-08
fix the report failure script

jakemac53:


jakemac53:
actually this is working I guess, it just doesn't catch build errors or something?

Issue #1534, opened 2018-06-07
FileBasedAssetReader.findAssets() stream is prematurely terminated following a Glob.list() exception


chalin:
When testing a new builder from [site-webdev PR #1624](https://github.com/dart-lang/site-webdev/pull/1624), I faced the following issue. The [build.doc.yaml](https://github.com/dart-lang/site-webdev/blob/bf62af2fe2e05e67577821ac46a815c1af84269e/build.doc.yaml) contains the following `targets.$default.sources.include` glob:

- `examples/html/{lib,test,web}/**`

The problem is that there is no `web` folder under `examples/html`. While `Glob.list()` reports the exception:

    FileSystemException: Directory listing failed, path = '.../site-webdev/examples/html/web/' (OS Error: No such file or directory, errno = 2)

The exception gets silently gobbled up, and the element generation for the asset stream stops.

---

The point at which the `FileSystemException` seems to get caught and dropped is in the `FileBasedAssetReader.findAssets()` function:

https://github.com/dart-lang/build/blob/526457f2da5c97b37e65b736bd807212fcc7d2e6/build_runner_core/lib/src/asset/file_based.dart#L41

In particular this `try-catch` block:

https://github.com/dart-lang/build/blob/526457f2da5c97b37e65b736bd807212fcc7d2e6/build_runner_core/lib/src/asset/file_based.dart#L51-L62

Because of the async nature of the directory tree generation, I was seeing some files from under `examples/html` but not always the same collection of files.

It would probably be better to handle async errors (and report them as warnings) using, say, [stream.handleError()](https://api.dartlang.org/dev/2.0.0-dev.60.0/dart-async/Stream/handleError.html), and to iterate with the `await for` over the error free stream. In this way, errors won't prevent any of the input targets from being listed.

cc @kevmoo, @kwalrath

jakemac53:
Ah! Thank makes sense - we should not just be swallowing that exception. Thanks for investigating!

PR #1545, opened 2018-06-08
Fix error handling with missing directories when globbing

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1534

The asset reader itself is now returning the stream from the glob api, just filtered a bit (but not handling the errors).

The `BuildDefinition` class which is the only thing that actually directly uses this api now swallows the errors explicitly. 

cc @chalin

### DIFF
--- a/build_runner_core/lib/src/generate/build_definition.dart
+++ b/build_runner_core/lib/src/generate/build_definition.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:build/build.dart';
 import 'package:glob/glob.dart';
 import 'package:logging/logging.dart';
+import 'package:stream_transform/stream_transform.dart';
 import 'package:watcher/watcher.dart';
 
 import '../asset/build_cache.dart';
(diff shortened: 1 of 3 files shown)


build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -374,26 +374,43 @@ class _Loader {
...
-      yield* _environment.reader
-          .findAssets(glob, package: targetNode.package.name)
-          .where((id) => !targetNode.excludesSource(id));
-    }
+  Stream<AssetId> _listAssetIds(TargetNode targetNode) {
+    var controller = new StreamController<AssetId>();
+    () async {
+      try {
+        for (final glob in targetNode.sourceIncludes) {
+          await controller.addStream(
```
natebosch: do we want to try to fire off multiple glob searches concurrently? I don't think we were before, but we could try it.

```dart
Stream<T> _mergeAll<T>(Iterable<Stream<T>> streams) => streams.first.transform(mergeAll(streams.skip(1)));

Stream<AssetId> _listAssetIds(TargetNode targetNode) {
  return _mergeAll(targetNode.sourceIncludes.map((glob) =>
      _listIdsSafe(glob, package: targetNode.package.name)
          .where((id) => !targetNode.excludesSource(id)));
}
```

jakemac53: done

Issue #1523, opened 2018-06-06
folder with hyphen --> ddc runtime error


nilsdoehring:
Area: webdev, dartdevc

Running a script from a folder with a hyphen in its name results in a ddc runtime error.

Example project layout
![image](https://user-images.githubusercontent.com/1327275/41024842-9be5cc8e-6970-11e8-881e-e4081801a56a.png)

Chrome console output:
```
Uncaught ReferenceError: folder__main is not defined
    at main.dart.bootstrap.js:153
    at Object.execCb (require.js:1696)
...
```

Steps to reproduce:
* Create a `Bare-bones Web App` in IDEA via `New Project -> Dart`
* Create a folder with a hyphen under `web/` (or anywhere else)
* Move files from `web/` to `web/hyphen-folder`
* Debug `web/hyphen-folder/index.html`

The error comes from the hyphen being sanitized incorrectly/differently in generated js,
* in _main.dart.bootstrap.js_: `app.hyphen-folder__main.main();`
* in _main.ddc.js_: `hyphen$45folder__main`

IDEA EAP 2018.2
SDK 2.0.0-dev.60.0
  build_runner: ^0.8.9
  build_test: ^0.10.2
  build_web_compilers: ^0.4.0+3
OSX, Chrome

jakemac53:
Should get a version published next week most likely

PR #1547, opened 2018-06-08
convert hyphens to $45 in ddc bootstrap code

jakemac53:
fixes https://github.com/dart-lang/build/issues/1523


jakemac53:
cc @jmesserly is there something more sane we can do here?

Basically when bootstrapping ddc we have to try and recreate the original module name today, which is fairly error prone and has weird edge cases like this.

jmesserly:
yes, DDC needs to export a good entry point: https://github.com/dart-lang/sdk/issues/31516

I'm looking into these issues generally. I have an old CL out that did address them (https://dart-review.googlesource.com/c/sdk/+/42885) ... but I need to find a migration path for internal code/anyone using the "legacy" module format.

it may be possible to fix #31516 first, if that would help you.

jmesserly:
FYI, the relevant functions in DDC for escaping the library names are https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/lib/src/analyzer/code_generator.dart#L6390 and https://github.com/dart-lang/sdk/blob/master/pkg/dev_compiler/lib/src/compiler/module_builder.dart#L324

jakemac53:
I knew this was familiar haha.

Yes, resolving  dart-lang/sdk#31516 would help here, but if you have other plans coming down the pipeline I am happy to just copy the code you pointed at for now as well.

jmesserly:
SGTM. Yeah I'm juggling a lot of things right now. :)

PR #1550, opened 2018-06-10
fix(runner_core): don't double-initialize builders

kevmoo:
expandedPhases is an Iterable that is accessed twice with a call to
.where causing double iteartion – and double initialization of each
builder

This code goes through the same items, but only once.

### DIFF
--- a/build_runner_core/lib/src/package_graph/apply_builders.dart
+++ b/build_runner_core/lib/src/package_graph/apply_builders.dart
@@ -258,19 +258,24 @@ Future<List<BuildPhase>> createBuildPhases(
             return targetGraph.allModules[key];
           })?.where((n) => n != null));
   final applyWith = _applyWith(builderApplications);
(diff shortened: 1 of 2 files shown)


kevmoo:
Oops! Not sure how to test this...

jakemac53:
You should be able to make a builder factory that increments a global counter and expect it to have a certain value afterwords?

kevmoo:
Can I land this and leave the test for a TODO for one of you guys w/ more context?

jakemac53:
How about I just push a test to this branch, it shouldn't take to long

jakemac53:
test added

natebosch:
tweaked it some to keep it closer to the old pattern and add a `toList()` to avoid the double instantiate.  In the longer term once we can use `whereType` we'll want it to look like this rather than a for loop.

jakemac53:
3 people on one PR! its a new record! :medal_sports:

kevmoo:
So I will land once travis is green? Thanks for your help here!

Issue #1513, opened 2018-06-04
Fix build_web_compilers test failure with Dart2


kevmoo:
https://travis-ci.org/dart-lang/build/jobs/388021356#L534

```
  Expected: contains AssetId:<a|web/index.dart.js>
    Actual: ?:[]
  Builder failed to write asset a|web/index.dart.js
  
  package:test                                    expect
  package:build_test/src/test_builder.dart 49:7   checkOutputs.<fn>
  dart:collection                                 __InternalLinkedHashMap&_HashVMBase&MapMixin&_LinkedHashMapMixin.forEach
  package:build_test/src/test_builder.dart 41:13  checkOutputs
  package:build_test/src/test_builder.dart 141:3  testBuilder
  ===== asynchronous gap ===========================
  dart:async                                      _AsyncAwaitCompleter.completeError
  package:build_test/src/test_builder.dart        testBuilder
  ===== asynchronous gap ===========================
  dart:async                                      _asyncThenWrapperHelper
  package:build_test/src/test_builder.dart        testBuilder
  test/dart2js_bootstrap_test.dart 54:11          main.<fn>
```

jakemac53:
https://github.com/dart-lang/build/pull/1517 just marks this as skipped for now, we can leave this issue open just so we remember to unskip it once the SDK is fixed.

PR #1553, opened 2018-06-11
unskip dart2js bootstrap test that passes now

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1513


jakemac53:
merging on red as we need to fix a bunch of deprecation issues

Issue #1370, opened 2018-04-26
Add a Global builder options section to build.yaml


natebosch:
Strawman:

```yaml
global_options:
  builder_package|builder_name:
    options:
      key: "applies anywhere this builder is already running"
    release_options:
      key: "applies in release mode anywhere this builder is already running"
   another_package|another_builder:
     enabled: true # acts as if this builder was `apply_to: all_packages` and overrides any package disabling it?
    third_package|third_builder:
      enabled: false # builder will never run anywhere?
targets:
  $default:
    builders:
      builder_package|builder_name:
        options:
          key: "never applies because it's overridden globally?"
```

Open questions:

- Should global config take highest precedence in terms of overrides? This is the case today with `--define`
- Should we allow enabling and disabling builders globally? Note that this would change the structure of the build so would cause an invalidation and full-rebuild if flipping with `--config`

matanlurey:
> Open questions:
> 
> - Should global config take highest precedence in terms of overrides? This is the case today with `--define`

Yes, I think that is the most understandable to most users.

> - Should we allow enabling and disabling builders globally? Note that this would change the structure of the build so would cause an invalidation and full-rebuild if flipping with `--config`

Yes. Though I think we should issue an `info` that makes this clear, i.e.:

`INFO: By using a global configuration option, the build cache is being cleared` (or better text)

FWIW, Bazel gets around this by having multiple caches, but I can understand not doing this.

jakemac53:
> Should global config take highest precedence in terms of overrides? This is the case today with --define

I think --define should still override it, but it should probably override everything else. The one case that gets weird is if you have a target level override in the root package, then you probably want that.......

jakemac53:
> FWIW, Bazel gets around this by having multiple caches, but I can understand not doing this.

It honestly might make sense for us to do this, for each --config and release/debug combination. The biggest problem I see is it makes the analysis server issue even more challenging, which cache should it read from? If we create a special `.packages` file for generated files, which cache should that point to? Create one per configuration/mode?

jakemac53:
> Should we allow enabling and disabling builders globally?

I think we should provide all the same options we do for targets, so yes. This gives people a way of disabling a bad builder if necessary to unblock themselves.

> Note that this would change the structure of the build so would cause an invalidation and full-rebuild if flipping with --config

I think that is fine, we shouldn't see many uses of `--config` in the wild once everybody migrates to `--release`.

natebosch:
[...]

PR #1560, opened 2018-06-12
Add build_config support for global options

natebosch:
Towards #1370

Not including `enabled` for now since it's not clear we'll want it.

- Add `GlobalBuilderOptions` class since it is slightly different than
  existing similar classes. Add parsing
- Document in `README.md` and `build_yaml_format.md`
- Include parsing and result checking in the parse test.

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -56,6 +56,26 @@ configuration may have the following keys:
   mode with `dev_options` or `release_options`.
 - **dev_options**: Map, Optional: A free-form map which will be passed to the
   `Builder` as a `BuilderOptions` when it is constructed. Usage varies depending
+  on the particular builder. The values in this map override Builder defaults or
+  non mode-specific options per-key when the build is done in dev mode.
+- **release_options**: Map, Optional: A free-form map which will be passed to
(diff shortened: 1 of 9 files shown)


build_config/README.md
```diff
@@ -54,6 +54,26 @@ configuration may have the following keys:
...
+- **dev_options**: Map, Optional: A free-form map which will be passed to the
+  `Builder` as a `BuilderOptions` when it is constructed. Usage varies depending
+  on the particular builder. The values in this map override Builder defaults or
+  non mode-specific options per-key when the build is done in dev mode.
+- **release_options**: Map, Optional: A free-form map which will be passed to
+  the `Builder` as a `BuilderOptions` when it is constructed. Usage varies
+  depending on the particular builder. The values in this map override Builder
+  defaults or non mode-specific options when the build is done in release mode.
+
+## Configuring `Builder`s globally
```
grouma: Discussed offline. It would be useful to have some sample configuration or use case for the global_options.

natebosch: yeah I think that will come once we get to https://github.com/dart-lang/build/issues/1308#issuecomment-388938427 we'll add an example to our docs which uses global options

PR #1561, opened 2018-06-13
Honor `global_options` from `build.yaml`

### LINKED ISSUE #1370: Add a Global builder options section to build.yaml

natebosch:
Closes #1370

- Add merging of the `global_options` before the overrides.
- Add a test using global options.
- Add a test for the error message when specifying an invalid key in the
  global options.

### DIFF
--- a/build_runner/test/generate/build_integration_test.dart
+++ b/build_runner/test/generate/build_integration_test.dart
@@ -409,7 +409,7 @@ main(List<String> args) async {
       return '${result.stdout}';
     }
 
-    test('warns on invalid builder key --define', () async {
+    test('warns on invalid builder key in target options', () async {
       await d.dir('a', [
         await pubspec('a', currentIsolateDependencies: [
           'build',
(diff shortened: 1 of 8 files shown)


build_runner/test/generate/build_integration_test.dart
```diff
@@ -409,7 +409,7 @@ main(List<String> args) async {
       return '${result.stdout}';
     }
 
-    test('warns on invalid builder key --define', () async {
+    test('warns on invalid builder key in target options', () async {
```
jakemac53: Why did we remove the `--define` test?

jakemac53: ahhh ok

PR #1562, opened 2018-06-13
replaced path.join with path.url.join

AlexanderJohr:
Replaced path.join with path.url.join because p.join uses platform specific separators, p.uri.join is for uris (and the build system only deals in uris)

### DIFF
--- a/build/lib/src/asset/id.dart
+++ b/build/lib/src/asset/id.dart
@@ -59,10 +59,10 @@ class AssetId implements Comparable<AssetId> {
     if (parsedUri.hasScheme) {
       if (parsedUri.scheme == 'package') {
         return new AssetId(parsedUri.pathSegments.first,
-            p.join('lib', p.joinAll(parsedUri.pathSegments.skip(1))));
+            p.url.join('lib', p.url.joinAll(parsedUri.pathSegments.skip(1))));
       } else if (parsedUri.scheme == 'asset') {
         return new AssetId(parsedUri.pathSegments.first,
-            p.joinAll(parsedUri.pathSegments.skip(1)));
+            p.url.joinAll(parsedUri.pathSegments.skip(1)));
       }
       throw new UnsupportedError(
           'Cannot resolve $uri; only "package" and "asset" schemes supported');
@@ -72,7 +72,7 @@ class AssetId implements Comparable<AssetId> {
           'An AssetId "from" must be specified to resolve a relative URI');
     }
     return new AssetId(
-        p.normalize(from.package), p.join(p.dirname(from.path), uri));
+        p.normalize(from.package), p.url.join(p.dirname(from.path), uri));
   }
 
   /// Parses an [AssetId] string of the form "package|path/to/asset.txt".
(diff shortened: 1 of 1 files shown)


jakemac53:
restarting tests now that a package:path fix has landed

build/lib/src/asset/id.dart
```diff
@@ -72,7 +72,7 @@ class AssetId implements Comparable<AssetId> {
           'An AssetId "from" must be specified to resolve a relative URI');
     }
     return new AssetId(
-        p.normalize(from.package), p.join(p.dirname(from.path), uri));
+        p.normalize(from.package), p.url.join(p.dirname(from.path), uri));
```
jakemac53: `p.normalize` and `p.dirname` also need to be changed to the `.url` equivalents I believe

jakemac53:
Also, note that the AssetId constructor itself looks like it should be fixing these up... https://github.com/dart-lang/build/pull/1562/files#diff-4adccb0af60d55da474afae8bc3ad5f6R33

AlexanderJohr:
The AssetId constructor does fix it up, that is correct. But if the _normalizePath should be used, then it is missing on the `resolve` factory constructor:
https://github.com/dart-lang/build/pull/1562/files#diff-4adccb0af60d55da474afae8bc3ad5f6R58

Uri.parse(uri) parses the uri with backslashes incorrectly and then Uri.parse("directory\file.txt").pathSegments returns ["directory\file.txt"] instead of ["directory", "file.txt"]:
https://github.com/dart-lang/build/pull/1562/files#diff-4adccb0af60d55da474afae8bc3ad5f6R62

But if _normalizePath is used in both constructors it is executed twice, which doesn't seem like a good design.

I think instead of fixing the path with _normalizePath and generating bad paths with path.x instead of path.url.x to let _normalizePath fix it later, it should rather use path.url. everywhere and check in the constructor if a bad path was used as an argument.

But what is your advice on this?

jakemac53:
I think you are correct that we should use `p.url` throughout to reduce the wasted effort here. See my one comment about the place that was missed though (it may not matter much but would be good for consistency to clean it up as well).

It looks like this method will fail given a uri like `package:foo\bar.dart` or `asset:foo\bar.dart`, so if you would like to add in a throw if/when we detect those situations that would also be totally reasonable.

AlexanderJohr:
What is wrong with the tests? It seems that compile-time errors come from files that I haven't touched. like this one: https://travis-ci.org/dart-lang/build/jobs/395049217

It complains:
```
[error] Expected a method, getter, setter or operator declaration. (/test/hello_world_test.dart, line 11, col 7)
[error] Expected a string literal. (/test/hello_world_test.dart, line 11, col 7)
[error] Expected to find ';'. (/test/hello_world_test.dart, line 11, col 7)
[error] Invalid URI syntax: ''. (/test/hello_world_test.dart, line 11, col 7)
[error] Unexpected text ':'. (/test/hello_world_test.dart, line 11, col 7)
```

But it is a comment:
https://github.com/AlexanderJohr/build/blob/master/_test/test/hello_world_test.dart#L11

PR #1563, opened 2018-06-13
Drop exception and stack trace from BuildResult

### LINKED ISSUE #1209: Drop the `exception` field from `BuildResult`
After https://github.com/dart-lang/build/pull/1208 this field is going to be a lot less meaningful - there are potentially multiple exceptions in disparate actions that cause a failing overall build.

We either need to expand this allow for multiple exceptions, or drop it and find another way to communicate why a build fails to the serve handler. Currently we only use this to render a page showing why a build failed - we could also do that through the AssetGraph or some other mechanism.

This would be technically breaking.

natebosch:
Closes #1209

- Migrate the one remaining use case for throwing a
  `FatalBuildException` to handle the logging itself and remove special
  handling for this exceptio type in the build result.
- Remove all uses of the `exception` field. Most uses were already
  irrelevant because exceptions should already be logged.

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -5,7 +5,6 @@
 import 'dart:async';
 import 'dart:collection';
 import 'dart:convert';
-import 'dart:io';
 
 import 'package:build/build.dart';
 import 'package:build_resolvers/build_resolvers.dart';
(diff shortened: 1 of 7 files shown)


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -204,31 +203,23 @@ class _SingleBuild {
...
+        result = _convertToFailure(result, failureType: FailureType.cantCreate);
       }
     }
     if (result.status == BuildStatus.success) {
       _logger.info('Succeeded after ${humanReadable(watch.elapsed)} with '
           '${result.outputs.length} outputs ($actionsCompletedCount actions)\n');
     } else {
-      if (result.exception is FatalBuildException) {
-        // TODO(???) Really bad idea. Should not set exit codes in libraries!
-        exitCode = 1;
```
jakemac53: Lets add this to the changelog as well

natebosch: Discussed offline, will followup with a PR that removes the one other place we set `exitCode` and update the changelog.

PR #1565, opened 2018-06-14
MetaModule: use parseDirectives instead of parseCompilationUnit

nickclmb:
The MetaModule builder collects dependencies of each asset, and it only needs import/export directives from dart file. So using the analyzer's parseCompilationUnit is overkill, and parseDirectives should be enough. 

As a result - less time for parsing, and less memory for analyzer cache.

Signed-off-by: Nikolay Samokhin <nickclmb@gmail.com>

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -262,8 +262,8 @@ Future<List<Module>> _computeModules(
   for (var asset in assets) {
     var content = await reader.readAsString(asset);
     // Skip errors here, dartdevc gives nicer messages.
-    var parsed = parseCompilationUnit(content,
-        name: asset.path, parseFunctionBodies: false, suppressErrors: true);
+    var parsed = parseDirectives(content,
+        name: asset.path, suppressErrors: true);
     parsedAssetsById[asset] = parsed;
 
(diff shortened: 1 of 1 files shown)


build_modules/lib/src/meta_module.dart
```diff
@@ -262,8 +262,8 @@ Future<List<Module>> _computeModules(
   for (var asset in assets) {
     var content = await reader.readAsString(asset);
     // Skip errors here, dartdevc gives nicer messages.
-    var parsed = parseCompilationUnit(content,
-        name: asset.path, parseFunctionBodies: false, suppressErrors: true);
+    var parsed =
+        parseDirectives(content, name: asset.path, suppressErrors: true);
```
jakemac53: Note that if `public` is false then we do actually need the whole compilation unit because `_isEntrypoint` checks for a `main` function. This could change to conditionally parse the entire compilation unit though only if `public` is `true`.

nickclmb: Ok, got you. Fixed that. Applying parseDirectives on lib/** files only still gives great memory win on big project :-)

jakemac53:
thanks @nickclmb !

PR #1566, opened 2018-06-14
add sdk sourcemaps to asset graph to let build_runner serve them

nickclmb:
Hi everyone,

Not sure if this pull request is valid, but anyway, i found that build_runner doesn't serve sdk source maps for me:
```
[INFO] WebServer 2018-06-14T18:11:07.063461  0:00:00.008181 GET     [404] /packages/$sdk/dev_compiler/amd/dart_sdk.js.map 
```

So i created this pull request to fix that.

Signed-off-by: Nikolay Samokhin <nickclmb@gmail.com>


jakemac53:
Ah yes we intentionally don't serve them right now because they are broken :(

Long story, but basically the sdk doesn't ship with all the original dart files - and there isn't a clear idea of where the ones that do exist would even be pulled from.

Additionally, the source map paths are totally broken absolute paths last I checked.

PR #1572, opened 2018-06-15
move to TypeMatcher

jakemac53:
fixes travis redness


jakemac53:
@natebosch forgot to update pubspecs initially, build_test also needs a version bump because it has a real dependency on package:test, but we don't need to release it right away.

PTAL

PR #1577, opened 2018-06-16
Convert build_config Matchers to TypeMatcher

natebosch:


build_config/test/build_config_test.dart
```diff
@@ -223,170 +223,106 @@ void expectPostProcessBuilderDefinitions(
...
-      equals(_expected.options.config).matches(item.options.config, _) &&
-      equals(_expected.devOptions.config).matches(item.devOptions.config, _) &&
-      equals(_expected.releaseOptions.config)
-          .matches(item.releaseOptions.config, _);
-
-  @override
-  Description describe(Description description) =>
-      description.addDescriptionOf(_expected);
-}
+Matcher _matchesBuilderDefinition(BuilderDefinition definition) =>
```
kevmoo: At what point do you just implement `==` on BuilderDefinition?

natebosch: The failure message isn't as nice.

I could be convinced that since these are data types they should implement equals anyway... but I'm not sure.

natebosch: It would also have to get into using DeepEquality and stuff

PR #1581, opened 2018-06-18
Add storage and re-reporting of previous errors

### LINKED ISSUE #910: `pub run build_runner build` returns exit code 0 even when SEVERE errors reported

natebosch:
Closes #910

The main benefit of the proposed `--rerun-failing-actions` is to be able
to see errors again from previous builds that have scrolled off screen.
The downside to rerunning is that it is a pattern encouraging retrying
non-hermetic actions. Instead we will store and re-print previous error
messages which more directly addresses the need.

This commit sketches out the interaction between the failure reporter
and the build - storage of errors to individual files will happen in a
followup.


build_runner_core/lib/src/logging/failure_reporter.dart
```diff
@@ -0,0 +1,74 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+
+import '../asset_graph/node.dart';
+
+/// A tracker for the errors which have already been reported during the current
```
jakemac53: nit: 80 chars (same below)

natebosch: This is at 80 :)

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -424,13 +426,12 @@ class _SingleBuild {
...
         .add(actionDescription);
 
-    var errorThrown = false;
     await tracker.track(
         () => runBuilder(builder, [input], wrappedReader, wrappedWriter,
                 new PerformanceTrackingResolvers(_resolvers, tracker),
                 logger: logger,
                 resourceManager: _resourceManager).catchError((_) {
-              errorThrown = true;
+              // Errors tracked throug logger
```
jakemac53: Do we need to clean up the previous errors before running this builder again? See the _cleanupStaleOutputs call on line 413 above

natebosch: oh yeah, I had put in `FailureReporter.clean` but forgot to call it. Calling it from within `_cleanUpStaleOutputs` now.

build_runner_core/lib/src/logging/failure_reporter.dart
jakemac53: 🤦‍♂️

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -621,11 +621,13 @@ class _SingleBuild {
...
-    await Future.wait(outputs.map((output) {
-      var node = _assetGraph.get(output) as GeneratedAssetNode;
-      if (node.wasOutput) return _delete(output);
+    var nodes =
+        outputs.map((o) => _assetGraph.get(o) as GeneratedAssetNode).toList();
+    await Future.wait(nodes.map((node) {
+      if (node.wasOutput) return _delete(node.id);
       return new Future.value(null);
     }));
+    await FailureReporter.clean(nodes.first);
```
jakemac53: lets leave a comment about why we only call this for the first node

natebosch: ended up removing this approach and it's cleaning based on the primary input now.

build_runner_core/lib/src/logging/failure_reporter.dart
```diff
@@ -0,0 +1,117 @@
...
+  Future<void> markSkipped(Iterable<GeneratedAssetNode> outputs) =>
+      Future.wait(outputs.map((output) async {
+        if (!_reportedActions.add(_actionKey(output))) return;
+        await clean(output.phaseNumber, output.primaryInput);
+      }));
+
+  /// Log stored errors for any build steps which would output nodes in
+  /// [failingNodes] which haven't already been reported.
+  Future<void> reportErrors(Iterable<GeneratedAssetNode> failingNodes) async {
+    for (final failure in failingNodes) {
```
jakemac53: it is probably worth parallelizing this

natebosch: Done. Do you think I should add a file handle pool? This isn't really in the critical path for a build so it might not make much difference.

jakemac53: hmm, you could just use the asset reader as well which would handle this? If somebody has a lot of cascading errors it could cause problems.

natebosch: I'd prefer to leave it using `dart:io` for now, for consistency.

Since we do directory level work we can't use asset reader/writer for everything.

Issue #1584, opened 2018-06-18
Add back option to pass in a custom Resolvers instance


natebosch:
We had removed this since it was unused, but a new use case has come up where we can pass in a more optimized resolver which might not work for all use cases.

natebosch:
Hmmm, part of the reason we removed this option may have been that we need to know inside `build_impl.dart` that it is specifically an `AnalyzerResolvers` instead of a general `Resolvers`.

PR #1585, opened 2018-06-18
Allow constructing with custom AnalysisOptions

natebosch:
Towards #1584


build_resolvers/lib/src/resolver.dart
```diff
@@ -83,8 +83,9 @@ class AnalyzerResolver implements ReleasableResolver {
   /// later). If this happens we don't want to hide the asset from the analyzer.
   final _seenAssets = new Set<AssetId>();
 
-  AnalyzerResolver(DartUriResolver dartUriResolver) {
```
kevmoo: Is this a non-exposed class?

natebosch: Yes, `AnalyzerResolvers` is the only exported class.

kevmoo:
Cool...

PR #1591, opened 2018-06-20
Caching for simultaneously resolved assets

yury-yufimov:
It seems, that in large projects in gap between 'updateDependencies' and 'updateContents' for asset, another 'updateDependencies' can be called from another resolve process. 
So, this is attempt to cache sources for calculation of dependencies immediately, for not calling analyzer second time


yury-yufimov:
CLA signed

jakemac53:
This looks safe for me - there is probably a bigger refactor that should eventually be done here but I think this is reasonable for now.

PR #1592, opened 2018-06-20
Drop build_barback package

natebosch:
- Switch to build_resolvers everywhere.
- Drop the package.
- Remove references in the README.
- Remove Travis config.


kevmoo:
In the spirit of good stewardship, should we publish one more and tell folks it's no longer supported?

natebosch:
> should we publish one more and tell folks it's no longer supported?

I don't think it's worth it. For use cases other than the Resolvers it's already been effectively dead for a while.

pulyaevskiy:
Oh... sorry guys, but if you drop a package you should definitely tell people about it. I can't stress enough how important this is.

It turns out that pointycastle depends on reflectable and reflectable still depends on build_barback. 

There are actually quite a few packages that depend on reflectable: https://pub.dartlang.org/packages?q=dependency%3Areflectable

If there is a replacement (you mentioned something named `build_resolvers` multiple times) then everyone would really appreciate if you provide an upgrade path as well.

Please don't forget about people outside of google if you decided to make this open source.

jakemac53:
This package effectively hit EOL and will not get further updates - primarily because barback itself is now deprecated/archived.

We did not delete the package from pub though - it is simply not being maintained any more because it is not feasible to do so.

jakemac53:
Also to clarify a bit it a large part of the reasoning here is that barback isn't getting migrated to dart 2 - so therefor this package can't be migrated to dart 2. Since the repo only supports dart 2 now we can't maintain it.

kevmoo:
Please file an issue on reflectable to update to pkg:build – that's the
direction we should be going!

jakemac53:
https://github.com/dart-lang/reflectable/issues/136 effectively covers that i believe

natebosch:
https://github.com/dart-lang/reflectable/issues/130

PR #1602, opened 2018-06-21
Add more missing integration test imports

natebosch:
Dart 1 treated these missing types as `dynamic`, but Dart 2 doesn't do
that and we get an invalid override.


jakemac53:
just cancel travis/appveyor and submit IMO

PR #1605, opened 2018-06-21
List valid inputs

natebosch:


build_runner_core/test/generate/build_test.dart
```diff
@@ -1171,5 +1171,29 @@ void main() {
...
+    test('the entrypoint cannot be read by a builder', () async {
+      var builders = [
+        applyToRoot(new TestBuilder(
+            buildExtensions: replaceExtension('.txt', '.hasEntrypoint'),
+            build: (buildStep, _) async {
+              var hasEntrypoint = await buildStep
+                  .findAssets(new Glob('**'))
+                  .contains(
+                      makeAssetId('a|.dart_tool/build/entrypoint/build.dart'));
+              await buildStep.writeAsString(
```
jakemac53: presumably you could just literally put an `expect` in here? but this is fine too :)

natebosch: yeah, would have to wrap the entire thing with an `expectAsync2` as well.

PR #1608, opened 2018-06-22
track digests in scratch_space to speed up ensureAssets

jakemac53:
This allows us to not have to clear the scratch space between builds, and also eliminates a lot of `file.existsSync()` calls.

Will publish scratch_space after this PR, then send a PR to publish build_modules.

### DIFF
--- a/scratch_space/lib/src/scratch_space.dart
+++ b/scratch_space/lib/src/scratch_space.dart
@@ -105,7 +105,8 @@ class ScratchSpace {
                   await file.writeAsBytes(await reader.readAsBytes(id));
                 }));
       } finally {
-        _pendingWrites..remove(id);
+        // ignore: unawaited_futures
+        _pendingWrites.remove(id);
       }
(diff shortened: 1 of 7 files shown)


scratch_space/lib/src/scratch_space.dart
```diff
@@ -80,30 +83,31 @@ class ScratchSpace {
...
+            () => _descriptorPool.withResource(() async {
+                  var file = fileFor(id);
+                  if (await file.exists()) {
+                    await file.delete();
+                  }
+                  await file.create(recursive: true);
+                  await file.writeAsBytes(await reader.readAsBytes(id));
+                }));
+      } finally {
+        _pendingWrites..remove(id);
```
natebosch: is the `..` necessary here?

jakemac53: It just gets rid of the unawaited futures hint

PR #1613, opened 2018-06-22
Discussion: Canonicalize AssetIds before storing them

natebosch:
If we get duplicate copies of Strings or AssetIds generated by Builders
and passed to the AssetReader we might inflate the size of the
AssetGraph.

### DIFF
--- a/build_runner_core/lib/src/asset_graph/graph.dart
+++ b/build_runner_core/lib/src/asset_graph/graph.dart
@@ -505,6 +505,8 @@ class AssetGraph {
   // TODO remove once tests are updated
   void add(AssetNode node) => _add(node);
   Set<AssetId> remove(AssetId id) => _removeRecursive(id);
+
+  AssetId canonicalize(AssetId id) => get(id)?.id ?? id;
 }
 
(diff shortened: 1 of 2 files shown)


kevmoo:
@natebosch – any measurements on memory impact?

natebosch:
Working on getting the impact now. Uploaded a branch so that multiple people could try it

kevmoo:
Cool!

natebosch:
I'm not seeing this make any difference in the number of AssetIds in the heap when I'm running a watch on angular_component_example and running a few builds. It could be that build doesn't exhibit the behavior of doing String manipulation to construct asset ids very much.

I'm also noticing that even on a no-op build without doing an work that could cause non-canonical AssetIds in the AssetGraph the size in the heap is much bigger than the size on disk. It's likely that more of the memory is going towards collection overhead than to the AssetIds or Strings

kevmoo:
Can close out?

natebosch:
yeah I think we should close. We'll keep looking for other places to reduce memory usage but as mentioned above I'm not able to find concrete evidence that this is helping.

Issue #1619, opened 2018-06-28
AssetGraph placed in folder, based on ABSOLUTE path to build file


yury-yufimov:
We are using dart2.0.0-dev.63.0 and build_runner 0.9.0 for building our project.

We have strong intention to distribute initial state of BuildCache (.dart_tool content) to all developers, for reducing initial build time.

After build is finished, asset graph is placed in 
/Users/MY_USER_NAME/projects/MY_PROJECT_FOLDER/.dart_tool/build/...MD5_FOR_BUILD_SCRIPT.../asset_graph.json

It turns out, that MD5_FOR_BUILD_SCRIPT depends on absolute path to build script, like:
/Users/MY_USER_NAME/projects/MY_PROJECT_FOLDER/tools/grind.dart

That makes it quite difficult to distribute suitable cache for all developers.

Proposal: calculate MD5_FOR_BUILD_SCRIPT only from relative path within project folder and, may be, project name from pubspec.yaml

DisDis:
@kevmoo @natebosch @jakemac53

jakemac53:
Seems reasonable :+1:

PR #1621, opened 2018-06-28
use package relative path for the asset graph hash

jakemac53:
Closes https://github.com/dart-lang/build/issues/1619

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 0.2.0+1
+## 0.2.1
 
 - The hash dir for the asset graph under `.dart_tool/build` is now based on a
   relative path to the build script instead of the absolute path.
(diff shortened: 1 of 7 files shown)


build_runner_core/lib/src/util/constants.dart
```diff
@@ -19,7 +19,7 @@ final errorCachePath =
     p.url.join(cacheDir, _scriptHashFor(_scriptPath), 'error_cache');
 
 final String _scriptPath = Platform.script.scheme == 'file'
-    ? Platform.script.toFilePath()
+    ? p.relative(Platform.script.toFilePath(), from: p.current)
```
natebosch: This is probably an existing problem... but I think we might get difference hashes here for the same relative path between Windows and posix. Should we normalize that?

jakemac53: hmmm, good point, `p.url.join(p.split(...))`?

build_runner_core/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 0.2.0+1
```
natebosch: synced offline about making this `0.2.1` and pushing a  new version of `build_runner` as well. Fine to merge on red since we'll have a dep on an unpublished package.

PR #1629, opened 2018-06-30
Add command pub run build_runner run

### LINKED ISSUE #1159: Add command pub run build_runner run
... to mirror `pub run` with builders applied.

thosakwe:
Resolves #1159.

```
(virtualenv_py3) Tobes-MacBook-Air:example thosakwe$ pbr run --help
[INFO] Generating build script completed, took 692ms
Performs a single build on the specified targets, and executes a Dart script with the given arguments.

Usage: build_runner run [arguments]
-h, --help                          Print this usage information.
    --[no-]assume-tty               Enables colors and interactive input when the script does not appear to be running directly in a terminal, for instance when it is a subprocess
    --delete-conflicting-outputs    By default, the user will be prompted to delete any files which already exist but were not known to be generated by this specific build script.
                                    
                                    Enabling this option skips the prompt and deletes the files. This should typically be used in continues integration servers and tests, but not otherwise.

    --low-resources-mode            Reduce the amount of memory consumed by the build process. This will slow down builds but allow them to progress in resource constrained environments.
-c, --config                        Read `build.<name>.yaml` instead of the default `build.yaml`
    --[no-]track-performance        Enables performance tracking and the /$perf page.
    --log-performance               A directory to write performance logs to, must be in the current package. Implies `--track-performance`.
-o, --output                        A directory to copy the fully built package to. Or a mapping from a top-level directory in the package to the directory to write a filtered build output to. For example "web:deploy".
-v, --verbose                       Enables verbose logging.
-r, --[no-]release                  Build with release mode defaults for builders.
    --define                        Sets the global `options` config for a builder by key.

Run "build_runner help" to see global options.
```


thosakwe:
Note - I just realized that arguments passed to the script in question, specifically those with flags, will throw an error because the command runner doesn't recognize them.

I'll have to change it so that you pass them after a `--`, similar to the `pbr test` command.

thosakwe:
One final thing - if the user specifies an `output` directory, is there still a pointing creating a temp directory? I personally don't think so, and it might be useful for debugging if users can see the generated files, instead of them being deleted.

thosakwe:
Never mind what I said later - it seems that `package:args` has this functionality built in. I didn't make any changes to my fork, and yet it seems to "just work":

```
(virtualenv_py3) Tobes-MacBook-Air:tobe thosakwe$ pbr run bin/hello.dart -- a b --c --d=efg
[INFO] Generating build script completed, took 699ms
[INFO] Reading cached asset graph completed, took 153ms
[INFO] Checking for updates since last build completed, took 483ms
[INFO] Running build completed, took 16ms
[INFO] Caching finalized dependency graph completed, took 69ms
[INFO] Creating merged output dir `/var/folders/md/xrrg3dn921l5_qw7v2l3cb9w0000gn/T/build_runner_run_scriptdmg6EB/` completed, took 1.7s
[INFO] Writing asset manifest completed, took 5ms
[INFO] Succeeded after 1.8s with 0 outputs (0 actions)
Args: [a, b, --c, --d=efg]
```

Other than that, I think it's ready to go.

jakemac53:
Thanks! Note that with July 4th this week we are not likely to review this until next week.

In the meantime you could add integration tests, you can see the `build_runner/test/integration_tests` for some examples, or the `_test` directory. Or you can wait until after our initial review if you prefer :).

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+      '${super.invocation.replaceFirst('[arguments]', '[build-arguments]')} '
+      '<executable> [-- [script-arguments]]';
+
+  @override
+  SharedOptions readOptions() {
+    // The default option parser will throw if we pass additional arguments,
+    // because it expects positional arguments to be build directories.
+    //
+    // Instead, pass an empty list, and we'll handle positional arguments ourselves.
+    return new SharedOptions.fromParsedArgs(
```
jakemac53: See the [Test command](https://github.com/dart-lang/build/blob/master/build_runner/lib/src/entrypoint/test.dart#L33) implementation of this, we should do something similar here to ensure that we only have a single extra argument before the `--` (the script to run), and then additional extra arguments are after the `--`.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+      onError.listen((e) {
+        isolate?.kill();
+        onExit.close();
+        onError.close();
+        Zone.current.handleUncaughtError(
+            e[0], new StackTrace.fromString(e[1].toString()));
+      });
+
+      onExit.listen((_) {
+        onExit.close();
```
jakemac53: nit: I would just move this cleanup into the `finally` block (and use `?.`). Then you can delete these same lines from the `catch` block as well.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+    // Instead, pass an empty list, and we'll handle positional arguments ourselves.
+    return new SharedOptions.fromParsedArgs(
+        argResults, [], packageGraph.root.name, this);
+  }
+
+  @override
+  FutureOr<int> run() async {
+    // Ensure that the user passed the name of a file to run.
+    if (argResults.rest.isEmpty) {
+      stderr
```
jakemac53: In general, we try not to write directly to stderr, except for unhandled exceptions.

Instead, you can create a new `Logger` (probably with `Run` as the name) and then use `_logger.severe(...)`.

Same goes for the usages of stderr below.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+        stdout.writeln('Skipping script run due to build failure');
+        return result.failureType.exitCode;
+      }
+
+      // Find the path of the script to run.
+      var scriptPath = p.setExtension(p.join(tempPath, scriptName), '.dart');
+      var packageConfigPath = p.join(tempPath, '.packages');
+
+      // Use a completer to determine the exit code.
+      var completer = new Completer<int>();
```
jakemac53: nit: more descriptive name (exitCodeCompleter?)

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+        packageConfig: p.toUri(packageConfigPath),
+      );
+
+      return await completer.future;
+    } on IsolateSpawnException catch (e) {
+      stderr.writeln(e);
+      onExit?.close();
+      onError?.close();
+      stderr.writeln(
+          'Could not spawn isolate. Ensure that your file is in a valid directory (i.e. "lib", "web", "test).');
```
jakemac53: The general recommended directories for vm apps are `bin`, `test`, `tool`, `benchmark`, `example`, but not `lib`. Basically, anything _but_ `lib` ;). Flutter violates this convention but we don't want to encourage it as standard practice.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+      Isolate isolate;
+      onExit = new ReceivePort();
+      onError = new ReceivePort();
+
+      // On an error, kill the isolate, and rethrow the error.
+      onError.listen((e) {
+        isolate?.kill();
+        onExit.close();
+        onError.close();
+        Zone.current.handleUncaughtError(
```
jakemac53: I think we probably just want to `_logger.severe('Unhandled error from script: $scriptName', e[0], e[1])` instead of rethrowing it (which is effectively what this does I think?).

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+      // Use a completer to determine the exit code.
+      var completer = new Completer<int>();
+
+      Isolate isolate;
+      onExit = new ReceivePort();
+      onError = new ReceivePort();
+
+      // On an error, kill the isolate, and rethrow the error.
+      onError.listen((e) {
+        isolate?.kill();
```
jakemac53: Are we sure we want this to exit on unhandled errors? I don't know if its clear either way what should be done.

thosakwe:
I'll be able to address the review tomorrow, woohoo!

jakemac53:
@thosakwe anything you need from us here?

jakemac53:
Note that we now also have the `build_vm_compilers` package - ideally we would actually use the precompiled kernel from that but I don't know how that would affect hot reload.

I do know its possible to do hot reload with precompiled kernel though because flutter does it.

thosakwe:
Sorry, had forgotten I had this PR open! I'll send my changes in a little bit.

thosakwe:
Also, that's a good point. Would `pbr run` require users to use `build_vm_compilers`? As-is, I think someone could just say `pbr run example/main.dill` (or whatever the kernel extension happens to be).

jakemac53:
> Also, that's a good point. Would pbr run require users to use build_vm_compilers? As-is, I think someone could just say pbr run example/main.dill (or whatever the kernel extension happens to be).

We do have some precedent for requiring a dep, for instance `pbr serve` warns if you don't have a build_web_compilers dep, and `test` errors I believe if you don't have a `build_test` dep.

That being said, I think `run` is a little different since it can work ok without build_vm_compilers, its just going to take longer to start up.

jakemac53:
Maybe we could make the command be smart and use precompiled dill files if available - there is precedent already set in the test package for this for instance.

jakemac53:
However, I wouldn't block this cl on that, just floating ideas :)

kevmoo:
Where are we here, @jakemac53 ?

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,145 @@
...
+import 'base_command.dart';
+import 'options.dart';
+
+class RunCommand extends BuildRunnerCommand {
+  @override
+  String get name => 'run';
+
+  @override
+  String get description =>
+      'Performs a single build on the specified targets, and executes a Dart script with the given arguments.';
```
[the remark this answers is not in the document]
thosakwe: Removed "on the specified targets" now.

Will address the second portion of this comment in a few minutes...

build_runner/lib/src/entrypoint/run_script.dart
thosakwe: Ok, changed it - I've more or less just copied the implementation from the test command at this point.

build_runner/lib/src/entrypoint/run_script.dart
thosakwe: In my head, it makes sense to kill the isolate after it crashes, mainly since the running script would have stopped executing, and we printed the error to the screen.

I can't think of a reason where you'd want to keep the isolate alive (plus it's probably already dead by this point), but maybe a flag (like `--no-kill-isolate-on-error`) might work out?

thosakwe: Should I mark this as resolved? IMO it makes sense to leave it this way.

build_runner/lib/src/entrypoint/run_script.dart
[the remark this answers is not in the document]
thosakwe: Do you think it will still be necessary for me to pass `buildDirs`, or are we good on that?

build_runner/lib/src/entrypoint/run_script.dart
thosakwe: Yes, I think so. I'll send a commit in right now.

thosakwe: Just LMK if there's anything else I should add or fix up!

natebosch: We'll take another look soon - sorry for the delay.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,167 @@
...
+
+import 'dart:async';
+import 'dart:io';
+import 'dart:isolate';
+
+import 'package:args/command_runner.dart';
+import 'package:build_runner_core/build_runner_core.dart';
+import 'package:io/io.dart';
+import 'package:path/path.dart' as p;
+import '../generate/build.dart';
```
natebosch: [nit] Add an extra blank line between the `package:` and relative imports

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,167 @@
...
+
+  @override
+  String get invocation =>
+      '${super.invocation.replaceFirst('[arguments]', '[build-arguments]')} '
+      '<executable> [-- [script-arguments]]';
+
+  @override
+  SharedOptions readOptions() {
+    // This command doesn't allow specifying directories to build, instead it
+    // always builds the `test` directory.
```
natebosch: I think this was copied from the test command and does not apply here

build_runner/lib/src/entrypoint/run_script.dart
natebosch: Since we're using `errorsAreFatal` now I think we can remove this line

jakemac53:
One thing i noticed when testing this out - it would be useful to have a way to completely silence the output from the build.

Today you get all the build logs and then it will run the script and give you any output from that, but if you want to use this in the context of piping the results to another process or logging the output to a file etc, you probably only want the output from running the script itself.

Maybe we should add a `--log-level` flag so you can set `--log-level=severe` or something, whatever makes the most sense for your use case?

jakemac53:
Note also that the process doesn't appear to terminate - there is probably a stream listener that isn't getting cancelled or something along those lines.

jakemac53:
Some other edge cases:

- passing an invalid/missing script path
- attempting to run a non-dart script (today it appears to just replace the extension with `.dart`, I would probably remove that and force the full `.dart` extension).

thosakwe:
Got most of the fixes done. What's left:
1. Adding a `--log-level` flag. Should this just correspond to the `logLevel` parameter on the `build` function?
2. Detecting when the user has, as per @jakemac53, "[passed] an invalid/missing script path." Not 100% sure how to do that, in case the script the user wants to run has not been generated. I'm thinking that the `IsolateSpawnException` might let the user know what's up if they pass a nonexistent path.

thosakwe:
Anything else left for me to do?

thosakwe:
Ping! 😄

thosakwe:
Ping!

jakemac53:
Hi :)

Sorry for the delay here, we will try and look at this soon

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -0,0 +1,175 @@
...
+    // Define these before starting the isolate, so that we can close
+    // them if there is a spawn exception.
+    ReceivePort onExit, onError;
+
+    // Use a completer to determine the exit code.
+    var exitCodeCompleter = Completer<int>();
+
+    try {
+      var outputMap = options.outputMap ?? {};
+      outputMap.addAll({tempPath: null});
```
jakemac53: We could pass the top level directory under which the script lives here instead of null - that way only that directory would be built (it would also be hoisted up one level, which would change the path for the script you run a bit).

thosakwe:
I was able to create 6 integration tests. Let me know if you think I should add any more.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -100,9 +100,7 @@ class RunCommand extends BuildRunnerCommand {
     var exitCodeCompleter = Completer<int>();
 
     try {
-      var outputMap = options.outputMap ?? {};
-      outputMap.addAll({tempPath: null});
-
+      var outputMap = (options.outputMap ?? {})..addAll({tempPath: '.'});
```
jakemac53: I don't think we want `.` for the output map here, not actually sure what that would do? I think this is supposed to be the top level directory to build, and `null` I believe means all directories, which is probably a decent default.

build_runner/test/generate/run_script_test.dart
```diff
@@ -0,0 +1,136 @@
+@Tags(['integration'])
```
jakemac53: need to add copyright

build_runner/test/generate/run_script_test.dart
```diff
@@ -0,0 +1,136 @@
...
+
+      // We don't need to run a build in the setUp() closure, because
+      // that's not the functionality we're testing.
+    });
+
+    test('at least one argument must be provided', () async {
+      // Should throw error 64.
+      var result = await runDart('a', 'tool/build.dart',
+          args: ['run', '--output', 'build']);
+      expect(result.exitCode, ExitCode.usage.code,
```
jakemac53: It would be good to do a basic `contains` check on the stdout/stderr as well to differentiate these test cases and make sure we are getting the correct error messages.

build_runner/test/generate/run_script_test.dart
```diff
@@ -0,0 +1,136 @@
...
+      expect(result.exitCode, 0, reason: result.stderr as String);
+      expect(lastLine, 'it works!', reason: result.stderr as String);
+    });
+
+    test('passes input args', () async {
+      // Run the generated script, and examine its output.
+      var result = await runDart('a', 'tool/build.dart', args: [
+        'run',
+        'bin/main.copy.dart',
+        '--output',
```
jakemac53: This should work without an explicit --output arg as well, please add a test for that as well :)

build_runner/lib/src/entrypoint/run_script.dart
jakemac53: Actually this should be able to just provide the same top level directory as the script it wants to run, so something like `p.split(scriptName).first`. There is a question as to whether this should be `p.url.split` or `p.split`.... presumably `p.split` though which will expect a platform specific path.

thosakwe: I'm actually a little confused on this. What do you by mean by "top level directory"? For example, if the user says `pbr run bin/main.dart`, would `tempPath` be `bin/`?

jakemac53: So `tempPath` would still be `bin/`, but the value at that key (currently `'.'`) would be `bin`. That would allow us to only build the bin directory.

This does mean we would also hoist that directory up to the top level though which could break some things... that decision was made to mimic what `pub build` used to do (hoist the web dir up to the top level).

thosakwe: So, in that case, would `p.dirname(scriptName)` work? Or should I just go ahead with `p.split(scriptName).first`?

natebosch: `p.split(scriptName).first` would be better. If they want to run `bin/something/foo.dart` the top-level directory is `bin/` and that's the granularity we're able to build.

build_runner/lib/src/entrypoint/run_script.dart
```diff
@@ -170,6 +174,8 @@ class RunCommand extends BuildRunnerCommand {
       if (!exitCodeCompleter.isCompleted) {
         exitCodeCompleter.complete(ExitCode.success.code);
       }
+
+      await logSubscription.cancel();
```
jakemac53: It looks like this should probably move to an outer try/finally - we have some other bailout conditions before this try block which could prevent this from running (which would make the vm hang on the subscription being open)

thosakwe:
Anything left for me to do?

jakemac53:
> Anything left for me to do?

I don't think so - I will take another look later today and maybe we can finally get this merged :crossed_fingers:

jakemac53:
I went ahead and resolved the conflict as it was an easy one :D

jakemac53:
Ok @thosakwe unfortunately there is one issue with the latest build_runner which will need to be addressed, see https://travis-ci.org/dart-lang/build/jobs/517985470#L993 which is related to [this recent change](https://github.com/dart-lang/build/pull/2091)

thosakwe:
Awesome, I'll fix it up tonight.

natebosch:
Thanks @thosakwe !

thosakwe:
Woohoo!

PR #1630, opened 2018-06-30
Use a shared AnalyzerResolvers by default

### LINKED ISSUE #1418: Make build_test `resolve*` methods faster
Previously we used a `const BarbackResolvers()` instance which shared a lot of state between tests. Since moving to `new AnalyzerResolvers()` in #1121 we lose state in between tests and things go quite a bit slower.

I think we can do something were we reuse the resolvers by default.

natebosch:
Closes #1418

This brings performance for tests back inline with what it was when this
code was using a `const BarbackResolvers()` instance.

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,6 +1,8 @@
-## 0.10.2+5-dev
+## 0.10.3
 
 - Require test version ^0.12.42 and use `TypeMatcher`.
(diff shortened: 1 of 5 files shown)


natebosch:
Tested internally, this gets all the previously timing out tests to pass

matanlurey:
I am OOO this week, but endorse this message.

PR #1634, opened 2018-07-03
Invalidate cache by sdk version

nickclmb:
Hello,

I'm trying to reuse dart_tool cache between different machines, and it appears that dart versions on linux and macos can be different (examples):
``` 
Dart VM version: 2.0.0-dev.63.0 (Unknown timestamp) on "linux_x64"
Dart VM version: 2.0.0-dev.63.0 (Fri Jun 15 00:42:43 2018 +0200) on "linux_x64"
Dart VM version: 2.0.0-dev.63.0 (Fri Jun 15 00:42:43 2018 +0200) on "macos_x64"
```

So in this pull request it compares only the version ignoring everything after first space.

Regards,
Nikolay Samokhin

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,3 +1,5 @@
+PATCH
+- The SDK version comparison in asset graph check function compare actual version ignoring everything after version value
 ## 0.2.1
 
 - The hash dir for the asset graph under `.dart_tool/build` is now based on a
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,5 +1,7 @@
-PATCH
-- The SDK version comparison in asset graph check function compare actual version ignoring everything after version value
+## 0.2.1+1
+
+- Allow reuse cache between machines with different OS
+
 ## 0.2.1
 
 - The hash dir for the asset graph under `.dart_tool/build` is now based on a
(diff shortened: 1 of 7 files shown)


build_runner_core/CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+PATCH
```
grouma: This should be 0.2.1+1 and the corresponding pubspec should be updated.

build_runner_core/pubspec.yaml
```diff
@@ -22,6 +22,7 @@ dependencies:
   meta: ^1.1.0
   path: ^1.1.0
   pool: ^1.0.0
+  pub_semver: ^1.4.1
```
grouma: We are trying to keep build_runner_core as lightweight as possible. Is this package actually necessary? Can we do simple string comparison instead given the use case?

nickclmb:
Hi @grouma 

Thanks for feedback. Updated the changelog and removed dep on pub_semver please review.

nickclmb:
dartfmt errors on files that were not changed

build_runner_core/lib/src/util/sdk_version_match.dart
```diff
@@ -0,0 +1,7 @@
+bool areSdkVersionsSame(String thisVersion, String thatVersion) {
```
natebosch: how about

```dart
bool isSameSdkVersion(String thisVersion, String thatVersion) =>
    thisVersion?.split(' ')?.first == thatVersion?.split(' ')?.first;
```

jakemac53:
I think its safe to assume the sdk is identical across platforms.

PR #1636, opened 2018-07-06
run dartfmt to fix travis errors

jakemac53:


jakemac53:
@munificent I noticed dartfmt now splits inside string interpolation blocks, is that intended? I didn't realize that was even allowed?

jakemac53:
tbr @natebosch

jakemac53:
actually I will go ahead and clean up these string interpolation changes so they don't wrap

PR #1640, opened 2018-07-09
Run precompiled vm tests for package:build

jakemac53:
Fix an issue with the kernel builder and package cycles, and use the new kernel pipeline to run the package:build tests with caching on travis.

### DIFF
--- a/build/build.yaml
+++ b/build/build.yaml
@@ -1,0 +1,6 @@
+targets:
+  $default:
+    builders:
+      build_modules|modules:
+        options:
+          strategy: coarse
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -85,7 +85,15 @@ Future _createKernel(
   await Future.wait(transitiveDeps.map((dep) async {
     var kernelId = dep.primarySource.changeExtension(outputExtension);
     if (await buildStep.canRead(kernelId)) {
-      transitiveKernelDeps.add(kernelId);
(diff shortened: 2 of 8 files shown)


build/build.yaml
```diff
@@ -0,0 +1,6 @@
+targets:
+  $default:
+    builders:
+      build_modules|modules:
+        options:
+          strategy: coarse
```
natebosch: I'm in favor of changing this to the default

build_modules/lib/src/kernel_builder.dart
```diff
@@ -76,39 +77,47 @@ Future _createKernel(
...
-    if (await buildStep.canRead(kernelId)) {
-      transitiveKernelDeps.add(kernelId);
-    } else {
-      transitiveSourceDeps.addAll(dep.sources);
-    }
-  }));
+  var request = new WorkRequest();
+  AssetId outputId;
+  File outputFile;
+  File packagesFile;
```
natebosch: why were these moved up?

jakemac53: They are used below but assigned inside the block so that the temporary sets within the block can be gc'd

natebosch: Ah, I missed that 84 to 152 is now a block

I think we need to split this out into some submethods since it's nearly 100 lines and the structure is getting hard to follow.

jakemac53: done

PR #1643, opened 2018-07-10
significantly reduce the memory consumed by createOutputDirs

jakemac53:
Wrap the read and write calls up together when grabbing from the pool of available descriptors, previously we would end up doing all reads first and blocking the writes, which effectively caused us to hold all assets in memory.

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Clarify wording for conflicting output directory options. No behavioral
   differences.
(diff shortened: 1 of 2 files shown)


kevmoo:
How significant?

jakemac53:
> How significant?

Scales with the size of the build, on the order of the size of all the assets that would be written. In current models that would only be large if you were to say build hundreds of tests in --release mode. With the vm compilation pipeline though this was bad enough to use all 64 gigs of ram on my machine :grimacing: . Now it uses 2.5.

Issue #1242, opened 2018-04-04
Add --symlink flag which changes the -o directory to be all symlinks


jakemac53:
This would speed up running tests or development scenarios where people need a merged output dir.

jakemac53:
To clarify, if we use symlinks not only will the first creation of the merged dir be faster but we also should be able to leave in place any existing links. We only have to delete any links to deleted files and add new ones for new files.

matanlurey:
Huge +1 (for compatible systems, at least).

I'd love to see this be the default for POSIX (with an opt-out as `--no-symlinks`)

jakemac53:
I think we will need to start this as opt-in only, since the main supported use case of the output directory is for deployment.

For the test command we will automagically use the symlink option though, which is the other code path that is affected by this (and the main motivation for fixing it too).

PR #1647, opened 2018-07-11
Support creating output directories that only contain symlinks

jakemac53:
Closes https://github.com/dart-lang/build/issues/1242

Added a new `--symlink` option, which is defaulted to true for `test` but no other commands.

This is unfortunately a breaking change in build_runner and build_runner_core as it required adding a new interface which must be implemented by any `RunnerAssetReader`, which is an exposed class from both of these packages.


jakemac53:
In terms of performance, when testing with the `pub` package and the new vm compilation pipeline it takes creating the merged directory from >40 seconds to <1 second. Note that the output directory is unnecessarily large since we don't clean up the modular dill files in that case, or the original dart files, but its still a good proxy I think for how this can affect large packages.

jakemac53:
Tried running this with angulars _tests and looks like web tests don't work :man_facepalming: . Investigating now... but will require a fix in package:test it looks like.

build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,23 @@
+## 0.10.0-dev
+
+### Breaking Changes
+
+- The `RunnerAssetReader` interface now requires that you implement the new
```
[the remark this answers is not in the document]
jakemac53: done

jakemac53:
cc @natebosch ok I updated this with the changes we talked about.

There is now a `finalizeBuild` method on `BuildEnvironment`, and we use that to create the merged output dir. Only the `IOEnvironment` knows about the new asset reader interface.

jakemac53:
Looks like symlinks don't work as expected on windows even though there is supposedly support for linking individual files according to windows docs, the VM always ends up creating a directory symlink.

jakemac53:
Filed https://github.com/dart-lang/sdk/issues/33966

build_runner_core/lib/src/environment/io_environment.dart
```diff
@@ -17,6 +17,8 @@ import '../package_graph/package_graph.dart';
 import 'build_environment.dart';
 import 'create_merged_dir.dart';
 
+final _logger = new Logger('IOEnvironment');
```
natebosch: need to include this at https://github.com/dart-lang/build/blob/a10c67f7ef87bae6e942cd8ccc7dfc1903e5afa1/build_runner/lib/src/logging/std_io_logging.dart#L69

jakemac53: done

Issue #1644, opened 2018-07-10
Add `<file>.dart.library` assets


natebosch:
Today if _any_ Dart file changes in a package we rerun the builder for the `.meta_module.raw` asset. This process involves at least re-reading _every_ Dart file in the package and using the analyzer to parse it's directives.

If we insert a step to go from `.dart` to `.dart.library` and just parse out the directives for a given file then:
1.  Only the `.dart.library` files are inputs to `.meta_module.raw` so that process only ever reruns when a set of imports actually changes.
2.  We only use the analyzer to parse the directives of the Dart file that changed, not all the other Dart files in the package.

PR #1524, opened 2018-06-06
meta module parse conditional imports

nickclmb:
Signed-off-by: Nikolay Samokhin <nickclmb@gmail.com>
Hi everyone,

At the moment if you have conditional imports in dart file, meta module will not have them in module dependencies:
```dart
import 'package:RAL/html_i.dart'
if (dart.library.html) 'dart:html'
if (dart.library.io) 'package:RAL/io/html_io.dart';
```
As a result the module will have dependency only on RAL/html_i.dart, missing RAL/io/html_io.dart
This pull request fixes that.

PR #1649, opened 2018-07-12
Add a `ModuleLibrary` class

natebosch:
Towards #1644

Explicitly separate out the step of parsing a Dart source file and
classifying directives from the other work happening in the meta module.
The `ModuleLibrary` class is essentially the `_AssetNode` class from
before, except it is also responsible for determining which files should
be skipped and which are entrypoints rather than that responsibilty
being split with `_computeModules`.

"internal" and "external" dependencies are also not tracked separately
since it is applied as a filter only for the strongly connected
components call.

Eventually we will add in a capability to store a `ModuleLibrary` which
will let us write it as a separate Asset and then restore it to
compute meta modules.

One minor change in behavior is that we no longer throw if a "part"
directive reaches across packages or top-level directories. This should
be resolved if there are package level cycles by the next iteration of
strongly connected components.

### DIFF
--- a/build_modules/lib/src/module_library.dart
+++ b/build_modules/lib/src/module_library.dart
@@ -1,0 +1,112 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:build/build.dart';
+import 'package:meta/meta.dart';
+
+class ModuleLibrary {
+  /// The AssetId of the original Dart source file.
+  final AssetId id;
+
+  /// Whether this library can be imported.
+  ///
+  /// This will be false if the source file is a "part of", or imports code that
+  /// can't be used outside the SDK.
+  final bool isImportable;
+
+  /// Whether this library is an entrypoint.
+  ///
+  /// True if the library is in `lib/` but not `lib/src`, or if it is outside of
+  /// `lib/` and contains a `main` method. Always false if this is not an
+  /// importable library.
+  final bool isEntryPoint;
+
+  /// The IDs of libraries that are imported or exported by this library.
+  ///
+  /// Null if this is not an importable library.
+  final Set<AssetId> deps;
+
+  /// The "part" files for this library.
+  ///
+  /// Null if this is not an importable library.
+  final Set<AssetId> parts;
+
+  ModuleLibrary._(this.id,
+      {@required this.isEntryPoint, @required this.deps, @required this.parts})
+      : isImportable = true;
+
+  ModuleLibrary._nonImportable(this.id)
+      : isImportable = false,
+        isEntryPoint = false,
+        deps = null,
+        parts = null;
+
+  factory ModuleLibrary._fromCompilationUnit(
+      AssetId id, bool isEntryPoint, CompilationUnit parsed) {
+    var deps = new Set<AssetId>();
+    var parts = new Set<AssetId>();
+    for (var directive in parsed.directives) {
+      if (directive is! UriBasedDirective) continue;
+      var path = (directive as UriBasedDirective).uri.stringValue;
+      if (Uri.parse(path).scheme == 'dart') continue;
+      var linkedId = new AssetId.resolve(path, from: id);
+      if (linkedId == null) continue;
+      if (directive is PartDirective) {
+        parts.add(linkedId);
+        continue;
+      }
+
+      List<Configuration> conditionalDirectiveConfigurations;
+
+      if (directive is ImportDirective && directive.configurations.isNotEmpty) {
+        conditionalDirectiveConfigurations = directive.configurations;
+      } else if (directive is ExportDirective &&
+          directive.configurations.isNotEmpty) {
+        conditionalDirectiveConfigurations = directive.configurations;
+      }
+      deps.add(linkedId);
+      if (conditionalDirectiveConfigurations != null) {
+        deps.addAll(conditionalDirectiveConfigurations
+            .map((c) => Uri.parse(c.uri.stringValue))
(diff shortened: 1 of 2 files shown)


build_modules/lib/src/module_library.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:analyzer/analyzer.dart';
+import 'package:build/build.dart';
+import 'package:meta/meta.dart';
+
+class ModuleLibrary {
```
grouma: Let's add a doc comment here.

build_modules/lib/src/module_library.dart
```diff
@@ -0,0 +1,112 @@
...
+  /// The AssetId of the original Dart source file.
+  final AssetId id;
+
+  /// Whether this library can be imported.
+  ///
+  /// This will be false if the source file is a "part of", or imports code that
+  /// can't be used outside the SDK.
```
grouma: Curious, what would be an example import that can't be used outside the SDK? Are you referring to `dart:_` type imports?

build_modules/lib/src/module_library.dart
```diff
@@ -0,0 +1,112 @@
...
+        parts.add(linkedId);
+        continue;
+      }
+
+      List<Configuration> conditionalDirectiveConfigurations;
+
+      if (directive is ImportDirective && directive.configurations.isNotEmpty) {
+        conditionalDirectiveConfigurations = directive.configurations;
+      } else if (directive is ExportDirective &&
+          directive.configurations.isNotEmpty) {
```
grouma: When can directive.configurations be non empty if it isn't an import or export directive? If it can't, then can we get rid of the check?

build_modules/lib/src/module_library.dart
```diff
@@ -0,0 +1,112 @@
...
+        conditionalDirectiveConfigurations = directive.configurations;
+      }
+      deps.add(linkedId);
+      if (conditionalDirectiveConfigurations != null) {
+        deps.addAll(conditionalDirectiveConfigurations
+            .map((c) => Uri.parse(c.uri.stringValue))
+            .where((u) => u.scheme != 'dart')
+            .map((u) => new AssetId.resolve(u.toString(), from: id)));
```
grouma: What's going on here?

build_modules/lib/src/module_library.dart
grouma: It would be good to have some tests for this class. There is some non trivial logic.

natebosch: None of this is new logic, it was just all private before (and therefore tested through `MetaModule.forAssets`, which it still is today).

I can add some tests though since it's now public within the package.

build_modules/lib/src/module_library.dart
natebosch: Yeah this is specifically referring to the checks at https://github.com/dart-lang/build/pull/1649/files#diff-08a717a90be39b7bdd847098ce7bc663R92

build_modules/lib/src/module_library.dart
natebosch: This is https://github.com/dart-lang/build/pull/1649/files#diff-e6beb981c2249722d51e60e34b104f88L97

It happens with config specific imports. When we add support for different modules based on platform this line will need to change - we'll need to consider only the the configuration that matches the platform.

build_modules/lib/src/module_library.dart
natebosch: Moved from https://github.com/dart-lang/build/pull/1649/files#diff-e6beb981c2249722d51e60e34b104f88L109

Same thing that's happening here: https://github.com/dart-lang/build/pull/1649/files#diff-08a717a90be39b7bdd847098ce7bc663R54

We'll have an import URI thats with either `package:something/something.dart`; or `relative/import.dart`, and so we take all the URIs and turn them into AssetIds (which are absolute) by resolving them against where the import comes from.

build_modules/lib/src/module_library.dart
grouma: I was confused by the line moved from https://github.com/dart-lang/build/pull/1649/files#diff-e6beb981c2249722d51e60e34b104f88L109

That looked like new logic to me. I'm fine with adding tests in a later PR.

build_modules/lib/src/module_library.dart
grouma: I don't follow. Maybe I didn't phrase my question clearly. Do we even need to check if `directive is ImportDirective` or `directive is ExportDirective`? My assumption is that `directive.configurations.isNotEmpty` can only be true when the directive is one of those two types. Thus the logic can be simplified to:

```
if (directive.configurations.isNotEmpty) {
   conditionalDirectiveConfigurations = directive.configurations;
}
```

natebosch: For some context on the code: https://github.com/dart-lang/build/pull/1524

The `.configurations` field only exists on [NamespaceDirective](https://pub.dartlang.org/documentation/analyzer/latest/analyzer.dart.ast.ast/NamespaceDirective-class.html), not on `UriBasedDirective` so we need check at least that. We could probably check only that and skip the separate branches for `ImportDirective` and `ExportDirective` though, so I'll take a look at that in a followup.

build_modules/lib/src/module_library.dart
```diff
@@ -6,6 +6,10 @@ import 'package:analyzer/analyzer.dart';
 import 'package:build/build.dart';
 import 'package:meta/meta.dart';
 
+/// A Dart library within a module.
+///
+/// Modules can be computed based on library dependencies (imports and expors)
```
grouma: typo: exports*

PR #1652, opened 2018-07-12
Move ModuleLibrary parsing to a separate builder

### LINKED ISSUE #1644: Add `<file>.dart.library` assets

natebosch:
Closes #1644

- Add capability to write and parse a String for a `ModuleLibrary`.
- Add a Builder to produce these files. This is a thin wrapper around
  the functionality in `ModuleLibrary`.
- Change `MetaModuleBuilder` to use the library assets instead of the
  source assets.
- Update tests to reflect new behavior.
- Add cleanup of the `.dart.library` files since they are not useful
  outside of the build.

### DIFF
--- a/build_modules/build.yaml
+++ b/build_modules/build.yaml
@@ -2,6 +2,7 @@ builders:
   modules:
     import: "package:build_modules/builders.dart"
     builder_factories:
+      - moduleLibraryBuilder
       - metaModuleBuilder
       - metaModuleCleanBuilder
       - moduleBuilder
@@ -12,6 +13,7 @@ builders:
         - .meta_module.raw
         - .meta_module.clean
       .dart:
+        - .dart.library
         - .module
         - .linked.sum
         - .unlinked.sum
--- a/build_modules/build.yaml
+++ b/build_modules/build.yaml
@@ -13,7 +13,7 @@ builders:
         - .meta_module.raw
         - .meta_module.clean
       .dart:
-        - .dart.library
+        - .module.library
         - .module
         - .linked.sum
         - .unlinked.sum
--- a/build_modules/lib/src/module_library.dart
+++ b/build_modules/lib/src/module_library.dart
@@ -105,6 +105,33 @@ class ModuleLibrary {
         (isLibDir && !id.path.startsWith('lib/src/')) || _hasMainMethod(parsed);
     return new ModuleLibrary._fromCompilationUnit(id, isEntryPoint, parsed);
   }
+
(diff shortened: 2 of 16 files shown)


build_modules/build.yaml
```diff
@@ -12,6 +13,7 @@ builders:
         - .meta_module.raw
         - .meta_module.clean
       .dart:
+        - .dart.library
```
jakemac53: not a huge fan of this name, it makes it sound like it has more information than it does? Also we haven't kept the `.dart` extension for other builders... with this generic of a name it kind of makes sense to keep it but a less generic name would allow you to drop it.

build_modules/lib/src/module_library.dart
```diff
@@ -105,6 +105,33 @@ class ModuleLibrary {
...
     return new ModuleLibrary._fromCompilationUnit(id, isEntryPoint, parsed);
   }
+
+  /// Parses the output of [toString] back into a [ModuleLibrary].
+  ///
+  /// Importable libraries can be round tripped to a String. Non-importable
+  /// libraries should not be printed or parsed.
+  static ModuleLibrary parse(String encoded) {
+    final lines = encoded.split('\n');
+    final id = new AssetId.parse(lines.first);
```
jakemac53: We shouldn't need to to encode this in the file, its easily reproducible from the path

build_modules/build.yaml
natebosch: do you have suggestions for alternatives? I had considered `.directives` but that naming felt a little specific to analyzer terminology to me...

jakemac53: `.directives` was my initial thought but it also contains a bit more info than that (whether its any entrypoint or not).

Maybe `.module.info` or something like that?

natebosch: discussed offline. Going with `.module.library`.

build_modules/lib/src/module_library.dart
natebosch: Done

natebosch:
Appveyor was a flake, merging on red.

PR #1653, opened 2018-07-13
Filter by non-hidden applied builders

### LINKED ISSUE #1578: Filter out the builders which "applies_builders" some builder which is "build_to: source" on non root packages
Today we filter out builders that won't work because they want to build to source and auto apply to some package other than the root. Once source_gen adds the combining builder the actual builder doing work would be build_to: cache and so would get needlessly run.

cc @kevmoo - we'll want this before launching the combining builder stuff because it will impact performance.

Blocking https://github.com/dart-lang/source_gen/issues/319

natebosch:
Closes #1578

Previously we filtered out any builders on non-root packages which would
attempt to write to source. Now we also filter out builders which write
to cache but want to apply some builder that would write to source,
under the assumption that if a builder can't be applied it isn't worth
doing the hidden work.

Add a test for the existing filtering, as well as the new filtering.

### DIFF
--- a/build_runner_core/lib/src/package_graph/apply_builders.dart
+++ b/build_runner_core/lib/src/package_graph/apply_builders.dart
@@ -280,9 +280,17 @@ Future<List<BuildPhase>> createBuildPhases(
             return targetGraph.allModules[key];
           })?.where((n) => n != null));
   final applyWith = _applyWith(builderApplications);
+  final allBuilders = new Map<String, BuilderApplication>.fromIterable(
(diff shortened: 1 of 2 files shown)


natebosch:
cc @kevmoo - I think this is the last blocking issue to moving forward with the new `source_gen` version.

kevmoo:
Can we publish this soonish? Is this a breaking change?

natebosch:
I don't think we consider this breaking.

Issue #1661, opened 2018-07-18
We should detect and throw if we detect ongoing async work after a builder returns


jakemac53:
Today inside your `build` method if you don't await any async operation it can potentially end up executing _after_ the build method completes at which point we won't be tracking file reads properly and all kinds of other things could get broken.

We could try to detect this using zones, or we could just make sure the SingleStepReader won't complete a future after the build step has completed (basically we just set a flag after the build call returns, and we complete with an error instead if that has been set).

natebosch:
As long as we don't accept _new writes_ after the build method completes I think we're safe. I don't mind if you complete the method, then after a delay start a new asset read and can't write a new output - by definition the input didn't affect the outputs you did write (or decided not to write) so it doesn't matter if we're tracking it.

That said - it doesn't look like we reject writes after the build step is complete so we should probably add that.

jakemac53:
We should at least disable new reads as well, because we won't be tracking those as dependencies properly and you can actually get concurrent modification errors as well.

jakemac53:
Also going back to the original LRU cache issue I was trying to work around, since the cache is synchronous we don't insert into it until after the read returns. That could be _after_ the entire build is done, which means we could insert stale values into the cache after a new build has started.

There are other weird situations as well that could occur if executing async code after you return from the build method (specifically with Resources, but probably other things).

The only real way to deal with all of that is to handle it in the zone I think. I think its fair to say you aren't allowed to have unawaited async code inside the `build` method at all.

natebosch:
Is the idea to track the number of started and completed microtasks in the zone? It seems like an easy thing to get wrong...

jakemac53:
> Is the idea to track the number of started and completed microtasks in the zone? It seems like an easy thing to get wrong...

I don't think it needs to be that complicated, I think we can just provide custom `run*` methods and at the top of them we check if we have already completed, if so we throw an error and don't invoke the actual callback.

jakemac53:
I think the current solution is good enough here, closing to reflect that

PR #1667, opened 2018-07-18
Throw when using a BuildStep after complete

natebosch:
Towards #1661

This should catch most bugs with builders that have unawaited work, but
there still could be visible side effects which arent caught through a
Resource or similar.


natebosch:
> I think that is probably the correct behavior though?

Yeah I think so. We will throw away any work you do anyway, may as well make it explicit.

PR #1672, opened 2018-07-19
Prepare to publish bazel_codegen

natebosch:
Compatible with the non-dev Dart 2 SDK


natebosch:
Probably not actually being used externally, but may as well publish.

kevmoo:
One user 😄 https://pub.dartlang.org/packages?q=dependency%3A_bazel_codegen

natebosch:
My guess is dazel is broken today... this could be depended on by other users setting up things with rules_dart manually (but probably isn't?)

PR #1674, opened 2018-07-20
Add a tag to the header line for cached errors

### LINKED ISSUE #1673: Warn a developer if an error message in their package is served from cache
This issue is a reaction to my confusion here: https://groups.google.com/a/dartlang.org/d/msg/misc/5KdzgsyGiAc/fCiNj9zfAgAJ

To avoid this confusion for other developers in the future, it could be better communicated when a builder is actually executed, and when the build output is retrieved from cache.

Right now, the output looks like this (with `--verbose`):
```
[SEVERE] regular_scanner|scanner_builder on example/main.dart:
Error running TableDrivenScannerGenerator
<stack trace>
```

I think it would be good to add something like: `[WARNING] regular_scanner|scanner_builder on example/main.dart is cached, showing previously generated output`

natebosch:
Closes #1673

Attempting to debug build actions which aren't actually running is
really confusing today - the builder will never throw and breakpoints
will not be hit, but we'll print logs as if it did. Making the header
line look different and adding an FAQ will hopefully clear this up.

### DIFF
--- a/build_runner_core/lib/src/logging/failure_reporter.dart
+++ b/build_runner_core/lib/src/logging/failure_reporter.dart
@@ -83,7 +83,7 @@ class FailureReporter {
     return Future.wait(errorFiles.map((errorFile) async {
       if (await errorFile.exists()) {
         final errorReports = jsonDecode(await errorFile.readAsString());
-        final actionDescription = (errorReports as List).first as String;
+        final actionDescription = '${(errorReports as List).first} (cached)';
         final logger = new Logger(actionDescription);
         for (final List error in errorReports.skip(1)) {
           final stackTraceString = error[2] as String;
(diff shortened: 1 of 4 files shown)


build_runner_core/lib/src/logging/failure_reporter.dart
```diff
@@ -83,7 +83,7 @@ class FailureReporter {
     return Future.wait(errorFiles.map((errorFile) async {
       if (await errorFile.exists()) {
         final errorReports = jsonDecode(await errorFile.readAsString());
-        final actionDescription = (errorReports as List).first as String;
+        final actionDescription = '${(errorReports as List).first} (cached)';
```
kevmoo: Nice!

jakemac53:
pubspec/changelog?

PR #1675, opened 2018-07-20
Stop using deprecated strongMode option

natebosch:


build_resolvers/lib/src/resolver.dart
```diff
@@ -85,7 +85,9 @@ class AnalyzerResolver implements ReleasableResolver {
 
   AnalyzerResolver(
       DartUriResolver dartUriResolver, AnalysisOptions analysisOptions) {
-    _context.analysisOptions = analysisOptions;
+    if (analysisOptions != null) {
```
jakemac53: I don't really like having apis that accept nulls in non-optional args... but I guess this does allow us to use one less thing from src...

jakemac53: Oh if this isn't public I would prefer we just make it optional.

natebosch: Done

PR #1677, opened 2018-07-23
clean up kernel modules by default

jakemac53:


jakemac53:
Considering holding off on this due to https://github.com/dart-lang/sdk/issues/33952, since we will likely be keeping these around in the future...

PR #1682, opened 2018-07-24
Use the coarse module by default, even on root

natebosch:
The coarse strategy has the advantage of not using a resolver, and in
many cases it seems to be the saner default.


jakemac53:
should get invalidated less often too :)

PR #1685, opened 2018-07-25
misc(modules,runner_core,config): update to json_serializable v1

kevmoo:


build_runner_core/tool/build.dart
```diff
@@ -15,11 +16,14 @@ main() async {
     options,
     environment,
     [
-      applyToRoot(jsonSerializable(BuilderOptions.forRoot),
-          hideOutput: false,
+      applyToRoot(
+          jsonSerializable(
+              new BuilderOptions({'generate_to_json_function': false})),
```
kevmoo: Turns out the mixin model works well here CC @jakemac53 – or at least I couldn't figure out how to use the functions in a sane way...

jakemac53: I was able to migrate to the functions using the normal patterns, pushed a commit here.

kevmoo:
Thanks, @jakemac53 – fixed dartfmt issue.

@natebosch @jakemac53 – LGTY?

jakemac53:
Ok, well I am not opposed in any case. We have already released versions of things that work with Dart 2.0, so there is no rush for people to update to the latest build packages if they want to wait a bit.

Issue #1462, opened 2018-05-17
Add a per-phase build pool


natebosch:
https://github.com/dart-lang/build/pull/1312 should help for most cases but there are still ways that we can end up running a lot of lazy builds concurrently which can mean very high memory usage.

We should add a resource pool which limits the number of in flight build actions per phase. It needs to be per phase because we can't limit an earlier lazy build when it would be blocking for an already ongoing build.

PR #1698, opened 2018-08-01
add a per-phase pool to limit the number of concurrent actions

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1462

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -369,7 +369,7 @@ class _SingleBuild {
 
   Future<Iterable<AssetId>> _runForInput(
       int phaseNumber, InBuildPhase phase, AssetId input) {
-    var pool = _buildPhasePool[phaseNumber] ??= new Pool(16);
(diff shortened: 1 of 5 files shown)


jakemac53:
This doesn't appear to have any impact on build times - tested on angular components example and our basic _test package.

It should help with memory usage on larger apps though.

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -365,73 +368,77 @@ class _SingleBuild {
...
-    hungActionsHeartbeat.ping();
-    pendingActions[phaseNumber].remove(actionDescription);
+      // Add `builderOutputs` to the primary outputs of the input.
+      var inputNode = _assetGraph.get(input);
+      assert(inputNode != null,
+          'Inputs should be known in the static graph. Missing $input');
+      assert(
+          inputNode.primaryOutputs.containsAll(builderOutputs),
+          'input $input with builder $builder missing primary outputs: \n'
+              'Got ${inputNode.primaryOutputs
```
natebosch: [nit] Consider picking a new manual string break point to avoid the line break inside of it

PR #1703, opened 2018-08-02
Add simple live-reload implementation

samogot:
Add live-reload option that turns on simple WebSocket broadcaster of rebuild finished events.
Inject simple JS snippet to bootstrap that listen for update message from WebSocket and reloads the whole page.

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -133,8 +133,9 @@ class BuildUpdatesWebSocketHandler {
   BuildUpdatesWebSocketHandler([this._handlerFactory = webSocketHandler]) {
     var untypedTearOff = (webSocket, protocol) =>
         _handleConnection(webSocket as WebSocketChannel, protocol as String);
-    _internalHandler = _handlerFactory(
-        untypedTearOff, protocols: [_buildUpdatesProtocol]) as shelf.Handler;
+    _internalHandler =
+        _handlerFactory(untypedTearOff, protocols: [_buildUpdatesProtocol])
+            as shelf.Handler;
   }
 
   shelf.Handler get handler => _internalHandler;
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -12,6 +12,8 @@ import 'package:logging/logging.dart';
 import 'package:mime/mime.dart';
 import 'package:path/path.dart' as p;
 import 'package:shelf/shelf.dart' as shelf;
+import 'package:shelf_web_socket/shelf_web_socket.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
 
 import '../generate/watch_impl.dart';
 import 'asset_graph_handler.dart';
@@ -19,6 +21,7 @@ import 'path_to_asset_id.dart';
 
 const _performancePath = r'$perf';
 final _graphPath = r'$graph';
+final _hotReloadPath = r'$hotreload';
 
 final _logger = new Logger('Serve');
 
@@ -43,10 +46,13 @@ class ServeHandler implements BuildState {
   final Future<AssetHandler> _assetHandler;
   final Future<AssetGraphHandler> _assetGraphHandler;
 
+  final WebSocketHandler _webSocketHandler = WebSocketHandler();
+
   ServeHandler._(this._state, this._assetHandler, this._assetGraphHandler,
       this._rootPackage) {
     _state.buildResults.listen((result) {
       _lastBuildResult = result;
+      _webSocketHandler.emitUpdateMessage(result);
     });
   }
 
@@ -55,14 +61,16 @@ class ServeHandler implements BuildState {
   @override
   Stream<BuildResult> get buildResults => _state.buildResults;
 
-  shelf.Handler handlerFor(String rootDir, {bool logRequests}) {
+  shelf.Handler handlerFor(String rootDir, {bool logRequests, bool liveReload}) {
+    liveReload ??= false;
     logRequests ??= false;
     if (p.url.split(rootDir).length != 1) {
       throw new ArgumentError.value(
           rootDir, 'rootDir', 'Only top level directories are supported');
     }
     _state.currentBuild.then((_) => _warnForEmptyDirectory(rootDir));
-    var cascade = new shelf.Cascade()
+    var cascade = new shelf.Cascade();
+    cascade = (liveReload ? cascade.add(_webSocketHandler.handler) : cascade)
         .add(_blockOnCurrentBuild)
         .add((shelf.Request request) async {
       if (request.url.path == _performancePath) {
@@ -107,6 +115,34 @@ class ServeHandler implements BuildState {
           'has no assets in the build. You may need to add some sources or '
           'include this directory in some target in your `build.yaml`');
     }
+  }
+}
+
+class WebSocketHandler {
+  final Set<WebSocketChannel> _connectionPull = Set();
+  final shelf.Handler _internalHandler;
+
+  // is it OK?
+  // ignore: implicit_this_reference_in_initializer
+  WebSocketHandler() : _internalHandler = webSocketHandler(_handleConnection);
+
+  Future<shelf.Response> handler(shelf.Request request) async {
+    if (!request.url.path.startsWith(_hotReloadPath)) {
+      return new shelf.Response.notFound('');
+    }
+    return _internalHandler(request);
+  }
+
+  void emitUpdateMessage(BuildResult buildResult) {
+    for (var webSocket in _connectionPull) {
+      webSocket.sink.add('update');
+    }
+  }
+
+  void _handleConnection(WebSocketChannel webSocket) async {
+    _connectionPull.add(webSocket);
+    await webSocket.stream.drain();
+    _connectionPull.remove(webSocket);
   }
 }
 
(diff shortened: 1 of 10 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -321,6 +322,36 @@ require.config({
...
+    var ws = new WebSocket(wsUrl);
+    ws.onmessage = function(event) {
+      console.log(event);
+      if(event.data === 'update'){
+        location.reload();
+      }
+    };
+  } catch (e) {
+    // Live reload might be turned off. Just ignore it for now
+    // TODO: Find a way to pass serve option here, to not inject this code at all, if live/hot reload is turned off
```
jakemac53: This is an interesting problem - ideally this package does not have to rely on the implementation details in the server (such as the `update` event, or the `$hotreload` path). We want to keep breaking changes in build_runner to a minimum, so ideally we wouldn't have other packages depending on these details. 

To that end, it might make sense to actually inject this code from the server itself, possibly into all `html` requests that it serves, although that could be challenging as well. It would be worth talking to @grouma about how we do this for DDR internally.

build_runner/lib/src/server/server.dart
```diff
@@ -110,6 +119,34 @@ class ServeHandler implements BuildState {
   }
 }
 
+class WebSocketHandler {
```
jakemac53: This is pretty specific to LiveReload right now, we might make it more generic later but for now lets rename it to `LiveReloadWebSocketHandler` - or maybe `BuildUpdatesWebSocketHandler`.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
samogot: It looks like non-trivial task to figure out if request is html page or not. Both `Content-type` header and `<html>` tag might be omitted.
Maybe it's easier to inject code to the end of any `*.dart.bootstrap.js` file?

jakemac53: > Maybe it's easier to inject code to the end of any *.dart.bootstrap.js file?

The problem with that is we are then backing knowledge of build_web_compilers into build_runner - which is maybe worse than the other way around haha.

build_runner/lib/src/server/server.dart
```diff
@@ -43,10 +46,13 @@ class ServeHandler implements BuildState {
   final Future<AssetHandler> _assetHandler;
   final Future<AssetGraphHandler> _assetGraphHandler;
 
+  final _webSocketHandler = BuildUpdatesWebSocketHandler();
```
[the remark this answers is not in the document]
jakemac53: I am in favor of just doing a `dartfmt --fix`, but can we also enable a lint then to make us not introduce any more `new` keywords?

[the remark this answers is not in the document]
natebosch: yes we could add `unnecessary_new` and `unnecessary_const` to avoid regressions.

build_runner/lib/src/server/server.dart
```diff
@@ -119,34 +123,64 @@ class ServeHandler implements BuildState {
...
+shelf.Handler _injectBuildUpdatesClientCode(shelf.Handler innerHandler) {
+  return (shelf.Request request) {
+    if (!request.url.path.endsWith('.js')) {
+      return innerHandler(request);
+    }
+    return Future.sync(() => innerHandler(request)).then((response) async {
+      // TODO: Find a way how to check and/or modify body without reading it whole
+      var body = await response.readAsString();
+      if (body.startsWith(_entrypointExtensionMarker)) {
+        body += _buildUpdatesInjectedJS;
```
jakemac53: I think we should replace the placeholder comment as opposed to appending

samogot: I don't think so.
1. Why do not remove comment - we might have other middlewares in future that will depend on the same extension point
2. Why append to the end of file - I feel it "safer" to append code to the end then to the begining

natebosch: What is the browser behavior if there is some valid JS source followed by invalid stuff - will the stuff at the beginning of the file still get executed?
In our particular setup we're pretty sure we're always sending valid parseable javascript, but if the browser would start running our injected stuff even if the rest of the file contains broken stuff that could be a reason to put it at the beginning.

samogot: AFAICR if there is fatal syntax error - it won't. if "invalid stuff" can somehow be parsed as JS, then file will be executed till the first unhanded error

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
grouma: DDR knows ahead of time what the main entry JS file is called. It then intercepts this request and returns the DDR application. This application will embed a request for the main entry along with a request parameter `forReal`. The logic can be found in `ddr_handler.dart`. This pattern has been helpful because we have completely changed our injected application logic a number of times. Is it possible to know what the root request will be in build_runner?

build_runner/test/server/serve_handler_test.dart
```diff
@@ -162,6 +164,97 @@ void main() {
       expect(await response.readAsString(), contains('--track-performance'));
     });
   });
+
+  group('build updates', () {
+    final _entrypointExtensionMarker = '/* ENTRYPOINT_EXTENTION_MARKER */';
+
+    test('injects client code if enabled', () async {
+      _addSource('a|web/some.js', _entrypointExtensionMarker);
```
jakemac53: nit: append some other js content to the end as well

build_runner/test/server/serve_handler_test.dart
```diff
@@ -162,6 +164,97 @@ void main() {
...
+    });
+
+    test('doesn\'t inject client code in non-marked files', () async {
+      _addSource('a|web/some.js', 'content');
+      var response = await serveHandler.handlerFor('web', liveReload: true)(
+          new Request('GET', Uri.parse('http://server.com/some.js')));
+      expect(await response.readAsString(), isNot(contains('\$livereload')));
+    });
+
+    test('emmits a message to all listners', () async {
```
jakemac53: nit: this test and the next one share a lot of setup logic, you can move them into a `group` (see 2nd code snippet [here](https://github.com/dart-lang/test#writing-tests)).

That lets you define a shared `setUp` and `tearDown` function for both tests.

build_runner/test/server/serve_handler_test.dart
```diff
@@ -162,6 +164,97 @@ void main() {
...
+      Function exposedOnConnect;
+      var mockHandlerFactory = (Function onConnect, {protocols}) {
+        exposedOnConnect = onConnect;
+      };
+      var buildUpdatesWebSocketHandler = BuildUpdatesWebSocketHandler(
+          mockHandlerFactory);
+
+      var streamControllerFirst1 = StreamController<List<int>>();
```
jakemac53: nit: can we name these a bit more descriptively?

It is also important to make sure these get closed (in a `tearDown` method ideally).

build_runner/test/server/serve_handler_test.dart
```diff
@@ -162,6 +164,97 @@ void main() {
...
+      var streamControllerFirst2 = StreamController<List<int>>();
+      var serverChannelFirst = WebSocketChannel(StreamChannel(streamControllerFirst1.stream, streamControllerFirst2.sink), serverSide: true);
+      var clientChannelFirst = WebSocketChannel(StreamChannel(streamControllerFirst2.stream, streamControllerFirst1.sink), serverSide: false);
+
+      var streamControllerSecond1 = StreamController<List<int>>();
+      var streamControllerSecond2 = StreamController<List<int>>();
+      var serverChannelSecond = WebSocketChannel(StreamChannel(streamControllerSecond1.stream, streamControllerSecond2.sink), serverSide: true);
+      var clientChannelSecond = WebSocketChannel(StreamChannel(streamControllerSecond2.stream, streamControllerSecond1.sink), serverSide: false);
+
+      var deferredExpect1 = clientChannelFirst.stream.toList().then((value) => expect(value, ['update', 'update']));
```
jakemac53: Instead of this you can use the [built in stream matchers](https://github.com/dart-lang/test#stream-matchers)

build_runner/lib/src/server/server.dart
```diff
@@ -110,6 +123,68 @@ class ServeHandler implements BuildState {
...
+/// Class that manages web socket connection handler to inform clients about
+/// build updates
+class BuildUpdatesWebSocketHandler {
+  final activeConnections = <WebSocketChannel>[];
+  final shelf.Handler Function(Function, {Iterable<String> protocols})
+      _handlerFactory;
+  shelf.Handler _internalHandler;
+
+  BuildUpdatesWebSocketHandler([this._handlerFactory = webSocketHandler]) {
+    var untypedTearOff = (webSocket, protocol) =>
```
jakemac53: one last nit, can you document why we need to do this, and ideally point at an issue in the original package?

build_runner/lib/src/server/server.dart
```diff
@@ -110,6 +123,68 @@ class ServeHandler implements BuildState {
...
+  };
+}
+
+/// Hot-reload config
+///
+/// Listen WebSocket for updates in build results
+///
+/// Now only live-reload functional - just reload page on update message
+final _buildUpdatesInjectedJS = '''\n
+(function() {
```
jakemac53: Oh also its probably worth adding a comment here `// Injected by build_runner for live reload support` or something along those lines

build_runner/lib/src/server/server.dart
samogot: Created dart-lang/shelf_web_socket#11

PR #1709, opened 2018-08-04
Add assets digests handler

samogot:
This entry point will be used for hot reloading, in order to initially
retrieve digests of all assets, as build_runner might not know list of
all of them, unlike client.

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -21,6 +21,7 @@ import 'path_to_asset_id.dart';
 
 const _performancePath = r'$perf';
 final _graphPath = r'$graph';
+final _assetsDigestPath = r'$assets';
 final _buildUpdatesProtocol = r'$livereload';
 final _buildUpdatesMessage = 'update';
 final entrypointExtensionMarker = '/* ENTRYPOINT_EXTENTION_MARKER */';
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/server/server.dart
```diff
@@ -112,6 +116,24 @@ class ServeHandler implements BuildState {
...
   }
 
+  Future<shelf.Response> _assetsDigestHandler(
+      shelf.Request request, String rootDir) async {
+    var assertPathList = jsonDecode(await request.readAsString()) as List;
+    var rootPackage = _state.packageGraph.root.name;
+    var resultMap = <String, String>{};
+    for (String path in assertPathList) {
+      try {
+        var assetId = pathToAssetId(rootPackage, rootDir, path.split('/'));
```
jakemac53: nit: `p.url.split()`

build_runner/lib/src/server/server.dart
```diff
@@ -112,6 +116,24 @@ class ServeHandler implements BuildState {
...
+        var digest = await _state.reader.digest(assetId);
+        resultMap[path] = digest.toString();
+      } on AssetNotFoundException {
+        resultMap[path] = null;
```
natebosch: do we need to store null? Should we instead `resultMap.remove(path)`?

build_runner/lib/src/server/server.dart
```diff
@@ -112,6 +116,24 @@ class ServeHandler implements BuildState {
         headers: {HttpHeaders.contentTypeHeader: 'text/html'});
   }
 
+  Future<shelf.Response> _assetsDigestHandler(
+      shelf.Request request, String rootDir) async {
+    var assertPathList = jsonDecode(await request.readAsString()) as List;
+    var rootPackage = _state.packageGraph.root.name;
+    var resultMap = <String, String>{};
```
natebosch: [nit] How about `results` or `digestsByPath`

PR #1717, opened 2018-08-06
misc: move travis to run on stable

kevmoo:
example and _test run stable and dev


jakemac53:
cc @natebosch how do you feel about running with the `dev` branch only for presubmits, but having the cron job run stable/dev/edge?

natebosch:
Only dev on presubmit sgtm

PR #1720, opened 2018-08-06
Log print messages at warning

natebosch:
This comes up repeatedly that Builder authors are confused when they add
print calls and can't see the messages because info lines get
overwritten without `--verbose`, and `--verbose` adds a bunch of extra
noise that drowns out the wanted messages.

Since we expect Builder authors to rely on `log.info` for long term
logging they want to keep at info and `print` to be used only for short
term stuff it should be safe to up the level to warning.


kevmoo:
Nice!

PR #1721, opened 2018-08-06
Run dartfmt --fix to drop `new` and some `const`

natebosch:
Add the `unnecessary_new` and `unnecessary_const` lints to prevent new
instances from showing up.


natebosch:
https://github.com/dart-lang/json_serializable/issues/275 makes this tricky because our presubmit checks that the generated files are correct. We can either avoid `dartfmt --fix` on these files, or wait until json_serializable isn't producing the `new` anymore...

jakemac53:
https://github.com/dart-lang/build/pull/1726 its not producing new any more!

Issue #1702, opened 2018-08-01
Globs don't get invalidated if a generated node changes whether it is output or not


jakemac53:
The way we track globs today is only based off of new or deleted source files, and the generated asset nodes that affects.

The problem is we don't require you to actually output a node that you declare as an output, and if you change your decision then anything doing a glob that would select that file will not be invalidated.

This is going to be pretty difficult to resolve unfortunately without regressing performance severely, so we will have to put some thought into it.

PR #1722, opened 2018-08-06
Glob nodes

jakemac53:
Fixes https://github.com/dart-lang/build/issues/1702

- Adds a new node type `GlobAssetNode`, and a new interface (`NodeWithInputs`) for nodes that have some inputs and an invalidation state.
- Results of the glob are cached on the node, and re-used across the same phase.
- When using `findAssets`, the glob node gets as inputs all the available nodes, and is invalidated just like any other generated node when those inputs are invalidated.
- On new/deleted sources we still need extra logic, but now it only iterates the glob nodes to check for globs which is more efficient than before in the case that a lot of nodes in the same phase performed the same globs.


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -367,12 +379,33 @@ class _SingleBuild {
...
+      if (node.phaseNumber >= phaseNum) return false;
+      if (!phase.hideOutput && node.isHidden && node.id.package != fromPackage)
+        return false;
+      return doAfter(
+          _ensureAssetIsBuilt(node), (_) => node.wasOutput && !node.isFailure);
+    }
+    return node.isReadable && node.isValidInput;
+  }
+
+  FutureOr<dynamic> _ensureAssetIsBuilt(AssetNode node) {
```
natebosch: should this be `FutureOr<void>`?

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -621,6 +653,53 @@ class _SingleBuild {
...
+      globNode = new GlobAssetNode(
+          globNodeId, glob, phaseNum, NodeState.definitelyNeedsUpdate);
+      _assetGraph.add(globNode);
+    }
+
+    return toFuture(
+        doAfter(_updateGlobNodeIfNecessary(globNode), (_) => globNode));
+  }
+
+  FutureOr<Null> _updateGlobNodeIfNecessary(GlobAssetNode globNode) {
```
natebosch: `FutureOr<void>` (we'll need to start fixing this elsewhere, but shouldn't introduce new cases)

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -621,6 +653,53 @@ class _SingleBuild {
...
+      var actualMatches = <AssetId>[];
+      for (var node in potentialNodes) {
+        if (await _isReadableNode(
+            node, globNode.phaseNumber, globNode.id.package)) {
+          actualMatches.add(node.id);
+        }
+      }
+      globNode.results = actualMatches;
+
+      globNode.state = NodeState.upToDate;
```
natebosch: cascade operator?

build_runner_core/lib/src/asset_graph/node.dart
```diff
@@ -260,3 +255,41 @@ class PostProcessAnchorNode extends AssetNode with _SyntheticAssetNode {
         builderOptionsId);
   }
 }
+
+/// A node representing a glob ran from a builder.
+///
+/// The [id] must always be unique to a given package, phase, and glob
```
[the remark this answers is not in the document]
jakemac53: it could but we need to generate these outside of the constructor to check for their existence in the graph as well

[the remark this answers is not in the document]
natebosch: maybe add it as a static method?

build_runner_core/lib/src/generate/build_impl.dart

build_runner_core/lib/src/generate/build_impl.dart

build_runner_core/lib/src/generate/build_impl.dart

build_runner_core/lib/src/asset_graph/node.dart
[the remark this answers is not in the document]
jakemac53: done

PR #1732, opened 2018-08-08
Move live reload client code to new hot_reload_client.dart file

samogot:
It is expected that client code will become complicated, so in separate
dart file it would be easier to test and maintain.


build_runner/lib/src/server/server.dart
```diff
@@ -226,12 +226,7 @@ shelf.Handler _injectBuildUpdatesClientCode(shelf.Handler innerHandler) {
 /// Now only live-reload functional - just reload page on update message
 final _buildUpdatesInjectedJS = '''\n
 // Injected by build_runner for live reload support
-(function() {
-  var ws = new WebSocket('ws://' + location.host, ['$_buildUpdatesProtocol']);
-  ws.onmessage = function(event) {
-    location.reload();
-  };
-}());
+requirejs(['packages/build_runner/src/server/hot_reload_client.dart'])
```
natebosch: I'm not sure if we should be relying on `requirejs` here... is there any other way we can handle this?

build_runner/lib/src/server/hot_reload_client.dart.js
```diff
@@ -0,0 +1,3857 @@
+// Generated by dart2js (full emitter, strong), the Dart to JavaScript compiler version: 2.0.0-dev.69.4.
```
natebosch: we should check if there are other flags we should be using. --minify, maybe --omit-implicit-checks

ideally we should get this code size down

.gitignore
```diff
@@ -21,3 +21,6 @@ bazel-*
 packages.bzl
 BUILD
 WORKSPACE
+
+# Deps of checked-in dart2js generated code
+*.dart.js.deps
```
jakemac53: Lest be a bit more specific here, also see https://github.com/dart-lang/build/pull/1734 for the custom build script which we can update to also run on this new entrypoint.

PR #1735, opened 2018-08-08
release build_web_compilers v0.4.2

jakemac53:


build_web_compilers/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 0.4.2
 
-- Add magic comment marker for build_runner to know where to inject 
+- Add magic comment marker for build_runner to know where to inject
```
kevmoo: nit: could we be a bit more clear here? Where is it added? In which contexts?

jakemac53: I don't know that it is really relevant for users of the package (beyond the fact that its now supported)?

It is added in the main *.dart.js script though.

jakemac53: ill add a note that this is only supported with the `dartdevc` compiler

jakemac53:
Looks like a new analyzer breakage with the latest dev release - submitting on red for now.

PR #1737, opened 2018-08-09
Fix new analyzer hints, which revealed a subtle bug

jakemac53:
It looks like the inference for return values of closures is better now and that caused a couple hints.

One was technically a bug although the way in which it was used it didn't actually matter :).

### DIFF
--- a/build_runner_core/lib/src/generate/build_definition.dart
+++ b/build_runner_core/lib/src/generate/build_definition.dart
@@ -257,7 +257,8 @@ class _Loader {
           deletedSources.add(id);
           return _environment.writer.delete(id);
         }
-      }).where((v) => v is Future));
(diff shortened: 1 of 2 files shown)


kevmoo:
FYI @devoncarew @bwilkerson @scheglov @stereotype441 @MichaelRFairhurst

build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -257,7 +257,8 @@ class _Loader {
           deletedSources.add(id);
           return _environment.writer.delete(id);
         }
-      }).where((v) => v is Future));
+        return null;
+      }).whereType<Future>());
```
kevmoo: 👍

PR #1739, opened 2018-08-09
Use whereType where possible

jakemac53:
Unfortunately it looks like we forgot to add Stream.whereType for dart 2 :(. Migrated those uses to a `.where` followed by a `.cast`.


build_modules/lib/src/errors.dart
```diff
@@ -94,7 +94,7 @@ Future<String> _missingImportMessage(
   var contents = await reader.readAsString(sourceId);
   var parsed = parseDirectives(contents, suppressErrors: true);
   var import = parsed.directives
-      .where((directive) => directive is UriBasedDirective)
+      .whereType<UriBasedDirective>()
       .cast<UriBasedDirective>()
```
natebosch: Can remove the cast now

jakemac53: oh whoops missed that one, done

PR #1741, opened 2018-08-09
Implement basic hot reload

samogot:
Reload any changed modules without any specific order, than reload main
module and call main func to hopefully apply changes.

Still need to get module dependency graph to reload in topological order
and find correct accessor chain where reloaded modules can be saved in
closures.


build_runner/test/server/hot_reload_client_test.dart
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+@TestOn('browser')
```
samogot: Do I need to add any new things to .travis.yml for this test to be run?

jakemac53: Yes - we can talk offline :)

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
+  ]);
+  return completer.future;
+}
+
+class ReloadHandler {
+  final String Function(String) _moduleIdByPath;
+  final FutureOr<JsObject> Function(String) _reloadModule;
+  final Map<String, String> _digests;
+
+  ReloadHandler(Map<String, String> this._digests,
```
jakemac53: you don't need to add the explicit type here, it will be the type of the field already (didn't even realize you could! haha)

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
 final _buildUpdatesProtocol = r'$livereload';
 
-main() {
+JsObject _dartLoader = context['\$dartLoader'];
+JsObject _urlToModuleId = _dartLoader['urlToModuleId'];
+
+List<String> get moduleUrls {
+  JsArray callMethod = context['Array']
+      .callMethod('from', [_urlToModuleId.callMethod('keys', [])]);
+  return List.castFrom(callMethod);
```
natebosch: Will we potentially be accessing elements from this list multiple times? We probably want `List.from` instead of `List.castFrom`. The distinction is that `castFrom` is lazy whereas `List.from` eagerly casts each element once and then doesn't need to check types again later.
https://www.dartlang.org/guides/language/effective-dart/usage#avoid-using-cast

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
+  final String Function(String) _moduleIdByPath;
+  final FutureOr<JsObject> Function(String) _reloadModule;
+  final Map<String, String> _digests;
+
+  ReloadHandler(Map<String, String> this._digests,
+      [this._moduleIdByPath = moduleIdByPath,
+      this._reloadModule = reloadModule]);
+
+  void listener(MessageEvent e) async {
+    Map updatedAssetDigests = json.decode(e.data as String);
```
jakemac53: nit: use `var ` on the left hand side and cast it to a Map explicitly:

```dart
var updatedAssetDigests = json.decode(e.data as String) as Map<String, dynamic>;
```

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
+import 'dart:convert';
 import 'dart:html';
+import 'dart:js';
```
natebosch: we generally prefer to use `package:js` instead of `dart:js` - can you see what it would look like doing that? It's possible not much will need to change since `package:js` exports parts of `dart:js`.

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
+  final Map<String, String> _digests;
+
+  ReloadHandler(Map<String, String> this._digests,
+      [this._moduleIdByPath = moduleIdByPath,
+      this._reloadModule = reloadModule]);
+
+  void listener(MessageEvent e) async {
+    Map updatedAssetDigests = json.decode(e.data as String);
+    var moduleIdsToReload = <String>[];
+    for (String path in updatedAssetDigests.keys) {
```
jakemac53: you can swap `String` for `var` here once you make the change above

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
+        if (moduleId.endsWith('.ddc')) {
+          moduleId = moduleId.substring(0, moduleId.length - 4);
+        }
+        moduleIdsToReload.add(moduleId);
+      }
+      _digests[path] = updatedAssetDigests[path] as String;
+    }
+    if (moduleIdsToReload.isNotEmpty) {
+      await Future.wait(moduleIdsToReload
+          .map((moduleId) => Future.value(_reloadModule(moduleId))));
```
natebosch: why do we need the `Future.value` here?

samogot: Because Future.wait expects `Future`, not `FutureOr`. And mock implementation of `_reloadModule` returns plain value.

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
 
+final _assetsDigestPath = r'$assetDigests';
 final _buildUpdatesProtocol = r'$livereload';
 
-main() {
+JsObject _dartLoader = context['\$dartLoader'];
+JsObject _urlToModuleId = _dartLoader['urlToModuleId'];
+
+List<String> get moduleUrls {
+  JsArray callMethod = context['Array']
```
jakemac53: nit: Instead of creating a new array on the javascript side, you can just use `List.from` to make a copy on the Dart side.

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,85 @@
...
+
+main() async {
+  var modulePaths = moduleUrls
+      .map((key) => key.replaceFirst(window.location.origin + '/', ''))
+      .toList();
+  var modulePathsJson = json.encode(modulePaths);
+
+  var request = await HttpRequest.request('/$_assetsDigestPath',
+      responseType: 'json', sendData: modulePathsJson, method: 'POST');
+  var digests =
```
jakemac53: you can use `.cast<String, String>` on the original map instead which is a bit cleaner:

```dart
var digests = (request.response as Map).cast<String, String>();
```

build_runner/test/server/hot_reload_client_test.dart
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  test('do not reload up to date modules', () {
+    digests['file1'] = 'hash1';
+    pathToModuleId['file1'] = 'module1';
+    handler.listener(MessageEvent('', data: '{"file1":"hash1"}'));
+    verifyZeroInteractions(methods);
+  });
+
+  test('do not reload random non-module files, but remember their digest', () {
```
jakemac53: should we remember the digests in this case? what would we do with them?

samogot: To handle new modules that may appear while developing. I'm not sure yet, if it will be possible to hot-load them, but if it will - we will need their digests to hot-reload them next time

build_runner/lib/src/server/hot_reload_client.dart
samogot: It don't exports `JsObject` and other types I use. And using `@JS` annotation will ends up with large amount of boilerplate code

build_runner/lib/src/server/hot_reload_client.dart
samogot: It doesn't work this way. `JsObject` isn't subtype of `Iterable`, and JS Iterator either

matanlurey: If you don't use `dart:js` you won't get a `JsObject`.

build_runner/lib/src/server/hot_reload_client.dart
matanlurey: `JsObject` is soft-deprecated:
https://github.com/dart-lang/sdk/issues/33350

We will likely be adding lints in the near future to prevent usage:
https://github.com/dart-lang/sdk/issues/33355

Using `package:js/js_util.dart` you should have access to some lower-level APIs that don't require much boiler-plate.

build_runner/lib/src/server/hot_reload_client.dart
samogot: But there is no class in Dart I can map JS' Map or Iterator to. Or at least I failed to find one

matanlurey: I don't understand the problem. What are you trying to do?

samogot: How to get access to JS' Map's keys?
```
@JS(r'$dartLoader.urlToModuleId')
external Map urlToModuleId;
```
Fails when I do `urlToModuleId.keys` (no some method with dart-mangled name)

```
@JS
class JSIteratorItem {
 external bool get done;
 external dynamic get value;
}

@JS
class JsIterator {
  external JSIteratorItem next();
}

@JS(r'$dartLoader.urlToModuleId.keys')
external JsIterator urlToModuleIdKeys();
```
Something like this should work, but it too verbose, and I need to iterate it manually.

I wish I don't understand something and there is a better way.

matanlurey: There is no such thing as a JS Map (well, other than ES6 `Map`, that is different).

Do you mean an object in JavaScript? If that is the case, you can write:

```dart
@JS(r'$dartLoader.urlToModuleId')
external Object get urlToModuleId;

@JS('Object.keys')
external List<dynamic> _jsObjectKeys(Object any);
List<String> jsObjectKeys(Object any) => List.from(_jsObjectKeys(any));

// Example use.
void main() {
  var keys = jsObjectKeys(urlToModuleId); // List<String>
}
```

samogot: No, I mean exactly ES6 `Map`

matanlurey: Ok, sorry about that. Then the following should work:

```dart
@JS(r'$dartLoader.urlToModuleId')
external JsMap get urlToModuleId;

@JS('Map')
abstract class JsMap {
  external Object keys();
}

@JS('Array.from')
external List<dynamic> jsArrayFrom(Object any);
```


... does that help?

samogot: Yeah, thanks

build_runner/test/server/hot_reload_client_test.dart
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
grouma: [nit] 2018

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,107 @@
...
+
+Future<Object> reloadModule(String moduleId) {
+  var completer = Completer<Object>();
+  dartLoader.forceLoadModule(moduleId, allowInterop(completer.complete));
+  return completer.future;
+}
+
+class ReloadHandler {
+  final String Function(String) _moduleIdByPath;
+  final FutureOr<Object> Function(String) _reloadModule;
```
grouma: Should this be a typedef to make it easier to mock? I find it a little odd in the test to have an abstract class of methods.

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,107 @@
...
+String moduleIdByPath(String path) =>
+    dartLoader.urlToModuleId.get(window.location.origin + '/' + path);
+
+Future<Object> reloadModule(String moduleId) {
+  var completer = Completer<Object>();
+  dartLoader.forceLoadModule(moduleId, allowInterop(completer.complete));
+  return completer.future;
+}
+
+class ReloadHandler {
```
grouma: [nit] This should probably get a doc comment.

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,107 @@
...
+external DartLoader get dartLoader;
+
+@JS('Array.from')
+external List jsArrayFrom(Object any);
+
+List<String> get moduleUrls {
+  return List.from(jsArrayFrom(dartLoader.urlToModuleId.keys()));
+}
+
+String moduleIdByPath(String path) =>
```
grouma: Does this need to be public? Same with moduleUrls?

build_runner/test/server/hot_reload_client_test.dart
```diff
@@ -0,0 +1,86 @@
...
+    await untilCalled(methods.main());
+    verifyInOrder([
+      methods.reloadModule('module1'),
+      methods.reloadModule('web/main'),
+      methods.main()
+    ]);
+    verifyNoMoreInteractions(methods);
+  });
+
+  test('dropes .ddc suffix from module id', () async {
```
grouma: [nit] typo

build_runner/lib/src/server/hot_reload_client.dart
```diff
@@ -2,14 +2,107 @@
...
+  var digests = (request.response as Map).cast<String, String>();
+
+  var handler = ReloadHandler(digests);
+
   var webSocket =
       WebSocket('ws://' + window.location.host, [_buildUpdatesProtocol]);
-  webSocket.onMessage.listen((MessageEvent e) {
-    window.location.reload();
-  });
+  webSocket.onMessage.listen(handler.listener);
```
grouma: Not for this PR but do we need some error handling on this connection? Should we attempt to reconnect if there are issues. I can go into detail for what we do in DDR.

build_runner/lib/src/server/hot_reload_client.dart
samogot: I haven't find any examples how to mock functions/typedefs directly outside of a class/interface. Also packing functions into class allows to use `verifyZeroInteractions` and `verifyNoMoreInteractions`

build_runner/pubspec.yaml
```diff
@@ -45,6 +45,8 @@ dev_dependencies:
   build_test: ^0.10.0
   build_modules: ^0.3.0
   build_web_compilers: ^0.4.0
+  js: ^0.6.1+1
```
jakemac53: This is imported from `lib` and so it should move to a real dependency.

Issue #1059, opened 2018-02-27
Bootstrap logic doesn't support WebWorkers


kevmoo:
https://github.com/dart-lang/build/blob/fc4bb8564290d429c704377f87ff341bcd907a19/build_web_compilers/lib/src/dev_compiler_bootstrap.dart#L137

Blocking https://github.com/dart-lang/angular/issues/938

natebosch:
Being able to have a single "entrypoint" js which pulls in multiple other scripts is pretty important to us. Without this I think we'd need to concatenate other files together or something along those lines.

I wonder if there is another approach we could use to get the same effect as a web worker...

natebosch:
`importScripts` should work in Chrome.
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts

kevmoo:
I got pretty far with this

```diff
diff --git a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
index 96cf057..a5c78a1 100644
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -134,19 +134,33 @@ String _entryPointJs(String bootstrapModuleName) => '''
 (function() {
   $_currentDirectoryScript
   $_baseUrlScript
-  var el;
-  el = document.createElement("script");
-  el.defer = true;
-  el.async = false;
-  el.src =
-    baseUrl + "packages/\$sdk/dev_compiler/web/dart_stack_trace_mapper.js";
-  document.head.appendChild(el);
-  el = document.createElement("script");
-  el.defer = true;
-  el.async = false;
-  el.src = baseUrl + "packages/\$sdk/dev_compiler/amd/require.js";
-  el.setAttribute("data-main", _currentDirectory + "$bootstrapModuleName");
-  document.head.appendChild(el);
+
+  var mapperUri = baseUrl + "packages/\$sdk/dev_compiler/web/dart_stack_trace_mapper.js";
+  var requireUri = baseUrl + "packages/\$sdk/dev_compiler/amd/require.js";
+  var mainUri = _currentDirectory + "$bootstrapModuleName";
+
+  if (typeof document != 'undefined') {
+    var el;
+    el = document.createElement("script");
+    el.defer = true;
+    el.async = false;
+    el.src = mapperUri;
[...]

PR #1116, opened 2018-03-07
Support bootstrapping DDC for web workers

kevmoo:
Fixes https://github.com/dart-lang/build/issues/1059

PR #1758, opened 2018-08-13
fix: use postMessage during bootstrap iff the current context is Window

kevmoo:
Fixes https://github.com/dart-lang/build/issues/1059

Improves on logic from https://github.com/dart-lang/build/pull/1116

Just checking for `postMessage` is not sufficient.
DedicatedWorkerGlobalScope also has `postMessage`, but with only one
parameter instead of two. This causes a runtime failure when loading
a WebWorking with DDC + build_web_compilers

Use a type test instead

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -267,7 +267,7 @@ $_baseUrlScript
         return dart.getSourceMap(module);
       });
   }
-  if (window.postMessage) {
+  if (typeof Window === 'function') {
     window.postMessage({ type: "DDC_STATE_CHANGE", state: "start" }, "*");
   }
 ''';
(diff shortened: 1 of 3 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -267,7 +267,7 @@ $_baseUrlScript
         return dart.getSourceMap(module);
       });
   }
-  if (window.postMessage) {
+  if (typeof Window === 'function') {
```
natebosch: synced offline - this feels fragile to me. I think we should do some research to see if there is a canonical way to determine if we're the rendering window or a worker.

kevmoo:
Will look into things from
https://stackoverflow.com/questions/7931182/reliably-detect-if-the-script-is-executing-in-a-web-worker

kevmoo:
@natebosch – PTAL

jakemac53:
Should this just be a + release?

kevmoo:
> Should this just be a + release?

Sure. Wasn't sure if ya'll wanted to publish right away.

Honestly, there's one more thing I want to dig in on first...

kevmoo:
Or https://github.com/dart-lang/build/blob/master/build_web_compilers/lib/src/dev_compiler_bootstrap.dart#L163-L165

We could do that, too...

PR #1764, opened 2018-08-15
Simplify module interface

samogot:


build_runner/lib/src/server/hot_reload_client/client.dart
```diff
@@ -69,26 +68,34 @@ class ModuleWrapper implements Module {
   ModuleWrapper(this._internal);
 
   @override
-  bool get hasOnDestroy =>
-      _internal != null && hasProperty(_internal, r'hot$onDestroy');
+  Object onDestroy() {
+    try {
```
jakemac53: I would rather check for the existence of the property instead of catching the error, these patterns can be really annoying when debugging if you set it to stop on all errors (even handled errors).

Issue #1762, opened 2018-08-14
Hot restart should restore the page to its initial state by default, or require a custom hook


jakemac53:
Today the default handler simply invokes `main()`, but that won't get rid of the existing application on the page. Thus, if you append some tags to the body in your `main()`, then it will keep appending new copies of the app.

Internally we deep clone the original `document.body` before initially calling `main`, and then restore it to that state before we re-invoke `main`. This definitely isn't a universal solution, but is somewhat reasonable and tends to work for a lot of apps.

We could either implement that as the default, or we could possibly just require some top level hook to be implemented before we will even attempt to hot restart.

samogot:
With #1765 it would be possible to implement client hook this way
```
var _id = 'hello';

void main() {
  var helloDiv = Element.div()
    ..id = _id
    ..text = 'Hello Dart';
  document.body.append(helloDiv);
}

Object hot$onDestroy() {
  document.body.querySelectorAll('#$_id').forEach((e) => e.remove());
}
```

It will work even if `_id` is changed between versions.

jakemac53:
I think that as long as this is well documented it should be sufficient - one nice thing about these magic top level hooks is dart2js should optimize them away as well :).

PR #1770, opened 2018-08-17
Refactor and improve HMR code

samogot:
- Get rid of `roughLibraryKeyDecode` - use library path exposed from
  dart runtime now.
- Get load modules from cache directly and throw if it isn't loaded
  instead of using require.js and hoping that module already loaded and
  no network request would be done.
- Don't mess with `.ddc` extension in `build_runner` code - encapsulate
  all of it on build_web_compilers side.

Fixes https://github.com/dart-lang/build/issues/1760
Fixes https://github.com/dart-lang/build/issues/1762


build_runner/lib/src/server/hot_reload_client/client.dart
```diff
@@ -149,27 +148,29 @@ List<K> keys<K, V>(JsMap<K, V> map) {
...
-        completer.complete(module);
-      }),
-          allowInterop((e) => completer.completeError(
-              HotReloadFailedException(e.message), stackTrace)));
-      return completer.future;
-    };
-
-var _reloadModule = _futurifyLoaderFunction(dartLoader.forceLoadModule);
-var _loadModule = _futurifyLoaderFunction(dartLoader.loadModule);
+Module _getModuleLibraries(String moduleId) {
```
natebosch: [nit] consider omitting `get` from the method name. `_moduleLibraries`.

https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

build_runner/lib/src/server/hot_reload_client/client.dart
```diff
@@ -149,27 +148,29 @@ List<K> keys<K, V>(JsMap<K, V> map) {
...
+  var moduleObj = dartLoader.getModuleLibraries(moduleId);
+  if (moduleObj == null)
```
natebosch: when the entire conditional doesn't fit on one line add the `{}`

https://www.dartlang.org/guides/language/effective-dart/style#do-use-curly-braces-for-all-flow-control-structures

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -235,31 +250,34 @@ if(!window.\$dartLoader) {
...
          if (typeof onError != 'undefined') {
            errorCallbacks.get(moduleName).delete(onError);
          }
          if (typeof callback != 'undefined') {
-           callback(module);
+           callback();
          }
        });
      },
+     getModuleLibraries: null,
```
jakemac53: Lets add a comment here pointing to when/how this is actually getting set up

PR #1775, opened 2018-08-21
Close Hot Reload listeners on exit

### LINKED ISSUE #1774: Hot Reload causes server to not exit on first ctrl-c
I noticed when playing around with --hot-reload today that it is currently causing the server process to not exit when you send the first terminate signal.

Generally this is a sign that there are some stream listeners or web socket connections still open. These are typically cleaned up here https://github.com/dart-lang/build/blob/master/build_runner/lib/src/generate/watch_impl.dart#L307, so we probably need to add some more things to that.

samogot:
Closes https://github.com/dart-lang/build/issues/1774

### DIFF
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -203,11 +203,10 @@ class BuildUpdatesWebSocketHandler {
     }
   }
 
-  Future<void> close() async {
-    await Future.wait(connectionsByRootDir.values
+  Future<void> close() {
+    return Future.wait(connectionsByRootDir.values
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/server/server.dart
```diff
@@ -202,6 +202,13 @@ class BuildUpdatesWebSocketHandler {
       connectionsByRootDir.remove(rootDir);
     }
   }
+
+  Future<void> close() async {
```
jakemac53: nit: instead of using async/await here just return the Future.wait

PR #1776, opened 2018-08-21
fix: don't log severe when an asset URI cannot be resolved

kevmoo:
Follow-up to https://github.com/dart-lang/json_serializable/issues/236

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -297,7 +297,6 @@ class _AssetUriResolver implements UriResolver {
     } else {
       assetId = _resolve(null, uri.toString());
       if (assetId == null) {
-        log.severe('Unable to resolve asset ID for "$uri"');
         return null;
       }
     }
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.2.2+3
+
+- Don't log a severe message when a URI cannot be resolved. Just return `null`.
+
 ## 0.2.2+2
 
 - Use sdk summaries for the analysis context, which makes getting the initial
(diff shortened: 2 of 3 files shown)


kevmoo:
So throwing for dart: URIs lets us succeed, but `log.severe` fails.

Weird?

natebosch:
>So throwing for dart: URIs lets us succeed, but `log.severe` fails.
> Weird?

If the builders catches the exception this is expected. We always fail for `log.severe` but we only fail for exceptions the build system catches, not those that the builder catches and swallows.

build_resolvers/lib/src/resolver.dart
```diff
@@ -289,17 +289,21 @@ class _AssetUriResolver implements UriResolver {
...
+      case 'dart':
+        throw ArgumentError.value(uri, 'uri', 'dart: scheme is not supported.');
+      case 'asset':
+        var parts = uri.pathSegments;
+        assetId = AssetId(parts[0], path.joinAll(parts.skip(1)));
+        break;
+      default:
+        assetId = _resolve(null, uri.toString());
+        if (assetId == null) {
+          log.severe('Unable to resolve asset ID for "$uri"');
```
jakemac53: Should we throw here also then?

kevmoo: 🤷‍♂️

jakemac53:
It looks like we should actually just be returning null if we can't handle the URI, can you try changing it to do that @kevmoo ?

kevmoo:
> can you try changing it to do that

Sure. Should we at least warn? Or nothing?

jakemac53:
> Sure. Should we at least warn? Or nothing?

I think nothing, my guess is the analyzer will end up throwing some sort of exception if nothing handles the uri, so we might want to handle that but in a different place.

kevmoo:
PTAL

PR #1778, opened 2018-08-21
Restore --live-reload option

samogot:


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -131,12 +132,14 @@ class SharedOptions {
 /// Options specific to the `serve` command.
 class ServeOptions extends SharedOptions {
   final String hostName;
+  final bool liveReload;
   final bool hotReload;
   final bool logRequests;
   final List<ServeTarget> serveTargets;
 
   ServeOptions._({
```
jakemac53: We should add some validation here that both `hotReload` and `liveReload` are not true.

build_runner/README.md
```diff
@@ -99,8 +99,9 @@ Some commands also have additional options:
 ##### serve
 
 - `--hostname`: The host to run the server on.
-- `--hot-reload`: Enables automatic module reloading on rebuilds. 
-See [hot-module-reloading](../docs/hot_module_reloading.md) for more info.
+- `--live-reload`: Enables automatic page reloading on rebuilds.
```
grouma: If you provide both which one takes precedence? Could potentially model this as an enum to prevent that issue.

Issue #1780, opened 2018-08-22
Error output references [filename].ddc.js


afpatmin:
When debugging a web app with `<base href="/">` set, error stacktraces only references .ddc.js files.

I've created a minimal repro here:
[https://github.com/afpatmin/sourcemap_issue_repro](https://github.com/afpatmin/sourcemap_issue_repro)

Here's the error output with `<base href="/">` set:
[https://www.dropbox.com/s/b00hdmyaavlnr10/with-base.png?dl=0](https://www.dropbox.com/s/b00hdmyaavlnr10/with-base.png?dl=0)

And here it is without:
[https://www.dropbox.com/s/bcx6rpn9kv7wiv5/without-base.png?dl=0](https://www.dropbox.com/s/bcx6rpn9kv7wiv5/without-base.png?dl=0)

"Enable JavaScript source maps" and "Enable custom formatters" are both checked in Chrome console settings.

Dart VM version: 2.0.0 (Unknown timestamp) on "linux_x64"
build_runner: 0.10.1+1
build_web_compilers: 0.4.2+2
Builders: default
Browser: Chrome Version 68.0.3440.106 (Official Build) (64-bit)

jakemac53:
Do you get general source maps or does that not work either?

afpatmin:
Yes source maps works properly right up to the point when I add the `base` tag

jakemac53:
Right, but do the source maps show up in dev tools at all with the `base` tag, or is it only for stack traces that it appears to be an issue?

afpatmin:
Aha, yes it seems only stack traces doesn't work. I can still browse the .dart files directly under the sources tab, set breakpoints etc. I created a minimal project to demonstrate the issue if you like to test it yourself

drbcode:
This PR will fix this issue:
https://github.com/dart-lang/build/pull/1781

PR #1781, opened 2018-08-23
Fix sourcemaps and stack trace issues while using different baseHref

drbcode:
At the moment source maps are not working properly when `<base>` tag is used.

See this issue for example :
https://github.com/dart-lang/build/issues/1780

```js
window.\$dartLoader.rootDirectories.push(window.location.origin + baseUrl);
```

This method requires `baseUrl` to be a relative path and not an absolute URL.

`stack_trace_mapper.dart` relies on correctness of `$dartLoader.rootDirectories` to generate stack trace.

Because of adding incorrect URLs to `$dartLoader.rootDirectories` sourcemaps are not working, 

Merging above PR will fix this issue, no further action is needed.


drbcode:
I signed it!

Issue #1752, opened 2018-08-13
DDC ModuleBuilder with fine strategy doesn't handle conditional imports


nickclmb:
When ModuleBuilder runs with fine strategy (as example packages with circular dependencies) then .module files are missing dependencies from conditional imports:

```
import 'A.dart'                            // this uri present in deps
if (dart.library.io) 'A_io.dart'           // <-- missing
if (dart.library.html) 'A_html.dart';      // <-- missing
```

So as a result, DDC fails with error "'A_html.dart' not found."

As i can see the LibraryElement from resolver doesn't contain any info about import configurations, but ModuleLibrary output already has all dependencies.
So is it possible to use ModuleLibrary to build modules in Fine strategy instead of using resolver?

jakemac53:
Not using a Resolver here should also improve build times for projects that don't otherwise require a resolver.

natebosch:
I had an experimental branch using `ModuleLibrary` for the fine strategy and hit some roadblocks - it got a little complex because it can't be done in the number of phases we're using today. I think it could be done with an additional intermediate output though.

jakemac53:
The fix is published now with build_modules 0.3.2

PR #1782, opened 2018-08-23
Change the fine module strategy to output to meta modules

jakemac53:
All strategies now create meta_module files, and the analyzer dependency is now removed from build_modules :tada:.

- Module strategies are now respected for all packages instead of just the root package
- Deleted a lot of fallback code to discover when a package is using the fine strategy
- Can now mix and match fine and coarse strategies at will, even within package cycles (although this may cause larger modules)
- Added an explicit notion of "missing" modules - these are warned about when they are discovered but won't fail until some other builder tries to compute transitive dependencies.
- Unblocks config specific imports work (we are always doing our own crawling now instead of relying on analyzers computation of library cycles).

Fixes https://github.com/dart-lang/build/issues/1752

### DIFF
--- a/build_modules/lib/src/common.dart
+++ b/build_modules/lib/src/common.dart
@@ -34,19 +34,19 @@ Future<File> createPackagesFile(Iterable<AssetId> allAssets) async {
 enum ModuleStrategy { fine, coarse }
 
 ModuleStrategy moduleStrategy(BuilderOptions options) {
-  if (options.isRoot) {
-    var config = options.config['strategy'] as String ?? 'coarse';
-    switch (config) {
-      case 'coarse':
-        return ModuleStrategy.coarse;
-      case 'fine':
-        return ModuleStrategy.fine;
-      default:
-        throw 'Unexpected ModuleBuilder strategy: $config';
-    }
-  } else {
-    return ModuleStrategy.coarse;
+  // if (options.isRoot) {
+  var config = options.config['strategy'] as String ?? 'coarse';
+  switch (config) {
+    case 'coarse':
+      return ModuleStrategy.coarse;
+    case 'fine':
+      return ModuleStrategy.fine;
+    default:
+      throw 'Unexpected ModuleBuilder strategy: $config';
   }
+  // } else {
+  //   return ModuleStrategy.coarse;
+  // }
 }
 
 /// Validates that [config] only has the top level keys [supportedOptions].
--- a/build_modules/lib/src/common.dart
+++ b/build_modules/lib/src/common.dart
@@ -41,7 +41,7 @@ ModuleStrategy moduleStrategy(BuilderOptions options) {
     case 'fine':
       return ModuleStrategy.fine;
     default:
-      throw 'Unexpected ModuleBuilder strategy: $config';
+      throw ArgumentError('Unexpected ModuleBuilder strategy: $config');
   }
 }
 
--- a/build_modules/lib/src/common.dart
+++ b/build_modules/lib/src/common.dart
@@ -34,7 +34,6 @@ Future<File> createPackagesFile(Iterable<AssetId> allAssets) async {
 enum ModuleStrategy { fine, coarse }
 
 ModuleStrategy moduleStrategy(BuilderOptions options) {
-  // if (options.isRoot) {
   var config = options.config['strategy'] as String ?? 'coarse';
   switch (config) {
     case 'coarse':
@@ -44,9 +43,6 @@ ModuleStrategy moduleStrategy(BuilderOptions options) {
     default:
       throw 'Unexpected ModuleBuilder strategy: $config';
   }
-  // } else {
-  //   return ModuleStrategy.coarse;
-  // }
 }
 
 /// Validates that [config] only has the top level keys [supportedOptions].
--- a/build_modules/lib/src/errors.dart
+++ b/build_modules/lib/src/errors.dart
@@ -51,11 +51,8 @@ class MissingModulesException implements Exception {
 
   MissingModulesException._(this.message);
 
-  static Future<MissingModulesException> create(
-      Module module,
-      Set<AssetId> missingSources,
-      List<Module> transitiveModules,
-      AssetReader reader) async {
+  static Future<MissingModulesException> create(Set<AssetId> missingSources,
+      Iterable<Module> transitiveModules, AssetReader reader) async {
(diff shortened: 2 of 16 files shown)


build_modules/lib/src/common.dart
```diff
@@ -34,18 +34,14 @@ Future<File> createPackagesFile(Iterable<AssetId> allAssets) async {
...
-      case 'coarse':
-        return ModuleStrategy.coarse;
-      case 'fine':
-        return ModuleStrategy.fine;
-      default:
-        throw 'Unexpected ModuleBuilder strategy: $config';
-    }
-  } else {
-    return ModuleStrategy.coarse;
+  var config = options.config['strategy'] as String ?? 'coarse';
```
grouma: Why the removal of the root conditional? Won't that slow things down?

build_modules/lib/src/meta_module.dart
```diff
@@ -209,23 +210,47 @@ class MetaModule {
...
+  for (var library in libraries) {
+    final dir = _topLevelDir(library.id.path);
+    if (!librariesByDirectory.containsKey(dir)) {
+      librariesByDirectory[dir] = <AssetId, ModuleLibrary>{};
+    }
+    librariesByDirectory[dir][library.id] = library;
+  }
+  final modules = librariesByDirectory.values.expand(_computeModules).toList();
+  // Deterministically output the modules.
+  modules.sort((a, b) => a.primarySource.compareTo(b.primarySource));
```
grouma: [optional] Pull this sort into a private function with a descriptive name so the comment here and below is no longer needed.

build_modules/lib/src/meta_module_clean_builder.dart
```diff
@@ -120,6 +102,12 @@ Future<Set<Module>> _transitiveModules(
         continue;
       }
       seenMetas.add(depMetaAsset);
+      if (!await buildStep.canRead(depMetaAsset)) {
+        log.warning('Unable to read module information for '
+            'package:${depMetaAsset.package}, make sure you have a dependency '
+            'in it in your pubspec.');
```
grouma: [typo] on* it

build_modules/lib/src/modules.dart
```diff
@@ -88,32 +87,15 @@ class Module {
       fromJson: _assetIdsFromJson)
   final Set<AssetId> directDependencies;
 
+  @JsonKey(name: 'm', nullable: true, defaultValue: false)
+  final bool isMissing;
```
grouma: Let's add a comment why it would be missing.

build_modules/lib/src/module_builder.dart
```diff
@@ -37,60 +36,10 @@ class _CleanMetaModuleCache {
   }
 }
 
-/// Updates dependencies from the provided [Module] so that they point to the
-/// primary resource of the dependencies corresponding [Module].
-///
-/// Note that this will process the clean meta modules for the module
-/// dependencies if necessary.
-Future<Module> _cleanModuleDeps(
```
grouma: I don't follow why this is no longer necessary. Can you quickly explain?

build_modules/lib/src/common.dart
jakemac53: Previously injecting a fine strategy module in a dependency somewhere would cause all packages that depend on that package to also get the fine strategy, which I believe is why we silently ignore the config before.

Now you can mix and match the strategies without that cascading behavior - so you don't get the same behavior. Only the specific packages that opted into the fine strategy would get it.

That does mean that users adding the fine strategy but really only intending it to be an option for local development will end up with the fine strategy when people depend on them though... so you would end up needing a second build.yaml or something for local dev. Hopefully that is a rare case, but its hard to really say what the right call is here, silently ignoring it doesn't seem right any more though.

build_modules/lib/src/module_builder.dart
jakemac53: This was all only used by the fine strategy fallback codepath - which no longer exists.

jakemac53:
> This looks like a great clean up but I don't follow how the mixing of strategy actually works. Specifically why we no longer need to clean the module dependencies.

We do still clean the module dependencies - we just only have to do it in the MetaModuleCleanBuilder - we used to duplicate the logic in the ModuleBuilder when using the fine strategy.

build_modules/lib/src/common.dart
```diff
@@ -34,18 +34,14 @@ Future<File> createPackagesFile(Iterable<AssetId> allAssets) async {
...
+  switch (config) {
+    case 'coarse':
+      return ModuleStrategy.coarse;
+    case 'fine':
+      return ModuleStrategy.fine;
+    default:
+      throw 'Unexpected ModuleBuilder strategy: $config';
```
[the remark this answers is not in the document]
jakemac53: We should be, but have other violations right now. Updated this to an ArgumentError and filed https://github.com/dart-lang/build/issues/1785

build_modules/lib/src/errors.dart
```diff
@@ -51,11 +51,8 @@ class MissingModulesException implements Exception {
 
   MissingModulesException._(this.message);
 
-  static Future<MissingModulesException> create(
-      Module module,
```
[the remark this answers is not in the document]
jakemac53: gotta love how people actively push back on this not being an error.... so useful

build_modules/lib/src/meta_module.dart
jakemac53: done

build_modules/lib/src/meta_module_clean_builder.dart
jakemac53: done

build_modules/lib/src/modules.dart
jakemac53: done

build_modules/pubspec.yaml
```diff
@@ -8,7 +8,6 @@ environment:
   sdk: ">=2.0.0 <3.0.0"
 
 dependencies:
-  analyzer: ">0.30.0 < 0.33.0"
```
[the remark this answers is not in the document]
jakemac53: done :(

kevmoo:
> analyzer dependency is now removed from build_modules

It is? I still see it there...

jakemac53:
> It is? I still see it there...

Forgot to update the comment, it is actually still there because we still parse the directives. We no longer use the element model at least though.

PR #1783, opened 2018-08-23
Fix sourcemaps and stack trace issues while using base tag

drbcode:
At the moment source maps are not working properly when `<base>` tag is used.

See this issue for example :
https://github.com/dart-lang/build/issues/1780

```js
window.\$dartLoader.rootDirectories.push(window.location.origin + baseUrl);
```

This method requires `baseUrl` to be a relative path and not an absolute URL.

`stack_trace_mapper.dart` relies on correctness of `$dartLoader.rootDirectories` to generate stack trace.

Because of adding incorrect URLs to `$dartLoader.rootDirectories` sourcemaps are not working, 

Merging above PR will fix this issue, no further action is needed.


natebosch:
cc @vsmenon @jmesserly 

There is some interaction between the base href and DDC stack trace mapping, along with our bootstrapping code, which we aren't fully sure about. Is this something you can help dig in with?

Some previous history around this code:
https://github.com/dart-lang/build/pull/1332
https://github.com/dart-lang/build/pull/1353

vsmenon:
We pulled the mapping logic out of DDC proper - it's now here under build:

https://github.com/dart-lang/build/blob/master/build_web_compilers/lib/src/dev_compiler_stack_trace/source_map_stack_trace.dart#L21

natebosch:
The two things changed in this CL are:

- changed ordering of the search - now checking for `test` in `location` before looking at the base tag.
- Added an `endsWith("/")` check in addition to the existing `startsWith("/")`

In the issue the specific case you mentioned is `<base href="/">`.  My assumption is you aren't hitting a `location` with `test` - is that correct?
Then `"/"` both starts and ends with `"/"` so the change in that conditional shouldn't have mattered.

I'm not quite understanding how this change fixes the situation for you.

drbcode:
Link to original issue:
https://github.com/dart-lang/build/issues/1780

a minimal repro here:
https://github.com/afpatmin/sourcemap_issue_repro

packages used:
name: build_web_compilers
version: 0.4.2+2

As described in above issue, stack traces work until a user introduces a  `<base href="/">` tag.
### Example that demonstrates this issue
isuue

"Enable JavaScript source maps" and "Enable custom formatters" are both checked in Chrome console settings.

As you can see in above image, source maps load, user can set breakpoints and debug things but stacktraces are only referring to  js files.

So this issue has to do with something not working well with `stack_trace_mapper.dart.js` created by `build_web_compilers`.

In first screenshot I have called incorrect `baseUrl` script used by `build_web_compilers version: 0.4.2+2`
it returns `"http://localhost:8080/"`. 

This value is then used by this line:

https://github.com/dart-lang/build/blob/1c111a71761603ba62dbff77b99ab42f1e1aaffc/build_web_compilers/lib/src/dev_compiler_bootstrap.dart#L232


If you take a look at `stack_trace_mapper.dart`, it has [this code](https://github.com/dart-lang/build/blob/master/build_web_compilers/lib/src/dev_compiler_stack_trace/stack_trace_mapper.dart#L39-L40):


```dart
@JS(r'$dartLoader.rootDirectories')
external List get rootDirectories;
```

In first screenshot I have called `$dartLoader.rootDirectories` and it logs this to the console:
```
List<dynamic> length 1
0: http://localhost:8080http://localhost:8080/
```
As you can see, this value should be `http://localhost:8080` and not `http://localhost:8080http://localhost:8080/`.

Because this incorrect `$dartLoader.rootDirectories` is being used by `stack_trace_mapper.dart`, stack traces are not working properly.

#### Working example:

non_issue

As you can see in this screenshot shown above, stack traces are working properly, `$dartLoader.rootDirectories` is set correctly and `baseUrl` script is also returning correct value.

#### Working example with a different base href 

non_issue_href_changed

Above project was built using this command:
```bash
pub run build_runner build -o build
```

A golang server was used to serve static content from build directory:
Contents of `main.go` place this file at project root, also change value of base href tag to `<base href="/abc/">` as shown in above screenshot:

```golang
package main
import (
  "net/http"
  "log"
)
func main() {
  fs := http.FileServer(http.Dir("build/web"))
  http.Handle("/abc/", http.StripPrefix("/abc/", fs))
  log.Println("Listening...")
  http.ListenAndServe(":3000", nil)
}
```

As you can see in this screenshot shown above, stack traces are working properly, `$dartLoader.rootDirectories` is set correctly and `baseUrl` script is also returning correct value, everything works as it is expected to work.

#### working baseUrl script

Here is the correct `baseUrl` script that will resolve the issue.
```js
var baseUrl = (function () {
  if (pathParts[0] == "") {
    pathParts.shift();
  }
  }
  // base href should start and end with "/"
    if (el && el[0] && el[0].getAttribute("href") && el[0].getAttribute
    ("href").startsWith("/") && el[0].getAttribute("href").endsWith("/")){
      return el[0].getAttribute("href");
    }
  }
  // return default value
  return "/";
}());
```

- changed ordering of the search - now checking for test in location before looking at the base tag. So if the scripts are running on a test page then we don't look it value of base tag, because if value of base tag is used in this situation then it will cause test to fail.

- Added an endsWith("/") check in addition to the existing startsWith("/"), because if a user sets a `href` value that does not end with `/` then require js can not resolve it.

- Previous script returned `return el[0].href`, this returns an absolute URL and not a relative path, new script returns `return el[0].getAttribute("href")`, and if this attribute value begins and ends with `/` then it can be treated as a relative path for properly setting `baseUrl`.

This PR is an improvement over script we had before.

natebosch:
cc @dineshbhosale - I think this change makes sense given how we use the `baseUrl` variable [here](https://github.com/dart-lang/build/blob/715f6ecc3c320b0192d5d6120642eabc92aaa22b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart#L232)

Can you confirm whether this change still works for you?

dineshbhosale:
@natebosch @drbcode and me are same people, @drbcode  is my new profile.
Yes this change works as expected, it is better than what we had before.

Also I feel sorry because I did not see whether sourcemaps are correct or not when I sent [this PR](https://github.com/dart-lang/build/pull/1332) for implementing a better baseUrl script.

natebosch:
LGTM. Will merge once travis finishes.

PR #1793, opened 2018-08-27
Adds support for config specific imports

jakemac53:
Most of the changes are to the build_modules package, and there is now a separate set of module files per platform.

- Each platform is set up with a builder in the build_modules package - but they are not applied by default. Any new platforms should also be configured in the build_modules package.
- Individual compiler packages (or other consumers of modules) must apply the module builder for the platforms they care about, using `applies_builders` config.

cc @pulyaevskiy the build_node_compilers package will need an update as well - see the build_web_compilers package updates here.


build_modules/lib/src/meta_module_builder.dart
```diff
@@ -11,35 +11,50 @@ import 'package:glob/glob.dart';
...
   @override
   Future build(BuildStep buildStep) async {
     var libraryAssets =
         await buildStep.findAssets(Glob('**$moduleLibraryExtension')).toList();
-    var metaModule =
-        await MetaModule.forLibraries(buildStep, libraryAssets, strategy);
-    var id = AssetId(buildStep.inputId.package, 'lib/$metaModuleExtension');
+    var platformLoader = await buildStep.fetchResource(platformsLoaderResource);
+    var platforms = await platformLoader.load(buildStep);
+    var platform = platforms[_platform];
```
grouma: Can we wrap this logic up in a method? It took a while to parse what was going on.

build_modules/lib/src/module_library.dart
```diff
@@ -106,32 +143,64 @@ class ModuleLibrary {
...
+      assert(selectedImport != null);
+      for (var condition in conditions.keys) {
+        if (condition == r'$default') continue;
+        if (!condition.startsWith('dart.library.')) {
+          throw UnsupportedError(
+              '$condition not supported for config specific imports. Only the '
+              'dart.library.<name> constants are supported.');
+        }
+        var library = condition.substring('dart.library.'.length);
+        var coreLibSupport = platform.libraries[library];
```
grouma: I'm not sure what's going on here with coreLibSupport.

build_modules/lib/src/platform.dart
```diff
@@ -0,0 +1,120 @@
...
+/// calls [invalidate] between builds.
+class PlatformsLoader {
+  Future<Platforms> _platforms;
+  Digest _previousDigest;
+  bool _invalidated = true;
+
+  PlatformsLoader._();
+
+  Future<Platforms> load(AssetReader reader) async {
+    if (_invalidated) {
```
natebosch: [optional - not worth a rebase if you planned on merging instead of squashing] invert the condition and bail out early to reduce 1 level of nesting?

build_modules/lib/src/module_library.dart
```diff
@@ -70,16 +96,27 @@ class ModuleLibrary {
...
-      deps.add(linkedId);
       if (conditionalDirectiveConfigurations != null) {
-        deps.addAll(conditionalDirectiveConfigurations
-            .map((c) => Uri.parse(c.uri.stringValue))
-            .where((u) => u.scheme != 'dart')
-            .map((u) => AssetId.resolve(u.toString(), from: id)));
+        var conditions = <String, AssetId>{r'$default': linkedId};
+        for (var condition in conditionalDirectiveConfigurations) {
+          if (Uri.parse(condition.uri.stringValue).scheme == 'dart') {
+            throw 'Unsupported conditional import of '
```
natebosch: Why is this unsupported?

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## 0.4.0
+
+### Improvements
+
+- Modules are now platform specific, and config specific imports are supported,
+  but only for `dart.library.*` constants.
```
natebosch: I think this is already an constraint on conditional imports in the CFE (though it's maybe not documented super well?)

I don't think we should phrase this as if it's a limitation. "config specific imports against `dart.library.*` constants are supported"

build_modules/build.yaml
```diff
@@ -1,26 +1,51 @@
...
-        - .unlinked.sum
-    is_optional: True
     auto_apply: all_packages
+    is_optional: True
     required_inputs: [".dart"]
     applies_builders: ["|module_cleanup"]
+  dartdevc:
+    import: "package:build_modules/builders.dart"
+    builder_factories:
+      - metaModuleBuilderFactoryForPlatform('dartdevc')
```
natebosch: [aside] If we had union types I would totally make the `BuilderFactory` typedef be `Builder|List<Builder> Function()`

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,36 @@
...
+
+```yaml
+targets:
+  $default:
+    build_modules|dartdevc:
+      options:
+        strategy: fine
+```
+
+  The supported platforms are currently `dart2js`, `dartdevc`, and `vm`.
```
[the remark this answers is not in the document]
natebosch: Because it can't be discovered in the same `libraries.json` file 😢

pulyaevskiy:
Thanks for tagging me @natebosch! Will make sure to update when this lands.

build_modules/CHANGELOG.md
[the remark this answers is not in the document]
grouma: Boo

_test/test/config_specific_import_test.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+// ignore: uri_does_not_exist
```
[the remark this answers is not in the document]
jakemac53: See https://github.com/dart-lang/sdk/issues/34177

jakemac53:
@natebosch @grouma I am reconsidering using the libraries.json file for a couple reasons, curious what you think.

- It is unlikely that a platform changes its dart: library support any time soon
- Flutter doesn't ship a libraries.json file in the same place
- It makes the tests easier to write (don't have to mock the file)

grouma:
I'm in favor of hard coding the mapping. As you said, it is unlikely that specific imports are added regularly. If that situation changes we can easily adjust the solution. I'd rather reduce the complexity for now.

_test/test/common/message_io.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
jakemac53: done

_test/test/config_specific_import_test.dart
```diff
@@ -0,0 +1,22 @@
...
+import 'common/message.dart'
+    // ignore: uri_does_not_exist
+    if (dart.library.io) 'common/message_io.dart'
+    // ignore: uri_does_not_exist
+    if (dart.library.html) 'common/message_html.dart';
+
+main() {
+  test('Message matches expected', () {
+    if (1.0 is int) {
```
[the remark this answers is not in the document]
jakemac53: Ya, its really weird how the `dart.library.*` things aren't real constants just some special identifier thing that compilers know about?

build_modules/CHANGELOG.md
jakemac53: done

jakemac53: > though it's maybe not documented super well?

or at all ;)

build_modules/build.yaml
jakemac53: Yep, would clean this up a bunch

build_modules/lib/src/meta_module_builder.dart
jakemac53: done

build_modules/lib/src/module_library.dart
jakemac53: I don't think it has any value (you can't use things from them anyways) - tracking it would have meant tracking sdk deps per platform which complicates things unnecessarily.

build_modules/lib/src/module_library.dart
jakemac53: This is basically just checking if the current platform explicitly supports the core library that is being referenced in the conditional import. It might either omit the library entirely or explicitly have it marked as unsupported.

build_modules/lib/src/platform.dart
jakemac53: done

natebosch:
@vsmenon - can we expect stability in inclusion/exclusion of `dart.library.*` for each platform throughout the lifetime of Dart 2.x?

I'm OK with hardcoding this if we are reasonably confident that it's not going to allow for incompatibilities against newer SDKs before there is a breaking version bump. I was more nervous about this before stable.

Other than old versions existing which are possible to pub solve against - do we think there are any reasons this isn't an easily reversible choice? We can hardcode for now and add back libraries.json parsing in the future if we decide it's warranted right?

jakemac53:
> We can hardcode for now and add back libraries.json parsing in the future if we decide it's warranted right?

Right, it should be a transparent change.

vsmenon:
> can we expect stability in inclusion/exclusion of dart.library.* for each platform throughout the lifetime of Dart 2.x?

@natebosch Yes, I think so.  We're not making breaking changes in 2.x here - and removing would clearly be breaking.  We *might* consider new libraries / platform.

We're hardcoding what's true and false in DDC today.

build_modules/lib/src/module_library.dart
natebosch: > you can't use things from them anyways

I don't understand, this seems valid:

```dart
import 'fake_io.dart'
  if (dart.library.io) 'dart:io';
```

jakemac53: Hmm I guess in that sense you could use it, I hadn't considered the fake io case but I don't think that is a common use case anyways?

My thoughts regarding this were that the different libraries don't have any overlapping types, so directly doing this would not have any value:

```
import 'default.dart'
  if (dart.library.io) 'dart:io'
  if (dart.library.html) 'dart:html';
```

natebosch: yeah I think the fake_io, fake_html thing is going to be a real use case...

I agree that there isn't much value in tracking it, and we can probably assume that they are "correct" in that you won't have something like `if (dart.library.io) 'dart:html'` - and let the failure be late if you do.

Is it OK to ignore them instead of throw here?

jakemac53: I don't actually think the fake_io case would be common though - in what scenario would that actually be a good pattern?

Primarily this feature is meant to be used by libraries I think, where they are going to want some real implementation and not just fall back on a fake. I don't think it is particularly cumbersome to simply force people to come up with a more generic api and have their actual platform specific imports hidden behind a conditional import.

In theory you could use a fake within a test but I don't understand the value that the feature provides there since you are typically only running on one platform or always running with a fake/mock.

jakemac53:
Ok, refactored the platform support handling:

- hardcoded now instead of read from libraries.json
- renamed Platform to DartPlatform to not conflict with dart:io's Platform
- most apis now take a DartPlatform instead of a string
- only specific top level constants are exposed for platforms, constructor is private

jakemac53:
Looks like the build_script_generate test is getting flaky (timeouts) - submitting with red appveyor

Issue #1792, opened 2018-08-27
ModuleBuilder with coarse strategy: primary input not in inputs


nickclmb:
Outputs of DDC ModuleBuilder have invalid `previousInputsDigest` state.

Module builder with coarse strategy doesn't read its primary input (.dart file) and only reads meta module asset, so when build runner finish build and executes `_setOutputsState`, it removes `primaryInput` asset from `inputs` collection, and calculates MD5 digest for `previousInputsDigest` without primaryInput asset.

But at next build_runner start, AssetGraph deserializer forces adding of `primaryInput` to `inputs` collection:
```dart
for (var output in node.outputs) {
  var inputsNode = graph.get(output) as NodeWithInputs;
  assert(inputsNode != null, 'Asset Graph is missing $output');
  inputsNode.inputs ??= SplayTreeSet<AssetId>();
  inputsNode.inputs.add(node.id);
}
```
As a result, digest of ModuleBuilder nodes is always different on first rebuild.

I assume the fix is to add buildStep.canRead(buildStep.inputId) to module builder with coarse strategy.

nickclmb:
Or build_runner should never remove primaryInput from inputs in `_setOutputsState` function

jakemac53:
This seems like a bug in build_runner - we shouldn't be adding the primary input I think (if you don't read it you shouldn't depend on its bytes).

nickclmb:
Ok, so situation looks like this.
For example for asset   package|path.dart  asset graph will contain:
1 **SourceAssetNode**   id: package|path.dart  
outputs:  package|path.module

2. **GeneratedAssetNode**  id: package|path.module
primaryInput:   package|path.dart
inputs:      package|.meta_module.clean    ~~package|path.dart~~

so the package|path.dart at this moment in inputs of module node after asset graph deserialized, but after first rebuild it will be gone

Doesn't it bring some inconsistency here? I mean  package|path.module is output of some node, but that node is not in inputs?  If thats ok, i will prepare the PR for that issue.

jakemac53:
> Doesn't it bring some inconsistency here? I mean package|path.module is output of some node, but that node is not in inputs? If thats ok, i will prepare the PR for that issue.

The primary input really only determines the outputs, there is no requirement that you read it (although usually you would). As of now even the fine strategy no longer will actually read that file though - it only reads the meta_module asset, and thus the builder shouldn't need to rerun any time the dart file changes, only if the meta_module file changes.

[...]

PR #1794, opened 2018-08-28
Don't assume primary input is an actual input

jakemac53:
- `outputs` is no longer always a superset of `primaryOutputs` - it only contains nodes that actually read the file.
- `inputs` is now allowed to not contain the primary input

Fixes https://github.com/dart-lang/build/issues/1792

### DIFF
--- a/build_runner_core/lib/src/asset_graph/graph.dart
+++ b/build_runner_core/lib/src/asset_graph/graph.dart
@@ -178,6 +178,9 @@ class AssetGraph {
       if (inputsNode != null) {
         inputsNode.inputs.remove(id);
       }
+      if (inputsNode.state == NodeState.upToDate) {
+        inputsNode.state = NodeState.mayNeedUpdate;
+      }
     }
     if (node is NodeWithInputs) {
       for (var input in node.inputs) {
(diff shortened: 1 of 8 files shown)


build_runner_core/test/generate/build_test.dart
```diff
@@ -1131,6 +1131,35 @@ void main() {
           },
           writer: writer);
     });
+
+    test('no implicit dependency on primary input contents', () async {
+      var builders = [applyToRoot(SiblingCopyBuilder())];
```
natebosch: [nit] We moved away from implementing these behaviors as individual classes.

```dart
applyToRoot(TestBuilder(buildExtensions: appendExtension('.new'),
    build: (step, _) async {
      var sibling = ...
    }))
```

jakemac53: I can't use that because it uses `canRead` always and you can't override it

natebosch: ah... I'm sure I had a good reason for that... 😬

jakemac53:
Hmm so this unfortunately is triggering some of our regression tests around primary inputs that change from synthetic nodes into real nodes, working on a fix but its a bit involved :sob:

build_runner_core/lib/src/asset_graph/graph.dart
```diff
@@ -292,20 +295,24 @@ class AssetGraph {
...
+
     void invalidateNodeAndDeps(AssetId id) {
       var node = get(id);
       if (node == null) return;
       if (!invalidatedIds.add(id)) return;
 
-      if (node is NodeWithInputs && node.state == NodeState.upToDate) {
-        node.state = NodeState.mayNeedUpdate;
+      if (node is NodeWithInputs) {
+        if (node.state == NodeState.upToDate) {
```
natebosch: how is this different?

jakemac53: Oh I originally had another condition in here but I removed it, reverting it back

nickclmb:
I assume that removing line 59 from build_runner_core/lib/src/asset_graph/serialization.dart would be enough for that?
```
55      for (var output in node.outputs) {
56        var inputsNode = graph.get(output) as NodeWithInputs;
57        assert(inputsNode != null, 'Asset Graph is missing $output');
58        inputsNode.inputs ??= SplayTreeSet<AssetId>();
59        inputsNode.inputs.add(node.id);
60      }
```

PR #1796, opened 2018-08-28
misc(build_config): update json_serializable to latest and regenerate

kevmoo:
Removes fields that should not be allowed - due to bug fix in v1.2.1
Also bump dependency on build_runner

### DIFF
--- a/build_config/lib/src/build_target.g.dart
+++ b/build_config/lib/src/build_target.g.dart
@@ -8,13 +8,8 @@ part of 'build_target.dart';
 
 BuildTarget _$BuildTargetFromJson(Map json) {
   return $checkedNew('BuildTarget', json, () {
-    $checkKeys(json, allowedKeys: const [
-      'builders',
-      'dependencies',
-      'sources',
-      'key',
-      'package'
-    ]);
+    $checkKeys(json,
+        allowedKeys: const ['builders', 'dependencies', 'sources']);
     var val = BuildTarget(
         sources: $checkedConvert(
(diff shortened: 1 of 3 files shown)


build_config/lib/src/build_target.g.dart
```diff
@@ -8,13 +8,8 @@ part of 'build_target.dart';
 
 BuildTarget _$BuildTargetFromJson(Map json) {
   return $checkedNew('BuildTarget', json, () {
-    $checkKeys(json, allowedKeys: const [
-      'builders',
-      'dependencies',
-      'sources',
-      'key',
```
kevmoo: `key` and `package` are properties on `BuildTarget`, but they are not set-able via the ctor – so they shouldn't be allowed.

Same below...

kevmoo:
> We don't really do the "misc():" stuff...

I was trying to be a bit more verbose w/ my commit messages – as someone once asked me 😄

natebosch:
> I was trying to be a bit more verbose w/ my commit messages – as someone once asked me

You're not adding information, you're adding noise. "misc" doesn't say anything other than "not a bug fix" so it isn't useful.

"build_config" doesn't tell us much since the included files are in part of the commit metadata. We probably do include the package name sometimes when there isn't much else to say, but I'd prefer to avoid it when it's not necessary.

"Use latest json_serializable" or "Use latest json_serializable from build_config" would both be fine.

kevmoo:
ack

PR #1797, opened 2018-08-28
Use unawaited from package:pedantic

natebosch:
Models the new best practices for ignoring unawaited futures lints. This
is largely to avoid encouraging use of `/ignore`, but also has benefits
like encolosing the full unawaited expression in parens which is more
clear than a comment on the line above.

### DIFF
--- a/build_runner/test/generate/serve_test.dart
+++ b/build_runner/test/generate/serve_test.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'package:async/async.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as path;
+import 'package:pedantic/pedantic.dart';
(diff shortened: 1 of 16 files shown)


build_runner/test/generate/serve_test.dart
```diff
@@ -61,36 +62,36 @@ a:file://fake/pkg/path
       await wait(100);
 
       var request = Request('GET', Uri.parse('http://localhost:8000/a.txt'));
-      // ignore: unawaited_futures
-      (webHandler(request) as Future<Response>).then((Response response) {
+      unawaited((webHandler(request) as Future<Response>)
+          .then(expectAsync1((Response response) {
```
natebosch: I also added some expectAsync since they looked missing - we want to ensure these expects actually run.

PR #1803, opened 2018-08-30
Don't try to substring from identical string

natebosch:


build_runner/lib/src/watcher/graph_watcher.dart
```diff
@@ -100,7 +100,8 @@ class PackageGraphWatcher {
   List<String> _nestedPaths(PackageNode rootNode) {
     return _graph.allPackages.values
         .where((node) {
-          return rootNode != node && node.path.startsWith(rootNode.path);
+          return node.path.length > rootNode.path.length &&
```
kevmoo: Is this safe? Could you have node.path of `foo/bar/baz` and rootNode.path of `foo/ba` ?

node is not really "in" root in that case, right?

natebosch: :shrug:

It's as least as safe as it was before?

@jakemac53 and I had discussed using `p.isWithin` but weren't sure if it was worth it.

PR #1805, opened 2018-08-31
Add timings package

samogot:


timings/test/timings_test.dart
```diff
@@ -0,0 +1,191 @@
...
+          time = time.add(const Duration(seconds: 4)); // Tracked async, delayed
+        });
+      }).then((_) {
+        time = time.add(const Duration(seconds: 8)); // Async, after tracking
+      });
+      time = time.add(const Duration(seconds: 16)); // Sync, inside tracking
+      await Future.wait([
+        future,
+        // Async, inside tracking
+        Future.microtask(() => time = time.add(const Duration(seconds: 32)))
```
jakemac53: This comment is a bit misleading - it gets interleaved before the main future returns but it is part of a different zone and is not tracked as part of the `innerDuration`.

Also, this is relying on some internals of the event loop - instead I would make a completer which this completes, and wait on that future in the main scopeLog call. That way you can make sure it actually waits for this to run before completing.

timings/lib/timings.dart
```diff
@@ -0,0 +1,225 @@
...
+
+/// Implementation of a real [AsyncTimeTracker].
+///
+/// Use [AsyncTimeTracker] factory to get an instance.
+class _AsyncTimeTrackerImpl extends AsyncTimings implements AsyncTimeTracker {
+  final internalFutures = <Future>[];
+
+  T _trackSyncSlice<T>(ZoneDelegate parent, Zone zone, T Function() action) {
+    // Don't track time in nested runs twice
+    if (slices.isEmpty || slices.last.stopTime != null) {
```
jakemac53: I think this will need to get a little bit more complicated, we need to figure out exactly how to handle the nested zones.

I think what we really want is to _not_ include the time from the nested zone at all in the parent zone. Or, possibly we could even represent that with a special type of TimeSlice?

The primary reasoning for this is that `build_runner` actually does a lot of lazy builds. Basically, inside your Builder code if you read an asset which is generated, that might actually cause it to be built, and we want that to be tracked separately from your Builder.

I don't know if that would be the common case though, or if we might need to provide multiple options?

timings/test/timings_test.dart
```diff
@@ -0,0 +1,191 @@
...
+      ]);
+      expect(asyncTracker.startTime, startTime);
+      expect(asyncTracker.stopTime.millisecondsSinceEpoch,
+          lessThan(time.millisecondsSinceEpoch));
+      expect(asyncTracker.duration, const Duration(seconds: 55));
+      expect(asyncTracker.innerDuration, const Duration(seconds: 7));
+      expect(asyncTracker.slices.length, greaterThan(1));
+    });
+  });
+}
```
jakemac53: Related to my comment about nested performance zones earlier, we should add some tests for the nested zones to confirm their behavior as well.

timings/README.md
```diff
@@ -0,0 +1,5 @@
+# [![Build Status](https://travis-ci.org/dart-lang/build.svg?branch=master)](https://travis-ci.org/dart-lang/build)
+
+## Timings
+
+Timings is a simple package for tracking perfomanse of both async and sync actions
```
jakemac53: nit: `performance`

timings/lib/src/clock.dart
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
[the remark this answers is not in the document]
samogot: this file wasn't changed. Just copied

timings/pubspec.yaml
```diff
@@ -0,0 +1,14 @@
+name: timings
```
kevmoo: Should we call this build_timings? Why is this in the build multi-repo?

timings/lib/src/timings.dart
```diff
@@ -0,0 +1,386 @@
...
+  TimeSlice._();
+
+  factory TimeSlice(DateTime startTime, DateTime stopTime) =>
+      _TimeSliceImpl(startTime, stopTime);
+
+  @override
+  String toString() => '($startTime + $duration)';
+}
+
+class _TimeSliceImpl extends TimeSlice {
```
natebosch: Can this be merged into `TimeSlice`? Why do we need a separate `Impl`?

timings/lib/src/timings.dart
```diff
@@ -0,0 +1,386 @@
...
+  @override
+  final DateTime startTime;
+
+  @override
+  final DateTime stopTime;
+
+  _TimeSliceImpl(this.startTime, this.stopTime) : super._();
+}
+
+/// The timings of an async operation, consist of several sync [slices] and
```
natebosch: do we make any guarantees about how `startTime` and `endTime` relate to the times in `slices`? Should we?

timings/lib/src/timings.dart
```diff
@@ -0,0 +1,386 @@
...
+
+class _TimeSliceGroupImpl extends TimeSliceGroup {
+  @override
+  final slices = [];
+
+  _TimeSliceGroupImpl() : super._();
+}
+
+abstract class TimeTracker implements TimeSlice {
+  /// Indicates if tracker is tracking right now.
```
natebosch: [nit] Should be a noun phrase. `/// Whether tracking is active`. Or possibly drop altogether? Or maybe add more details?

```
/// Whether tracking is active.
///
/// Tracking is only active after `isStarted` and before `isFinished`.
```

timings/lib/src/timings.dart
```diff
@@ -0,0 +1,386 @@
...
+}
+
+class _TimeSliceGroupImpl extends TimeSliceGroup {
+  @override
+  final slices = [];
+
+  _TimeSliceGroupImpl() : super._();
+}
+
+abstract class TimeTracker implements TimeSlice {
```
natebosch: It's strange to me that a Tracker would be a Slice...

Overall this library has a lot more `extends` and `implements` than I would hope for. Why is a group of slices also a single slice? Why do we need all the `*Impl` classes?

kevmoo:
FYI: https://github.com/dart-lang/timing

Maybe get the code fully reviewed here then landed it in the other repo?

PR #1806, opened 2018-09-04
change default dart2js args to the new -O2

jakemac53:
This arg is intended to contain all "safe" optimizations, which today includes `--minify` and `--lax-runtime-type-to-string`.

### DIFF
--- a/build_web_compilers/pubspec.yaml
+++ b/build_web_compilers/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 0.4.3
+version: 0.4.4-dev
 description: Builder implementations wrapping Dart compilers.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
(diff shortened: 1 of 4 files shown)


jakemac53:
Making it a special option would require checking for a conflicting arg in dart2js_args though... but it would make it harder to accidentally override it.

natebosch:
Yeah, we might even be able to get away with not checking for a conflict.

I just think this one is special enough that's it's worth supporting directly. With config set alias it'll also make it possible to do with a command line flag to the build system

build_web_compilers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 0.4.3
+version: 0.4.4-dev
```
kevmoo: Don't we need to update the minimum SDK constraint to require `2.1.0-dev..??` for when this was added? It's not in dart2 stable

jakemac53: Good catch!

jakemac53:
Going to just close this for now, and wait on a 2.1.0 release as I would prefer to not have the minimum bound be a dev release.

kevmoo:
@jakemac53 cleanup this branch?

PR #1809, opened 2018-09-04
Add pkg:pedantic lints to analysis_options

kevmoo:


kevmoo:
> It'd be nice if package:pedantic published it's analysis_options.yaml so we could import it...

ack

PR #1812, opened 2018-09-05
Use timing package for performance tracking

samogot:
Should not be merged before https://github.com/dart-lang/timing/pull/1 and releasing package

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Added `--hot-reload` cli option and appropriate functionality.
   See [hot-module-reloading](../docs/hot_module_reloading.md) for more info.
+- Use timings package for performance tracking.
 
 ## 0.10.1+1
 
--- a/build_runner/lib/src/server/server.dart
+++ b/build_runner/lib/src/server/server.dart
@@ -335,11 +335,11 @@ String _renderPerformance(BuildPerformance performance, bool hideSkipped) {
     var rows = StringBuffer();
     for (var action in performance.actions) {
       if (hideSkipped &&
-          !action.phases.any((phase) => phase.label == 'Build')) {
+          !action.stages.any((phase) => phase.label == 'Build')) {
         continue;
       }
       var actionKey = '${action.builderKey}:${action.primaryInput}';
-      for (var phase in action.phases) {
+      for (var phase in action.stages) {
         var start = phase.startTime.millisecondsSinceEpoch -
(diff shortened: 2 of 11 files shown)


build_runner/lib/src/server/server.dart
```diff
@@ -335,11 +335,11 @@ String _renderPerformance(BuildPerformance performance, bool hideSkipped) {
     var rows = StringBuffer();
     for (var action in performance.actions) {
       if (hideSkipped &&
-          !action.phases.any((phase) => phase.label == 'Build')) {
+          !action.stages.any((phase) => phase.label == 'Build')) {
```
jakemac53: s/phase/stage for the parameter name too

build_runner/pubspec.yaml
```diff
@@ -55,3 +55,10 @@ dev_dependencies:
   test_process: ^1.0.0
```
jakemac53: we also need to change the minimum version for build_runner_core to be the new version with this change

build_runner_core/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.3.2
```
jakemac53: This is a breaking change, so we need to do 0.4.0.

build_runner_core/lib/src/generate/performance_tracker.dart
```diff
@@ -10,31 +10,19 @@ import 'dart:async';
 import 'package:build/build.dart';
 import 'package:json_annotation/json_annotation.dart';
 import 'package:meta/meta.dart';
+import 'package:timing/src/timing.dart';
```
jakemac53: We should avoid this import, I think the other changes we talked about should allow us to expose publicly everything we need though?

samogot:
@jakemac53 ptal :)

build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 0.10.3
```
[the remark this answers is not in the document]
samogot: Technically `performance` in `BuildResult` is marked as `@experimental`, so may be not considered as breaking change. But OK)

jakemac53:
I didn't realize we had marked it as experimental... in that case we could consider doing a non-breaking release, @natebosch wdyt?

natebosch:
SGTM, `@experimental` is definitely enough.

samogot:
So, is it ready to be approved then? @jakemac53  @natebosch

PR #1813, opened 2018-09-07
remove all remaining uses of cli_util

jakemac53:


jakemac53:
> Not sure it's worth a changelog entry...

I don't want empty versions in the changelog, sure its not very important but :man_shrugging:

build_resolvers/lib/src/resolver.dart
```diff
@@ -427,7 +427,9 @@ class AnalyzerResolvers implements Resolvers {
     _initAnalysisEngine();
     var resourceProvider = PhysicalResourceProvider.INSTANCE;
     var sdk = FolderBasedDartSdk(
-        resourceProvider, resourceProvider.getFolder(cli_util.getSdkPath()))
+        resourceProvider,
+        resourceProvider.getFolder(native_path
+            .dirname(native_path.dirname(Platform.resolvedExecutable))))
```
natebosch: oops, this one _is_ used internally and is broken by this... we have an internal only patch on `cli_util` that make it work in the build system...

PR #1824, opened 2018-09-12
Visualize performance slices, add sorting and filtering

samogot:


build_runner/lib/src/server/server.dart
```diff
@@ -428,11 +605,20 @@ String _renderPerformance(BuildPerformance performance, bool hideSkipped) {
...
+        margin: 0.5em;
+      }
       </style>
     </head>
     <body>
       <div class="controls-header">
         <p>$showSkippedLink</p>
+        <p>$showSlicesLink</p>
+        <p><b>Sort by</b>: </p>
+        <p>Start Time (<a href="$sortStartTimeAsc">Asc</a>, <a href="$sortStartTimeDesc">Desc</a>)</p>
```
jakemac53: I think a dropdown might be nicer, its a little bit more code to handle the selection but makes it more obvious these are mutually exclusive and uses less space in the UI

jakemac53:
I also have a more general comment, I am not sure if it is super clear what "async duration" means. I think we should come up with a better name, maybe "cpu duration" or something like that?

natebosch:
> I also have a more general comment, I am not sure if it is super clear what "async duration" means. I think we should come up with a better name, maybe "cpu duration" or something like that?

That's also misleading. I think this would capture things like `Process.runSync` and `someFile.readAsStringSync` which aren't necessarily time spent on the CPU.

build_runner/lib/src/server/server.dart
```diff
@@ -119,11 +132,31 @@ class ServeHandler implements BuildState {
 
   shelf.Response _performanceHandler(shelf.Request request) {
     var hideSkipped = false;
+    var detailedSlices = false;
+    var slicesResolution = 5;
+    var sortOrder = PerfSortOrder.startTimeAsc;
+    String filter;
```
natebosch: [nit] does `request.url.queryParameters['filter']` return null if `containsKey('filter')` returns `false`? If so, should we initialize this here?

build_runner/lib/src/server/server.dart
```diff
@@ -330,37 +363,177 @@ class AssetHandler {
   }
 }
 
-String _renderPerformance(BuildPerformance performance, bool hideSkipped) {
+String _getPerformancePath(bool hideSkipped, bool detailedSlices,
```
natebosch: [nit] avoid prefixing method names with `get` or `_get`.

build_runner/lib/src/server/server.dart
```diff
@@ -330,37 +363,177 @@ class AssetHandler {
...
+
+      if (min > start) min = start;
+      if (max < end) max = end;
+
+      rows.writeln(
+          '          ["$actionKey", "${stage.label}", "$tooltip", $start, $end],');
+      ++count;
+    }
+
+    RegExp filterRE;
```
natebosch: [optional] since we care about the _value_ not the _side effects_ I like to use the ternary conditional when possible.

```dart
final filterRegex = filter != null ? RegExp(filter) : null;
```

build_runner/lib/src/server/server.dart
```diff
@@ -119,11 +132,31 @@ class ServeHandler implements BuildState {
 
   shelf.Response _performanceHandler(shelf.Request request) {
     var hideSkipped = false;
+    var detailedSlices = false;
+    var slicesResolution = 5;
```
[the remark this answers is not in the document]
samogot: It is confusing but would be hard to track real duration.

build_runner/lib/src/server/server.dart
```diff
@@ -330,37 +363,177 @@ class AssetHandler {
...
+          writeRow(action, stage, stage);
+          continue;
+        }
+        var slices = stage.slices.fold<List<TimeSlice>>([], (list, slice) {
+          if (list.isNotEmpty &&
+              slice.startTime.difference(list.last.stopTime) < resolution) {
+            // concat with previous if gap less than resolution
+            list.last = TimeSlice(list.last.startTime, slice.stopTime);
+          } else {
+            if (list.length > 1 && list.last.duration < resolution) {
```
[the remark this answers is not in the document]
samogot: It first collapses near slices and than deletes ones that still small after collapsing. It fill never delete first or last slice.

Basically now I can see deleting doesn't give much improvements, so we can remove this code

build_runner/lib/src/server/server.dart
samogot: `info: Don't explicitly initialize variables to null`

natebosch: what I mean is `var filter = request.url.queryParameters['filter'];`.

samogot: Ah... sure :)

samogot:
@jakemac53 @natebosch ptal :)

samogot:
We also may use `User time`, but as misleading as `CPU time`, except that we don't have exact definition what "user" means in our context :)

Issue #1838, opened 2018-09-20
DDC error: "NoSuchMethodError: The getter 'inputs' was called on null."


giorgiofran:
Thank you for taking the time to file an issue!

In order to route, prioritize, and act on this as soon as possible please include:

* [ ] Dart SDK Version 2.0.0
* [ ] `build_runner "0.10.2"`
* [ ]  Windows
* [ ] Chrome

I'm not sure what could have cause this error, but here is what I did;
I tried to move the template of a component in an html file. (forgetting an \$event that in the next build gave an error).
Then I created a component derived from the one above pointing to the same html generated before.
This done I got the below error. Also if I tried to clean (pub run build_runner clean) the cash.
I tried to rollback my changes but there was no solution, the same error was there.
Finally I deleted the ".dart_tool/build/098d3ee73e6cc294616d3a2e2c3c81ad/asset_graph.json"
and then the build completed regularly.
-----

NoSuchMethodError: The getter 'inputs' was called on null.
Receiver: null
Tried calling: inputs
dart:core                                                                                                                                  Object.noSuchMethod
C:\Users\Giorgio\AppData\Roaming\Pub\Cache\hosted\pub.dartlang.org\build_runner_core-0.3.1+5\lib\src\asset_graph\serialization.dart 58:20  _AssetGraphDeserializer.deserialize
package:build_runner_core/src/asset_graph/graph.dart 42:48                                                                                 new AssetGraph.deserialize
package:build_runner/src/entrypoint/clean.dart 45:24                                                                                       CleanCommand.run.<fn>
package:build_runner_core/src/logging/logging.dart 25:30                                                                                   logTimed
Async
package:build_runner/src/entrypoint/clean.dart 38:11                                                                                       CleanCommand.run
package:args/command_runner.dart 194:27                                                                                                    CommandRunner.runCommand
package:args/command_runner.dart 109:29                                                                                                    CommandRunner.run.<fn>
dart:async                                                                                                                                 new Future.sync
package:args/command_runner.dart 109:11                                                                                                    CommandRunner.run
package:build_runner/src/entrypoint/run.dart 22:31                                                                                         run
.dart_tool\build\entrypoint\build.dart 132:22                                                                                              main

jakemac53:
Duplicate of https://github.com/dart-lang/build/issues/1804, unfortunately we don't have a repro case still for this :(

PR #1839, opened 2018-09-20
Catch asset graph deserialize failure during clean

natebosch:
See #1838

If we fail to deserialize an asset graph during a build the next logical
step for the user is to try the `clean` command, which will also fail
and not fix the issue.

### DIFF
--- a/build_runner/lib/src/entrypoint/clean.dart
+++ b/build_runner/lib/src/entrypoint/clean.dart
@@ -39,7 +39,7 @@ class CleanCommand extends Command<int> {
       var assetGraphFile = File(assetGraphPath);
       if (!assetGraphFile.existsSync()) {
         logger.warning('No asset graph found. '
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/entrypoint/clean.dart
```diff
@@ -38,23 +38,29 @@ class CleanCommand extends Command<int> {
...
+        logger.warning('No asset graph found. '
+            'Skipping cleanup of generated file in source directories.');
+        return;
+      }
+      AssetGraph assetGraph;
+      try {
+        assetGraph = AssetGraph.deserialize(await assetGraphFile.readAsBytes());
+      } catch (_) {
+        logger.warning('Failed to deserialize AssetGraph. '
+            'Skipping cleanup of generated file in source directories.');
```
jakemac53: s/file/files

Issue #535, opened 2017-10-20
Change Builder.build to FutureOr<void>


matanlurey:
Prepares for `FutureOr<void>`, and doesn't require adding `async` to your method body.

... though maybe all of this is not too important if post-Barback we can make methods sync?

jakemac53:
Not sure what you meant re: post-barback, but in general I think this SGTM.

We already made a similar change in source_gen because somebody requested it https://github.com/dart-lang/source_gen/issues/250.

matanlurey:
Seems like this isn't really worth it since `Future<void>` is rolling out.

We can wait until then.

jakemac53:
Ping on this, @natebosch @matanlurey @grouma  should we implement this for 1.0.0 or should we close the issue?

natebosch:
Yes, I'd like to get this in for 1.0.0

matanlurey:
Not: should be void now not Null :)

PR #1847, opened 2018-09-24
change Builder.build to return FutureOr<void>

jakemac53:
Closes https://github.com/dart-lang/build/issues/535

Waiting to merge until after https://github.com/dart-lang/build/pull/1846

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,4 +1,11 @@
-## 0.12.8-dev
+## 1.0.0-dev
+
+### Breaking Changes
+
+- Changed the return type of `Builder.build` from `Future<dynamic>` to
+  `FutureOr<void>`.
+
+## 0.12.8
 
 - Added the `T trackStage<T>(String label, T Function() action);` method to `BuildStep`.
   Actions that are tracked in this way will show up in the performance timeline at `/$perf`.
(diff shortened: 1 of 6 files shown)


build/CHANGELOG.md
```diff
@@ -1,4 +1,11 @@
-## 0.12.8-dev
+## 1.0.0-dev
+
+### Breaking Changes
+
+- Changed the return type of `Builder.build` from `Future<dynamic>` to
```
natebosch: I think this is mostly breaking for build system implementations right? Might want to call that out that most Builder authors should't be impacted.

build/lib/src/builder/multiplexing_builder.dart
```diff
@@ -17,10 +17,17 @@ class MultiplexingBuilder implements Builder {
...
 
   @override
-  Future build(BuildStep buildStep) => Future.wait(_builders
-      .where((builder) => builder.buildExtensions.keys
-          .any((extension) => buildStep.inputId.path.endsWith(extension)))
-      .map((builder) => builder.build(buildStep)));
+  Future<List<dynamic>> build(BuildStep buildStep) {
+    var futures = _builders
+        .where((builder) => builder.buildExtensions.keys
+            .any((extension) => buildStep.inputId.path.endsWith(extension)))
```
natebosch: while changing this file anyway - looks like this could be a tear off?

Issue #1830, opened 2018-09-18
Implement toJson/fromJson on pkg:build - AssetId


kevmoo:
Even with pending cleanup, we have 2 seperate packages that have to write custom to/from logic. Better to just add these to AssetId itself

https://github.com/dart-lang/build/blob/08232558d2c583985a74a9cab3060e0187241520/build_modules/lib/src/modules.dart#L179-L187

PR #1870, opened 2018-10-01
build_modules: use new JsonConvert logic for AssetId, DartPlatform

kevmoo:
Longterm, we should consider #1830


analysis_options.yaml
```diff
@@ -77,7 +77,8 @@ linter:
     - unnecessary_brace_in_string_interps
     #- unnecessary_const
     - unnecessary_getters_setters
-    - unnecessary_lambdas
+    # pkg:json_serializable needs a big update to eliminate these
```
natebosch: can restore this now?

kevmoo: Yes! oversight!

PR #1888, opened 2018-10-25
Remove "dev channel or fresher" requirement

kwalrath:


kwalrath:
The Travis build succeeded long ago, and the Appveyor one doesn't seem to be starting, so I'm going to merge this since it should be a safe change.

Issue #1745, opened 2018-08-10
Re-run build script when changed (instead of killing)


matanlurey:
The use case here is debugging a builder, I'd like to be able to do:

```bash
$ pub run build_runner watch -v
```

... and change my build script/compiler, instead of re-running after any change.

(If this isn't a good default behavior, a flag is fine with me too)

Issue #1496, opened 2018-05-29
Snapshot the generated build.dart script


natebosch:
With Dart 2 things are going to get very slow to start if we're parsing from source on every run. We should start snapshotting the Dart script.

- [ ] Figure out what criteria to use to invalidate it (we unconditionally invalidate today)
- [ ] Add support for passing down the "apparent" script name to use for the asset graph path. Today we compute it based on the file path of the entrypoint, but that will very depending on whether we're running from source (if we're producing an app-jit snapshot) or from the snapshot.

jakemac53:
We could solve this instead with a longer lived daemon process? It would use existing logic to kill itself if any files it imports are changed, but for most users that shouldn't ever happen (outside of pub get/upgrade).

jakemac53:
Closing in favor of https://github.com/dart-lang/build/issues/1656

jakemac53:
So I did a bit of testing here, and the snapshot does can actually check itself for changes appropriately (based on the original sources it came from).

Given that, we could add a simple hidden flag like `--exit-if-changed`, or make it somehow recognize it is a snapshot and automatically exit with a special exit code in this scenario. Then, we would re-snapshot from the wrapper script and regenerate the snapshot.

jakemac53:
This could also be tied in with https://github.com/dart-lang/build/issues/1745

jakemac53:
Also related https://github.com/dart-lang/build/issues/1201

Issue #1201, opened 2018-03-23
Exit non-zero (75?) when terminating due to config change


kevmoo:
75 seems meant for this https://github.com/dart-lang/io/blob/080b68a704133d6462215ac89da5fccb8655b335/lib/src/exit_code.dart#L61

This will make hooking restart in webdev (https://github.com/dart-lang/webdev/issues/18) much easier, too!

natebosch:
Is this a blocker to dropping pub build?

kevmoo:
Sorry. Strictly not. Makes it easier to implement the restart thing in webdev...

PR #1892, opened 2018-10-30
Run build_runner scripts from snapshot.

jakemac53:
- Adds a new exception type (BuildScriptUpdated) which is thrown when (shocker) the build script is updated. When this exception is caught at the top level, we return an exit code 75.
- The wrapper script now generates build script snapshots and runs those in the isolate. When it sees the exit code 75 it will clean up the asset graph and old snapshot, and re-generate a new snapshot.

Closes https://github.com/dart-lang/build/issues/1496
Closes https://github.com/dart-lang/build/issues/1745
Closes https://github.com/dart-lang/build/issues/1201

cc @kevmoo as this will affect webdev as well, they would ideally do the same logic

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -60,17 +60,18 @@ Future<Null> main(List<String> args) async {
   if (commandName != _generateCommand) {
     logListener = Logger.root.onRecord.listen(stdIOLogListener());
   }
-  var buildScript = await generateBuildScript();
-  var scriptFile = File(scriptLocation)..createSync(recursive: true);
-  scriptFile.writeAsStringSync(buildScript);
-  if (commandName == _generateCommand) {
-    print(p.absolute(scriptLocation));
-    return;
-  }
 
   var shouldRerun = true;
   while (shouldRerun) {
     shouldRerun = false;
+
+    var buildScript = await generateBuildScript();
+    var scriptFile = File(scriptLocation)..createSync(recursive: true);
+    scriptFile.writeAsStringSync(buildScript);
+    if (commandName == _generateCommand) {
+      print(p.absolute(scriptLocation));
+      return;
+    }
 
     if (!await _createSnapshotIfMissing()) return;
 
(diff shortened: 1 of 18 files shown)


jakemac53:
I updated this to also handle changing of build.yaml files gracefully, it will regenerate the script and re-run without re-snapshotting. If there were any changes to the build script that will be handled on the next build, which will end up causing a `BuildScriptChangedException` and a re-run + re-snapshot.

build_runner/bin/build_runner.dart
```diff
@@ -56,54 +60,89 @@ Future<Null> main(List<String> args) async {
...
-    exitPort.sendPort.send(null);
-  }
-  StreamSubscription exitCodeListener;
-  exitCodeListener = messagePort.listen((isolateExitCode) {
-    if (isolateExitCode is! int) {
-      throw StateError(
-          'Bad response from isolate, expected an exit code but got '
-          '$isolateExitCode');
+  var shouldRerun = true;
+  while (shouldRerun) {
```
natebosch: what do you think about factoring this out into a method that returns the exit code from the isolate - then the loop can be `while (await generateAndRun() == ExitCode.tempFail.code);`

build_runner/lib/src/entrypoint/run.dart
```diff
@@ -18,19 +20,59 @@ import 'runner.dart';
...
-    // A message should have already been logged.
-    return ExitCode.config.code;
+  var resultCompleter = Completer<int>();
+
+  void safeComplete(int exitCode) {
+    if (resultCompleter.isCompleted) return;
+    resultCompleter.complete(exitCode);
+  }
+
+  unawaited(runZoned(() async {
```
natebosch: looks like the surrounding method doesn't need to be `async` and so wouldn't need the `unawaited`.

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -240,6 +240,10 @@ class WatchImpl implements BuildState {
             .hasBeenUpdated(mergedChanges.keys.toSet())) {
           _terminateCompleter.complete();
           _logger.severe('Terminating builds due to build script update');
+
+          /// Intentional unhandled async error here, which will be caught at the
```
natebosch: [nit] double slash comment

I'm a little uncertain about this trick... I wonder if it would be more explicit to have a global variable through which we can report the status?

PR #1897, opened 2018-11-05
handle sdk updates with retry logic

jakemac53:
- If we get an isolate spawn exception, we will always resnapshot and rerun once to handle sdk updates.
- Exposes a utility for generating and snapshotting the build script (under `src`, only intended for use by `webdev`).
- Adds `Bootstrap` logger to the list of known loggers to hide its name from output.

cc @kevmoo see the `generateAndRun` method in `lib/src/build_script_generate/bootstrap.dart`, which webdev can hopefully use. Let me know if it needs to change at all.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### Bug Fixes
 
 - Handle re-snapshotting the build script on SDK updates.
-- Add `Bootstrap` to the list of known loggers.
+- Suppress header for the `Bootstrap` logger.
 
 ## 1.1.0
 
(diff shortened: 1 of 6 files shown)


build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.1.1
+
+### Bug Fixes
+
+- Handle re-snapshotting the build script on SDK updates.
+- Add `Bootstrap` to the list of known loggers.
```
natebosch: "known loggers" is an implementation detail - I'd maybe  go with "Suppress header for the `Bootstrap` logger."

Issue #1886, opened 2018-10-18
Migrate to AnalysisDriver


natebosch:
- [x] `_bazel_codegen`
- [ ] `build_resolvers`

Migrated from https://github.com/dart-lang/bazel/issues/183

natebosch:
This will be breaking for Builders in a similar way that the bazel_codgen migration to summaries was breaking - certain APIs will no longer work when starting from a `LibraryElement`.

During the migration to summaries we had to move away from `computeNode()` (and, unknown at the time, away from `.unit`) on any library that was coming from a summary. These APIs were giving an escape hatch into the AST model which doesn't work backed by a summary.

During the migration to `AnalysisDriver`, or eventually `AnalysisSessions` we'll also have to move away from these APIs for the libraries that we have the source for. It's still possible to to get to an AST model but _not_ through the `LibraryElement`. Today the option is to start by accessing the AST model for a URI and then crawling that model. Eventually there may be new ways to jump to the AST model for a given Element that don't involve calling APIs on that element.

yury-yufimov:
@natebosch , am I right, that this is strategic / long-term plans, and these changes won't be available in DartSDK 2.1 / 2.2 ?

natebosch:
Moving forward with this is not tied to a particular SDK version.

We are blocked on moving forward until we can make sure we update known usages of the Resolver API so that they work with the `LibraryElement` instance that will be produced by the driver model.

jakemac53:
Note we also need to update to return empty sources for uris that we don't have a source for, see discussion at https://github.com/dart-lang/build/pull/1910

PR #1898, opened 2018-11-08
Add an option to use the analyis driver

natebosch:
Towards #1886

This is hidden behind a flag for now and should not be used, the
migration causes a number of breakinge changes in the behavior of the
`LibraryElement` instances that are returned by the resolver.

- Add a flag `--use-analysis-driver` defaulted to false.
- Add an implementation of the `Resolvers` class backed by an
  `AnalysisDriver`. There is a fair amount of duplication here but it
  will be removed when the migration is made as a hard flip.
- Switch a few argument lists to use trailing commas so future changes
  have single line diffs.

### DIFF
--- a/bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
+++ b/bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
@@ -1,6 +1,7 @@
-// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
 import 'dart:async';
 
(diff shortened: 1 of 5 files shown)


bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2017, the Dart project authors. Please see the AUTHORS file
```
grouma: [nit] 2018 here and elsewhere

natebosch:
> Are there tests to go along with this code path?

No, there is stuff that we expect to not work. We'll get some tests in place before do we a flip, but for now I just want this flag to be able to do some experiments with code and see how it fails.

bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
natebosch: Done

PR #1908, opened 2018-12-01
Prepare for breaking change to UriResolver.

stereotype441:
Analyzer version 0.34.0 contains a breaking change to UriResolver: the
addition of a `clearCache` method - see
https://dart-review.googlesource.com/c/sdk/+/85726/1/pkg/analyzer/lib/src/generated/source.dart.

Even though we are not ready to upgrade to analyzer 0.34.0 yet, we can
prepare for the change by adding implementations of the method.  It's
not necessary for the implementations to do anything, since our
UriResolver implementations don't cache any URI resolution.

### DIFF
--- a/bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
+++ b/bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
@@ -29,6 +29,9 @@ class BuildAssetUriResolver implements UriResolver {
   }
 
   @override
+  void clearCache() {}
+
+  @override
   Source resolveAbsolute(Uri uri, [Uri actualUri]) {
     final cachedId = _lookupAsset(uri);
     if (cachedId == null) return null;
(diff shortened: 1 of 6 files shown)


stereotype441:
@scheglov FYI

bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart
```diff
@@ -28,6 +28,9 @@ class BuildAssetUriResolver implements UriResolver {
     }
   }
 
+  @override
+  void clearCache() {}
```
kevmoo: Or extends? Then you can skip the blank override...

stereotype441: > Or extends? Then you can skip the blank override...

Done.  PTAL

jakemac53:
Please also update changelogs/pubspecs

PR #1910, opened 2018-12-03
return empty sources for non-dart uris in BuildAssetUriResolver

jakemac53:
This allows missing parts to appear in `LibraryElement#parts`.

cc @scheglov

### DIFF
--- a/bazel_codegen/lib/src/summaries/build_asset_uri_resolver.dart
+++ b/bazel_codegen/lib/src/summaries/build_asset_uri_resolver.dart
@@ -23,13 +23,17 @@ class BuildAssetUriResolver implements UriResolver {
     for (var asset in assets.where((asset) => asset.path.endsWith('.dart'))) {
       var uri = assetUri(asset);
       if (!_knownAssets.containsKey(uri)) {
-        _knownAssets[uri] = AssetSource(asset, await read(asset));
+        _knownAssets[uri] = AssetSource(asset, await read(asset), true);
       }
     }
   }
 
   @override
-  Source resolveAbsolute(Uri uri, [Uri actualUri]) => _knownAssets[uri];
+  Source resolveAbsolute(Uri uri, [Uri actualUri]) => _knownAssets.putIfAbsent(
+      uri,
+      () => uri.scheme == 'dart'
+          ? null
+          : AssetSource(AssetId.resolve(uri.toString()), null, false));
 
   @override
   Uri restoreAbsolute(Source source) {
@@ -40,11 +44,14 @@ class BuildAssetUriResolver implements UriResolver {
 class AssetSource implements Source {
   final AssetId _assetId;
   final String _content;
+  final bool _exists;
 
-  AssetSource(this._assetId, this._content);
+  AssetSource(this._assetId, this._content, this._exists);
 
   @override
-  TimestampedData<String> get contents => TimestampedData(0, _content);
+  TimestampedData<String> get contents => _exists
+      ? TimestampedData(0, _content)
+      : throw StateError('Attempted to read non-existent source');
 
   @override
   String get encoding => '$_assetId';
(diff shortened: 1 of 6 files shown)


bazel_codegen/test/build_asset_uri_resolver_test.dart
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  test('returns null for dart: uris', () {
+    expect(resolver.resolveAbsolute(Uri.parse('dart:async')), isNull);
+  });
+
+  test('returns an empty source for unknown assets', () {
+    var source = resolver.resolveAbsolute(Uri.parse('package:foo/bar.dart'));
+    expect(source.exists(), isFalse);
+    expect(() => source.contents, throwsA(const TypeMatcher<StateError>()));
```
scheglov: There is already `throwsStateError` in `package:test`.

bazel_codegen/lib/src/summaries/build_asset_uri_resolver.dart
```diff
@@ -23,13 +23,17 @@ class BuildAssetUriResolver extends UriResolver {
...
       if (!_knownAssets.containsKey(uri)) {
-        _knownAssets[uri] = AssetSource(asset, await read(asset));
+        _knownAssets[uri] = AssetSource(asset, await read(asset), true);
       }
     }
   }
 
   @override
-  Source resolveAbsolute(Uri uri, [Uri actualUri]) => _knownAssets[uri];
+  Source resolveAbsolute(Uri uri, [Uri actualUri]) => _knownAssets.putIfAbsent(
```
natebosch: I worry about this returning empty sources for URIs that should be handled by summaries.

This resolver comes [earlier in the list](https://github.com/dart-lang/build/blob/fc63e34c54e2194b00d7c8d524ee1bd71d13a0b3/bazel_codegen/lib/src/summaries/analysis_context.dart#L32-L35) so I'm confused how that isn't already a problem...

jakemac53: Good catch - I needed to move this to the end of the list.

jakemac53: The api for that looks a lot different... @scheglov do you know what we should be doing for analysis driver?

Do we need to use `resolver.newFile` with null content or will it handle this better by default?

jakemac53:
> Does the `AnalysisDriver` code path require the same behavior?

The api for that looks a lot different, @scheglov do we need to use `resolver.addFile` with null content or will analysis driver do something better by default here?

jakemac53:
@natebosch lets follow up with the analysis driver changes

scheglov:
I don't see the code about which you are asking the question, not sure how to navigate to it on GitHub. AnalysisDriver needs just URI to file path resolution (and back). The file does not have to exist, if a file does not exist, it is treated as empty.

jakemac53:
@scheglov The new uri resolver is here https://github.com/dart-lang/build/blob/master/bazel_codegen/lib/src/resolver/build_asset_uri_resolver.dart

It will end up returning null for `resolveAbsolute` today but I wasn't sure how to create an empty source since we don't control that any more.

scheglov:
If a URI cannot be resolved to a file path, it will be considered invalid and skipped.
Corresponding import / export / part directives will not be reflected in the element model of the library.

jakemac53:
> If a URI cannot be resolved to a file path, it will be considered invalid and skipped.
> Corresponding import / export / part directives will not be reflected in the element model of the library.

Ok - so back to my original question then, how do we create/return an empty source? We can create a uri for these files, it just isn't obvious how to get the source.

More specifically it looks like today we effectively do `resourceProvider.newFile(path, content)` with the known files, and then `resourceProvider.getFile(path).createSource(uri)`.

If we know we haven't ever called `newFile` before with a path can we still just call `getFile` with it and then subsequently call `createSource` on that object? Or will we get a null back from `getFile` or an exception?

scheglov:
`resourceProvider.getFile(path)` return a handle object.
The files does not have to exist, you will always get a not `null` object.
When you attempt to read it, and it still does not exist, then you will get an exception.
But otherwise you can work with it - ask its path,  create `Source`.

PR #1920, opened 2018-12-06
Include part files in build script update checks

### LINKED ISSUE #1912: build script change detection doesn't check parts
I noticed today that if you modify a part file that is imported by the build script we don't pick that up and invalidate the build.

natebosch:
Fixes #1912

Iterate over all declarations and pull out the source URIs for each.
This prevents the issue of only including the root of each library as an
asset used by the build script.

If a part file has no declarations it may be skipped, however an edit to
such a part file should have no impact on the build until an edit is
also made to another file to use the new declaration.

### DIFF
--- a/build_runner_core/lib/src/changes/build_script_updates.dart
+++ b/build_runner_core/lib/src/changes/build_script_updates.dart
@@ -52,8 +52,13 @@ class BuildScriptUpdates {
     return BuildScriptUpdates._(supportsIncrementalRebuilds, allSources);
   }
 
-  static Iterable<Uri> get _urisForThisScript =>
-      currentMirrorSystem().libraries.keys;
+  static Set<Uri> get _urisForThisScript => currentMirrorSystem()
+      .libraries
+      .values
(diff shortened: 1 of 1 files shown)


build_runner_core/lib/src/changes/build_script_updates.dart
```diff
@@ -52,8 +52,13 @@ class BuildScriptUpdates {
     return BuildScriptUpdates._(supportsIncrementalRebuilds, allSources);
   }
 
-  static Iterable<Uri> get _urisForThisScript =>
-      currentMirrorSystem().libraries.keys;
+  static Set<Uri> get _urisForThisScript => currentMirrorSystem()
```
jakemac53: any idea if this visibly impacts performance? I would guess not - even though it could potentially do a lot more work in practice there shouldn't be to many top level declarations in each library...

natebosch:
Filed https://github.com/dart-lang/sdk/issues/35396 to track

PR #1928, opened 2018-12-11
try using Platform.executable instead of Platform.resolvedExecutable

jakemac53:
Will update with changelog/pubspec updates if travis is happy - internally we need to use Platform.executable so hopefully this can let us eliminate the custom patches.


natebosch:
I worry this may work on Travis but may not work on some users setups.

jakemac53:
This definitely doesn't work - it looks like `Platform.executable` is not a relative or absolute path necessarily - its literally how you invoked it. In many cases this means just `dart`, which we can't reasonably turn into a path to that executable itself.

PR #1930, opened 2018-12-12
Update to latest package:graphs

natebosch:
Will roll this out package by package instead of using
dependency_overrides to get through CI.

### DIFF
--- a/build_runner_core/lib/src/package_graph/apply_builders.dart
+++ b/build_runner_core/lib/src/package_graph/apply_builders.dart
@@ -268,9 +268,8 @@ Future<List<BuildPhase>> createBuildPhases(
         (globalOptions[key] ?? BuilderOptions.empty).overrideWith(overrides);
   }
 
-  final cycles = stronglyConnectedComponents<String, TargetNode>(
+  final cycles = stronglyConnectedComponents<TargetNode>(
       targetGraph.allModules.values,
-      (node) => node.target.key,
       (node) => node.target.dependencies?.map((key) {
             if (!targetGraph.allModules.containsKey(key)) {
               _logger.severe('${node.target.key} declares a dependency on $key '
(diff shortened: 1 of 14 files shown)


build_runner_core/lib/src/package_graph/apply_builders.dart
```diff
@@ -268,17 +268,18 @@ Future<List<BuildPhase>> createBuildPhases(
...
             if (!targetGraph.allModules.containsKey(key)) {
               _logger.severe('${node.target.key} declares a dependency on $key '
                   'but it does not exist');
               throw CannotBuildException();
             }
             return targetGraph.allModules[key];
-          })?.where((n) => n != null));
+          })?.where((n) => n != null),
+      equals: (a, b) => a.target.key == b.target.key,
+      hashCode: (node) => node.target.key.hashCode);
```
kevmoo: nit: try a trailing common at the end of the param list, I think it makes it easier to read

natebosch: It has almost zero impact here... I don't think it's worth the extra line

kevmoo:
ack

natebosch:
Turns out `_test` would need a bunch of hacks to not upgrade all at once... will do it all in 1 go with overrides.

kevmoo:
ack

kevmoo:
So green travis then publish them all right away?

natebosch:
> So green travis then publish them all right away?

Yeah that's my plan. I'll submit this with a green travis, then submit a change with a red travis that drops the overrides and publish everything.

build_runner_core/lib/src/package_graph/apply_builders.dart
jakemac53: Bobs fears have now been fully realized :sob: :sob: :sob:

PR #1933, opened 2018-12-13
Drop bazel_codegen package

natebosch:
It's moving internal only.


jakemac53:
Do we actually need to move this package internal? It is generally useful for any build system integration and is rarely touched.

natebosch:
It'll have more work done as we change analysis stuff. It's unusable as is because it's completely undocumented, the build rules were effectively the only documentation.

jakemac53:
Fair enough

PR #1938, opened 2018-12-18
enable and fix a few more lints

kevmoo:


build_modules/lib/src/scratch_space.dart
```diff
@@ -38,7 +38,7 @@ final scratchSpaceResource = Resource<ScratchSpace>(() {
   // to shut down we might get file system exceptions on windows for an
   // arbitrary amount of time, so do retries with an exponential backoff.
   var numTries = 0;
-  while (true) {
+  for (;;) {
```
natebosch: I don't know that this is an improvement...

Did this lint catch any cases that we do care about?

kevmoo: > Did this lint catch any cases that we do care about?

It did not. I do like having **one** way to do the infinite loop, though.

Always do `for (;;)` seems a reasonable suggestion.

None of these are a big deal, though. Happy to revert.

natebosch: Internal code has `while (true)` 13 times as often as `for (;;)`.

I like the other lints, but let's leave these loop as is for now.

kevmoo:
@natebosch – PTAL

PR #1939, opened 2018-12-19
Allow PackageAssetReader without package as cwd

natebosch:
We can't rely on a pub layout for all packages because when running
tests under a `build_runner` generated output directory we omit the
`lib/` directory for non-root packages. It should be a safe assumption,
however, that the root package uses a pub layout whether we're running
from the actual package, a `build_runner` output, or a bazel test
environment.

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 0.10.4
 
-- Allow using `PackageAssetReader` when the current working directory is root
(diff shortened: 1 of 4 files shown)


natebosch:
Ugh, I think I was wrong about this, we don't have a pub style layout for the root package with `pub run build_runner test`.

natebosch:
Now it checks whether the package has a `lib/` directory at the `packagePath` and uses it as a heuristic to detect pub style packages.

@grouma PTAL

PR #1947, opened 2019-01-04
Add assetIdForElement

natebosch:
This replaces some functionality that builders were implementing outside
of the resolver, which then depended on details of the resolver
implementation.

- Add `UnresolavableAssetException`. Eventually we should update
  `AssetId.resolve` to use this exception.
- Add `assetIdForElement` API and implementations in all places
  necessary.
- Bump `build` to version `1.1.0` and update dependencies accordingly.

### DIFF
--- a/build_resolvers/pubspec.yaml
+++ b/build_resolvers/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_resolvers
-version: 0.2.2+7
+version: 0.2.3
 description: Resolve Dart code in a Builder
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_resolvers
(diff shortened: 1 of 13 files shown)


build_resolvers/pubspec.yaml
```diff
@@ -9,11 +9,15 @@ environment:
```
jakemac53: Need to update the version also (0.2.3-dev?)

Same for build_runner/build_runner_core (and update changelogs for those)

build_runner/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
 dependencies:
   args: ">=1.4.0 <2.0.0"
   async: ">=1.13.3 <3.0.0"
-  build: ">=1.0.0 <1.1.0"
+  build: ">=1.1.0 <1.0.0"
```
jakemac53: <1.2.0

natebosch: Done. Also moved minimum back to 1.0.0 since I didn't have to change any Dart code here.

natebosch:
Added a test. PTAL

PR #1949, opened 2019-01-04
Initial Build Daemon Support

grouma:
Initial export of Build Daemon logic.

Notable changes from the internal version:
- Merged a number of disjoint libraries into a single package
- Updated the client to take an optional command path
  - Internally this command path will be the prebuilt daemon vm binary
- Updated the daemon binary and example to work externally
- Cleaned up a number of lints

Additional changes:
- Updated mono repo settings to include build daemon
- Removed `avoid_returning_null` lint

Remaining work:
Provide a working implementation of `DaemonBuilder` which makes use of `build_runner_core`.

### DIFF
--- a/build_daemon/bin/build_daemon.dart
+++ b/build_daemon/bin/build_daemon.dart
@@ -1,0 +1,30 @@
+import 'dart:async';
+
+import 'package:build_daemon/constants.dart';
+import 'package:build_daemon/src/daemon.dart';
+import 'package:build_daemon/daemon_builder.dart';
+import 'package:watcher/watcher.dart';
+
+/// Entrypoint for the Dart Build Daemon.
+Future main(List<String> args) async {
+  var workingDirectory = args[0];
+
+  var daemon = Daemon(workingDirectory);
+
+  if (!daemon.tryGetLock()) {
+    if (runningVersion(workingDirectory) == currentVersion) {
+      print('Daemon is already running.');
+      print(readyToConnectLog);
+    } else {
+      print(versionSkew);
+    }
+  } else {
+    print('Starting daemon...');
+    // TODO(grouma) - Create a real builder for package:build
+    var builder = DaemonBuilder();
+    var watcher = Watcher(workingDirectory);
+    await daemon.start(builder, watcher.events);
+    print(readyToConnectLog);
+    await daemon.onDone;
+  }
+}
--- a/build_daemon/bin/build_daemon.dart
+++ b/build_daemon/bin/build_daemon.dart
@@ -7,7 +7,7 @@ import 'package:watcher/watcher.dart';
 
 /// Entrypoint for the Dart Build Daemon.
 Future main(List<String> args) async {
-  var workingDirectory = args[0];
+  var workingDirectory = args.first;
 
   var daemon = Daemon(workingDirectory);
 
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -16,15 +16,13 @@ import 'data/server_log.dart';
 class BuildDaemonClient {
   IOWebSocketChannel _channel;
 
-  final _buildResultsStreamController =
-      StreamController<BuildResults>.broadcast();
+  final _buildResults = StreamController<BuildResults>.broadcast();
 
   final _serverLogStreamController = StreamController<ServerLog>.broadcast();
 
   BuildDaemonClient._();
 
-  Stream<BuildResults> get buildResultsStream =>
-      _buildResultsStreamController.stream;
+  Stream<BuildResults> get buildResults => _buildResults.stream;
   Future<void> get finished async {
     if (_channel != null) await _channel.sink.done;
   }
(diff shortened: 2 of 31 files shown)


build_daemon/bin/build_daemon.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:async';
+
+import 'package:build_daemon/constants.dart';
+import 'package:build_daemon/src/daemon.dart';
+import 'package:build_daemon/daemon_builder.dart';
+import 'package:watcher/watcher.dart';
+
+/// Entrypoint for the Dart Build Daemon.
+Future main(List<String> args) async {
+  var workingDirectory = args[0];
```
natebosch: [nit] args.first?

build_daemon/lib/client.dart
```diff
@@ -0,0 +1,129 @@
...
+import 'data/build_status.dart';
+import 'data/build_target_request.dart';
+import 'data/log_to_paths_request.dart';
+import 'data/serializers.dart';
+import 'data/server_log.dart';
+
+class BuildDaemonClient {
+  IOWebSocketChannel _channel;
+
+  final _buildResultsStreamController =
```
natebosch: [optional] Is the `StreamController` part of the name adding value?

build_daemon/lib/client.dart
```diff
@@ -0,0 +1,129 @@
...
+      StreamController<BuildResults>.broadcast();
+
+  final _serverLogStreamController = StreamController<ServerLog>.broadcast();
+
+  BuildDaemonClient._();
+
+  Stream<BuildResults> get buildResultsStream =>
```
natebosch: Is the `Stream` part of the name adding value?

build_daemon/lib/client.dart
```diff
@@ -0,0 +1,129 @@
...
+    var request = LogToPathsRequest((b) => b..paths.replace(paths));
+    _channel.sink.add(jsonEncode(serializers.serialize(request)));
+  }
+
+  /// Registers a build target to be built upon any file change.
+  ///
+  /// Changes that match the patterns in [blackListPattern] will be ignored.
+  ///
+  /// Note that you should provide the "fully qualified" target name. For
+  /// example one should provide //some-target:test_run instead of
```
natebosch: [nit] This is a detail of the Dart test rules on bazel - should it be included here?

build_daemon/lib/client.dart
```diff
@@ -0,0 +1,129 @@
...
+        .transform(utf8.decoder)
+        .transform(const LineSplitter())
+        .firstWhere((line) => line == versionSkew || line == readyToConnectLog);
+
+    if (result == versionSkew) {
+      throw VersionSkew();
+    }
+
+    var port = _existingPort(workingDirectory);
+    if (port == null) return null;
```
natebosch: Based on the example usage, maybe this should throw an exception instead?

build_daemon/lib/src/managers/build_target_manager.dart
```diff
@@ -0,0 +1,86 @@
...
+      ? '/${splitPath.sublist(3).join('/')}'
+      : splitPath.join('/');
+  if (blackListedPatterns.any((pattern) => path.contains(pattern))) {
+    return true;
+  }
+  return false;
+}
+
+class BuildTarget {
+  // The target to build.
```
natebosch: Maybe "The name of the target to build" or "The build system identifier for the target"?

Otherwise, how is this different from the object in whole which is also a "BuildTarget"?

build_daemon/lib/src/managers/build_target_manager.dart
```diff
@@ -0,0 +1,86 @@
...
+        _buildTargets.firstWhere((b) => b == buildTarget, orElse: () => null);
+    if (existingTarget != null) {
+      existingTarget.listeners.add(channel);
+    } else {
+      _buildTargets.add(buildTarget);
+    }
+  }
+
+  void removeChannel(WebSocketChannel channel) {
+    var toRemove = [];
```
natebosch: Should this get a specified generic on the literal?

build_daemon/lib/src/server.dart
```diff
@@ -0,0 +1,129 @@
...
+  Future<int> listen() async {
+    var handler = webSocketHandler((WebSocketChannel channel) async {
+      _channels.add(channel);
+
+      channel.stream.listen((message) async {
+        dynamic request;
+        try {
+          request = serializers.deserialize(jsonDecode(message as String));
+        } catch (_) {
+          print('Unable to parse message: $message');
```
natebosch: I wonder if printing is appropriate in the library code... Should we look at using a Logger instead?

build_daemon/bin/build_daemon.dart
grouma: Done.

build_daemon/lib/client.dart
grouma: Changed.

build_daemon/lib/client.dart
grouma: Done.

build_daemon/lib/client.dart
grouma: Removed.

build_daemon/lib/client.dart
grouma: Changed.

build_daemon/lib/src/managers/build_target_manager.dart
grouma: Updated the comment.

build_daemon/lib/src/managers/build_target_manager.dart
grouma: I added the type but I guess `removeAll` requires only `Iterable<Object>`.

build_daemon/lib/src/server.dart
grouma: Yes. We should eventually make use of the logger package. I'll open an issue.

PR #1952, opened 2019-01-07
Copyrights, readme, changelog and other cleanup

grouma:
Clean up to prepare for initial publish.

The README.md will be expanded in the future as we nail down the client API.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,0 +1,3 @@
+### 0.1.0
+
+- Initial Build Daemon support.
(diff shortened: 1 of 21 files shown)


build_daemon/README.md
```diff
@@ -0,0 +1,3 @@
+# Build Daemon
```
kevmoo: Not needed! Headers on readmes are redundant in all places where they are displayed.

build_daemon/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+### 0.1.0
```
jakemac53: really useless nit but I think two `##`s is better, that allows us to use the `###` for subheadings instead of `####` which is barely recognizable as a header. Also at least in github `##` has an underline which helps separate each version a bit better

## Two

### Three

####  Four

PR #1956, opened 2019-01-08
Make it possible to add builder-specific data to BuildResult.

iinozemtsev:
- Add builderDetails field to BuildResult
- Allow overriding serializers in client and server

### DIFF
--- a/build_daemon/lib/data/build_status.dart
+++ b/build_daemon/lib/data/build_status.dart
@@ -35,6 +35,10 @@ abstract class BuildResult implements Built<BuildResult, BuildResultBuilder> {
   String get error;
   @nullable
   bool get isCached;
+
+  /// Arbitrary additional information from daemon builder.
+  @nullable
+  Object get builderDetails;
 }
 
 abstract class BuildResults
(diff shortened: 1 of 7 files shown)


iinozemtsev:
I signed it!

grouma:
Update the changelog and daemon version. I'll merge this and pull it into Google3.

build_daemon/lib/data/build_status.dart
```diff
@@ -35,6 +35,10 @@ abstract class BuildResult implements Built<BuildResult, BuildResultBuilder> {
   String get error;
   @nullable
   bool get isCached;
+
+  /// Arbitrary additional information from daemon builder.
+  @nullable
+  Object get builderDetails;
```
grouma: If we are injecting custom serializers is this even necessary? Is it possible for the serializer to override this object and just work?

iinozemtsev: If I understand you correctly, this is basically composition vs inheritance (custom builders return BuildResult subclasses), and I don't have strong opinions regarding this, but usage of Built values add some nuances - BuildResult should be converted to an interface, and this affects builders generated for other built values which include BuildResult field.

I've created an alternative pull request #1959, it's hard to me to choose which approach is better without having other builder implementations in mind

grouma:
The subclassing approach appears to be the better solution. Closing this one.

PR #1961, opened 2019-01-09
Use xor to compute combined input hashes instead of md5

jakemac53:
This should still be unique enough for our purposes (we don't use these digests in hash maps), and it allows us to use unordered sets which are more efficient, as well as being very fast on its own.

On angular gallery for edits to the material button example I saw incremental build time improvements of about 10%.

This will also allow us to add parallelism to the computeCombinedDigest function in a followup PR, without having to retain ordering.

There is one known potential bug this would introduce, which is if you were to swap the contents of two files the resulting combined digest would not change. In order to resolve that issue we include the `AssetId.hashCode` in the digest in the default `AssetReader.digest` implementation, and add a comment that custom implementations should include the asset id in their digest as well.

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,13 +1,10 @@
-## 1.1.1
-
-- Include the AssetId hash code in the default digest implementation.
-  - Any custom implementations of the `AssetReader.digest` method should do the
-    same, and a comment has been added to that effect.
-
 ## 1.1.0
 
 - Add `Resolver.assetIdForElement` API. This allows finding the Dart source
   asset which contains the definition of an element found through the analyzer.
+- Include the AssetId hash code in the default digest implementation.
+  - Any custom implementations of the `AssetReader.digest` method should do the
+    same, and a comment has been added to that effect.
 
 ## 1.0.2
 
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.1.1
+
+- Include the AssetId hash code in the default digest implementation.
+  - Any custom implementations of the `AssetReader.digest` method should do the
+    same, and a comment has been added to that effect.
+
 ## 1.1.0
 
 - Add `Resolver.assetIdForElement` API. This allows finding the Dart source
--- a/build/lib/src/asset/reader.dart
+++ b/build/lib/src/asset/reader.dart
@@ -45,10 +45,10 @@ abstract class AssetReader {
   /// may differ based on the current build system being used.
   Future<Digest> digest(AssetId id) async {
     var digestSink = AccumulatorSink<Digest>();
-    var bytesSink = md5.startChunkedConversion(digestSink);
-    bytesSink.add(await readAsBytes(id));
-    bytesSink.add([id.hashCode]);
-    bytesSink.close();
+    md5.startChunkedConversion(digestSink)
+      ..add(await readAsBytes(id))
+      ..add(id.toString().codeUnits)
+      ..close();
     return digestSink.events.first;
   }
 }
--- a/build/lib/src/asset/reader.dart
+++ b/build/lib/src/asset/reader.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:convert';
 
 import 'package:crypto/crypto.dart';
+import 'package:convert/convert.dart';
 import 'package:glob/glob.dart';
 
 import 'id.dart';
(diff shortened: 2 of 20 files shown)


build_runner_core/lib/src/asset/reader.dart
```diff
@@ -43,8 +42,8 @@ class SingleStepReader implements AssetReader {
   final FutureOr<GlobAssetNode> Function(
       Glob glob, String package, int phaseNum) _getGlobNode;
 
-  /// The assets read during this step in sorted order.
-  final assetsRead = SplayTreeSet<AssetId>();
+  /// The assets read during this step.
+  final assetsRead = Set<AssetId>();
```
kevmoo: Use `HashSet`? It does not preserve order and is faster than the default `LinkedHashSet`.

build_runner_core/lib/src/asset_graph/node.dart
```diff
@@ -175,9 +174,7 @@ class GeneratedAssetNode extends AssetNode implements NodeWithInputs {
     @required this.isFailure,
     @required this.primaryInput,
     @required this.builderOptionsId,
-  })  : inputs = inputs != null
-            ? SplayTreeSet.from(inputs)
-            : SplayTreeSet<AssetId>(),
+  })  : inputs = inputs != null ? Set.from(inputs) : Set(),
```
kevmoo: HashSet?

build_runner_core/lib/src/asset_graph/serialization.dart
```diff
@@ -55,7 +55,7 @@ class _AssetGraphDeserializer {
       for (var output in node.outputs) {
         var inputsNode = graph.get(output) as NodeWithInputs;
         assert(inputsNode != null, 'Asset Graph is missing $output');
-        inputsNode.inputs ??= SplayTreeSet<AssetId>();
+        inputsNode.inputs ??= Set<AssetId>();
```
kevmoo: HashSet?

build/lib/src/asset/reader.dart
```diff
@@ -36,11 +37,19 @@ abstract class AssetReader {
...
+  ///
   /// This should be treated as a transparent [Digest] and the implementation
   /// may differ based on the current build system being used.
   Future<Digest> digest(AssetId id) async {
-    var bytes = await readAsBytes(id);
-    return md5.convert(bytes);
+    var digestSink = AccumulatorSink<Digest>();
+    var bytesSink = md5.startChunkedConversion(digestSink);
+    bytesSink.add(await readAsBytes(id));
+    bytesSink.add([id.hashCode]);
```
natebosch: `ByteConversionSink` does take a `List<int>`, but afaik each of those ints must fit into a byte and we don't have a guarantee that `id.hashCode` does.

We could convert it, I think it would look like:

```dart
Uint8List intBytes(int number) {
  var result = new Uint8List(8);
  for (int i = 0; i < 8; i++) {
    result[7 - i] = number & 0xFF;
    number = number >> 8;
  }
  return result;
}
```

Alternatively do `bytesSink.add(utf8.encode('$id'))` or similar.

build/lib/src/asset/reader.dart
```diff
@@ -36,11 +37,19 @@ abstract class AssetReader {
...
+  /// different, even if their content is the same.
+  ///
   /// This should be treated as a transparent [Digest] and the implementation
   /// may differ based on the current build system being used.
   Future<Digest> digest(AssetId id) async {
-    var bytes = await readAsBytes(id);
-    return md5.convert(bytes);
+    var digestSink = AccumulatorSink<Digest>();
+    var bytesSink = md5.startChunkedConversion(digestSink);
+    bytesSink.add(await readAsBytes(id));
```
natebosch: [nit] How about cascade operator?

build_runner_core/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.1.4-dev
```
natebosch: This is also unpublished.

build/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.1.1
```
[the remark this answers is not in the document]
jakemac53: Either way should be fine I think

jakemac53:
@natebosch re: the current unpublished versions lets discuss this AM?

I don't really have an opinion here - I don't remember the state of things or why they aren't published yet..

build/CHANGELOG.md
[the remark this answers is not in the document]
natebosch: Synced offline, let's roll this in with 1.1.0 and publish everything together.

PR #1962, opened 2019-01-09
Throw an UnsupportedError on unresolvable URIs

jbdeboer:
Fall through to the throw statement if _packageResolver can not resolve a URI.

Also, fix a typo in the error message.


natebosch:
Looks good, can you update the changelog and pubspec as well?

jbdeboer:
done.

PR #1964, opened 2019-01-10
use a pool to limit the number of digests we ask for at a time

jakemac53:
The pool size of 8 is pretty arbitrary, but it does improve build times significantly (cuts api-affecting edits down by ~50% in some basic tests).

I tried using `Pool.forEach` but it was slower - probably because of all the throwaway streams it creates? I don't actually care about the values returned in this case.

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -717,23 +717,22 @@ class _SingleBuild {
 
     // Limit the total number of digests we are computing at a time. Otherwise
     // this can overload the event queue.
-    var pool = Pool(8);
-    await Future.wait(ids.map((id) => pool.withResource(() async {
-          var node = _assetGraph.get(id);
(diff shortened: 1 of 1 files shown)


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -715,22 +715,25 @@ class _SingleBuild {
...
-        _combine(md5.convert(id.toString().codeUnits).bytes as Uint8List);
-        continue;
-      } else {
-        node.lastKnownDigest ??= await reader.digest(id);
-      }
-      _combine(node.lastKnownDigest.bytes as Uint8List);
-    }
+    // Limit the total number of digests we are computing at a time. Otherwise
+    // this can overload the event queue.
+    var pool = Pool(8);
```
natebosch: Discussed offline - since we have per-phase pools we should be fine using `Future.wait` here without additional pooling.

PR #1965, opened 2019-01-10
Support for OptionsSkew

grouma:
- Remove support for passing options after a daemon has begun running
- Options are now provided upon connecting
- Throw when there is an options skew
- No longer pass options to the DaemonBuilder `build` method
- Update examples (note that the daemon binary will soon be folded into package:build_runner)


build_daemon/bin/build_daemon.dart
```diff
@@ -12,22 +12,32 @@ import 'package:watcher/watcher.dart';
...
   if (!daemon.tryGetLock()) {
-    if (runningVersion(workingDirectory) == currentVersion) {
+    var runningOptions = currentOptions(workingDirectory);
+    var version = runningVersion(workingDirectory);
+    if (version != currentVersion) {
+      print('Running Version: $version');
+      print('Current Version: $currentVersion');
+      print(versionSkew);
+    } else if (!(runningOptions.containsAll(requestedOptions) &&
+        runningOptions.length == requestedOptions.length)) {
```
natebosch: [nit] the `.length` check is probably faster and could come first

build_daemon/lib/src/daemon.dart
```diff
@@ -21,6 +21,14 @@ String runningVersion(String workingDirectory) {
   return versionFile.readAsStringSync();
 }
 
+/// Returns the current options of the running build daemon. Null if one isn't
```
natebosch: https://www.dartlang.org/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

PR #1968, opened 2019-01-12
Initial build runner support of build daemon

grouma:
`package:build_daemon` changes:
- Prevent calling shutdown twice
- Clean up exception messages 
- Update example to work with the example directory

`package:build_runner_core` changes:
- Update `run` method to take in `buildDirs`
- Rev package as this is a breaking change

`package:build_runner` changes:
- Create new `daemon` command
- Implement `DaemonBuilder` interface using `BuildImpl`.
- Split out shared logic from `watcher` command into `delete_writer.dart`, `collect_changes.dart` and `change_filter.dart`.


build/pubspec.yaml
```diff
@@ -24,3 +24,15 @@ dev_dependencies:
   build_vm_compilers: ^0.1.0
   pedantic: ^1.0.0
   test: ^1.2.0
+
+dependency_overrides:
+  build_daemon:
+    path: ../build_daemon
+  build_runner:
```
jakemac53: You can likely get rid of at least some of these now that I published all the packages

build_daemon/example/example.dart
```diff
@@ -9,10 +9,17 @@ import 'package:build_daemon/client.dart';
 
 void main(List<String> args) async {
   BuildDaemonClient client;
-  var workingDirectory = Directory.current.path;
+  var workingDirectory =
+      Directory.fromUri(Uri(path: Directory.current.path + '/../example')).path;
```
jakemac53: I would recommend using the `path` package here, something like `p.normalize(p.join(p.current, '../example'))`. I am also curious where the `/../example` is coming from (cruft?).

build_daemon/example/example.dart
```diff
@@ -9,10 +9,17 @@ import 'package:build_daemon/client.dart';
...
+
   try {
-    client = await BuildDaemonClient.connect(workingDirectory,
-        ['pub', 'run', 'build_daemon', workingDirectory, 'some-option']);
+    client = await BuildDaemonClient.connect(workingDirectory, [
+      'pub',
+      'run',
+      'build_runner',
+      'daemon',
+      '--delete-conflicting-outputs',
```
jakemac53: This isn't defaulted today so this would be a change from current behavior

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -0,0 +1,160 @@
...
+  @override
+  Future<void> build(Set<String> targets, Set<String> logToPaths) async {
+    _logMessage(Level.INFO, 'About to build $targets...');
+    try {
+      var mergedChanges = collectChanges([_changesFromLastBuild]);
+      var result = await _builder.run(mergedChanges, targets.toList());
+      var results = <daemon.BuildResult>[];
+      for (var target in targets) {
+        if (result.status == BuildStatus.success) {
+          // TODO(grouma) - Can we notify if a target was cached?
```
jakemac53: you could check if `result.outputs.isEmpty`?

grouma: Discussed offline. We need to know if a given target is cached. This isn't a huge concern until Bolt onboards. I think there are alternatives.

build_daemon/example/example.dart
grouma: This is just an example.

build_daemon/example/example.dart
grouma: Done.

build/pubspec.yaml
grouma: Done.

build_daemon/lib/client.dart
```diff
@@ -102,7 +102,8 @@ class BuildDaemonClient {
     var result = await stream.firstWhere(_isActionMessage, orElse: () => null);
 
     if (result == null) {
-      throw Exception('Unable to start build daemon: ${await output}');
+      throw Exception(
```
natebosch: Didn't notice this before - but in general we don't want to row a bare `Exception` - better to subclass it with something specific. We don't need to resolve in this PR though.

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -0,0 +1,161 @@
...
+    this._builder,
+    this._buildOptions,
+    this._outputStreamController,
+    this._graphEvents,
+  );
+
+  @override
+  Stream<daemon.BuildResults> get builds => _buildResults.stream;
+
+  Stream<WatchEvent> get changes => _graphEvents.map((data) {
```
natebosch: would it be better to set this as a field in the constructor? It looks like we're relying on this getter only being called once.

build_runner/lib/src/entrypoint/daemon.dart
```diff
@@ -0,0 +1,60 @@
...
+  Future<int> run() async {
+    var workingDirectory = Directory.current.path;
+    var options = readOptions();
+    var daemon = Daemon(workingDirectory);
+    var requestedOptions = argResults.arguments.toSet();
+    if (!daemon.tryGetLock()) {
+      var runningOptions = currentOptions(workingDirectory);
+      var version = runningVersion(workingDirectory);
+      if (version != currentVersion) {
+        print('Running Version: $version');
```
natebosch: Normally we expect output to go through the logger - but that could impact some of the things that need to get parsed precisely by process which started the daemon...

I _think_ it's OK to use print here - @jakemac53 do you have any concerns?

build_runner/lib/src/entrypoint/daemon.dart
```diff
@@ -0,0 +1,60 @@
...
+      var builder = await BuildRunnerDaemonBuilder.create(
+        packageGraph,
+        builderApplications,
+        options,
+      );
+      await daemon.start(requestedOptions, builder, builder.changes);
+      print(readyToConnectLog);
+      await daemon.onDone;
+    }
+    return 0;
```
natebosch: This turns into an exit code. Do we want a failure in the case where there was version skew or anything like that?

jakemac53: Agreed we probably want other exit codes for failures (ensures that travis will fail for instance).

PR #1974, opened 2019-01-15
Add buildDirs to build method

grouma:
- Update `run` method to take in `buildDirs`
- Remove `buildDirs` from `BuildOptions`
- Rev package as this is a breaking change

### DIFF
--- a/_test_common/lib/test_phases.dart
+++ b/_test_common/lib/test_phases.dart
@@ -118,7 +118,6 @@ Future<BuildResult> testBuilders(
       skipBuildScriptCheck: true,
       overrideBuildConfig: overrideBuildConfig,
       enableLowResourcesMode: enableLowResourcesMode,
-      buildDirs: buildDirs,
       logPerformanceDir: logPerformanceDir);
 
   BuildResult result;
@@ -129,7 +128,7 @@ Future<BuildResult> testBuilders(
     builderConfigOverrides,
     isReleaseBuild: false,
   );
-  result = await build.run({}, options.buildDirs);
+  result = await build.run({}, buildDirs);
   await build.beforeExit();
   await options.logListener.cancel();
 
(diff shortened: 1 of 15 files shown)


_test_common/lib/test_phases.dart
```diff
@@ -129,7 +129,7 @@ Future<BuildResult> testBuilders(
     builderConfigOverrides,
     isReleaseBuild: false,
   );
-  result = await build.run({});
+  result = await build.run({}, options.buildDirs);
```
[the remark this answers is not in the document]
grouma: The `OptionalOutputTracker` requires a list of buildDirs. This becomes somewhat problematic due to the order of construction.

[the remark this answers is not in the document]
natebosch: I think that probably means that the `OptionalOutputTracker` won't work when we need it for the build daemon?

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -85,11 +84,12 @@ class BuildImpl {
         _verbose = options.verbose,
         _environment = buildDefinition.environment,
         _trackPerformance = options.trackPerformance,
-        _buildDirs = options.buildDirs,
         _logPerformanceDir = options.logPerformanceDir;
 
-  Future<BuildResult> run(Map<AssetId, ChangeType> updates) =>
-      _SingleBuild(this).run(updates)..whenComplete(_resolvers.reset);
+  Future<BuildResult> run(
+          Map<AssetId, ChangeType> updates, List<String> buildDirs) =>
```
jakemac53: Why not make `buildDirs` an optional (named) argument instead (with [] as the default)? Then i don't think we have to deal with a breaking change (and its a sane default anyways).

natebosch: I think it would still be a breaking behavior change

grouma: This is still a breaking change regardless as we expose `BuildOptions`. I do like making the buildDirs optional though.

Issue #1972, opened 2019-01-14
Make it possible to override the `.dart_tool/generated` directory


natebosch:
This could make it possible to create and snapshot a tool that does a specific build which doesn't depend on packages in the pub solve and run it so it doesn't step on the toes of another build.

jakemac53:
I started looking at this, today this is a constant which is referenced in a few locations although not an insane number.

The obvious easy solution is to simply make that a mutable global variable... feels dirty but definitely the easiest approach. It also allows users to simply hardcode this in their custom build script instead of using a command line arg, which actually more directly solves this use case.

We could make it a public method you have to go through to set it, (overrideGeneratedDir?) or we could just remove the `const` from the field.

We can do custom validation logic if we want either way... possibly enforcing it can only be set once or something, maybe even enforce you do it before creating a `BuildRunner` instance? 

wdyt @natebosch

jakemac53:
The alternative is basically plumbing this through `BuildOptions` and down through anything that wants access to this constant (and migrate the constant to a method that takes an optional arg).

natebosch:
I have a slight preference for adding it to `BuildOptions`, but if it's going to cascade into a huge change I'd be fine with a top-level `overrideGeneratedDir` method.

jakemac53:
This is resolved now

PR #1977, opened 2019-01-15
add overrideGeneratedOutputDirectory and lockGeneratedOutputDirectory

jakemac53:
Will add tests/etc, for now just floating this as a potential resolution to https://github.com/dart-lang/build/issues/1972.

This is extremely defensive, so it should be quite safe without causing much churn elsewhere in the code.

### DIFF
--- a/build_runner_core/lib/src/util/constants.dart
+++ b/build_runner_core/lib/src/util/constants.dart
@@ -26,10 +26,48 @@ final String _scriptPath = Platform.script.scheme == 'file'
 /// Directory containing automatically generated build entrypoints.
 ///
 /// Files in this directory must be read to do build script invalidation.
-const entryPointDir = '.dart_tool/build/entrypoint';
+const entryPointDir = '$cacheDir/entrypoint';
 
 /// The directory to which hidden assets will be written.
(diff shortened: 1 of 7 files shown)


jakemac53:
Looking at the test failures I think I will relax the reading of the generated dir before it has been locked, but still make sure to lock it inside of BuildImpl.create.

build_runner_core/lib/src/util/constants.dart
```diff
@@ -26,10 +26,42 @@ final String _scriptPath = Platform.script.scheme == 'file'
 /// Directory containing automatically generated build entrypoints.
 ///
 /// Files in this directory must be read to do build script invalidation.
-const entryPointDir = '.dart_tool/build/entrypoint';
+const entryPointDir = '$cacheDir/entrypoint';
 
 /// The directory to which hidden assets will be written.
-const generatedOutputDirectory = '$cacheDir/generated';
+///
+/// Can not be accessed prior to being locked with
```
natebosch: I think this is no longer true?

natebosch:
I think it's probably safe to allow reading early. We'll have to be careful in the future not to read that in anyway that is dangerous before locking it.

PR #1978, opened 2019-01-15
Add support for color logging

grouma:
Messages from the build daemon will now have appropriate colors based on the LogRecord.

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -10,6 +10,7 @@ import 'package:build_daemon/daemon_builder.dart';
 import 'package:build_daemon/data/build_status.dart' as daemon;
 import 'package:build_daemon/data/server_log.dart';
 import 'package:build_runner/src/entrypoint/options.dart';
+import 'package:build_runner/src/logging/std_io_logging.dart';
 import 'package:build_runner/src/package_graph/build_config_overrides.dart';
 import 'package:build_runner/src/watcher/asset_change.dart';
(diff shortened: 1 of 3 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -98,11 +100,12 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
     await _buildOptions.logListener.cancel();
   }
 
-  void _logMessage(Level level, String message) {
-    // TODO(grouma) - use the logic in std_io_logging.dart to include color.
-    _outputStreamController.add(ServerLog((b) => b.log =
-        LogRecord(level, message, 'BuildRunnerBuildDaemon').toString()));
-  }
+  void _logMessage(Level level, String message) => overrideAnsiOutput(
+      true,
```
jakemac53: I think we want to retain the `assumeTty == true || ansiOutputEnable` logic if possible

grouma: I don't think that makes sense here because this message is being sent over a websocket. The daemon client is running in another process which presumably will be printing these messages.

natebosch: It's not outrageously expensive, but each call to `overrideAnsiOutput` creates a `Zone` to run inside. Should we be doing that a higher level than each log message?

grouma: I was looking at that but couldn't think of a better location. Thoughts?

PR #1979, opened 2019-01-16
Forward filesystem changes to daemon builder

iinozemtsev:
This should give builders more flexibility in terms of reporting/ignoring changes


iinozemtsev:
I saw it, but  probably not close enough. My main concern against listening to changes is that builder doesn't know what changes should be ignored (currently build_target_manager tracks this).

PR #1980, opened 2019-01-17
Pass build changes to build_runner

grouma:
- Use the `fileChanges` passed by the build daemon (which can be filtered) to initiate the builds
- Remove now unused internal state e.g. `_changesFromLastBuild`

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -12,7 +12,6 @@ import 'package:build_daemon/data/server_log.dart';
 import 'package:build_runner/src/entrypoint/options.dart';
 import 'package:build_runner/src/logging/std_io_logging.dart';
 import 'package:build_runner/src/package_graph/build_config_overrides.dart';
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -161,7 +161,9 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
               expectedDeletes,
             ));
 
-    return BuildRunnerDaemonBuilder._(
-        builder, buildOptions, outputStreamController, graphEvents);
+    var changes =
+        graphEvents.map((data) => WatchEvent(data.type, data.id.path));
```
natebosch: I should have noticed this before - this won't work when you have changes in the non-root package with a path dependency. This looks like it's assuming all changes are in the root package only.

grouma: I am now serializing the AssetId. This shouldn't be a problem as we are consuming the watch events. It does make some of the blacklist patterns more difficult to write but for the external world I don't think those will be used very often.

PR #1981, opened 2019-01-17
Daemon asset serving

grouma:
- Make the `rootDir` optional for the `AssetHandler`
- Create a basic `AssetServer` which uses the `AssetHandler`
- Spin up an asset server from within the build daemon
- Log the port for which this asset server is listening (webdev will know where to look)
- Add a handful of integration tests

### DIFF
--- a/build_runner/lib/src/daemon/asset_server.dart
+++ b/build_runner/lib/src/daemon/asset_server.dart
@@ -7,7 +7,6 @@ import 'dart:io';
 
 import 'package:build_runner/src/server/server.dart';
 import 'package:build_runner_core/build_runner_core.dart';
-import 'package:shelf/shelf.dart';
 import 'package:shelf/shelf_io.dart' as shelf_io;
 
 class AssetServer {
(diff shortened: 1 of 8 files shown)


build_runner/lib/src/daemon/asset_server.dart
```diff
@@ -0,0 +1,36 @@
...
+  static Future<AssetServer> run(
+      FinalizedReader reader, String rootPackage) async {
+    var server = await HttpServer.bind(InternetAddress.anyIPv6, 0);
+    shelf_io.serveRequests(server, _handler(reader, rootPackage));
+    return AssetServer._(server);
+  }
+
+  static Handler _handler(FinalizedReader reader, String rootPackage) {
+    var cascade = Cascade().add((request) {
+      return AssetHandler(reader, rootPackage).handle(request);
```
natebosch: I suppose they are relatively cheap - but do we want to construct an `AssetHandler` on every request?

Should this get pulled out to a variable in the surrounding method?

And actually, now that I'm seeing it - why are we going through a `Pipeline` and `Cascade`? There is no `Middleware` and only a single handler?

Does it work if line 25 is `shelf_io.serveRequests(server, AssetHandler(reader, rootPackage).handle);`

build_runner/lib/src/entrypoint/daemon.dart
```diff
@@ -56,9 +58,13 @@ class DaemonCommand extends BuildRunnerCommand {
...
           options,
         );
+        var server =
+            await AssetServer.run(builder.reader, packageGraph.root.name);
+        File(assetServerPortFilePath(workingDirectory))
+            .writeAsStringSync('${server.port}');
         await daemon.start(requestedOptions, builder, builder.changes);
         stdout.writeln(readyToConnectLog);
-        await daemon.onDone;
+        await daemon.onDone.then((_) => server.stop());
```
natebosch: can avoid the empty closure since we don't care about the result.

`await daemon.onDone.whenComplete(server.stop);`

grouma: Didn't know about the `whenComplete` method. Neat!

PR #1982, opened 2019-01-18
Change logToPaths to ClientOptions

iinozemtsev:
- Added an empty interface `ClientOptions`
- Replaced `ChannelOptionsManager` with `Map<WebSocketChannel, ClientOptions> in server
- Replaced `LogToPathsRequest` with `SetClientOptions` request

Also I believe fixing two subtle bugs:

1. The original code was passing `logToPaths` from all connected clients, new code passes client options of only those clients, who are interested in targets to be built.
1. The original code didn't call ChannelOptionsManager.removeChannel.


iinozemtsev:
Closing as #1985 is a better alternative for this

PR #1983, opened 2019-01-18
Migrate to AnalysisDriver

### LINKED ISSUE #1886: Migrate to AnalysisDriver

natebosch:
Closes #1886


jakemac53:
I am going to push a commit (or a couple) here which update to the latest plan - we will do a non-breaking release of package build which requires analyzer 0.35.0 (the first release with no task model).

So, this will be package:build version 1.1.1, and I am also going to change package:build_resolvers to version 1.0.0 since its a breaking change anyways.

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -0,0 +1,158 @@
+// Copyright (c) 2018, the Dart project authors. Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+// ignore: deprecated_member_use
+import 'package:analyzer/analyzer.dart' show parseDirectives;
```
jakemac53: Will this still exist in analyzer 0.35? @stereotype441

kevmoo:
@jakemac53 @natebosch – made another push here

jakemac53:
cc @scheglov see https://travis-ci.org/dart-lang/build/jobs/486622648#L621, I think we had a similar issue internally also, but I don't remember the resolution.

Basically [this test](https://github.com/dart-lang/build/blob/master/build_resolvers/test/resolver_test.dart#L64) is failing because the uri doesn't exist and it isn't being reported in the imported libraries list as a result, when it would have previously.

scheglov:
Can `package:b/missing.dart` be resolved to the file path?

If it cannot be resolved, then there is no `Source` for it, so there is the corresponding `ImportElement` with `uri`, but its `importedLibrary` is `null`.

jakemac53:
In some cases maybe it could be resolved to an expected path (or at least a reasonable expected path), assuming the package is a dependency at least. But even if the package isn't a dependency I think we still want to be able to see the import?

scheglov:
You see the import, in form of `ImportElement`.
But its URI cannot be resolved.
So, `importedLibrary` for it is `null`.

If the URI can be resolved, so, there is file path for it, even if the file does not exist, then the `importedLibrary` is not `null` (but the `LibraryElement` is basically empty - single unit, no declarations).

jakemac53:
Ok, I have pushed what is hopefully the last commit to this branch.

Assuming travis is happy I will go ahead and merge, and then publish the following packages, which should unblock the world:

build 1.1.1
build_modules 1.0.5
build_resolvers 1.0.0
build_test 0.10.6
build_vm_compilers 0.1.1+5
build_web_compilers 1.0.1

PR #1984, opened 2019-01-18
Ignore `uri_does_not_exist` in analyzer/DDC

natebosch:
- Drop `strong-mode` option - it is the default.
- Add ignore for `uri_does_not_exist`.

Since imports are crawled while finding modules we have separate error
reporting already for missing imports. We don't need DDC to find these
for us.

### DIFF
--- a/_test/test/config_specific_import_test.dart
+++ b/_test/test/config_specific_import_test.dart
@@ -4,11 +4,8 @@
 
 import 'package:test/test.dart';
 
-// ignore: uri_does_not_exist
 import 'common/message.dart'
-    // ignore: uri_does_not_exist
     if (dart.library.io) 'common/message_io.dart'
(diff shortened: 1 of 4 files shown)


jakemac53:
I think this is reasonable, can you also remove the comments here just to confirm all is good https://github.com/dart-lang/build/blob/master/_test/test/config_specific_import_test.dart#L7

natebosch:
Ugh, DDC seems to ignore this. I don't know if it ignores `--options` altogether, or only ignores if you try to change error levels.

In any case I don't think we can hack around it on our end.

natebosch:
On the plus side, the original request from https://github.com/dart-lang/sdk/issues/29713 is now essentially granted. On the downside we can't hack around https://github.com/dart-lang/sdk/issues/34177 using this strategy.

PR #1985, opened 2019-01-21
Add support for custom build targets

iinozemtsev:
- Add BuildTarget interface
- Remove ClientOptions
- Make it possible to override `shouldBuild` in BuildTargetManager
- Remove custom path manipulations from default `shouldBuild`

The predicate in BuildTargetManager constructor could be just a method
in BuildTarget interface, but that would create an unwanted dependency
from data to WatchEvent.

I'm intentionally making this PR on top of #1982, so that they both can be merged, but if we decide to go with this one only, I'll squash commits.


build_daemon/lib/data/client_options.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Marker interface for any client-related information to be passed to builder.
+abstract class ClientOptions {}
```
grouma: Is this necessary anymore? We can always add it in the future but I don't think it will be used both internally and externally given the changes to the build_target.

build_daemon/lib/src/server.dart
```diff
@@ -16,16 +16,16 @@ import 'package:web_socket_channel/web_socket_channel.dart';
...
 import '../data/serializers.dart';
 import 'managers/build_target_manager.dart';
-import 'managers/channel_options_manager.dart';
 
 class Server {
   final _isDoneCompleter = Completer();
-  final _buildTargetManager = BuildTargetManager();
-  final _logPathsManager = ChannelOptionsManager();
+  final BuildTargetManager _buildTargetManager;
+  final _clientOptions = <WebSocketChannel, ClientOptions>{};
```
grouma: This doesn't look to be used.

PR #1994, opened 2019-01-25
Ensure reader is reset prior to build

grouma:
- Update version of `package:build_runner_core` to contain the reader fix
  - Will publish once this is in
- Update `package:build_runner` to depend on the latest `package:build_runner_core`.
- Fix a typo in the changelog

### DIFF
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_runner
-version: 1.2.1
+version: 1.2.0
 description: Tools to write binaries that run builders.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_runner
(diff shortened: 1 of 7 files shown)


build_runner/pubspec.yaml
```diff
@@ -14,7 +14,7 @@ dependencies:
   build_config: ^0.3.1
   build_daemon: ^0.2.0
   build_resolvers: ^0.2.0
-  build_runner_core: ^2.0.0
+  build_runner_core: ^2.0.1
```
jakemac53: You don't need to bump this - its just a bug fix. A pub upgrade will fix things for people either way.

jakemac53: (no need to release a new build_runner at all)

grouma: Not sure what I was thinking. Fixed.

grouma: Actually I was right initially. I removed the reset in `build_runner` so the dot dep is necessary.

PR #1997, opened 2019-01-25
Change Daemon Builder log message

grouma:
Now that the build target is a built value the `toString` value contains additional information which isn't relevant for users.


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -57,7 +57,8 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
         .map<AssetChange>(
             (change) => AssetChange(AssetId.parse(change.path), change.type))
         .toList();
-    _logMessage(Level.INFO, 'About to build $targets...');
+    var targetNames = targets.map((t) => t.target);
+    _logMessage(Level.INFO, 'About to build $targetNames...');
```
natebosch: Note that this won't print out all targets if there are too many. You'll want `${targetNames.toList()}` if you want to be sure to see them all.

PR #2002, opened 2019-01-28
Add path dep

grouma:
Explicitly add path dependency before publishing

This dep was added in https://github.com/dart-lang/build/pull/2001


kevmoo:
We need a lint for this! 

https://github.com/dart-lang/linter/issues/30

PR #2006, opened 2019-01-30
Serve output digests through build daemon

grouma:
Create a handler for `/$outputDigests` which returns the digests for all outputs. These digests will be consumed by the hot reload client to determine which modules should be reloaded.

Discussion:
The complete output digests contain significantly more data than necessary. We can do something similar to https://github.com/dart-lang/build/blob/master/build_runner/lib/src/server/server.dart#L165 which returns digests for just requested assets. I'm not sure if this is strictly necessary but it's an optimization we can do later.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.3
+
+- `daemon` command now serves output asset digests under `/$outputDigests`. 
+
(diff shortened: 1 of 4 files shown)


grouma:
Discussed offline. We'll provide similar logic in build_web_modules instead which should be cleaner and more performant.

PR #2010, opened 2019-01-31
Fix expected output of generated_build_script to use latest code_builder

kevmoo:
Also tweaked generator to include //ignore comment to make analyzer happy


_test/test/goldens/generated_build_script.dart
```diff
@@ -1,3 +1,5 @@
+// ignore_for_file: directives_ordering
```
kevmoo: I could have bumped the min version of code builder used in build_runner – then enabled the order option. But that seems silly to push down to our users now...

PR #2013, opened 2019-01-31
Shut down the daemon if no client connects

grouma:


build_daemon/test/daemon_test.dart
```diff
@@ -46,6 +46,13 @@ void main() {
       expect(await getOutput(daemon), 'RUNNING');
     });
 
+    test('shuts down if no client connects within 30 seconds', () async {
+      var workspace = uuid.v1() as String;
+      var daemon = await _runDaemon(workspace);
+      testDaemons.add(daemon);
+      expect(await daemon.exitCode, isNotNull);
+    }, timeout: Timeout(Duration(minutes: 1)));
```
jakemac53: Can we instead make it configurable and then reduce it for the test? That is really painful to add 30 seconds to our CI just for this.

PR #2014, opened 2019-01-31
remove overrides after publishing

jakemac53:
Everything should be good now, the only override left is an analyzer override for a couple packages due to source_gen and json_serializable dev dependencies.


jakemac53:
@grouma this should be g2g now

PR #2015, opened 2019-02-01
Transitive module digests output for an entrypoint

grouma:
Output the transitive module digests for an entrypoint. These will be consumed during hot reload.


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -99,6 +99,20 @@ Future<Null> bootstrapDdc(BuildStep buildStep,
...
       dartEntrypointId.changeExtension(jsEntrypointExtension),
       entrypointJsContent);
+
+  // Output the digests for transitive modules.
+  // These can be consumed for hot reloads.
+  var moduleDigests = <String, String>{};
+  for (var dep in transitiveDeps) {
+    var assetId = dep.jsId(jsModuleExtension);
+    moduleDigests[
+            assetId.path.replaceFirst('lib/', 'packages/${assetId.package}')] =
```
jakemac53: I think we need to strip the top level directory off as well, if it doesn't start with `lib/`.

Issue #2022, opened 2019-02-04
Unhandled exception in the AngularDart compiler


afpatmin:
Dart SDK: 2.2.0-dev.1.1
OS: Linux (Ubuntu 18.04.1)
build_runner: 1.2.5
build_web_compilers: 1.1.0

Running webdev serve --live-reload crashes after upgrading build_runner and build_web_compiler dependences. Output (with --verbose flag set)

```
...
[FINEST] Watch:Got modify event for "asset:onboarding_client/.dart_tool/build/098d3ee73e6cc294616d3a2e2c3c81ad/asset_graph.json in /home/patrick/Documents/Clients/Boardon/boardon.io/client
[SEVERE] angular on package:angular_components/src/material_tooltip/tooltip.dart (cached):
Unhandled exception in the AngularDart compiler!

Please report a bug: https://github.com/dart-lang/angular/issues/new
NoSuchMethodError: The getter 'scheme' was called on null.
Receiver: null
Tried calling: scheme
dart:core                                                                       Object.noSuchMethod
package:front_end/src/base/resolve_relative_uri.dart 18:27                      resolveRelativeUri
package:analyzer/src/dart/analysis/file_state.dart 667:21                       FileState._fileForRelativeUri
package:analyzer/src/dart/analysis/file_state.dart 541:24                       FileState.refresh
package:analyzer/src/dart/analysis/file_state.dart 897:12                       FileSystemState.getFileForPath
package:analyzer/src/dart/analysis/driver.dart 949:33                           AnalysisDriver.getSourceKind
package:build_resolvers/src/resolver.dart 83:24                                 AnalyzerResolver.isLibrary
package:build_resolvers/src/resolver.dart 59:56                                 PerActionResolver.isLibrary
package:build                                                                   _DelayedResolver.isLibrary
package:angular/src/source_gen/template_compiler/template_processor.dart 95:10  _resolveReflectables.<fn>
package:angular_compiler/src/analyzer/reflector.dart 208:18                     ReflectableReader._needsInitReflector
package:angular_compiler/src/analyzer/reflector.dart 174:17                     ReflectableReader._resolveNeedsReflector.<fn>
dart:async                                                                      Future.wait
package:angular_compiler/src/analyzer/reflector.dart 173:18                     ReflectableReader._resolveNeedsReflector
dart:async                                                                      _AsyncAwaitCompleter.start
[...]

PR #2024, opened 2019-02-05
ensure restoreAbsolute never returns null

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2022


natebosch:
Ouch, I think I even had to make this fix in the old code path and never applied it to my WIP cl for the migration... Sorry about that!

PR #2030, opened 2019-02-08
Simplify daemon client logging

grouma:
Issue: Critical information is logged to the STDIO prior to the daemon client being fully initialized. We want to provide an easy way for clients to listen for this information and display it to the user.

Changes:
- Remove the `ServerLog` stream from the `BuildDaemonClient`.
- Add an optional `daemonLogHandler` to be passed in during `BuildDaemonClient.connect`.
- Simplify the `BuildDaemonClient.connect` method into a handful of appropriately named helper methods. 
- No longer assumeTty within the daemon command, simply log messages to the STDIO. Note that webdev will handle pretty printing and related logic.
- Rev `package:build_daemon` and `package:build_runner`.

### DIFF
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -19,9 +19,7 @@ import 'data/server_log.dart';
 
 int _existingPort(String workingDirectory) {
   var portFile = File(portFilePath(workingDirectory));
-  if (!portFile.existsSync()) {
-    throw FileSystemException('Unable to read port file.');
-  }
+  if (!portFile.existsSync()) throw MissingPortFileException();
   return int.parse(portFile.readAsStringSync());
 }
 
@@ -122,6 +120,8 @@ class BuildDaemonClient {
         _existingPort(workingDirectory), daemonSerializers, logHandler);
   }
 }
+
+class MissingPortFileException extends Error {}
 
 class OptionsSkew extends Error {}
 
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -19,7 +19,7 @@ import 'data/server_log.dart';
 
 int _existingPort(String workingDirectory) {
   var portFile = File(portFilePath(workingDirectory));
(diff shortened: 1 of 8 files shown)


build_daemon/lib/client.dart
```diff
@@ -17,22 +17,80 @@ import 'data/build_target_request.dart';
 import 'data/serializers.dart';
 import 'data/server_log.dart';
 
-class BuildDaemonClient {
-  IOWebSocketChannel _channel;
-  Serializers _serializers;
+int _existingPort(String workingDirectory) {
+  var portFile = File(portFilePath(workingDirectory));
+  if (!portFile.existsSync()) throw Exception('Unable to read port file.');
```
natebosch: Use a more specific type, create a new type if you need to.

https://api.dartlang.org/stable/2.1.0/dart-core/Exception-class.html

> Creating instances of Exception directly with `new Exception("message")` is discouraged, and only included as a temporary measure during development, until the actual exceptions used by a library are done.

Same thing below.

build_daemon/lib/client.dart
```diff
@@ -17,22 +17,80 @@ import 'data/build_target_request.dart';
...
+  }
+  await sub.cancel();
+}
 
-  final _serverLogStreamController = StreamController<ServerLog>.broadcast();
+bool _isActionMessage(String line) =>
+    line == versionSkew || line == readyToConnectLog || line == optionsSkew;
 
-  BuildDaemonClient._();
+typedef void DaemonLogHandler(ServerLog serverLog);
```
natebosch: can we inline the function types instead of adding a typedef?

build_daemon/lib/client.dart
```diff
@@ -123,6 +121,8 @@ class BuildDaemonClient {
   }
 }
 
+class MissingPortFileException extends Error {}
```
natebosch: why Error and not Exception?

Come to think of it, OptionsSkew and VersionSkew should probably be Exception as well. They are things that may happen outside of the programmer's control and don't indicate a bug.

PR #2032, opened 2019-02-08
Force cascades when possible

natebosch:
Enable the `cascade_invocations` lint and fix voilations.

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -83,20 +83,19 @@ Future<Null> bootstrapDdc(BuildStep buildStep,
       bootstrapId.path,
       from: _context.dirname(dartEntrypointId.path)));
 
-  var bootstrapContent =
-      StringBuffer('$_entrypointExtensionMarker\n(function() {\n');
-  bootstrapContent.write(_dartLoaderSetup(
-      modulePaths,
-      _p.url.relative(appDigestsOutput.path,
-          from: _p.url.dirname(bootstrapId.path))));
-  bootstrapContent.write(_requireJsConfig);
-
   // Strip top-level directory
   var appModuleSource =
       _context.joinAll(_context.split(module.primarySource.path).sublist(1));
 
-  bootstrapContent.write(_appBootstrap(
-      bootstrapModuleName, appModuleName, appModuleScope, appModuleSource));
+  var bootstrapContent =
+      StringBuffer('$_entrypointExtensionMarker\n(function() {\n')
+        ..write(_dartLoaderSetup(
+            modulePaths,
+            _p.url.relative(appDigestsOutput.path,
+                from: _p.url.dirname(bootstrapId.path))))
+        ..write(_requireJsConfig)
+        ..write(_appBootstrap(bootstrapModuleName, appModuleName,
+            appModuleScope, appModuleSource));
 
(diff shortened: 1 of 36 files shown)


natebosch:
Opening this to start a discussion and so we have some concrete diffs to consider. I'm not 100% convinced this is a good lint to enable.

I think some of the changes were good and the lint pushed to better code, some of the cases feel a bit convoluted though.

@jakemac53 @grouma - thoughts?

grouma:
I'm generally indifferent to this change. To be honest I don't really see the benefit. Do you have a particular example where it is beneficial? There are a few cases where it creates a long line which I don't like.

jakemac53:
I guess I would slightly lean towards not adding this lint, just because there is some incremental overhead of conforming to it for very little benefit.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -83,20 +83,19 @@ Future<Null> bootstrapDdc(BuildStep buildStep,
...
-  bootstrapContent.write(_appBootstrap(
-      bootstrapModuleName, appModuleName, appModuleScope, appModuleSource));
+  var bootstrapContent =
+      StringBuffer('$_entrypointExtensionMarker\n(function() {\n')
+        ..write(_dartLoaderSetup(
+            modulePaths,
+            _p.url.relative(appDigestsOutput.path,
+                from: _p.url.dirname(bootstrapId.path))))
+        ..write(_requireJsConfig)
+        ..write(_appBootstrap(bootstrapModuleName, appModuleName,
```
natebosch: Here the lint encouraged me to notice the difference between where it did and didn't make me use cascades. With a slight reordering that doesn't impact behavior the overall intent becomes much more clear because all the writes are together nicely.

build_web_compilers/lib/src/dev_compiler_stack_trace/stack_trace_mapper.dart
```diff
@@ -78,8 +78,7 @@ class LazyMapping extends Mapping {
       var rawMap = _provider(uri);
       if (rawMap != null) {
         var strMap = rawMap is String ? rawMap : _stringify(rawMap);
-        var mapping = parse(strMap) as SingleMapping;
-        mapping
+        var mapping = parse(strMap) as SingleMapping
```
natebosch: cases like this where we were already using cascades, but on the variable rather than within the assignment, are more idiomatic after applying the lint

natebosch:
> Do you have a particular example where it is beneficial?

> for very little benefit.

I added some comments with what I consider to be some of the highlights and lowlights. Some of these are cases that we should have noticed in code reviews, but the lint makes them more obvious.

I think the overhead is pretty small, the fixes are very mechanical and quick when you're not trying to do the entire repo at once. IMO the benefit outweighs the cost.

> There are a few cases where it creates a long line which I don't like.

That is a downside, but we also had cases like that without the lint where we used short cascades anyway. At least with the lint it's consistent and we'll get used to it quicker.

grouma:
Thanks for calling out some of the beneficial examples. I'm still pretty indifferent to this change. Feel free to turn it on unless @jakemac53 feels strongly otherwise.

natebosch:
We synced offline and while none of us feel overwhelmingly one way or the other we'll give it a shot and we always have the option of turning it back off it it becomes too annoying day to day.

PR #2034, opened 2019-02-08
Implement checkOutputs as a Matcher

### LINKED ISSUE #376: Implement `checkOutputs` as a Matcher
We don't currently have unit tests of `checkOutputs` but it is public. Adding tests against code which has it's own `expect` isn't straightforward - it would be a lot easier to test this independently if it was implemented as a Matcher.

The downside is that implementing the Matcher api isn't as easy as the code we have now.

natebosch:
Closes #376

Add a method to create a Matcher against a map of output contents.
Deprecate `checkOutputs`.

This shifts the burden of handling the `mapAssetIds` concept to a more
localized place where it is needed. The implementation is also more
idiomatic by reusing existing matcher concepts like Map equality.

### DIFF
--- a/_test_common/lib/test_phases.dart
+++ b/_test_common/lib/test_phases.dart
@@ -167,14 +167,13 @@ void checkBuild(BuildResult result,
     }
   }
 
-  AssetId mapHidden(AssetId id, String expectedGeneratedDir) =>
-      unhiddenAssets.contains(id)
-          ? AssetId(rootPackage,
-              '.dart_tool/build/$expectedGeneratedDir/${id.package}/${id.path}')
-          : id;
-
   if (status == BuildStatus.success) {
-    checkOutputs(unhiddenOutputs, result.outputs, writer,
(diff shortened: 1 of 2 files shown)


natebosch:
I'd be fine skipping this change and closing the issue as wont fix if we don't like this.

There is a big downside of this implementation where if the asset is expected to be a String the failure message will still print as a list of bytes...

jakemac53:
> There is a big downside of this implementation where if the asset is expected to be a String the failure message will still print as a list of bytes...

Hmm ya I think that would be a deal breaker, can that not be resolved?

natebosch:
> can that not be resolved?

I don't think so. The `actual` printing is done by the test package and doesn't involve the `Matcher` instance. The reason we could get strings printed before was because we were decoding them before the call to `expect`. It's the same reason we don't get nice printing when we use the `decodedMatches()` helper we have instead of comparing against a String directly - seeing a String as the expected value is the only way we know to decode it before the expect instead of after.

I think the UX we have now is good enough so I'll close the original issue.

PR #2035, opened 2019-02-08
Drop some unnecessary typedefs

### LINKED ISSUE #1086: Consider ditching custom typedefs for new Function syntax?
Since we're all on Dart 2 – a nice way to clean things up?

natebosch:
Closes #1086

Remove some typedefs that are not saving enough space or used often
enough to justify a unique name.

Keep typedefs that are giving a significantly shorter or more clear name
to a function type.

Deprecate the `OnDelete` typedef in `build_runner_core` since it is
public, even though it's not used anywhere removing it would technically
be a breaking change.

### DIFF
--- a/build_runner_core/lib/src/asset/writer.dart
+++ b/build_runner_core/lib/src/asset/writer.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:build/build.dart';
 
+@deprecated
 typedef void OnDelete(AssetId id);
 
 abstract class RunnerAssetWriter implements AssetWriter {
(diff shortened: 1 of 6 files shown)


build_runner_core/lib/src/asset/writer.dart
```diff
@@ -6,6 +6,7 @@ import 'dart:async';
 
 import 'package:build/build.dart';
 
+@deprecated
```
kevmoo: Tis nice to give a reason `@Deprecated(...)`

natebosch: nothing is using it, and it's extremely unlikely anyone would use it. This shouldn't have been made public at all.

I'm only deprecating instead of deleting out of an abundance of caution, I don't think we really need a reason.

kevmoo:
ack

PR #2037, opened 2019-02-08
Warn instead of create an output without assets

### LINKED ISSUE #1297: Warn when attempting to create a merged output from a top-level directory that doesn't exist
`pub run build_runner build --output not_here:build` should print a warning and a non-zero exit code.

We _should_ support the case where `not_here` doesn't exist until the build happens though.

natebosch:
Fixes #1297

Since we don't know which assets will exist (only those which may exist)
before the build, save this check until we are outputting the directory.
The build that already happen may be a waste - but it will get reused if
the command is corrected to specify an output from a root that does have
assets in it.

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.0
+
+- Warn when there are no assets to write in a specified output directory.
+
 ## 2.0.2
 
 - Update `build_resolvers` to version `1.0.0`.
--- a/build_runner_core/lib/src/environment/create_merged_dir.dart
+++ b/build_runner_core/lib/src/environment/create_merged_dir.dart
@@ -43,9 +43,7 @@ Future<bool> createMergedOutputDirectories(
   for (var output in outputMap.keys) {
     if (!await _createMergedOutputDir(output, outputMap[output], packageGraph,
         environment, reader, finalizedAssetsView, outputSymlinksOnly)) {
-      _logger.severe('Unable to create merged directory for $output.\n'
-          'Choose a different directory or delete the contents of that '
-          'directory.');
+      _logger.severe('Unable to create merged directory for $output.');
       return false;
     }
   }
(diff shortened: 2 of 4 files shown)


natebosch:
There is some weirdness in the output:

```
[INFO] Deleting previous outputs in `build` completed, took 3ms
[SEVERE] No assets exist in not_a_dir, skipping output
[INFO] Creating merged output dir `build` completed, took 24ms
[SEVERE] Unable to create merged directory for build.
Choose a different directory or delete the contents of that directory.
[SEVERE] Failed after 890ms
```

Note that:

- `Creating merged output dir `build` completed, took 24ms` doesn't really make sense
- `Choose a different directory or delete the contents of that directory.` doesn't make sense, this is no longer the only reason to fail

Do we think these issues are critical? I'm not sure the best way to resolve them...

jakemac53:
Hmm ya this is a bit weird but we would have to change the logTimedAsync method or something to fix it... that is where the `completed` text is coming from.

I think we should remove the `Choose a different directory or delete the contents of that directory.` text and possibly move it earlier if possible.

grouma:
I missed the earlier comment on this one. I'd remove both of the non-relevant info logs. Alternatively, I'd merge some of the information if possible:

```
[INFO] Deleting previous outputs in `build` completed, took 3ms
[SEVERE] No assets exist in not_a_dir, skipping output
[Severe] Creating merged output dir `build` failed, took 24ms
[SEVERE] Failed after 890ms
```
It still would be nice to get the exact reason why the creation failed.

jakemac53:
> [Severe] Creating merged output dir `build` failed, took 24ms

This would require a deeper change, we could try and follow up though. The thing handling this logging is generic and probably shouldn't trap async errors.

natebosch:
OK, I've made some tweaks. Here's what it looks like now when there is an existing directory:

```
[INFO] Caching finalized dependency graph completed, took 132ms

Found existing directory `build` but no manifest file.
Please choose one of the following options:
1 - Leave the directory unchanged and skip writing the build output
2 - Delete the directory and all contents
3 - Leave the directory in place and write over any existing files
1
[SEVERE] Skipped creation of the merged output directory.
[SEVERE] Unable to create merged directory for build.
[SEVERE] Failed after 2.7s
```

Here's what it looks like when there are no matching outputs:

```
[INFO] Caching finalized dependency graph completed, took 137ms
[SEVERE] No assets exist in not_a_dir, skipping output
[SEVERE] Unable to create merged directory for build.
[SEVERE] Failed after 248ms
```

natebosch:
Here's what it looks like when you run and we're unable to prompt:

```
[INFO] Caching finalized dependency graph completed, took 151ms
[SEVERE] Unable to create merged directory at build.
Choose a different directory or delete the contents of that directory.
[SEVERE] Unable to create merged directory for build.
[SEVERE] Failed after 4.6s
```

PR #2038, opened 2019-02-08
Prevent terminals launching on Windows when using the daemon

grouma:
The daemon command is typically launched in a process with mode `detachedWithStdio`. Due to the behavior outlined in https://github.com/dart-lang/sdk/issues/35809 we need to launch nested programs with mode `detachedWithStdio` as well. Also we must run the snapshot directly, otherwise the `.bat` file will create a terminal. Ideally we simply run these snapshots as isolates, but that change can come at a further date. Given the check on `stdin.hasTerminal` this should only impact the `daemon` command and tests.

Towards https://github.com/dart-lang/webdev/issues/111

### DIFF
--- a/build_modules/lib/src/workers.dart
+++ b/build_modules/lib/src/workers.dart
@@ -122,13 +122,11 @@ final dartdevcDriverResource =
 /// Manages a shared set of persistent dartdevk workers.
 BazelWorkerDriver get _dartdevkDriver {
   _dartdevkWorkersAreDoneCompleter ??= Completer<Null>();
-  var packages = p.joinAll([sdkDir, '..', '..', '..', '.packages']);
   return __dartdevkDriver ??= BazelWorkerDriver(
       () => Process.start(
           p.join(sdkDir, 'bin', 'dart'),
           [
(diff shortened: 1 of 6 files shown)


build_modules/lib/src/workers.dart
```diff
@@ -66,8 +69,14 @@ BazelWorkerDriver get _analyzerDriver {
   _analyzerWorkersAreDoneCompleter ??= Completer<Null>();
   return __analyzerDriver ??= BazelWorkerDriver(
       () => Process.start(
-          p.join(sdkDir, 'bin', 'dartanalyzer$_scriptExtension'),
-          ['--build-mode', '--persistent_worker'],
+          p.join(sdkDir, 'bin', 'dart'),
+          [
+            p.join(sdkDir, 'bin', 'snapshots', 'dartanalyzer.dart.snapshot'),
```
natebosch: the fact that we're also running from snapshots now needs to be called out in the commit message

grouma: Done.

PR #2042, opened 2019-02-09
Better explain build_to cache vs source

devoncarew:
- re-order the `build_to: source` and `build_to: cache` recommendations to emphasize the choice between building for packages vs building for applications

### DIFF
--- a/docs/builder_author_faq.md
+++ b/docs/builder_author_faq.md
@@ -3,13 +3,13 @@
 Outputs to `source` will generally be published with the package on pub. **This
 can be risky.** If the generated code depends on _any_ details from the current
 pub solve - that is it reads information from dependencies - then a consumer of
-the package which got different versions of dependencies in their pub solve
+the package which has different versions of dependencies in their pub solve
 **may be broken**. If the generated code imports any libraries, including from
 the package providing the builder, it must only use the API surface area which
 is guaranteed to not break without a major version bump.
 
-Outputs to `cache` will never be published with the package and if they are
-required to compile or run then the build system must be used by every consumer
+Outputs to `cache` will never be published with the package, and if they are
+required to compile or run, then the build system must be used by every consumer
 of the code. **This is always safe**, but may place limitations on the end user.
 Any outputs which are used _during_ a build to produce other outputs, but don't
 need to be compiled or seen by the user, should also be built to `cache`.
--- a/docs/builder_author_faq.md
+++ b/docs/builder_author_faq.md
@@ -1,21 +1,18 @@
 ## When should a Builder build to `cache` vs `source`?
 
-Use `build_to: source` when:
-- The generated code should be `pub publish`ed with the package and used by
-  downstream packages without running a build themselves.
-- The generated code _does not_ depend on details outside of the containing
-  package that might change with a different version of its dependencies. Any
-  details from the _generating_ package, or any other dependency used by the
-  generated code must not change without a major version bump.
+Outputs to `source` will generally be published with the package on pub. **This
+can be risky.** If the generated code depends on _any_ details from the current
+pub solve - that is it reads information from dependencies - then a consumer of
+the package which got different versions of dependencies in their pub solve may
+be broken. If the generated code imports any libraries, including from the
+package providing the builder, it must only use the API surface area which is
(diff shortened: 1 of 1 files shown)


jakemac53:
@natebosch how do you feel about this?

Personally I think the existing first bullet really is the most important thing (whether the code can realistically be published at all).

The note about publishing to pub is unfortunately a bit of a red herring - it sounds like a good idea at first but its very easy to write builders that rely on for instance the exact fields existing in a super class. If that class happens to be from another package which adds new fields in a non-breaking version the generated code is no longer valid.

It is actually very difficult to write a builder which can safely publish its code in all scenarios.

devoncarew:
For context, this was from a related issue, where the relevant point for them was the distinction between a package you intend to publish and an application w/o any dependencies on it.

natebosch:
> the relevant point for them was the distinction between a package you intend to publish and an application w/o any dependencies on it.

Part of the problem here is that builder authors are the only ones who can make the choice today, not the consumers of the builder.

Another problem is that users want to build to the cache for aesthetic instead of functional reasons. Its possible that if we give better support for generating into a separate source directory the users wouldn't care as strongly about it.

kevmoo:
What's the story here?

natebosch:
I think I'm OK with this change. I agree with Jake that the prior order was more in importance order - I think this new order is more pedagogical.

jakemac53:
Imo we should just close this pr, the existing ordering while harder to understand does surface the more important aspects first. I don't think we should prioritize for the incorrect but easier to teach model?

natebosch:
After some offline discussion I think maybe the bullet point format isn't going to be able to express the risk of making the wrong decision while still being approachable. I'm going to take a stab at rewriting this.

natebosch:
@jakemac53 I rewrote as a paragraph. WDUT?

docs/builder_author_faq.md
```diff
@@ -1,21 +1,18 @@
...
-- The generated code _does not_ depend on details outside of the containing
-  package that might change with a different version of its dependencies. Any
-  details from the _generating_ package, or any other dependency used by the
-  generated code must not change without a major version bump.
-- The generated code should be `pub publish`ed with the package and used by
-  downstream packages without running a build themselves.
+Outputs to `source` will generally be published with the package on pub. **This
+can be risky.** If the generated code depends on _any_ details from the current
+pub solve - that is it reads information from dependencies - then a consumer of
+the package which got different versions of dependencies in their pub solve may
```
jakemac53: nit: bold **may be broken**

natebosch: Done

natebosch:
@devoncarew - what do you think of the change?

docs/builder_author_faq.md
```diff
@@ -1,21 +1,18 @@
...
-- The generated code _does not_ depend on details outside of the containing
-  package that might change with a different version of its dependencies. Any
-  details from the _generating_ package, or any other dependency used by the
-  generated code must not change without a major version bump.
-- The generated code should be `pub publish`ed with the package and used by
-  downstream packages without running a build themselves.
+Outputs to `source` will generally be published with the package on pub. **This
+can be risky.** If the generated code depends on _any_ details from the current
+pub solve - that is it reads information from dependencies - then a consumer of
+the package which got different versions of dependencies in their pub solve
```
devoncarew: `which got different` ==> `which has different`?

docs/builder_author_faq.md
```diff
@@ -1,21 +1,18 @@
...
 
-Use `build_to: cache` when:
-- The generated code uses details from outside of the package with the generated
-  code, including from the _generating_ package which may change without a major
-  version bump.
-- The builder generates lots of files that the user might not expect to see.
-- The builder targets use cases for "application" packages that are unlikely to
-  be published rather than packages which will be published and become
-  dependencies.
+Outputs to `cache` will never be published with the package and if they are
```
devoncarew: `package, and if they`

docs/builder_author_faq.md
```diff
@@ -1,21 +1,18 @@
...
+required to compile or run then the build system must be used by every consumer
```
devoncarew: `compile or run, then the`

Issue #2045, opened 2019-02-12
Build runner fails on libraries that indirectly import `dart-ext`


simolus3:
I'm using `Dart VM version: 2.1.0 (Tue Nov 13 18:22:02 2018 +0100) on "linux_x64"` with `build_runner 1.2.3` and `build 1.1.0`. I'm applying `built_value_generator` on a project that includes a file like this (called `vm_database.dart`)
```dart
import 'package:sqlite2/sqlite.dart';
// code that does not use built_value at all
```
`package:sqlite2/sqlite.dart` imports `package:sqlite2/src/native.dart`, which in turn has a
`import 'dart-ext:dart_sqlite';` directive.

When I run `pub run build_runner build --verbose`, I get an error on my file:
```
[INFO] Generating build script completed, took 213ms
[INFO] BuildDefinition:Initializing inputs
[INFO] BuildDefinition:Reading cached asset graph...
[INFO] BuildDefinition:Reading cached asset graph completed, took 44ms

[INFO] BuildDefinition:Checking for updates since last build...
[INFO] BuildDefinition:Checking for updates since last build completed, took 415ms

[INFO] Build:Running build...
[SEVERE] built_value_generator|built_value on lib/vm_database.dart:

Unsupported operation: Cannot resolve dart-ext:dart_sqlite; only "package" and "asset" schemes supported
package:build                                               new AssetId.resolve
package:build_resolvers/src/resolver.dart 357:18            _resolve
package:build_resolvers/src/resolver.dart 229:21            AssetBasedSource.updateDependencies.<fn>
dart:core                                                   Iterable.toSet
package:build_resolvers/src/resolver.dart 231:10            AssetBasedSource.updateDependencies
package:build_resolvers/src/resolver.dart 145:16            AnalyzerResolver._performResolve.processAsset.<fn>
dart:async                                                  _completeOnAsyncReturn
package:pool/pool.dart                                      Pool.withResource
package:build_runner_core/src/asset/file_based.dart 34:39   FileBasedAssetReader.readAsBytes.<fn>
dart:async                                                  _completeOnAsyncReturn
package:pool/pool.dart                                      Pool.withResource
package:build_runner_core/src/asset/file_based.dart 133:26  _fileForOrThrow
package:build_runner_core/src/asset/file_based.dart 33:48   FileBasedAssetReader.readAsBytes
package:build_runner_core/src/asset/build_cache.dart 39:17  BuildCacheReader.readAsBytes
[...]

PR #2046, opened 2019-02-12
explicitly skip dart-ext: uris

jakemac53:
At least partially fixes https://github.com/dart-lang/build/issues/2045

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.0.7
+
+- Explicitly skip dart-ext uris during module creation.
+  - Filed Issue #2047 to track real support for native extensions.
+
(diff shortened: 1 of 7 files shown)


natebosch:
Do you think we need to handle this in `build_modules` as well?
https://github.com/dart-lang/build/blob/11d158038e8a00e0f23b1d8df37650ddac125917/build_modules/lib/src/module_library.dart#L81-L85

jakemac53:
> Do you think we need to handle this in `build_modules` as well?

Looks like we will need to yes, good catch

jakemac53:
@natebosch pushed a commit to skip these uris during module creation as well, and filed https://github.com/dart-lang/build/issues/2047 to track native extension support (but there are no plans to actually work on that unless we have customers for it)

PR #2051, opened 2019-02-14
Fix modules and run E2E tests with stable

grouma:
https://github.com/dart-lang/build/pull/2038 required sdk change https://github.com/dart-lang/sdk/commit/b484aada210c9d47fafdbe1809e8a466ead45ec9#diff-a56982f5d01aac6ee40c43140a933a0c. This change did not show up until the first version after 2.1.0 stable. 

Changes:
- Fix `package:build_modules` to call the dart2js with the old argument (`--library-root`)
- Restrict `package:build_modules` to the stable sdk version
- Only test with stable (will revert in a follow up PR)

In a follow up PR:
- Remove stable switch (maybe keep stable for E2E tests?)
- Revert argument back to `--libraries-spec`
- Update SDK constraint on `package:build_modules`

### DIFF
--- a/build_modules/mono_pkg.yaml
+++ b/build_modules/mono_pkg.yaml
@@ -1,5 +1,5 @@
 dart:
-  - stable 
+  - stable
 
 stages:
   - analyze_and_format:
--- a/build_modules/mono_pkg.yaml
+++ b/build_modules/mono_pkg.yaml
@@ -1,5 +1,5 @@
 dart:
-  - dev
+  - stable 
 
 stages:
   - analyze_and_format:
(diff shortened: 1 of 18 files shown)


grouma:
The SDK constraint needs to change because the dart2js snapshot contained in any SDK version past 2.1.0 does not support `--library-root`

build_modules/mono_pkg.yaml
```diff
@@ -1,5 +1,5 @@
 dart:
-  - dev
+  - stable
```
jakemac53: We should still run on dev I think also

PR #2052, opened 2019-02-14
Update dart2js snapshot arguments

grouma:
- Revert testing back to dev
- Update dart2js snapshot arguments to work with the dev SDK
- Properly restrict the SDK

### DIFF
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -1,11 +1,11 @@
 name: build_modules
-version: 1.0.7+1
+version: 1.0.7+2
 description: Builders for Dart modules
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_modules
 
 environment:
-  sdk: "2.1.0"
+  sdk: ">=2.1.1 <3.0.0"
 
 dependencies:
   analyzer: '>0.30.0 <0.36.0'
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -5,7 +5,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_modules
 
 environment:
-  sdk: ">=2.1.1 <3.0.0"
+  sdk: ">=2.1.1-dev <3.0.0"
 
 dependencies:
   analyzer: '>0.30.0 <0.36.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,55 +11,55 @@ after_failure:
 jobs:
   include:
     - stage: analyze_and_format
-      name: "SDK: stable - DIR: _test - TASKS: dartanalyzer --fatal-infos --fatal-warnings ."
+      name: "SDK: dev - DIR: _test - TASKS: dartanalyzer --fatal-infos --fatal-warnings ."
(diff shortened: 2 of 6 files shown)


natebosch:
We don't consider changing these arguments breaking in the SDK since they are automatically set within the shell scripts, should we worry about being broken again by this?

grouma:
That's a good question. This is an issue we should have been concerned about regardless because we weren't always calling the shell scripts. I believe one of the workers was calling the snapshot directly. That being said, I took a look at the various scripts. Some of them haven't been touched in years. Others were touched 8+ months ago to turn on Dart 2 by default, but weren't touched for a very long time before that. The only one that was edited recently was the Dart2JS script. I think it's safe to assume we won't get a breaking change in the near future, and if we do, we will catch it early given that we are testing with dev.

natebosch:
@sigmundch @jmesserly - do you think it's reasonable for us to run the ddc and dart2js snapshots instead of through the shell script wrappers? Should we anticipate any breaking changes in argument handling that would otherwise be hidden by the wrappers?

sigmundch:
I think it's ok to do it directly from the snapshot.

It is very rare that we change things that require updates to the wrapper. In fact, the wrapper practically only sets the location of the sdk libraries and whether the terminal supports colors for diagnostics. The change from libraries-root to libraries-spec was a rare exception that was related to the migration to the CFE. It is possible that this flag will change one more time (it might become optional because of how dart2js consumes a dill file for the sdk already).

If you think it provides value, we could look into finding a way to run a test of package:build in the main repo, so we can detect any breakages as soon as we make changes.

jakemac53:
> If you think it provides value, we could look into finding a way to run a test of package:build in the main repo, so we can detect any breakages as soon as we make changes.

We run some integration tests as a cron job against edge releases which should be sufficient I think (it won't be caught early, but we should notice it before dev releases).

natebosch:
Given the discussion I'm comfortable with this change. A breaking change should be rare, and as Jake mentions we should catch it in our cron before it gets to stable.

build_modules/pubspec.yaml
```diff
@@ -1,11 +1,11 @@
 name: build_modules
-version: 1.0.7+1
+version: 1.0.7+2
 description: Builders for Dart modules
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_modules
 
 environment:
-  sdk: "2.1.0"
+  sdk: ">=2.1.1-dev <3.0.0"
```
jakemac53: do we know which dev release specifically introduced the change?

grouma: Yes. The first dev release after 2.1.0. See https://github.com/dart-lang/sdk/commit/b484aada210c9d47fafdbe1809e8a466ead45ec9#diff-a56982f5d01aac6ee40c43140a933a0c

jmesserly:
> do you think it's reasonable for us to run the ddc and dart2js snapshots instead of through the shell script wrappers? Should we anticipate any breaking changes in argument handling that would otherwise be hidden by the wrappers?

Practically, we never change DDC's launch scripts. The most recent change was me removing some unnecessary arguments, before that it was changes related to the VM's dart 2 support flag, before that it was changes in 2016 when they were originally added. :)

PR #2053, opened 2019-02-14
Clean up some trailing whitespace

natebosch:


natebosch:
@grouma - https://stackoverflow.com/a/30884298/374798

PR #2054, opened 2019-02-15
allow configuring build script updates via environment

jonahwilliams:
Allows configuring how to do the build script checks via the environment, allowing flutter tools to skip them when running build scripts. 

---

More general topic of discussion: I've drifted towards using build_runner more directly because incorporating it into the flutter_tool doesn't have very high costs in terms of dependencies, and also because it exposes a ton of very useful functionality. In that case, I am not sure the best route to allow configuring this build_script check from build_runner, or whether it should be done at all.

For example, daemon_builder has a ton of useful functionality that I don't really want to duplicate. We would like to find a way to disable the script check in its environment though.

The second commit contains an example of piping an `overrideEnvironment` this through build_runner itself. This could be provided directly via the flutter tool or by generating a slightly different build script.

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -128,12 +128,10 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
     _buildResults.add(daemon.BuildResults((b) => b..results.addAll(results)));
   }
 
-  static Future<BuildRunnerDaemonBuilder> create(
-    PackageGraph packageGraph,
-    List<BuilderApplication> builders,
-    SharedOptions sharedOptions, {
-    OverrideableEnvironment Function(BuildEnvironment) overrideEnvironment
-  }) async {
+  static Future<BuildRunnerDaemonBuilder> create(PackageGraph packageGraph,
+      List<BuilderApplication> builders, SharedOptions sharedOptions,
+      {OverrideableEnvironment Function(BuildEnvironment)
+          overrideEnvironment}) async {
     var expectedDeletes = Set<AssetId>();
     var outputStreamController = StreamController<ServerLog>();
 
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -132,13 +132,13 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
     PackageGraph packageGraph,
     List<BuilderApplication> builders,
     SharedOptions sharedOptions,
-    BuildEnvironment overrideEnvironment,
+    OverrideableEnvironment Function(BuildEnvironment) overrideEnvironment,
   ) async {
     var expectedDeletes = Set<AssetId>();
     var outputStreamController = StreamController<ServerLog>();
 
     var environment = OverrideableEnvironment(
-        overrideEnvironment ?? IOEnvironment(packageGraph,
+        IOEnvironment(packageGraph,
             assumeTty: true,
             // TODO(grouma) - This should likely moved to the build_impl command
             // so that different daemon clients can output to different
@@ -148,6 +148,9 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
         onLog: (record) {
       outputStreamController.add(ServerLog((b) => b.log = record.toString()));
     });
+    if (overrideEnvironment != null) {
+      environment = overrideEnvironment(environment);
+    }
 
     var daemonEnvironment = OverrideableEnvironment(environment,
         writer: OnDeleteWriter(environment.writer, expectedDeletes.add));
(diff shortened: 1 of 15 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -132,12 +132,13 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
     List<BuilderApplication> builders,
     SharedOptions sharedOptions,
+    BuildEnvironment overrideEnvironment,
   ) async {
     var expectedDeletes = Set<AssetId>();
     var outputStreamController = StreamController<ServerLog>();
 
     var environment = OverrideableEnvironment(
-        IOEnvironment(packageGraph,
+        overrideEnvironment ?? IOEnvironment(packageGraph,
```
jonahwilliams: Not sure what the best API would be to combine these two would be.

jakemac53: We could make an optional named arg, which is a factory that takes a BuildEnvironment and returns an OverrideableEnvironment: `{OverrideableEnvironment Function(BuildEnvironment) overrideEnvironment}`?

Then, you could conditionally wrap the environment in an additional layer:

``dart
if (overrideEnvironment != null) {
  environment = overrideEnvironment(original);
}
``

build_runner_core/lib/src/environment/overridable_environment.dart
```diff
@@ -35,9 +44,15 @@ class OverrideableEnvironment implements BuildEnvironment {
     Future<BuildResult> Function(BuildResult result,
             FinalizedAssetsView finalizedAssetsView, AssetReader reader)
         finalizeBuild,
+    Future<BuildScriptUpdates> Function(
+    AssetReader assetReader,
```
jakemac53: Is this dartformatted? We should file a bug if so that doesn't look right.

You could try removing the trailing comma here too, that might make it behave better

build_runner_core/lib/src/environment/io_environment.dart
```diff
@@ -86,6 +88,11 @@ class IOEnvironment implements BuildEnvironment {
     }
     return buildResult;
   }
+
+  @override
+  Future<BuildScriptUpdates> buildScriptUpdates(PackageGraph packageGraph, AssetGraph graph) {
```
jakemac53: We should instead just `extend BuildEnvironment` to pick up the default impl I think

build_runner_core/lib/src/environment/overridable_environment.dart
jonahwilliams: Fixed by removing the trailing comma

jonahwilliams:
Updated to address comment and refactored overrideEnvironment to be a factory.

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -132,13 +132,13 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
     PackageGraph packageGraph,
     List<BuilderApplication> builders,
     SharedOptions sharedOptions,
-    BuildEnvironment overrideEnvironment,
+    OverrideableEnvironment Function(BuildEnvironment) overrideEnvironment,
```
jakemac53: Since this is nullable please make it an optional (named) arg

jakemac53:
@jonahwilliams looks like some files need to be formatted, then travis should hopefully be happy

natebosch:
Sorry for the late response after we've already churned on this...

Have you tried passing the flag `--skip-build-script-check`? Does that accomplish what you want?

jonahwilliams:
Passing in that flag seems to work, on a rebuild it doesn't run every builder again (though it does still warn about the dependency).  If we can skip checks via this route it works for me

```
[+5345 ms] [INFO] Initializing inputs
[   +8 ms] [INFO] Reading cached asset graph...
[ +148 ms] [INFO] Reading cached asset graph completed, took 148ms
[ +235 ms] [INFO] Checking for updates since last build...
[ +666 ms] [INFO] Checking for updates since last build completed, took 666ms
[  +12 ms] [WARNING] build_runner_core|lib/build_runner_core.dart was not found in the asset graph, incremental builds will not work.
[        ]  This probably means you don't have your dependencies specified fully in your pubspec.yaml.
[  +12 ms] [INFO] Running build...
[ +180 ms] [INFO] Running build completed, took 180ms
[        ] [INFO] Caching finalized dependency graph...
[  +87 ms] [INFO] Caching finalized dependency graph completed, took 87ms
[   +3 ms] [INFO] Succeeded after 275ms with 0 outputs (0 actions)
[  +45 ms] running builders... (completed in 6.7s)
[   +1 ms] "flutter generate" took 6,921ms.
[        ] "flutter generate" took 6,921ms.
```

jonahwilliams:
I think that will work for now. I can always come back to this later if need be - thanks all!

PR #2058, opened 2019-02-16
Refactor PackageGraphWatcher

natebosch:
Use `StreamGroup.merge` to combine all the individual node watchers
instead of dropping down to StreamController directly. This saves us
from needing to handle things like cancel forwarding.

- Make the `node` field public on `PackageNodeWatcher` so we can more
  easily `.map()` over the instances and set up the filtering.
- Add a `_nestedPathFilter` helper method to get a closure over the
  nested paths without needing a separate variable higher up.
- Fix the test for nested path filtering. The old test did not exercise
  the filtering code path since it did not send the change through the
  "outer" watcher like what would happen in a real file system.
- Enforce that a single call is made to `watch()` explicitly.
- Remove the `relative` concept from the node watcher since it was
  unused outside of tests.

### DIFF
--- a/build_runner/lib/src/watcher/graph_watcher.dart
+++ b/build_runner/lib/src/watcher/graph_watcher.dart
@@ -41,7 +41,8 @@ class PackageGraphWatcher {
   Stream<AssetChange> watch() {
     assert(!_isWatching);
     _isWatching = true;
-    return logTimedSync(_logger, 'Setting up file watchers', _watch);
+    return LazyStream(
+        () => logTimedSync(_logger, 'Setting up file watchers', _watch));
   }
 
   Stream<AssetChange> _watch() {
--- a/build_runner/lib/src/watcher/graph_watcher.dart
+++ b/build_runner/lib/src/watcher/graph_watcher.dart
@@ -47,8 +47,7 @@ class PackageGraphWatcher {
 
   Stream<AssetChange> _watch() {
     final allWatchers = _graph.allPackages.values
-        .where((node) => !(node.dependencyType == DependencyType.hosted ||
-            node.dependencyType == DependencyType.github))
+        .where((node) => node.dependencyType == DependencyType.path)
         .map(_strategy)
         .toList();
     final filteredEvents = allWatchers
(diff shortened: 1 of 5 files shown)


natebosch:
There is a behavior change here in that the listeners are started on the call to `watch()` while before they were started when the stream was listened to. In practice this shouldn't make much difference since we don't call `watch()` unless we intend to listen to the stream.

natebosch:
I take that back - it's really easy to retain the old behavior with a `LazyStream` so I may as well do that.

build_runner/lib/src/watcher/graph_watcher.dart
```diff
@@ -38,57 +39,34 @@ class PackageGraphWatcher {
...
-      onListen: () {
-        subscriptions = logTimedSync(
-          _logger,
-          'Setting up file watchers',
-          () => _watch(controller),
-        );
-      },
-      onCancel: () {
-        for (final subscription in subscriptions) {
-          subscription.cancel();
```
grouma: Does lazy stream cancel subscriptions upon cancel?

build_runner/lib/src/watcher/graph_watcher.dart
```diff
@@ -38,57 +39,34 @@ class PackageGraphWatcher {
...
-        // TODO: Consider a faster filtering strategy.
-        if (nestedPackages.any((path) => event.id.path.startsWith(path))) {
-          return;
-        }
-        sink.add(event);
-      }));
-    });
+  Stream<AssetChange> _watch() {
+    final allWatchers = _graph.allPackages.values
+        .where((node) => !(node.dependencyType == DependencyType.hosted ||
```
jakemac53: Not directly related, but we may want to consider changing this to just `node.dependencyType == DependencyType.path`? For instance the there are more dependency types, like SDK (for the flutter package).

build_runner/lib/src/watcher/graph_watcher.dart
natebosch: Yes

build_runner/lib/src/watcher/graph_watcher.dart
natebosch: Good idea

PR #2060, opened 2019-02-19
Fix link for build_reslovers README

gendalf7771:


gendalf7771:
I've added alternate email to sign CLI

Issue #2063, opened 2019-02-21
Bump dependency on uuid to 2.0.0


srawlins:
uuid 2.0.0 was released, and build_daemon depends on uuid. It's only a breaking change for build_daemon in that `v1()` now returns a String, so the casts in `daemon_test.dart` are now unnecessary.

PR #2064, opened 2019-02-21
Depend on latest uuid

grouma:
Closes https://github.com/dart-lang/build/issues/2063

### DIFF
--- a/build_daemon/pubspec.yaml
+++ b/build_daemon/pubspec.yaml
@@ -26,4 +26,4 @@ dev_dependencies:
   mockito: ^4.0.0
   test: ^1.3.3
   test_descriptor: ^1.1.1
-  uuid: ^1.0.3
+  uuid: ^2.0.0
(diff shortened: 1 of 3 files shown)


build_daemon/pubspec.yaml
```diff
@@ -26,4 +26,4 @@ dev_dependencies:
   mockito: ^4.0.0
   test: ^1.3.3
   test_descriptor: ^1.1.1
-  uuid: ^1.0.3
```
jakemac53: Given that all the casts are actually in test code only, we could leave this at `>=1.0.3 <3.0.0`

srawlins: Does this package have enforced static analysis in presubmits? bumping to uuid 2.0.0 should introduce UNNECESSARY_CAST Hints.

grouma: It does have unnecessary cast hints. Since this is only a dev dep though it should be fine to bump the minimum.

PR #2068, opened 2019-02-26
Daemon client shutdown

grouma:
Support closing a Daemon client. Fix related issue where shutting down the build daemon mid build would cause the process to hang.

### DIFF
--- a/build_daemon/lib/src/managers/build_target_manager.dart
+++ b/build_daemon/lib/src/managers/build_target_manager.dart
@@ -22,21 +22,22 @@ class BuildTargetManager {
 
   bool Function(BuildTarget, Iterable<WatchEvent>) shouldBuild;
 
-  bool get isEmpty => _buildTargets.isEmpty;
-
-  Set<BuildTarget> get targets => _buildTargets.keys.toSet();
-
-  Set<WebSocketChannel> channels(BuildTarget target) => _buildTargets[target];
-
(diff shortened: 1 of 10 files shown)


build_daemon/lib/src/managers/build_target_manager.dart
```diff
@@ -22,22 +22,23 @@ class BuildTargetManager {
 
   bool Function(BuildTarget, Iterable<WatchEvent>) shouldBuild;
 
-  bool get isEmpty => _buildTargets.isEmpty;
-
-  Set<BuildTarget> get targets => _buildTargets.keys.toSet();
-
-  Set<WebSocketChannel> channels(BuildTarget target) => _buildTargets[target];
```
grouma: Just did an auto sort here.

build_daemon/lib/src/server.dart
```diff
@@ -30,7 +30,6 @@ class Server {
 
   HttpServer _server;
   DaemonBuilder _builder;
-  final _channels = Set<WebSocketChannel>();
```
grouma: This was never used :/

build_daemon/lib/src/managers/build_target_manager.dart
jakemac53: Try to keep these in separate commits if you can, it makes reviewing a lot easier :D

PR #2072, opened 2019-03-02
Allow configuring Dart sdk location in KernelBuilder

jonahwilliams:
Currently the KernelBuilder allows configuring the location of the platform dill file via `sdkKernelPath`, but always joins this path with location of the Dart SDK used to execute the builder. For flutter (and eventually flutter web), the platform dill is located in another directory. 

I think we could use relative paths in the `sdkKernelPath` to work around this for now, but it will be more resilient to allow configuring the SDK directory as well.

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.0.9
 
-- Allow configuring the dart SDK directory in the `KernelBuilder` builder.
+- Allow configuring the platform SDK directory in the `KernelBuilder` builder.
 
 ## 1.0.8
 
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.0.9
+
+- Allow configuring the dart SDK directory in the `KernelBuilder` builder.
+
 ## 1.0.8
 
 - Don't follow `dart.library.isolate` conditional imports for the DDC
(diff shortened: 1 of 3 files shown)


jakemac53:
I am a bit concerned this is misleading, because it doesn't change the location of the sdk generally as people might expect. For instance the kernel worker itself is still going to run from the shared `sdkDir` location.

I would like to understand a bit more why the current `sdkDir` doesn't work also - that _should_ be pointing at the root of the SDK. If flutter has a different layout so that doesn't point at the root of the SDK, then that is a bigger problem that we could address more generally I think.

jonahwilliams:
In this case, its not so much that the files are outside the dart SDK, but that they're in another dart SDK.

The dart SDK included by flutter under `bin/cache/dart-sdk` is unmodified, this is important because the flutter tool runs on the vm and needs to work normally. It does not contain any references to flutter, and is missing the platform dill files (and eventually the web platform dill files, libraries.json, and web sources.

The flutter SDK is built in a different location (/bin/cache/artifacts/engine/common/flutter_patched_sdk/) instead of being merged with the dart SDK. This contains the flutter platform dill and outline files. Additionally, adding `dart:ui` web support will involve modifying parts of the SDK which runs the risk of breaking parts of the normal SDK, such as the compiled dartdevc sdk, libraries.dart, et cetera.

jakemac53:
Thanks for the explanation, that clarifies things.

I wonder if we should name the variable `platformSdk` or something along those lines, just to make it more obvious that this is only for selecting certain files that are specific to the platform, and not a general override?

jonahwilliams:
platformSdk sounds good to me, I'll update the PR and provide flutter as an example in the documentation.

PR #2084, opened 2019-03-14
Clear statics during hot restart

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/205

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -155,6 +155,8 @@ $_initializeTools
   var bootstrap = {
       hot\$onChildUpdate: function(childName, child) {
         if (childName === "$appModuleSource") {
+          // Clear static caches.
+          dart_sdk.dart.hotRestart();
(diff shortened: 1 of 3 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -155,6 +155,8 @@ $_initializeTools
   var bootstrap = {
       hot\$onChildUpdate: function(childName, child) {
         if (childName === "$appModuleSource") {
+          // Clear static caches.
+          dart_sdk.dart.hotRestart();
```
natebosch: How long has `hotRestart()` existed in the SDK? Do we need to bump the minimum constraint in the pubspec?

jakemac53: good question, @jmesserly do you know when this was added?

jakemac53: I did some digging it looks like 2.1 was the stable release so I pinned to that

PR #2085, opened 2019-03-14
Main injection hook

grouma:
- Add a new marker to inject code before the application's main method is called
- Update the comments to be a little more clear

We need a hook right before the main is called so that we have a guarantee that DDC is set up. We also will be able to pause the application before it actually begins executing.

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -425,11 +425,11 @@ requirejs.onResourceLoad = function (context, map, depArray) {
 };
 ''';
 
-/// Marker comment used by build_runner to identify the entrypoint file,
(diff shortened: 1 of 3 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -151,6 +151,7 @@ define("$bootstrapModuleName", ["$moduleName", "dart_sdk"], function(app, dart_s
   dart_sdk.dart.setStartAsyncSynchronously(true);
   dart_sdk._isolate_helper.startRootIsolate(() => {}, []);
 $_initializeTools
```
natebosch: Want to fix the indentation while you're in here?

grouma: Done.

Issue #2080, opened 2019-03-12
Daemon command should support multiple output options


grouma:
Right now if you try to connect to an existing daemon with different options and error is thrown. This is desired for most options. However, providing multiple `output` options is a non-issue in most cases.

cc @liuming0

PR #2091, opened 2019-03-19
Build Runner Core Output Locations

### LINKED ISSUE #2092: Outputting a subdirectory should not stop from outputting full package
Our behavior today for `pbr build -o everthing -o web:foo` is broken. We will be missing the compiled `js` output under `everything/test/` for instance. This is because we output `test`, but never built it - `buildDirs` was `web` so we skipped targets under `test` and likely only output source files and non-optional outputs. `pbr serve -o everything web:8080` is likely similarly broken.

Having a `--ouput` that wants to emit the entire package should also mean to build the entire package.

grouma:
- Update `run` method to now take an optional list of `BuildDirectory`
  - Each `BuildDirectory` contains an optional `OutputLocation`
- Validate `OutputLocation`s in `createMergedOutputDirectories`
- Add corresponding tests and rev version

Towards https://github.com/dart-lang/build/issues/2080
Fixes Fixes #2092


build_runner_core/lib/src/environment/build_environment.dart
```diff
@@ -40,11 +40,18 @@ abstract class BuildEnvironment {
   /// completes, it will expire afterwords since it can no longer guarantee a
   /// consistent state.
   ///
+  /// The [outputMap] is a map of input paths to a list of output paths. If the
+  /// input path is an empty string it means all input paths. If the list of
+  /// output paths is empty it means only output to the default location.
```
natebosch: What does "default location" mean here? If the paths are empty there won't be any output in the sense that we use the term elsewhere. We don't consider the generated directory under `.dart_tool` to be an output.

build_runner_core/lib/src/environment/build_environment.dart
```diff
@@ -40,11 +40,18 @@ abstract class BuildEnvironment {
...
   /// By default this returns the original result.
   ///
   /// Any operation may be performed, as determined by environment.
-  Future<BuildResult> finalizeBuild(BuildResult buildResult,
-          FinalizedAssetsView finalizedAssetsView, AssetReader assetReader) =>
+  Future<BuildResult> finalizeBuild(
+          BuildResult buildResult,
+          FinalizedAssetsView finalizedAssetsView,
+          AssetReader assetReader,
+          Map<String, List<String>> outputMap) =>
```
natebosch: I dislike the name `Map` here. We can use it if we can't think of anything better, but in general an argument or variable that repeats it's type is a code smell.

`outputDirectories`? `outputLocations`?

natebosch: Should this be `Map<String, Set<String>>`?

build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -39,17 +39,39 @@ Future<bool> createMergedOutputDirectories(
...
+          environment, reader, finalizedAssetsView, outputSymlinksOnly)) {
+        _logger.severe('Unable to create merged directory for $output.');
+        return false;
+      }
     }
   }
   return true;
 }
 
+Map<String, List<String>> _outputToInput(Map<String, List<String>> outputMap) {
```
natebosch: What do you think about this?

```dart
var conflictingOutputs = _findConflicts(outputLocations);
...

Set<String> _findConflicts(Map<String, List<String>> outputLocations) {
  final seen = <String>{};
  final conflicts = <String>{};
  for (var location in outputLocations.values.expand((l) => l)) {
    if (seen.contains(location)) {
      conflicts.add(location);
    } else {
      seen.add(location);
    }
  }
  return conflicts;
}
```

Or if you want to golf it:

```dart
Set<String> _findConflicts(Map<String, List<String>> outputLocations) {
  final seen = <String>{};
  final conflicts = <String>{};
  for (var location in outputLocations.values.expand((l) => l)) {
    (seen.contains(location) ? conflicts : seen).add(location);
  }
  return conflicts;
}
```

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -85,21 +85,23 @@ class TestCommand extends BuildRunnerCommand {
     try {
       _ensureBuildTestDependency(packageGraph);
       options = readOptions();
-      var outputMap = (options.outputMap ?? {})..addAll({tempPath: null});
+      var outputLocations = (options.outputLocations ?? {})
+        ..addAll({
+          '': {tempPath}
```
natebosch: What if `outputLocations` already had an empty key?

I think this needs the same handling as other places.

```dart
outputLocations[''] ??= {};
outputLocations[''].add(tempPath);
```

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -59,6 +59,7 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
   @override
   Future<void> build(
       Set<BuildTarget> targets, Iterable<WatchEvent> fileChanges) async {
+    var defaultTargets = targets.cast<DefaultBuildTarget>();
```
jakemac53: What about marking `targets` as covariant instead and then just overriding the type. This method does not accept anything but a DefaultBuildTarget so that would be give better static checking. 

If not, then I would not call this `defaultTargets`, which sounds like some sort of fallback set of targets instead of the ones that were passed in.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -315,7 +306,8 @@ Map<String, String> _parseOutputMap(ArgResults argResults) {
         throw ArgumentError.value(
             option, '--output', 'Input root can not be nested');
       }
-      result[output] = split.first;
+      result[split.first] ??= <String>{};
```
jakemac53: nit: `result.putIfAbsent(split.first, () => {}).add(output)`

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -85,21 +85,23 @@ class TestCommand extends BuildRunnerCommand {
     try {
       _ensureBuildTestDependency(packageGraph);
       options = readOptions();
-      var outputMap = (options.outputMap ?? {})..addAll({tempPath: null});
+      var outputLocations = options.outputLocations ?? {};
+      outputLocations[''] ??= <String>{};
```
jakemac53: nit: putIfAbsent here too

build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -39,17 +39,34 @@ Future<bool> createMergedOutputDirectories(
...
     }
   }
   return true;
 }
 
+Set<String> _conflicts(Map<String, Set<String>> outputLocations) {
+  final seen = <String>{};
+  final conflicts = <String>{};
+  for (var location in outputLocations.values.expand((l) => l)) {
+    (seen.contains(location) ? conflicts : seen).add(location);
```
jakemac53: nit: This reads better imo: `if (!seen.add(location)) conflicts.add(location);`

natebosch: ah, I completely forgot that `.add` returns a bool. Yeah this is way better.

build_runner/lib/src/daemon/daemon_builder.dart
grouma: Looks like this works well. This requires a change in `package:build_daemon` so I will follow up with that change.

build_runner/test/entrypoint/run_test.dart
```diff
@@ -146,7 +145,7 @@ main() {
     expect(await runSingleBuild(command, args), ExitCode.success.code);
     expectOutput('web/main.dart.js', exists: false);
     expectOutput('test/hello_test.dart.js', exists: true);
-  });
+  }, solo: true);
```
jakemac53: note this probably snuck in by accident

grouma: Ahh missed that one. Good thing we have lints.

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -184,7 +184,9 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
               change,
               builder.assetGraph,
               buildOptions,
-              sharedOptions.outputLocations.keys?.isNotEmpty == true,
+              sharedOptions.outputLocations.keys.any((input) =>
+                      sharedOptions.outputLocations[input].isNotEmpty) ==
+                  true,
```
natebosch: the `== true` is no longer necessary since `.any()` should never return null.

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -85,8 +85,11 @@ class TestCommand extends BuildRunnerCommand {
     try {
       _ensureBuildTestDependency(packageGraph);
       options = readOptions();
-      var outputLocations = options.outputLocations ?? <String, Set<String>>{};
-      outputLocations.putIfAbsent('test', () => <String>{}).add(tempPath);
+      var outputLocations = (options.outputLocations ?? <String, Set<String>>{})
+        // Build test by default.
```
natebosch: My impression was that if `''` was present in the keys that we would build everything. Having to include `test` here worries me.

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -211,7 +212,9 @@ class _SingleBuild {
     var watch = Stopwatch()..start();
     var result = await _safeBuild(updates);
     var optionalOutputTracker = OptionalOutputTracker(
-        _assetGraph, _outputLocations.keys.toList(), _buildPhases);
+        _assetGraph,
+        _outputLocations.keys.where((k) => k.isNotEmpty).toList(),
```
natebosch: This feels wrong to me. I'd expect something like

`_outputLocations.keys.contains('') ? [] : _outputLocations.keys.toList(),`

_test_common/lib/test_phases.dart
```diff
@@ -90,7 +88,7 @@ Future<BuildResult> testBuilders(
   bool enableLowResourcesMode = false,
   Map<String, Map<String, dynamic>> builderConfigOverrides,
   bool verbose = false,
-  List<String> buildDirs,
+  List<BuildTarget> buildTargets,
```
natebosch: Can this be a `Set` to handle some deduping ahead of time?

build_runner_core/lib/src/generate/build_target.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class BuildTarget {
```
natebosch: Both of these need an override for `hashCode` and `operator ==` now.

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -85,21 +85,27 @@ class TestCommand extends BuildRunnerCommand {
     try {
       _ensureBuildTestDependency(packageGraph);
       options = readOptions();
-      var outputMap = (options.outputMap ?? {})..addAll({tempPath: null});
+      print(tempPath);
```
kevmoo: Is print here a dev left-over?

grouma: Yup. I'll remove.

build_runner_core/lib/src/generate/build_target.dart
```diff
@@ -0,0 +1,64 @@
...
```
jakemac53: I didn't realize this was in build_runner_core (I thought it was the daemon?). That makes me less comfortable about the duplicate name with build_config...

grouma: I'm open to changing it. I do really like BuildTarget as it is very descriptive even if the term is overloaded. What about `BuildDestination`?

grouma: @natebosch thoughts?

build_runner_core/lib/src/generate/build_target.dart
```diff
@@ -0,0 +1,64 @@
...
+  int get hashCode {
+    var hash = 0;
+    hash = _hashCombine(hash, directory.hashCode);
+    hash = _hashCombine(hash, outputLocation.hashCode);
+    return _hashComplete(hash);
+  }
+}
+
+class OutputLocation {
+  final String output;
```
natebosch: How about `path` here?

build_runner_core/lib/src/generate/build_target.dart
natebosch: `BuildDirectory` would be consistent with `buildDirs`...

grouma: Noted. I'll update the PR in a few.

grouma:
Renamed `BuildTarget` to `BuildDirectory`. Also updated `OutputLocation` to have the member `path`. PTAL.

build_runner_core/lib/src/generate/build_directory.dart
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+class BuildDirectory {
+  final String directory;
+  final OutputLocation outputLocation;
+  BuildDirectory(this.directory, this.outputLocation);
```
jakemac53: nit: move `outputLocation` to an optional named arg since its optional, and document the behavior for when its null

_test_common/lib/test_phases.dart
```diff
@@ -4,21 +4,19 @@
 import 'dart:async';
 
 import 'package:build/build.dart';
-import 'package:build_config/build_config.dart';
+import 'package:build_config/build_config.dart' hide BuildTarget;
```
natebosch: `hide` is no longer necessary?

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -8,7 +8,7 @@ import 'dart:io';
 import 'package:build/build.dart';
 import 'package:build_daemon/daemon_builder.dart';
 import 'package:build_daemon/data/build_status.dart' as daemon;
-import 'package:build_daemon/data/build_target.dart';
+import 'package:build_daemon/data/build_target.dart' as daemon;
```
natebosch: no longer necessary?

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -191,9 +182,10 @@ class ServeOptions extends SharedOptions {
       }
     }
 
-    var outputMap = _parseOutputMap(argResults);
-    var buildDirs = _buildDirsFromOutputMap(outputMap)
-      ..addAll(serveTargets.map((t) => t.dir));
+    var buildDirs = _parseBuildDirs(argResults);
+    for (var target in serveTargets) {
+      buildDirs.add(BuildDirectory(target.dir));
```
natebosch: [nit] FWIW I preferred the old `..addAll(serveTargets.map)` since it's about collecting, not side effects.

We can revisit this though when we'll be able to do:

```dart
var buildDirs = {
    ..._parseBuildDirs(argResults),
    ...serveTargets.map((t) => BuildDirectory(t.dir)),
};
```

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -283,30 +271,32 @@ Map<String, Map<String, dynamic>> _parseBuilderConfigOverrides(
   return builderConfigOverrides;
 }
 
-/// Returns a map of output directory to root input directory to be used
-/// for merging.
+/// Returns a list of BuiltTargets with output information.
```
natebosch: BuiltTargets -> BuildDirectory.

[nit] Remove the "list of" from the wording.

`/// Returns the BuildDirectories requested for output arguments.`

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -4,7 +4,7 @@
 import 'dart:async';
 
 import 'package:build/build.dart';
-import 'package:build_config/build_config.dart';
+import 'package:build_config/build_config.dart' as config;
```
natebosch: Drop the `as config`?

build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -39,32 +40,66 @@ Future<bool> createMergedOutputDirectories(
...
+      }
     }
   }
   return true;
 }
 
+Set<String> _conflicts(Set<BuildDirectory> buildDirs) {
+  final seen = <String>{};
+  final conflicts = <String>{};
+  var outputLocations = buildDirs.fold<List<String>>([], (a, b) {
```
natebosch: `var outputLocations = buildDirs.map((d) => d.outputLocation).where((d) => d != null));`

PR #2093, opened 2019-03-20
Build everything when outputting everything

### LINKED ISSUE #2092: Outputting a subdirectory should not stop from outputting full package

natebosch:
Fixes #2092 
Fixes https://github.com/dart-lang/build/issues/2095

### DIFF
--- a/build_runner/lib/src/entrypoint/test.dart
+++ b/build_runner/lib/src/entrypoint/test.dart
@@ -82,10 +82,16 @@ class TestCommand extends BuildRunnerCommand {
         .absolute
         .uri
         .toFilePath();
+    var testPath = (Directory.fromUri(Uri.parse(p.join(tempPath, 'test')))
+          ..createSync())
+        .absolute
+        .uri
+        .toFilePath();
+
     try {
       _ensureBuildTestDependency(packageGraph);
       options = readOptions();
-      var outputMap = (options.outputMap ?? {})..addAll({tempPath: 'test'});
+      var outputMap = (options.outputMap ?? {})..addAll({testPath: 'test'});
       var result = await build(
         builderApplications,
         deleteFilesByDefault: options.deleteFilesByDefault,
@@ -107,6 +113,12 @@ class TestCommand extends BuildRunnerCommand {
         stdout.writeln('Skipping tests due to build failure');
         return result.failureType.exitCode;
       }
+
+      // Set up the temp directory as `package:test` expects.
+      Link(p.join(tempPath, '.packages'))
+          .createSync(p.join(testPath, '.packages'));
+      Link(p.join(tempPath, 'packages'))
+          .createSync(p.join(testPath, 'packages'), recursive: true);
 
       return await _runTests(tempPath);
     } on _BuildTestDependencyError catch (e) {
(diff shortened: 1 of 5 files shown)


jakemac53:
I was talking with Gary about this a bit this morning - the existing behavior does allow some things that aren't allowed any more with this change.

Specifically, after this change you can no longer build a single top level dir (`test` for example) without hoisting that directory to the top level. The `test` command currently relies on this behavior.

Previously, `-o` would only cause additional directories to be built if you asked explicitly for it with the `<dir-to-build>:<output-dir>` syntax.

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -108,13 +107,20 @@ class TestCommand extends BuildRunnerCommand {
...
+          .createSync(p.join(testDir.path, 'packages'), recursive: true);
+
+      return await _runTests(tempDir.absolute.uri.toFilePath());
     } on _BuildTestDependencyError catch (e) {
       stdout.writeln(e);
       return ExitCode.config.code;
     } finally {
       // Clean up the output dir.
-      await Directory(tempPath).delete(recursive: true);
+      await Directory(tempDir.absolute.uri.toFilePath())
```
natebosch: `await tempDir.delete(recursive: true);`

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -77,15 +77,14 @@ class TestCommand extends BuildRunnerCommand {
   Future<int> run() async {
     SharedOptions options;
     // We always run our tests in a temp dir.
-    var tempPath = Directory.systemTemp
-        .createTempSync('build_runner_test')
-        .absolute
-        .uri
-        .toFilePath();
+    var tempDir = Directory.systemTemp.createTempSync('build_runner_test');
+    var testDir = Directory.fromUri(Uri.parse(p.join(tempDir.path, 'test')));
```
natebosch: `var testDirPath = p.join(tempDir.path, 'test');`

Or _maybe_

`var testDirPath = p.join(tempDir.absolute.uri.toFilePath(), 'test');`

build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -140,7 +140,7 @@ Future<AssetId> _writeAsset(AssetId id, Directory outputDir, String root,
     var outputId = AssetId(packageGraph.root.name, assetPath);
     try {
       if (symlinkOnly) {
-        await Link(_filePathFor(outputDir, id)).create(
+        await Link(_filePathFor(outputDir, outputId)).create(
```
natebosch: I think this should go out on it's own first.

grouma:
Closing in favor of https://github.com/dart-lang/build/pull/2091

Issue #2094, opened 2019-03-20
Daemon mode doesn't detect build script updates


jakemac53:
Using the latest webdev I updated my pubspec to depend on a different version of build_web_compilers but it didn't notice the change.

I killed webdev and restarted and then it saw the change and rebuilt as expected.

It looks like `watch` mode handles this type of update by watching the `.packages` file.

PR #2103, opened 2019-03-25
Daemon shutdown

grouma:
- Support stopping the daemon with a notification

Towards https://github.com/dart-lang/build/issues/2094

### DIFF
--- a/build_daemon/test/daemon_test.dart
+++ b/build_daemon/test/daemon_test.dart
@@ -41,7 +41,7 @@ void main() {
       testWorkspaces.add(workspace);
       var daemon = Daemon('$workspace');
       expect(daemon.tryGetLock(), isTrue);
-      await daemon.start(<String>{}, DaemonBuilder(), Stream.empty());
+      await daemon.start(Set<String>(), DaemonBuilder(), Stream.empty());
       expect(daemon.onDone, completes);
       await daemon.stop();
     });
--- a/build_daemon/test/daemon_test.dart
+++ b/build_daemon/test/daemon_test.dart
@@ -6,9 +6,6 @@ import 'dart:convert';
 import 'dart:io';
 
 import 'package:build_daemon/constants.dart';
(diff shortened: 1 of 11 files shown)


build_daemon/test/managers/build_target_manager_test.dart
```diff
@@ -30,6 +30,24 @@ void main() {
...
+      ..addBuildTarget(targetB, channelB);
+    expect(manager.allChannels, containsAll([channelA, channelB]));
+    expect(manager.allChannels.length, 2);
+
+    manager.removeChannel(channelA);
+
+    expect(manager.allChannels, contains(channelB));
+    expect(manager.allChannels.length, 1);
+  });
+
```
jakemac53: A test for the shutdown command would be good as well.

build_daemon/test/daemon_test.dart
```diff
@@ -39,6 +36,18 @@ void main() {
       }
     });
 
+    test('can be stopped', () async {
+      var workspace = uuid.v1();
+      testWorkspaces.add(workspace);
+      var daemon = Daemon('$workspace');
+      expect(daemon.tryGetLock(), isTrue);
+      var timeout = Duration(seconds: 30);
```
jakemac53: 30 seconds seems like a lot, if it waits that long to actually shut down I would reduce this to make the test run faster (5 seconds?)

grouma: It shuts down immediately. This is just in case the shutdown logic doesn't work which actually should be handled by the test infra so I'm gonna remove it.

jakemac53: kk

jakemac53:
@grouma is this ready?

grouma:
Hmm I thought I would need to merge master again but no conflicts are showing up. I'm going to merge again regardless, rebuild, and submit shortly.

PR #2104, opened 2019-03-25
Minor code cleanup

grouma:
Clean up code since I was trigger happy on merging https://github.com/dart-lang/build/pull/2091

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -20,7 +20,8 @@ import 'package:build_runner/src/watcher/graph_watcher.dart';
 import 'package:build_runner/src/watcher/node_watcher.dart';
 import 'package:build_runner_core/build_runner_core.dart'
     hide BuildResult, BuildStatus;
-import 'package:build_runner_core/build_runner_core.dart' as core;
+import 'package:build_runner_core/build_runner_core.dart' as core
+    show BuildStatus;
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -18,14 +18,16 @@ import 'package:build_runner/src/watcher/collect_changes.dart';
 import 'package:build_runner/src/watcher/delete_writer.dart';
 import 'package:build_runner/src/watcher/graph_watcher.dart';
 import 'package:build_runner/src/watcher/node_watcher.dart';
-import 'package:build_runner_core/build_runner_core.dart';
+import 'package:build_runner_core/build_runner_core.dart'
+    hide BuildResult, BuildStatus;
+import 'package:build_runner_core/build_runner_core.dart' as core;
```
natebosch: how about `as core show BuildResult, BuildStatus;`

That way we don't accidentally have the same type with and without a prefix anywhere.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -271,7 +271,7 @@ Map<String, Map<String, dynamic>> _parseBuilderConfigOverrides(
   return builderConfigOverrides;
 }
 
-/// Returns a list of BuiltTargets with output information.
+/// Returns BuiltTargets with output information.
```
natebosch: BuildDirectory not BuiltTarget 

I also still think that something like "parsed from output arguments." would make this more clear.

Issue #2098, opened 2019-03-21
If running a "build script snapshot" with a "failed to instantiate builder" error, try rebuilding


kevmoo:
Situation

- Make a mistake in `lib/builder.dart`
- run build
- Build fails with `Failed to instantiate builder for json_serializable with configuration` error (fine)
- Fix mistake in `lib/builder.dart`
- Rerun build
- Build fails with `Failed to instantiate builder for json_serializable with configuration` error – because it's running the snapshot!

Work-around: run clean first

jakemac53:
There are a whole class of edge cases like this actually that mostly only come up for builder authors but we should fix in a general way.

One potential solution - if there is an existing snapshot but no serialized asset graph then delete the snapshot and recreate it. We can't check if it was up to date anyways in this case, and all the edge cases I know of are related to the first build failing for one reason or another, resulting in no asset graph being serialized.

natebosch:
If we invalidate the snapshot do we also delete the asset graph? I think you can still hit a problem if you build successfully so you have an asset graph file, then change something that would cause a failure that doesn't invalidate the snapshot and get stuck in this loop...

kevmoo:
`lib/builder.dart` – sorry. Fixed.

PR #2105, opened 2019-03-25
Don't re-use snapshots if we have no asset graph

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2098

cc @kevmoo

### DIFF
--- a/_test/test/build_integration_test.dart
+++ b/_test/test/build_integration_test.dart
@@ -6,8 +6,8 @@
 import 'dart:io';
 
 import 'package:build_runner_core/src/util/constants.dart';
-import 'package:test/test.dart';
 import 'package:path/path.dart' as p;
+import 'package:test/test.dart';
 
 import 'common/utils.dart';
 
@@ -100,7 +100,7 @@ void main() {
       await File(assetGraph).delete();
 
       var nextBuild = await runBuild();
-      print(nextBuild.stdout);
+      printOnFailure(nextBuild.stdout.toString());
       expect(
           nextBuild.stdout.split('\n'),
           containsAllInOrder([
--- a/_test/test/build_integration_test.dart
+++ b/_test/test/build_integration_test.dart
@@ -100,7 +100,6 @@ void main() {
       await File(assetGraph).delete();
 
       var nextBuild = await runBuild();
-      printOnFailure(nextBuild.stdout.toString());
       expect(
           nextBuild.stdout.split('\n'),
           containsAllInOrder([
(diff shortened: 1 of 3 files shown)


_test/test/build_integration_test.dart
```diff
@@ -92,5 +93,23 @@ void main() {
...
       expect(dartSource.existsSync(), true);
     });
+
+    test('Re-snapshots if there is no asset graph', () async {
+      var assetGraph = assetGraphPathFor(p.url
+          .join('.dart_tool', 'build', 'entrypoint', 'build.dart.snapshot'));
+      await File(assetGraph).delete();
+
+      var nextBuild = await runBuild();
+      print(nextBuild.stdout);
```
kevmoo: oops – print?

_test/test/build_integration_test.dart
```diff
@@ -92,5 +93,23 @@ void main() {
       await runBuild(trailingArgs: ['--output', 'build']);
       expect(dartSource.existsSync(), true);
     });
+
+    test('Re-snapshots if there is no asset graph', () async {
+      var assetGraph = assetGraphPathFor(p.url
+          .join('.dart_tool', 'build', 'entrypoint', 'build.dart.snapshot'));
+      await File(assetGraph).delete();
```
natebosch: This looks non-hermetic....

_test/test/build_integration_test.dart
natebosch: `printOnFailure`?

build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -116,7 +116,16 @@ Future<int> generateAndRun(List<String> args, {Logger logger}) async {
 /// Returns zero for success or a number for failure which should be set to the
 /// exit code.
 Future<int> _createSnapshotIfMissing(Logger logger) async {
+  var assetGraphFile = File(assetGraphPathFor(scriptSnapshotLocation));
   var snapshotFile = File(scriptSnapshotLocation);
+
+  /// If we failed to serialize an asset graph for the snapshot, then we don't
```
natebosch: double slash, this isn't a doc comment

_test/test/build_integration_test.dart
grouma: We run a build in the setup so this should not be an issue.

_test/test/build_integration_test.dart
```diff
@@ -92,5 +93,23 @@ void main() {
...
+
+      var nextBuild = await runBuild();
+      printOnFailure(nextBuild.stdout.toString());
```
natebosch: Turns out we always do this anyway. This can be removed.
https://github.com/dart-lang/build/blob/708c302227c8fb254550883aaae25236cb8c29cc/_test/test/common/utils.dart#L63-L64

PR #2110, opened 2019-03-27
add lower bound sdk contraint analyzer tests

jakemac53:
This should fail for build_runner_core, will follow up with a fix for that here soon (probably just remove the set literals first)

### DIFF
--- a/build_config/mono_pkg.yaml
+++ b/build_config/mono_pkg.yaml
@@ -6,9 +6,9 @@ stages:
     - group:
         - dartfmt: sdk
         - dartanalyzer: --fatal-infos --fatal-warnings .
-    - dartanalyzer: --fatal-infos --fatal-warnings .
+    - dartanalyzer: --fatal-warnings .
       dart:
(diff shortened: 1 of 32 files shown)


build_config/mono_pkg.yaml
```diff
@@ -6,6 +6,9 @@ stages:
     - group:
         - dartfmt: sdk
         - dartanalyzer: --fatal-infos --fatal-warnings .
+    - dartanalyzer: --fatal-infos --fatal-warnings .
+      dart:
+        - 2.0.0-dev.54
```
kevmoo: Just bump this and the pubspec to 2.0.0

jakemac53: We can do that separately, I want to have this tested at what is actually published

kevmoo: ack

kevmoo: use `2.0.0-dev.54.0` 😄

PR #2117, opened 2019-03-29
Log the output from summary steps

natebosch:
Analyzer output can be seen when building with `--verbose`.


natebosch:
It's at `log.fine` because _every_ step seems to have at least some output :

```
[FINE] build_modules|dartdevc on test/sub-dir/subdir_test.dartdevc.module:Analyzing file:///test/sub-dir/subdir_test.dart|/tmp/scratch_spaceDDYVHC/test/sub-dir/subdir_test.dart...
```

I'm not sure if having that output always makes sense or not, even with `--verbose`, since it's pretty noisy.

jakemac53:
I think this is fine - it will be a bit noisy on first build but incremental builds should be less noisy and its kind of useful to see all the modules being analyzed after a change.

PR #2118, opened 2019-03-30
Allow configuring SDK summary locations for Linked, Unlinked, DartDevc builders

jonahwilliams:
For building on the web, flutter will require an analyzer summary which contains dart:ui. This adds a BuilderOptions configuration of the SDK summary location for the Linked and Unlinked summary builders - since the location may not be known ahead of time to support `--local-engine` configurations. For the DevCompilerBuilder, we can add a similar platformSdk constructor parameter to what has been added for the kernel builders.

Follow up question:
 - Do we still need to pass the default analysis options files to the analyzer? They only contain the strong mode declaration, and I would rather not require a dev dependency on build_modules.

Fixes https://github.com/dart-lang/sdk/issues/36346

### DIFF
--- a/build_modules/lib/builders.dart
+++ b/build_modules/lib/builders.dart
@@ -13,7 +13,7 @@ typedef _BuilderFactory = Builder Function(BuilderOptions options);
 
 Builder moduleLibraryBuilder(_) => const ModuleLibraryBuilder();
 
-/// A builder config used to configure the location of the dart SDK summary.
+/// A builder config for the location of the dart SDK analyzer summary.
 ///
 /// This is used by the [UnlinkedSummaryBuilder] and [LinkedSummaryBuilder].
 const String dartSdkSummaryConfig = 'dart-sdk-summary';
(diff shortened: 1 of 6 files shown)


jakemac53:
> * Do we still need to pass the default analysis options files to the analyzer? They only contain the strong mode declaration, and I would rather not require a dev dependency on build_modules.

Yes I think so - otherwise analyzer ends up trying to enforce your project specific lints on all your dependencies. As far as the build is concerned we only want to enforce the basic language requirements.

Note that kernel doesn't support lints etc anyways, so with the switch to kernel/ddk this won't be an issue.

jakemac53:
Went ahead and approved - I don't think the issues I mentioned need to be blockers

jonahwilliams:
Sorry for the delay on this - For now I am mostly concerned with the ability to run flutter tests as part of the un-forking process, and only exposed via some repo specific tooling for flutter developers. I would imagine that we will switch to kernel by the time we have an alpha ready.

I'm also okay with holding off on this until we're actually ready to start un-forking. It is possible that doesn't happen before kernel is ready to go.

jakemac53:
Note that we actually deleted the summary builders now - and are publishing the new kernel builder stuff soon. At head, kernel is all that is currently supported.

jonahwilliams:
Awesome, closing then!

PR #2119, opened 2019-04-01
Fix exponential blowup in the kernel builder impl

jakemac53:
Specifically the problem this logic is solving is that any dependency which is provided as sources instead of kernel is essentially a "poison pill" node - all parents of those modules must also be provided as sources and not dill files.

The previous implementation worked around this in a weird way, checking all transitive deps (of each transtive dep) for package cycles. This was both inefficient and relied upon knowledge of the module structure and how it deals with package cycles which wasn't ideal.

The new logic handles this while collecting the transitive deps initially - it tracks module parents and any time it finds a module that has to be provided via source it marks all transitive parents as also being provided by source. It also bails early if the parent is already marked, preventing exponential behavior during marking.

### DIFF
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -3,6 +3,7 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
 import 'dart:convert';
 import 'dart:io';
 
@@ -101,18 +102,12 @@ Future<void> _createKernel(
 
   File packagesFile;
 
-  {
-    var transitiveDeps = await module.computeTransitiveDependencies(buildStep);
+  await buildStep.trackStage('CollectDeps', () async {
     var transitiveKernelDeps = <AssetId>[];
     var transitiveSourceDeps = <AssetId>[];
 
-    await Future.wait(transitiveDeps.map((dep) => _addModuleDeps(
-        dep,
-        module,
-        transitiveKernelDeps,
-        transitiveSourceDeps,
-        buildStep,
-        outputExtension)));
+    await _addModuleDeps(module, transitiveKernelDeps, transitiveSourceDeps,
+        buildStep, outputExtension);
 
     var allAssetIds = Set<AssetId>()
       ..addAll(module.sources)
@@ -124,7 +119,7 @@ Future<void> _createKernel(
 
     _addRequestArguments(request, module, transitiveKernelDeps, dartSdkDir,
         sdkKernelPath, outputFile, packagesFile, summaryOnly);
-  }
+  });
 
   // We need to make sure and clean up the temp dir, even if we fail to compile.
   try {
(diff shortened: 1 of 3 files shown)


build_modules/lib/src/kernel_builder.dart
```diff
@@ -148,28 +143,68 @@ Future<void> _createKernel(
...
+  final toCrawl = Queue.of([root.primarySource]);
+  final nodes = {root.primarySource: _ModuleNode(root, true)};
+
+  // Helper for any time we encounter a module which is provided via source and
+  // not kernel.
+  //
+  // For each of those we have to ensure that all parents of that module are
+  // provided via source as well.
+  void Function(AssetId) updateParents;
+  updateParents = (AssetId nodeId) {
```
natebosch: This doesn't look like it's reassigned so It doesn't need to be a variable.

```dart
void updateParents(AssetId nodeId) {
  ...
}
```

build_modules/lib/src/kernel_builder.dart
```diff
@@ -148,28 +143,68 @@ Future<void> _createKernel(
...
+
+  // Builds the module graph, the `sourceOnly` property of nodes gets
+  // incrementally updated so we need to wait until we are done to actually
+  // provide the deps.
+  while (toCrawl.isNotEmpty) {
+    var moduleId = toCrawl.removeLast();
+    var node = nodes[moduleId];
+    for (final dep in node.module.directDependencies) {
+      var depNode = nodes[dep];
+      if (depNode == null) {
```
natebosch: [optional] might match intent better as `var depNode = nodes.putIfAbsent(dep, () {`

build_modules/lib/src/kernel_builder.dart
```diff
@@ -148,28 +143,68 @@ Future<void> _createKernel(
...
+        parentNode.sourceOnly = true;
+        parentNode.parents.forEach(updateParents);
+      }
+    }
+  };
+
+  // Builds the module graph, the `sourceOnly` property of nodes gets
+  // incrementally updated so we need to wait until we are done to actually
+  // provide the deps.
+  while (toCrawl.isNotEmpty) {
```
natebosch: [optional] It's plausible we could accomplish this with https://pub.dartlang.org/documentation/graphs/latest/graphs/crawlAsync.html and still crawl up the `parents` to mark them each time we find a parent relationship.

build_modules/lib/src/kernel_builder.dart
jakemac53: Because it recursively calls itself it has to be, kind of annoying but :man_shrugging:

jakemac53: nevermind I guess I am just crazy :man_facepalming:

build_modules/lib/src/kernel_builder.dart
jakemac53: Unfortunately I couldn't because it does async work but isn't inserting a `Future<Module>`

build_modules/lib/src/kernel_builder.dart
jakemac53: I think it would end up feeling a bit weird - feels better to be more explicit given how specialized this is.

PR #2122, opened 2019-04-01
Add shared Module Resources

natebosch:
Extract the behavior of `_CleanMetaModuleCache` into a shared class and
use it consistently when reading in Module or MetaModule instances from
assets.

Any time we read a `Module` or `MetaModule` from a json serialized asset
we consider them to be immutable and safe to share for the duration of a
build. We have a consistent pattern of deserialization and construction.

- Add `DecodingCache` to handle the logic of caching and invalidating
  instances of classes that are serialized to an asset.
- Add a single resource for each of `Module` and `MetaModule` so that
  they can be used consistently.
- Update the places where we were json decoding and constructing these
  classes to all use the shared resources.

### DIFF
--- a/build_modules/lib/src/module_cache.dart
+++ b/build_modules/lib/src/module_cache.dart
@@ -10,23 +10,38 @@ import 'package:build/build.dart';
 import 'meta_module.dart';
 import 'modules.dart';
 
-final metaModuleCache = Resource<DecodingCache<MetaModule>>(
-    () => DecodingCache((m) => MetaModule.fromJson(m)),
-    dispose: (c) => c.dispose());
+final metaModuleCache = DecodingCache.resource((m) => MetaModule.fromJson(m));
 
-final moduleCache = Resource<DecodingCache<Module>>(
-    () => DecodingCache((m) => Module.fromJson(m)),
-    dispose: (c) => c.dispose());
+final moduleCache = DecodingCache.resource((m) => Module.fromJson(m));
 
+/// A cache of objects decoded from written assets suitable for use as a
+/// [Resource].
+///
+/// Instances that are decoded will be cached throughout the duration of a build
+/// and invalidated between builds. Instances that are shared through the cache
+/// should be treated as immutable to avoid leaking any information which was
+/// not loaded from the underlying asset.
 class DecodingCache<T> {
+  /// Create a [Resource] which can decoded instances of [T] serialized via json
+  /// to assets.
+  static Resource<DecodingCache<T>> resource<T>(
+          T Function(Map<String, dynamic>) _fromJson) =>
+      Resource<DecodingCache<T>>(() => DecodingCache._(_fromJson),
+          dispose: (c) => c._dispose());
+
   final _cached = <AssetId, Future<Result<T>>>{};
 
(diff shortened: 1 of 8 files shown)


natebosch:
@jakemac53 - one thing that makes me a little nervous is that these classes _aren't_ actually immutable, but we treat them as if they are.

I wonder if we should explore making them immutable for real.

build_modules/lib/src/module_cache.dart
```diff
@@ -0,0 +1,40 @@
...
+
+final metaModuleCache = Resource<DecodingCache<MetaModule>>(
+    () => DecodingCache((m) => MetaModule.fromJson(m)),
+    dispose: (c) => c.dispose());
+
+final moduleCache = Resource<DecodingCache<Module>>(
+    () => DecodingCache((m) => Module.fromJson(m)),
+    dispose: (c) => c.dispose());
+
+class DecodingCache<T> {
```
jakemac53: nit: add some doc comments describing this class

build_modules/lib/src/module_cache.dart
```diff
@@ -0,0 +1,40 @@
...
+  final _cached = <AssetId, Future<Result<T>>>{};
+
+  final T Function(Map<String, dynamic>) _construct;
+
+  DecodingCache(this._construct);
+
+  void dispose() => _cached.clear();
+
+  Future<T> find(AssetId id, AssetReader reader) async {
```
jakemac53: document that this returns null if `id` can't be read

build_modules/lib/src/module_cache.dart
```diff
@@ -0,0 +1,40 @@
...
+    dispose: (c) => c.dispose());
+
+final moduleCache = Resource<DecodingCache<Module>>(
+    () => DecodingCache((m) => Module.fromJson(m)),
+    dispose: (c) => c.dispose());
+
+class DecodingCache<T> {
+  final _cached = <AssetId, Future<Result<T>>>{};
+
+  final T Function(Map<String, dynamic>) _construct;
```
jakemac53: nit: `_fromJson`?

build_modules/lib/src/module_cache.dart
natebosch: Done.

build_modules/lib/src/module_cache.dart
natebosch: Done.

build_modules/lib/src/module_cache.dart
natebosch: Done.

build_modules/lib/src/module_cache.dart
```diff
@@ -10,31 +10,46 @@ import 'package:build/build.dart';
...
-  DecodingCache(this._construct);
+  DecodingCache._(this._fromJson);
 
-  void dispose() => _cached.clear();
+  void _dispose() => _cached.clear();
 
+  /// Find and deserialize a [T] stored in [id].
+  ///
+  /// If the asset at [id] is unreadable the returned future will resolve to
+  /// null. If the isntance is cached it will not be decoded again, but the
```
jakemac53: nit: `null` and `instance`

PR #2124, opened 2019-04-02
Make kernel/ddk the default, use new/faster kernel apis

jakemac53:
- Pass '--reuse-compiler-result' and '--use-incremental-compiler' along with input digests to the kernel worker and ddc worker
- Add the kernel builder for the ddc platform to build_web_compilers (we could add this to build_modules?)
- Change the ddc builder to use kernel by default in build_web_compilers
- Various minor kernel fixes


build_web_compilers/lib/src/common.dart
```diff
@@ -18,8 +18,7 @@ String defaultAnalysisOptionsArg(ScratchSpace scratchSpace) =>
 
 // TODO: better solution for a .packages file, today we just create a new one
 // for every kernel build action.
-Future<File> createPackagesFile(
-    Iterable<AssetId> allAssets, ScratchSpace scratchSpace) async {
```
grouma: I guess `scratchSpace` was never used? Is there a lint we can turn on for that?

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -186,16 +214,31 @@ Future _createDevCompilerModule(
...
   // status code if something failed. Today we just make sure there is an output
   // JS file to verify it was successful.
+  var message = response.output.replaceAll('${scratchSpace.tempDir.path}/', '');
   if (response.exitCode != EXIT_CODE_OK || !jsOutputFile.existsSync()) {
-    var message =
-        response.output.replaceAll('${scratchSpace.tempDir.path}/', '');
     throw DartDevcCompilationException(jsId, '$message}');
   } else {
+    if (message.isNotEmpty) {
+      log.info(message);
```
grouma: Will this be noisy? Also you can probably fit this on one line.

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -186,16 +214,31 @@ Future _createDevCompilerModule(
...
+    }
     // Copy the output back using the buildStep.
     await scratchSpace.copyOutput(jsId, buildStep);
     if (debugMode) {
-      await scratchSpace.copyOutput(
-          module.jsSourceMapId(jsSourceMapExtension), buildStep);
+      /// We need to modify the sources in the sourcemap to remove the custom
```
natebosch: [nit] block comment not doc comment

build_modules/lib/src/kernel_builder.dart
```diff
@@ -237,19 +238,43 @@ void _addRequestArguments(
...
   ]);
+  request.inputs.add(Input()
+    ..path = '${Uri.file(p.join(sdkDir, sdkKernelPath))}'
+    ..digest = [0]);
 
   // Add all summaries as summary inputs.
+  //
+  // Also request the digests for each and add those to the inputs.
+  //
+  // This runs synchronously and the digest futures will be awaited later on.
```
natebosch: what about something like this:

```dart
  var inputs = await Future.wait(transitiveKernelDeps.map((id) async {
    var relativePath = p.url.relative(scratchSpace.fileFor(id).uri.path,
        from: scratchSpace.tempDir.uri.path);

    return Input()
      ..path = '$multiRootScheme:///$relativePath'
      ..digest = await reader.digest(id);
  }));
  request.arguments.addAll(inputs
      .map((i) => '--input-${summaryOnly ? 'summary' : 'linked'}=${i.path}'));
  request.inputs.addAll(inputs);
```

build_web_compilers/lib/src/common.dart
natebosch: It's a tough lint to write - there would be false positives when you write a function that will be used as a tear-off and must fit a certain signature.

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -186,16 +214,31 @@ Future _createDevCompilerModule(
   // TODO(jakemac53): Fix the ddc worker mode so it always sends back a bad
   // status code if something failed. Today we just make sure there is an output
   // JS file to verify it was successful.
+  var message = response.output.replaceAll('${scratchSpace.tempDir.path}/', '');
   if (response.exitCode != EXIT_CODE_OK || !jsOutputFile.existsSync()) {
-    var message =
-        response.output.replaceAll('${scratchSpace.tempDir.path}/', '');
     throw DartDevcCompilationException(jsId, '$message}');
```
natebosch: [out of scope] `message` is already a String right? Maybe remove the extra wrapping in a string literal here?

build_web_compilers/lib/src/common.dart
grouma: Yeah I thought about that use case but if we had a pattern of naming unused parameters with underscores that could resolve the issue?

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: ddc does not output anything under normal circumstances

build_modules/lib/src/kernel_builder.dart
jakemac53: done, much cleaner for sure :+1:

build_modules/CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## 1.1.0-dev
 
 - Update the kernel worker to pass input digests, along with
-  --reuse-compiler-result and --use-incremental-compiler.
+  `--reuse-compiler-result` and -`-use-incremental-compiler`.
```
grouma: [nit] missed a -

build_modules/lib/src/kernel_builder.dart
```diff
@@ -236,18 +237,27 @@ void _addRequestArguments(
...
     '--exclude-non-sources',
     summaryOnly ? '--summary-only' : '--no-summary-only',
+    '--libraries-file',
+    p.join(sdkDir, 'lib', 'libraries.json'),
+    '--reuse-compiler-result',
+    '--use-incremental-compiler',
   ]);
+  request.inputs.add(Input()
+    ..path = '${Uri.file(p.join(sdkDir, sdkKernelPath))}'
+    ..digest = [0]);
```
natebosch: Maybe add a comment that this file will never change without invalidating the entire worker

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -117,19 +121,33 @@ Future _createDevCompilerModule(
     request.arguments.add('--no-source-map');
   }
 
-  // Add all the linked summaries as summary inputs.
+  // Add linked summaries as summary inputs.
+  //
+  // Also request the digests for each and add those to the inputs.
+  //
+  // This runs synchronously and the digest futures will be awaited later on.
```
natebosch: Once we get rid of the `useKernel` branches we should be able to clean this up with the same pattern as above...

jakemac53: yep, will do

PR #2125, opened 2019-04-02
Make Module instances immutable

natebosch:
This makes it safer to share instances across a build. We were already
treating them as immutable once they were read from a serialized asset.

- Wrap the collections in `UnmodifiableSetView` for safety. Maintain the
  behavior of copying them.
- Remove the `merge` instance method in favor a static method merging an
  `List<Module>`.
- Update the code calling `merge` to track lists of modules and merge
  them all at the last moment.

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -55,10 +55,10 @@ Module _moduleForComponent(
   return Module(primaryId, sources, directDependencies, platform, isSupported);
 }
 
-Map<AssetId, Module> _entryPointModules(
+Map<AssetId, List<Module>> _entryPointModules(
         Iterable<Module> modules, Set<AssetId> entrypoints) =>
     Map.fromIterable(modules.where((m) => m.sources.any(entrypoints.contains)),
-        key: (m) => (m as Module).primarySource);
+        key: (m) => (m as Module).primarySource, value: (m) => [m as Module]);
 
 /// Gets the local (same top level dir of the same package) transitive deps of
 /// [module] using [assetsToModules].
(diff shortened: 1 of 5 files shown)


build_modules/lib/src/meta_module.dart
```diff
@@ -55,10 +55,10 @@ Module _moduleForComponent(
   return Module(primaryId, sources, directDependencies, platform, isSupported);
 }
 
-Map<AssetId, Module> _entryPointModules(
+Map<AssetId, List<Module>> _entryPointModules(
```
jakemac53: This seem weird that the values are lists of modules - at least deserving of a comment if the idea is that they are module groups that will be added to later on? Or maybe rename it?

build_modules/lib/src/meta_module.dart
```diff
@@ -144,18 +144,19 @@ List<Module> _mergeModules(Iterable<Module> modules, Set<AssetId> entrypoints) {
     if (entrypointIds.length > 1) {
       var mId = (entrypointIds.toList()..sort()).map((m) => m.path).join('\$');
       if (mergedModules.containsKey(mId)) {
```
jakemac53: nit: use putIfAbsent

build_modules/lib/src/modules.dart
```diff
@@ -25,6 +26,26 @@ part 'modules.g.dart';
...
+  static Module merge(List<Module> modules) {
+    assert(modules.isNotEmpty);
+    if (modules.length == 1) return modules.single;
+    final allSources = Set.of(modules.expand((m) => m.sources));
+    final allDependencies = Set.of(modules.expand((m) => m.directDependencies))
+      ..removeAll(allSources);
+    final primarySource =
+        allSources.reduce((a, b) => a.compareTo(b) < 0 ? a : b);
+    return Module(primarySource, allSources, allDependencies,
+        modules.first.platform, modules.first.isSupported,
```
jakemac53: This was broken before but `isSupported` should actually be `modules.every((m) => m.isSupported)`

build_modules/pubspec.yaml
```diff
@@ -33,3 +34,7 @@ dev_dependencies:
     path: test/fixtures/a
   b:
     path: test/fixtures/b
+
+dependency_overrides:
```
jakemac53: cruft?

natebosch: Necessary because there is a `dev_dependency` on `build_vm_compilers` which is presently incompatible with this new version bump.

PR #2132, opened 2019-04-03
Print logs on failure by default

natebosch:
When a test build fails print the log messages to make it easier to
understand the failure.


natebosch:
Still needs the extra wrapper because `printOnFailure` takes a `String` and we need a function that takes a `LogRecord`.

PR #2134, opened 2019-04-03
Fix sourcemaps with kernel

jakemac53:
Please forgive all the commits here - I couldn't easily do a clean rebase so I just did a merge - will squash when merging.

Fixes https://github.com/dart-lang/sdk/issues/36236

- Fixes the inline source maps to strip the schemes from the sources
- Removes a ddc specific hack that was appending `.ddc` to module names before looking up the source map - this breaks kernel. (will remove entirely in a followup that drops ddc support)
- Cleaned up how the stack trace mapper app is built, following the same strategy as the injected client in webdev (you can now run `pub run build_runner watch web` in `build_web_compilers` and get a decent dev cycle).


build_web_compilers/build.yaml
```diff
@@ -68,6 +81,15 @@ builders:
         compiler: dart2js
     applies_builders:
       - build_web_compilers|dart2js_archive_extractor
+  _stack_trace_mapper_copy:
```
natebosch: This is going to pollute generated build scripts for everyone. Not a huge impact but I wonder if we should avoid it?

Should we make "private" builders a first class concept?

Should we create a separate package for this in the mono_repo which is a `dev_dependency` instead?

build_web_compilers/tool/copy_builder.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import 'package:build/build.dart';
+
+/// Factory for the build script.
+Builder copyBuilder(_) => _CopyBuilder();
+
+/// Copies the [_stackTraceMapperJs] file to [_stackTraceMapperCopyJs].
+class _CopyBuilder extends Builder {
+  @override
+  Map<String, List<String>> get buildExtensions => {
```
natebosch: [optional] 

```dart
@override
final buildExtensions = {
  _stackTraceMapperJs.path: [_stackTraceMapperCopyJs.path],
};
```

build_web_compilers/build.yaml
jakemac53: It doesn't actually pollute them - the builder is not even importable outside this project.

natebosch: Ah I forgot about that setup.

PR #2141, opened 2019-04-04
Drop support for building with analyzer summaries

natebosch:
- Remove the `useKernel` flag and all branches where it was false.
- Remove `build_root_app_summary` config.
- Rename some variables from "summary" to "kernel".

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -4,6 +4,7 @@
   analyzer summaries.
 - Increased the upper bound for `package:analyzer` to `<0.37.0`.
 - Removed support for `build_root_app_summary` configuration option.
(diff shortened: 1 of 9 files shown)


build_web_compilers/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Update to run DDC in kernel mode, and consume kernel outlines instead of
   analyzer summaries.
 - Increased the upper bound for `package:analyzer` to `<0.37.0`.
+- Removed support for `build_root_app_summary` configuration option.
```
jakemac53: We should mention that you can't build using analyzer summaries any more as well (technically the builder was public)

natebosch: Done

PR #2142, opened 2019-04-04
Stop writing compilation error files

natebosch:


natebosch:
cc @liuming0 - is it feasible to drop the dependency on these files from bolt?

liuming0:
Last time when I check, there is no other way to tie a build error message to a suite, so this is why I am using this.
Is there an alternative that I can try it out?

natebosch:
I'm not sure if we're ready to add new dependencies on the error handling code right now, so unfortunately I don't think we'll be able to remove these files in this release.

@liuming0 - no change for now - bolt can continue to read these files as it is today.

PR #2145, opened 2019-04-04
only build valid entrypoints in build_web_compilers and build_vm_compilers

jakemac53:
This is blocked on https://github.com/dart-lang/test/pull/1011


build_modules/lib/src/errors.dart
```diff
@@ -102,3 +104,36 @@ Future<String> _missingImportMessage(
   var lineInfo = parsed.lineInfo.getLocation(import.offset);
   return '`$import` from $sourceId at $lineInfo';
 }
+
+/// An [Exception] that is thrown when there are some unsupported modules.
+class UnsupportedModulesException implements Exception {
```
natebosch: I think this one is also readable as `UnsupportedModules` and the suffix is noise.

build_modules/lib/src/errors.dart
```diff
@@ -102,3 +104,36 @@ Future<String> _missingImportMessage(
...
+
+  Stream<ModuleLibrary> exactLibraries(AssetReader reader) async* {
+    for (var module in unsupportedModules) {
+      for (var source in module.sources) {
+        var libraryId = source.changeExtension(moduleLibraryExtension);
+        ModuleLibrary library;
+        if (await reader.canRead(libraryId)) {
+          library = ModuleLibrary.deserialize(
+              libraryId, await reader.readAsString(libraryId));
+        } else {
```
natebosch: why would we need this branch?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -108,10 +125,14 @@ Future<void> bootstrapDdc(BuildStep buildStep) async {
...
 /// Ensures that all transitive js modules for [module] are available and built.
+///
+/// Throws an [UnsupportedModulesException] exception if there are any
+/// unsupported modules.
 Future<List<Module>> _ensureTransitiveModules(
     Module module, AssetReader reader) async {
   // Collect all the modules this module depends on, plus this module.
-  var transitiveDeps = await module.computeTransitiveDependencies(reader);
+  var transitiveDeps = await module.computeTransitiveDependencies(reader,
+      throwIfUnsupported: true);
```
natebosch: where is `computeTransitiveDependencies` used without `throwIfUnsupported`?

I wonder if we should change the check to not use exceptiosn?

```dart
var transitiveDeps = await module.computeTransitiveDependencies(reader);
var unsupported = transitiveDeps.where((d) => !d.isSupported).toList();
if (unsupported.isNotEmpty) {
  // log warning
}
```

build_modules/lib/src/errors.dart
jakemac53: For part files - we don't output a `.library` file. However i just realized we are skipping them anyways so I will change the branch to a `continue` instead.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: The main reason I use the exception is its a nice place to hang off utilities for finding the actual `ModuleLibrary` instances (the modules are not useful on their own really).

jakemac53: Oh and also it makes it easier to handle errors at the top level and bail out - otherwise for instance this method can't make the overall builder bail out without making a more convoluted return type or returning null or something gross.

jakemac53:
@natebosch this should be unblocked now - please re-review

Issue #590, opened 2017-11-08
Enforce `build_extensions` config from `build.yaml` to be accurate


natebosch:
We don't currently read this section of the config since it all happens based on the `Builder` instances. We only really require this section for `dazel` since it's needed statically in the bazel BUILD files.

If we want to (eventually) advertise compatibility across build systems we may want to enforce accuracy here so we can have some confidence that a config that works with `build_runner` will also work with `dazel`.

Example from @kevmoo:

- Just hit weirdness where the Builder was build.g.part and the yaml config had build.g.dart. (https://github.com/dart-lang/build/issues/1686)

matanlurey:
+1. Definitely we should document that it's expected to be in sync, and we will enforce in future.

natebosch:
We discussed this a little bit. There is some nuance to this decision.

We have two places this is configured:

1. static config
2. Dart code

In both `build_runner` and `bazel` (2) is a *limit* - this will always be true.

In `bazel` (1) is a *minimum* (and `build_runner` ignores it). `bazel` uses this to decide what is visible to dependencies.

If we want to add all the enforcement of `bazel` we'd also want to introduce `default_content` and require that every Asset is either generated, or has default content.
We'd also need to decide whether the default content should go ahead and get created (potentially wasteful) or ignored (potentially different behavior if a dependency tries to read it in a later builder).
Supporting `default_content` would also require some refactoring in `build_runner` to work like it does in `_bazel_codegen`.

In `bazel` we don't really care that we end up with a lot of useless files, in `build_runner` we might.

matanlurey:
Does this overlap a bit with the work required for `transient` files for build_runner?

natebosch:
I don't believe this overlaps with work to support `transient` - other than transient outputs must exist in "Dart code config" and *not* in "static config"

matanlurey:
But the longer term goal still being to be consistent, right?

natebosch:
I think we'd like to push for consistency, but some of this is a UX decision that @kevmoo might make differently from me. My ideal scenario is to be able to tell builder authors "If it works with build_runner today it will work with bazel tomorrow"

The questions are whether we want to start with those guarantees/restrictions, and how much waste we're willing to tolerate to enforce them.

matanlurey:
Are the restrictions considered cumbersome? I guess I don't understand _why not_.

natebosch:
The restrictions are potentially inefficient.

[...]

PR #2149, opened 2019-04-05
Add a doctor command to check build extensions

natebosch:
Towards #590

Adds a `doctor` command which can check, but not yet enforce, that build
extensions are accurately defined in `build.yaml`.

This will let us start experimenting to see how well the validation
works without spamming warnings for existing users.

### DIFF
--- a/build_runner/lib/src/entrypoint/doctor.dart
+++ b/build_runner/lib/src/entrypoint/doctor.dart
@@ -1,0 +1,123 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:build/build.dart';
+import 'package:build_config/build_config.dart';
+import 'package:build_runner_core/build_runner_core.dart';
+import 'package:build_runner_core/src/generate/phase.dart';
+import 'package:io/io.dart';
+import 'package:logging/logging.dart';
+
+import '../logging/std_io_logging.dart';
+import '../package_graph/build_config_overrides.dart';
+import 'base_command.dart';
+
+/// A command that validates the build environment.
+class DoctorCommand extends BuildRunnerCommand {
+  @override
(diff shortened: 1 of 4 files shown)


natebosch:
Output for the `_test` directory:

```
[INFO] Generating build script completed, took 231ms
[WARNING] Deleted previous snapshot due to missing asset graph.
[INFO] Creating build script snapshot... completed, took 8.1s
[WARNING] Builder build_web_compilers|ddc_kernel uses input extension .ddc.module which is not specified in the `build.yaml`
[WARNING] Builder build_web_compilers|ddc uses input extension .ddc.module which is not specified in the `build.yaml`
[WARNING] Builder build_web_compilers|entrypoint outputs .digests  from .dart which is not specified in the `build.yaml`
[WARNING] Builder build_vm_compilers|vm uses input extension .vm.module which is not specified in the `build.yaml`
```

This leads to https://github.com/dart-lang/build/pull/2147

natebosch:
cc @kevmoo

build_runner/lib/src/entrypoint/doctor.dart
```diff
@@ -0,0 +1,123 @@
...
+  /// If there are any problems they will be logged and `false` returned.
+  bool _checkBuildExtensions(BuilderApplication builderApplication,
+      Map<String, BuilderDefinition> config) {
+    var phases = builderApplication.buildPhaseFactories
+        .map((f) => f(PackageNode(null, null, null, isRoot: true),
+            BuilderOptions.empty, InputSet.anything, InputSet.anything, true))
+        .whereType<InBuildPhase>()
+        .toList();
+    if (phases.isEmpty) return true;
+    if (!config.containsKey(builderApplication.builderKey)) return false;
```
jakemac53: why is this check necessary?

natebosch: The `BuilderApplications` we have are from the build script snapshot, and there could be slight variance in the builders that are in the loaded config.

There's going to be some weirdness around shapshots and stuff that we will want to work out here - we should iterated on the doctor command for a while before we expose it in the help output.

PR #2155, opened 2019-04-09
Move integration tests for build invalidation

natebosch:
Use a `test_descriptor` directory so that we have only a single builder
and can perform the builds faster than in the `_test` package.


natebosch:
No more `ensureCleanGitClient` for these tests 😄 

Between that and faster builds with only 1 simple builder the time on my machine went from 3:17 to 1:31.

PR #2158, opened 2019-04-09
Allow using `:` as separator for builders

### LINKED ISSUE #1127: Pick a different character than `|` for separators
At least for builder keys... On linux it causes us to need extra quoting and seems to also trigger some weirdness with `Process.run` on windows (which behaves differently than linux).

This would be very breaking - `build.yaml` files would need to change. We could maybe do it in a backwards compatible way and translate the `|` into the new character though?

natebosch:
Closes #1127 

Update the `normalizeBuilderKey` methods so that they first translate a
`|` into `:` and then continue normally from there.

Both `build.yaml` parsing, and `--define` argument parsing, use this
normalization so they will continue to accept existing keys.

### DIFF
--- a/_test/test/build_integration_test.dart
+++ b/_test/test/build_integration_test.dart
@@ -37,13 +37,11 @@ void main() {
       var content = 'cool';
       await runBuild(trailingArgs: [
         '--define',
-        'provides_builder|some_post_process_builder=default_content=$content'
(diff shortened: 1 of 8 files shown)


natebosch:
We'll want to separately consider adding a warning for using the old style so that folks start to clean up their usage.

We should immediately be able to clean up docs and our usage as long as we bump the minimum version of `build_config`.

PR #2160, opened 2019-04-10
Indicate that compiler output is generated

natebosch:


natebosch:
See https://github.com/dart-lang/webdev/pull/263

kevmoo:
Screen Shot 2019-04-10 at 12 00 05 PM

That's what I'm talkin about!

PR #2163, opened 2019-04-10
Update to new builder key format in tests

natebosch:
After the update to `build_config` version `0.3.2` the keys that we see
in the code no longer match what they were before. When we have
expectations against builders keys they need to use `:` as the
separator.

Manual calls to `apply` also need to use the new format since this is
what gets generated in the build script.

Also update the uses which mimic `build.yaml` files even though it isn't
strictly required.

### DIFF
--- a/build_runner/test/build_script_generate/builder_ordering_test.dart
+++ b/build_runner/test/build_script_generate/builder_ordering_test.dart
@@ -25,7 +25,7 @@ void main() {
               'build_extensions': {},
               'target': '',
               'import': '',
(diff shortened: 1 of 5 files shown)


natebosch:
> Do we still validate support for the `|` somewhere?

Not really. We could add a quick test in `build_config` if you want.

kevmoo:
> > Do we still validate support for the `|` somewhere?
> 
> Not really. We could add a quick test in `build_config` if you want.

Until we drop support, we should test it, right?

PR #2170, opened 2019-04-12
Dont prompt in daemon mode

jakemac53:
Also handle startup logs in daemon mode better, we didn't log anything before if it failed to start up.

Fixes https://github.com/dart-lang/webdev/issues/259

### DIFF
--- a/build_runner/lib/src/entrypoint/daemon.dart
+++ b/build_runner/lib/src/entrypoint/daemon.dart
@@ -8,6 +8,7 @@ import 'dart:io';
 import 'package:build_daemon/constants.dart';
 import 'package:build_daemon/src/daemon.dart';
 import 'package:build_runner/src/daemon/constants.dart';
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/entrypoint/daemon.dart
```diff
@@ -53,11 +54,17 @@ class DaemonCommand extends BuildRunnerCommand {
       }
     } else {
       stdout.writeln('Starting daemon...');
-      var builder = await BuildRunnerDaemonBuilder.create(
+      BuildRunnerDaemonBuilder builder;
+      // Ensure we capture any logs that happen during startup.
+      var startupLogSub =
+          Logger.root.onRecord.listen((LogRecord l) => stdout.writeln('$l\n'));
```
natebosch: [nit] Shouldn't need the type in the function argument.

PR #2177, opened 2019-04-16
Use the platform name for SDK dill digest

natebosch:
Prevents a potential bug where we fail to invalidate stored state in the
incremental compiler. This isn't an issue today since the 'vm' platform
is never able to use the incremental compiler due to passing
`--input-linked` arguments.

In the long term we'll want to consider using separate workers, or
setting up the kernel worker to be able to avoid a full invalidation
when hopping between platforms.

### DIFF
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -78,6 +78,7 @@ class KernelBuilder implements Builder {
           buildStep: buildStep,
           summaryOnly: summaryOnly,
           outputExtension: outputExtension,
+          platform: platform,
           dartSdkDir: platformSdk,
           sdkKernelPath: sdkKernelPath);
     } on MissingModulesException catch (e) {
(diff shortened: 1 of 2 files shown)


build_modules/lib/src/kernel_builder.dart
```diff
@@ -265,7 +266,7 @@ Future<void> _addRequestArguments(
   request.inputs.add(Input()
     ..path = '${Uri.file(p.join(sdkDir, sdkKernelPath))}'
     // Sdk updates fully invalidate the build anyways.
-    ..digest = [0]);
+    ..digest = md5.convert(utf8.encode(platform.name)).bytes);
```
jakemac53: Should we add this as a getter on DartPlatform with caching? probably doesn't matter much though.

natebosch: Discussed offline - probably not worth it.

PR #2179, opened 2019-04-16
Move the clean command to runner binary

natebosch:
Since the clean process does not need instantiated Builders it can run
in the outer binary rather than from the generated script. This should
make it faster.

- Move the non build script commands to a `bin/src/commands` directory.
- Make a list of "local" commands.
- Start logging before running local commands, explicitly disable
  logging in the one command which must not have any other stdout.

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -15,8 +15,8 @@ import 'package:build_runner/src/build_script_generate/bootstrap.dart';
 import 'package:build_runner/src/entrypoint/runner.dart';
 import 'package:build_runner/src/logging/std_io_logging.dart';
 
-import 'src/commands/generate_build_script.dart';
 import 'src/commands/clean.dart';
+import 'src/commands/generate_build_script.dart';
 
 Future<Null> main(List<String> args) async {
   // Use the actual command runner to parse the args and immediately print the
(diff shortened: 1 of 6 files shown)


natebosch:
> Should these commands live under lib instead of bin?

I could go either way. I kind of like at least separating them out of `lib/src/entrypoint` for clarity. I could see an argument for putting them somewhere else in `lib/`.

> How should clean interact with daemon mode long term?

Good question. I think it should probably shut it down if it's running. I'm pretty sure it will still make sense to do that without running from a generated script/snapshot.

jakemac53:
> I could go either way. I kind of like at least separating them out of `lib/src/entrypoint` for clarity. I could see an argument for putting them somewhere else in `lib/`.

It generally feels wrong to have much code under `bin/`, but it also feels wrong to have libraries shipped under `lib/` that aren't accessible via any public entrypoint, :man_shrugging:. I think you could argue some concrete benefits of the current approach (makes the builds of downstream users a tiny bit more efficient to not have to read them).

> Good question. I think it should probably shut it down if it's running. I'm pretty sure it will still make sense to do that without running from a generated script/snapshot.

Ya, conceptually I agree. I think it might be a bit difficult to actually do but we can cross that bridge when we come to it.

natebosch:
I restructured this a bit. There are now 2 implementations of `CleanCommand` which defer to a single implementation of `cleanFor`.  There are slight differences in how it works - the one `lib/` sets up it's own logging, and uses `assetGraphPath`. The one in `bin/` relies on logging already being set up and uses `assetGraphPathFor(buildScriptPath)`.

@jakemac53 - think this looks OK?

Issue #1804, opened 2018-08-31
NoSuchMethodError: The getter 'inputs' was called on null.


qiang437587687:
**I get this message :** 

You have hit a bug in build_runner
Please file an issue with reproduction steps at https://github.com/dart-lang/build/issues

**so, my build environment:**

* [2.0.0-edge.be6309690fd60284a87f3258a740c7c30efb1092 ] Dart SDK Version (`dart --version`)

* [build_runner ^0.10.0 ] What package(s) from this repo you are using, and the version (i.e. `build_runner 0.7.12`)

* [Just demo at https://flutter.io/json/ ] What builder(s) you are using (or writing yourself). Try to give a short summary of what they do.

* [ MacOS] Whether you are using Windows, MacOSX, or Linux (if applicable)

* [ iOS Simulator] Whether you are using Chrome, Safari, Firefox, Edge (if applicable)

* [dependencies:   
 cupertino_icons: ^0.1.2
  transparent_image: ^0.1.0
  pull_to_refresh: ^1.1.5
  json_annotation: ^1.0.0 

dev_dependencies:
  flutter_test:
    sdk: flutter
  build_runner: ^0.10.0
  json_serializable: ^1.0.0
] Any other packages or constraints we should know about

**error message:**

NoSuchMethodError: The getter 'inputs' was called on null.
Receiver: null
Tried calling: inputs
dart:core                                                                                                                                Object.noSuchMethod
/Users/zhangxianqiangmacbook/.pub-cache/hosted/pub.flutter-io.cn/build_runner_core-0.3.1+5/lib/src/asset_graph/serialization.dart 58:20  _AssetGraphDeserializer.deserialize
package:build_runner_core/src/asset_graph/graph.dart 42:48                                                                               new AssetGraph.deserialize
package:build_runner_core/src/generate/build_definition.dart 209:38                                                                      _Loader._tryReadCachedAssetGraph.<fn>
package:build_runner_core/src/logging/logging.dart 25:30                                                                                 logTimedAsync
package:build_runner_core/src/generate/build_definition.dart 207:12                                                                      _Loader._tryReadCachedAssetGraph
package:build_runner_core/src/generate/build_definition.dart 79:28                                                                       _Loader.prepareWorkspace
package:build_runner_core/src/generate/build_definition.dart 64:50                                                                       BuildDefinition.prepareWorkspace
package:build_runner_core/src/generate/build_impl.dart 106:49                                                                            BuildImpl.create
[...]

PR #2187, opened 2019-04-17
Report null assetId during deserialization

natebosch:
Towards #1804
Followup on #2181

In a reported case where we were able to inspect the serialized asset
graph the problem exhibited is that a `null` was tracked through
`outputs`, rather than an ID which points to a missing node. Catch that
case specifically as well because the error message can be more direct.

### DIFF
--- a/build_runner_core/lib/src/asset_graph/serialization.dart
+++ b/build_runner_core/lib/src/asset_graph/serialization.dart
@@ -62,6 +62,13 @@ class _AssetGraphDeserializer {
       if (node is BuilderOptionsAssetNode) continue;
 
       for (var output in node.outputs) {
+        if (output == null) {
+          log.severe('Found a null output from ${node.id} which is a '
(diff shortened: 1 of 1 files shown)


natebosch:
> I think we could do a better error if we do it during serialization? My guess is we have a non-null AssetId but when we look up the integer id in the map (for serialization) it doesn't exist for some reason?

Maybe? without knowing the repro to get to this state it's a little hard to build the code paths that could find it during serialization. It's also not clear what we should do with the error at that point - we'd probably still want the build to succeed right?

jakemac53:
> Maybe? without knowing the repro to get to this state it's a little hard to build the code paths that could find it during serialization. It's also not clear what we should do with the error at that point - we'd probably still want the build to succeed right?

I am not sure we really do want it to succeed - the asset graph it serialized would be invalid and something could be wrong with the build itself too.

PR #2190, opened 2019-04-17
make the incremental compiler optional in the kernel builder

jakemac53:


build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.0.1
```
natebosch: `2.1.0`. This adds new API surface area that a package would want to opt-in to.

PR #2194, opened 2019-04-17
Drop hardcoded list of known logger names

natebosch:
Now that there is a split with some of these logger names coming from
`build_runner_core` it's not as easy to roll out changes which introduce
new logger names. The only cases where we do want to show the logger
name are the ones where we are running actions and will always fit the
pattern of `'$builderLabel on $input'`.

### DIFF
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -61,7 +61,7 @@ void _stdIOLogListener(LogRecord record, {bool verbose}) =>
 /// splits the header for levels >= WARNING.
 String _recordHeader(LogRecord record, bool verbose) {
   var maybeSplit = record.level >= Level.WARNING ? '\n' : '';
-  return verbose || record.loggerName.contains(' on ')
+  return verbose || record.loggerName.contains(' ')
       ? '${record.loggerName}:$maybeSplit'
       : '';
 }
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -26,7 +26,7 @@ StringBuffer colorLog(LogRecord record, {bool verbose}) {
   final level = color.wrap('[${record.level}]');
   final eraseLine = ansiOutputEnabled && !verbose ? '\x1b[2K\r' : '';
   var lines = <Object>[
-    '$eraseLine$level ${_loggerName(record, verbose)}${record.message}'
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/logging/std_io_logging.dart
```diff
@@ -57,34 +57,11 @@ StringBuffer colorLog(LogRecord record, {bool verbose}) {
...
-    'serve',
-    'test',
-    'watch',
-  ];
+/// Filter out the Logger names which aren't coming from specific builders and
+/// splits the header for levels >= WARNING.
+String _recordHeader(LogRecord record, bool verbose) {
   var maybeSplit = record.level >= Level.WARNING ? '\n' : '';
-  return verbose || !knownNames.contains(record.loggerName)
+  return verbose || record.loggerName.contains(' on ')
```
jakemac53: Not a big fan of the pattern matching here... if we instead removed all the custom loggers and used a single common logger we could check for that one specifically and it would be a fair bit better I think?

jakemac53: Oh although then we wouldn't get the names in verbose mode...

jakemac53: Another option could be creating a custom LogRecord type? That would be safer but also has other disadvantages, although that class hasn't changed in ages so it is unlikely we would run into issues.

natebosch: I'd be in favor of dropping the names in verbose mode and simplifying down to a single common logger, but they can be handy sometimes so I can see an argument for keeping them.

I'm not excited by the idea of a custom `LogRecord` implementation, though I think all the cases for logs that we _do_ want to print the name go through [`BuildForInputLogger`](https://github.com/dart-lang/build/blob/83ea53f8dcfef04e17784698656c376f7e3e58c4/build_runner_core/lib/src/logging/build_for_input_logger.dart) which might give us some leverage to handle it...

natebosch: https://github.com/dart-lang/webdev/pull/305 added a dependency on the only logger names we want printed containing a space. I'll use the same criteria here.

In the long term we'll likely want to prefix all the non-shown logger names with `build.`

jakemac53: > In the long term we'll likely want to prefix all the non-shown logger names with `build.`

Although now that I think about it this would break for webdev haha

PR #2195, opened 2019-04-18
Add use-incremental-compiler option to the ddc builder

jakemac53:
This has some basic checking that the ddc builder is always configured with exactly the same options, which should keep people on the rails although it isn't perfect.

### DIFF
--- a/build_web_compilers/lib/builders.dart
+++ b/build_web_compilers/lib/builders.dart
@@ -45,7 +45,10 @@ PostProcessBuilder dartSourceCleanup(BuilderOptions options) =>
         ? const FileDeletingBuilder(['.dart', '.js.map'])
         : const FileDeletingBuilder(['.dart', '.js.map'], isEnabled: false);
 
-const _useIncrementalCompilerOption = 'use-incremental-compiler';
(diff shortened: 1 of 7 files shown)


natebosch:
Are we going to consider it breaking to change the default? Do we plan on dropping this option long term?

build_web_compilers/lib/builders.dart
```diff
@@ -40,3 +44,24 @@ PostProcessBuilder dartSourceCleanup(BuilderOptions options) =>
     (options.config['enabled'] as bool ?? false)
         ? const FileDeletingBuilder(['.dart', '.js.map'])
         : const FileDeletingBuilder(['.dart', '.js.map'], isEnabled: false);
+
+const _useIncrementalCompilerOption = 'use-incremental-compiler';
+bool _readUseIncrementalCompilerOption(BuilderOptions options) {
```
natebosch: Can we add a doc comment here about the global checking

jakemac53:
> Are we going to consider it breaking to change the default? Do we plan on dropping this option long term?

I don't believe we would, unless we anticipated that change actually breaking users for some reason. In terms of the global checking etc it shouldn't matter (as long as users are using the global options).

Issue #2197, opened 2019-04-18
Document the build daemon classes


jakemac53:
There are almost no doc comments on any of the build daemon related classes right now, which makes them difficult to understand and use. We need to go through and thoroughly document everything.

Specifically all the data classes are really bare and the BuildDaemonClient class also.

As an example, `client.buildResults` is a `Stream<BuildResults>`, and `BuildResults.results` is a `List<BuildResult>`. Why are there multiple build results for each item in the stream?

grouma:
Great idea. I can start tackling this today.

PR #2199, opened 2019-04-18
Add doc comments

grouma:
Closes https://github.com/dart-lang/build/issues/2197

### DIFF
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -60,6 +60,9 @@ Future<void> _handleDaemonStartup(
 bool _isActionMessage(String line) =>
     line == versionSkew || line == readyToConnectLog || line == optionsSkew;
 
+/// A client of the build daemon.
+///
+/// Handles starting and connecting to the build daemon.
 class BuildDaemonClient {
   final _buildResults = StreamController<BuildResults>.broadcast();
   final Serializers _serializers;
@@ -96,7 +99,7 @@ class BuildDaemonClient {
   void registerBuildTarget(BuildTarget target) => _channel.sink.add(jsonEncode(
       _serializers.serialize(BuildTargetRequest((b) => b..target = target))));
 
-  /// Builds all registered targets.
+  /// Builds all registered targets, including those not from this client.
   void startBuild() {
     var request = BuildRequest();
     _channel.sink.add(jsonEncode(_serializers.serialize(request)));
@@ -104,6 +107,9 @@ class BuildDaemonClient {
 
   Future<void> close() => _channel.sink.close();
 
+  /// Connects to the current daemon instance.
+  ///
+  /// If one is not running, a new daemon instance will be started.
   static Future<BuildDaemonClient> connect(
     String workingDirectory,
     List<String> daemonCommand, {
@@ -133,8 +139,13 @@ class BuildDaemonClient {
   }
 }
 
+/// Thrown when the port file for the running daemon instance can't be found.
 class MissingPortFile implements Exception {}
 
+/// Thrown if the client requests conflicting options with the current daemon
+/// instance.
 class OptionsSkew implements Exception {}
 
+/// Thrown if the current daemon instance version does not match that of the
+/// client.
 class VersionSkew implements Exception {}
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -63,6 +63,9 @@ bool _isActionMessage(String line) =>
 /// A client of the build daemon.
 ///
 /// Handles starting and connecting to the build daemon.
+///
+/// Example:
+///   https://pub.dartlang.org/packages/build_daemon#-example-tab-
 class BuildDaemonClient {
   final _buildResults = StreamController<BuildResults>.broadcast();
   final Serializers _serializers;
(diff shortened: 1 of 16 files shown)


build_daemon/lib/data/server_log.dart
```diff
@@ -7,6 +7,7 @@ import 'package:built_value/serializer.dart';
 
 part 'server_log.g.dart';
 
+/// A log
```
natebosch: remove this one since it's not adding new information.

build_daemon/lib/src/managers/build_target_manager.dart
```diff
@@ -30,15 +30,18 @@ class BuildTargetManager {
 
   Set<BuildTarget> get targets => _buildTargets.keys.toSet();
 
+  /// Returns all tracked channels of the manager.
```
natebosch: [nit] document like a field. Also avoid repeating information which is already clear from context.

/// All the tracked channels.

build_daemon/lib/data/server_log.dart
grouma: Oops. Meant to fill it in more but ultimately decided it really didn't need a comment.

build_daemon/lib/client.dart
```diff
@@ -96,14 +99,17 @@ class BuildDaemonClient {
   void registerBuildTarget(BuildTarget target) => _channel.sink.add(jsonEncode(
       _serializers.serialize(BuildTargetRequest((b) => b..target = target))));
 
-  /// Builds all registered targets.
+  /// Builds all registered targets, including those not from this client.
```
jakemac53: Probably worth noting the behavior if there is an ongoing build already (I assume it waits)

build_daemon/lib/daemon_builder.dart
```diff
@@ -10,6 +10,10 @@ import 'data/build_status.dart';
 import 'data/build_target.dart';
 import 'data/server_log.dart';
 
+/// A builder for the daemon.
+///
+/// Intended to be used as an interface for specific builder implementations
+/// which actually do the building.
```
jakemac53: Should this be abstract and have no implementations?

build_daemon/lib/data/build_request.dart
```diff
@@ -7,6 +7,9 @@ import 'package:built_value/serializer.dart';
 
 part 'build_request.g.dart';
 
+/// A build request.
```
jakemac53: nit: can remove this line

build_daemon/lib/client.dart
```diff
@@ -60,6 +60,9 @@ Future<void> _handleDaemonStartup(
 bool _isActionMessage(String line) =>
     line == versionSkew || line == readyToConnectLog || line == optionsSkew;
 
+/// A client of the build daemon.
+///
+/// Handles starting and connecting to the build daemon.
```
jakemac53: We should add some more docs on general usage here, maybe a full on basic example of registering a target and building it. This could highlight the differences between doing a single build and multiple builds.

build_daemon/lib/data/build_status.dart
```diff
@@ -44,6 +45,10 @@ abstract class DefaultBuildResult
   DefaultBuildResult._();
 }
 
+/// The group of [BuildResult]s for a single build done by the daemon.
+///
+/// Since the daemon can build multiple targets in parallel, the results are
+/// grouped together.
```
jakemac53: Do clients get results for only the targets they registered or all targets?

build_daemon/lib/data/build_target.dart
```diff
@@ -4,6 +4,7 @@ import 'package:built_value/serializer.dart';
 
 part 'build_target.g.dart';
 
+/// The build target, e.g. folder.
```
jakemac53: Add a note that what a target means is specific to the builder implementation/build system. It is a transparent identifier as far as the protocol is concerned.

build_daemon/lib/data/build_target.dart
```diff
@@ -20,12 +21,16 @@ abstract class DefaultBuildTarget
 
   DefaultBuildTarget._();
 
+  /// A set of patterns to match changes against.
+  ///
+  /// If a change matches a pattern this target will not be built.
```
jakemac53: nit: mention what this pattern is matching against specifically (file paths?)

build_daemon/lib/data/build_target.dart
```diff
@@ -37,7 +42,12 @@ abstract class OutputLocation
 
   String get output;
 
+  /// Whether to use symlinks for build outputs.
   bool get useSymlinks;
 
+  /// Whether to hoist the build output.
+  ///
+  /// Hoisted outputs will not contain the build target folder within their
+  /// path.
```
jakemac53: probably worth giving an example here

build_daemon/lib/data/build_status.dart
grouma: A client will only get results if they are interested in a target contained in the Build Result set. This may be an oversight and maybe we should update it to only return results of targets they are interested in.

jakemac53: I don't think it is a big deal (although they only care about results for their targets presumably). But we should call out the behavior in the docs.

PR #2207, opened 2019-04-22
Put a tight upper bound on SDK

natebosch:
Since there are open questions around enabling const inlining in CFE
we're not confident we won't need a new flag or something used for the
kernel worker. Add a tight upper bound constraint so that we can adjust
if necessary.

### DIFF
--- a/build_web_compilers/pubspec.yaml
+++ b/build_web_compilers/pubspec.yaml
@@ -5,7 +5,7 @@ author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
 
 environment:
-  sdk: ">=2.2.1-dev.4.0 <3.0.0"
+  sdk: ">=2.2.1-dev.4.0 <2.3.1"
 
 dependencies:
   analyzer: ">=0.30.0 <0.37.0"


kevmoo:
Do we want this, or just to publish a new release when the SDK updates?

The problem w/ this? When someone updates to Dart 2.4.0 or similar, they will roll BACK to an older version, which is really weird...right?

natebosch:
> Do we want this, or just to publish a new release when the SDK updates?

Depends on how quickly we can react to it I guess?

> When someone updates to Dart 2.4.0 or similar, they will roll BACK to an older version, which is really weird...right?

Not if they have `build_web_compilers: ^2.0.0` which we would hope for. They'd instead get an eager failure.

kevmoo:
We'll have to scramble the second a dev SDK comes out or our bots will all
start failing.

I guess it depends on how paranoid we are...

jakemac53:
We should do this for build_modules too? Although I guess it is to late for that and this can sort of handle it.

natebosch:
> We should do this for build_modules too? Although I guess it is to late for that and this can sort of handle it.

Yeah my thinking was that it would get handled by this package since that's where the usage matters.

> what will this do to our cron job runs?

It would make them fail once a new SDK is released.

jakemac53:
> It would make them fail once a new SDK is released.

I think it will make them fail as soon as any build on the bots happens with the next sdk version? (basically immediately)

kevmoo:
> > It would make them fail once a new SDK is released.
> 
> I think it will make them fail as soon as any build on the bots happens with the next sdk version? (basically immediately)

Then we bump it and see if it's not broken, etc...

I think this gives us way more hassle than it's worth...

jakemac53:
I wonder if we shouldn't leave the upper constraint higher in the repo but reduce it when we actually publish? It would be a bit interesting to try and enforce that though (and a single mistake ruins it forever).

jakemac53:
aaand I released 2.0 already but we never got this in so I guess we should just close it :grimacing:

Issue #2210, opened 2019-04-23
build_daemon returns Version skew in a race condition


grouma:
If two instances of the build daemon are started around the same time, only one will get a hold of the lock file. It should write a version file immediately after this. However, the second instance can try to read this version file before it is written, which will cause it to return a Version Skew. We can fix this by introducing a retry on the Version File if the value is null.

grouma:
Alternatively we can write the version file inside the `tryGetLock` method instead of inside the `daemon.start` method which is called an indeterminable while after the lock is grabbed.

PR #2216, opened 2019-04-23
Retry build_daemon state file reads

grouma:
Closes https://github.com/dart-lang/build/issues/2210

Changes:
 - Add retry logic to reading the various state files, e.g. portFile.
 - Update test daemon to better mimic actual implementations i.e. read version file
 - Add test to ensure there is no race condition
 - Rev version
   - Note that this isn't technically a breaking change as we haven't exposed the `src/daemon.dart` yet but it should be treated as one.


build_daemon/test/daemon_test.dart
```diff
@@ -128,27 +138,37 @@ void main() {
   });
 }
 
-Future<String> getOutput(Process daemon) async {
-  return await daemon.stdout
-      .transform(utf8.decoder)
-      .transform(const LineSplitter())
-      .firstWhere((line) => line == 'RUNNING' || line == 'ALREADY RUNNING');
-}
+Future<String> getOutput(Process daemon) async =>
```
natebosch: https://www.dartlang.org/guides/language/effective-dart/design#avoid-starting-a-method-name-with-get

Also, should it be private?

build_daemon/lib/client.dart
```diff
@@ -17,8 +17,13 @@ import 'data/build_target_request.dart';
 import 'data/serializers.dart';
 import 'data/server_log.dart';
 
-int _existingPort(String workingDirectory) {
+Future<int> _existingPort(String workingDirectory) async {
   var portFile = File(portFilePath(workingDirectory));
+  var retryCount = 0;
+  while (!portFile.existsSync() && retryCount < maxRetries) {
```
jakemac53: Thinking about this more we should probably model it as a timeout instead just to reduce the potential for accidentally increasing the amount of time we will wait.

Essentially, make `maxRetries` instead be a `Duration`, and use a stopwatch to bail out if the elapsed time > maxDuration.

build_daemon/lib/constants.dart
```diff
@@ -24,6 +24,11 @@ String daemonWorkspace(String workingDirectory) {
   return p.joinAll(segments);
 }
 
+/// The number of times to retry reading the various state files.
+///
+/// This prevents the likelihood of race conditions.
```
jakemac53: s/prevents/reduces :D

build_daemon/lib/constants.dart
```diff
@@ -24,6 +24,11 @@ String daemonWorkspace(String workingDirectory) {
...
+const maxRetries = 50;
```
jakemac53: also seems like this probably shouldn't be a public thing?

build_daemon/lib/src/daemon.dart
```diff
@@ -17,17 +17,27 @@ import 'server.dart';
 /// Returns the current version of the running build daemon.
 ///
 /// Null if one isn't running.
-String runningVersion(String workingDirectory) {
+Future<String> runningVersion(String workingDirectory) async {
   var versionFile = File(versionFilePath(workingDirectory));
+  var retryCount = 0;
```
jakemac53: might as well abstract this into a function to ensure consistent logic etc

build_daemon/test/daemon_test.dart
jakemac53: maybe `outputOf`?

build_daemon/test/daemon_test.dart
```diff
@@ -128,27 +138,37 @@ void main() {
...
+    await daemon.stdout
```
jakemac53: This is pretty difficult to follow with the indentation that dartfmt chose and the size of the expressions. I would create some local variables to make it more clear.

build_daemon/test/daemon_test.dart
jakemac53: Also add a comment about the behavior since it isn't necessarily intuitive

build_daemon/test/daemon_test.dart
```diff
@@ -128,27 +138,37 @@ void main() {
...
     import 'package:build_daemon/src/daemon.dart';
     import 'package:build_daemon/daemon_builder.dart';
+    import 'package:build_daemon/client.dart';
 
     main() async {
       var daemon = Daemon('$workspace');
       if (daemon.tryGetLock()) {
         var options = ['foo'].toSet();
         var timeout = Duration(seconds: $timeout);
+        await Future.delayed(Duration(seconds: 1));
```
jakemac53: add a comment for why this exists

build_daemon/lib/client.dart
natebosch: I don't know if it's necessary here since the work is synchronous and we only have async for the delay - but if we need it here's a retry function that we could either stick here or in `package:aynsc`. https://gist.github.com/natebosch/b038dd928e7bc9292d592b841f277471

jakemac53: TIL: `Future.timeout` is a thing :D

build_daemon/lib/constants.dart
grouma: Yeah it really shouldn't but I didn't know where else we should shove this so I thought it was pretty harmless. I'll see where it ends up with the suggested refactors.

PR #2221, opened 2019-04-23
Prepare to publish build_modules

natebosch:


build_modules/mono_pkg.yaml
```diff
@@ -12,7 +12,8 @@ stages:
   - unit_test:
     # Run the script directly - running from snapshot has issues for packages
     # that are depended on by build_runner itself.
-    - command: dart $(pub run build_runner generate-build-script) test --delete-conflicting-outputs-- -P presubmit
+    #- command: dart $(pub run build_runner generate-build-script) test --delete-conflicting-outputs-- -P presubmit
```
kevmoo: Maybe add a note here about why we're commenting this out?

build_modules/pubspec.yaml
```diff
@@ -26,10 +26,9 @@ dependencies:
 
 dev_dependencies:
   build_runner: ^1.0.0
-  build_test: ^0.10.0
-  build_vm_compilers: ">=0.1.0 <2.0.0"
+  #build_test: ^0.10.0
```
kevmoo: note about commented-out things? Or just remove em?

natebosch: Decided to restore them and bump the min SDK instead.

I realized there isn't any value in supporting an older SDK than is supported by the only 2 packages that allow this version of `build_modules` 😝 

Technically `build_node_compilers` could end up supporting this version and older SDK than `build_vm_compilers`, but I think that's unlikely...

kevmoo:
no worries

PR #2223, opened 2019-04-24
Experiment - drop $sdk

### LINKED ISSUE #2358: Drop `$sdk` package
It was always weird to have hardcoded magic specific to `build_web_compilers` (and `build_node_compilers`. Those builders should be fetching the SDK files themselves.

natebosch:
Closes #2358

We should not merge this until we're ready for a breaking change
release.
Checking travis...


jakemac53:
cc @pulyaevskiy for fyi as well, we plan to make this change soon and build_node_compilers will need to do something like what build_web_compilers now does

natebosch:
Specifically, `build_node_compilers` would need a change like https://github.com/dart-lang/build/pull/2215

pulyaevskiy:
@jakemac53 @natebosch Thanks for keeping me in the loop, really appreciate it!

natebosch:
Dropping this for now. I don't think it's worth the breaking change on it's _own_, and we have some blockers for getting done the other breaking changes we want. The issue is tagged for the next breaking release already so it should be easy to pick up once we have other reason for that.

PR #2226, opened 2019-04-24
Migrate from manual build script to copy builder

natebosch:
- Remove the `tool/build.dart` file with hardcoded builders.
- Update to the latest `build_web_compilers`.
- Add a copy builder to move the `.js` files from `web/` to `lib/` and
  into the source tree.
- Change to use `-O4` dart2js arg.


build_runner/web/hot_reload_client.dart
```diff
@@ -12,9 +12,9 @@ import 'dart:html';
 import 'package:js/js.dart';
 import 'package:js/js_util.dart';
 
-import 'module.dart';
-import 'reload_handler.dart';
-import 'reloading_manager.dart';
+import 'package:build_runner/src/server/build_updates_client/module.dart';
```
jakemac53: We actually might need to move these out to web?

natebosch: They are imported by some other stuff too I think. I didn't look _super_ closely so maybe they are just all in a little import cycle...

natebosch: I think the only blocker from moving these is some imports from tests.

We discussed offline and we'll ship this as is. In the next breaking release of `build_runner` we'll drop the hot reload code entirely in favor of the implementation in `webdev`

PR #2230, opened 2019-04-25
Fix strict-raw-types violations in build

natebosch:
Makes types more explicit. Change more `Future` to `Future<void>`, this
is not breaking for implementations of these interfaces since
`Future<dynamic>` is a subtype of `Future<void>`.

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 1.1.4
 
-- Use "strict raw types".
+- Internal cleanup: use "strict raw types".
+- Some return types for interfaces to implement changed from `Future<dynamic>`
+  to `Future<void>`.
 
 ## 1.1.3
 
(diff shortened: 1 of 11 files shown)


build/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.1.4
+
+- Use "strict raw types".
```
kevmoo: worth a changelog entry?

natebosch: 🤷‍♂️ I was thinking I'd publish to flush out any problems that this may cause so we know soon if we need to roll back. I don't expect any problems since this works internally...

natebosch:
> How'd you turn this on?

`analysis_options.yaml`

```yaml
analyzer:
  language:
    strict-raw-types: true
```

Can't turn it on in the repo cause there are a bunch of violations and it'll take me a while to clean it up all. Thought it would be worthwhile to clean up by package and these all seems like fairly uncontroversial improvements.

build/CHANGELOG.md
jakemac53: We should have a changelog entry for every version no matter what, so either this or `internal cleanup` would be fine.

PR #2232, opened 2019-04-26
Fix strict-raw-types in AssetGraph serialization

natebosch:
For discussion only


jakemac53:
Lets make a decision to merge this or not - I am totally fine with landing it :D

natebosch:
This was one of the noiser places if we enabled this lint which is why I opened this PR first.

We can either:

- Drop this and decide not to enforce.
- Submit this for now but not enforce.
- Clean up all other violations and also enforce.

I think my preference is for clean + enforce. I'll see how much work that would take.

natebosch:
@jakemac53 @grouma - check out the changes in `build_config` - I'm _pretty_ sure these need to all be `Map<dynamic, dynamic>` because yaml. Any concerns?

grouma:
Why `Map<dynamic, dynamic>` and not `Map<string, dynamic>`?

natebosch:
Yaml I think is `Map<dynamic, dynamic>`. It's one of the ways that it's a superset of json (and one of the ways I wish they hadn't chosen to extend json).

We are currently constrained by the function type here:
https://github.com/dart-lang/json_serializable/blob/4e287e7e76fdb2b582285afe69b74977b25b819b/checked_yaml/lib/checked_yaml.dart#L23

So we could cast that argument, but I'm not sure if we'd need do do it with `as Map<String, dynamic>` or `.cast<String, dynamic>()` because I don't know if the runtime types are reliably going to match.

We do later cast the keys to `String` in all cases I'm pretty sure, but I'm not sure what the difference in error behavior would be if we change it.

https://github.com/dart-lang/build/blob/83cdcc15b2b4ae0cf841ce2d6dbba38e54950d7f/build_config/lib/src/build_config.dart#L175

Do you want me to dig more? Would you prefer the `.cast<String, dynamic>()` over having `Map<dynamic, dynamic>` everywhere?

I'm also not sure if it would be breaking to make some of these signatures `Map<String, dynamic>`. The breakage would likely be minor, but we also don't have a super solid story on major version bumps in `build_config` so we'd have to choose whether we want to go through that.

grouma:
> Do you want me to dig more? Would you prefer the .cast<String, dynamic>() over having Map<dynamic, dynamic> everywhere?

I don't think it's necessary. As you have said, to properly type this it would likely require a breaking change. Not worth the effort given the minimal gain.

build_config/lib/src/build_config.g.dart
```diff
@@ -6,7 +6,7 @@ part of 'build_config.dart';
 // JsonSerializableGenerator
 // **************************************************************************
 
-BuildConfig _$BuildConfigFromJson(Map json) {
+BuildConfig _$BuildConfigFromJson(Map<dynamic, dynamic> json) {
```
jakemac53, line 9: You had to manually update this ya? That is a bit concerning... I can't remember if we ever got the support to add ignore comments back

natebosch: oh yeah, I wasn't paying attention to realize I was in a generated file here and I completely forgot that the reason I never got back to this was that we are blocked by `json_serializable`.

Hopefully if I properly link the issue and close out this PR I won't forget again 😜 

https://github.com/google/json_serializable.dart/issues/517

PR #2235, opened 2019-04-30
Wrap usage to terminal width

natebosch:


natebosch:
> pubspec/changelog?

Will follow up separately with a changelog entry for the `run` command and prep for publish.

PR #2236, opened 2019-04-30
Prepare to publish with `run` command.

natebosch:
Add a changelog entry and bump version.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.4.0
 
-- Add a `run` command to execute VM entrypoints from compiled kernel.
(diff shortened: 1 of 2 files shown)


build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.4.0
+
+- Add a `run` command to execute VM entrypoints from compiled kernel.
```
jakemac53: Does it actually support running from kernel yet? I didn't think it did, although it should be trivial to add if you want to add that here?

natebosch: Ah that's right, we never added that though we had discussed it... Updated the changelog entry.

PR #2238, opened 2019-04-30
Document the module algorithms we use

natebosch:
Write out in prose in the README a description of how this package
works.

### DIFF
--- a/build_modules/README.md
+++ b/build_modules/README.md
@@ -46,7 +46,7 @@ The overall process for emitting modules follows these steps:
    have an output. This step is mainly separated out for better invalidation
    behavior since the output will change less frequently than the code. The
    outputs from this step are non differentiated by platform.
-1. Emit package level `.meta_module` assets. This is an aggregation of the
+2. Emit package level `.meta_module` assets. This is an aggregation of the
    `.module.library` information with a first run at creating a module
    structure. The output depends on the algorithm, described below. The outputs
    from this step are specific to a platform and will exclude any libraries
@@ -54,7 +54,7 @@ The overall process for emitting modules follows these steps:
    will also be resolved to a concrete dependency based on the platform. In this
    step the dependencies of modules references the imported library, which may
    not match the primary source of the module containing that library.
-1. Emit package level `.meta_module.clean` assets. This step performs an extra
+3. Emit package level `.meta_module.clean` assets. This step performs an extra
    run of the strongly connected components algorithm so that any libraries
    which are in an import cycle are grouped into a single module. This is done
    as a separate step so that it can be sure it may read all the `.meta_module`
@@ -63,7 +63,7 @@ The overall process for emitting modules follows these steps:
    module becomes the union of their sources and dependencies. Dependencies are
    resolved to the "primary source" of the module containing the imported
    library.
-1. Emit library level `.module` assets which contain a filtered view of the
+4. Emit library level `.module` assets which contain a filtered view of the
    package level meta-module specific to a single module. These are emitted for
    the "primary source" of each module, as well as each library which is an
    entrypoint.
--- a/build_modules/README.md
+++ b/build_modules/README.md
@@ -33,7 +33,7 @@ building your Dart code.
 [`build_web_compilers`]: https://pub.dartlang.org/packages/build_web_compilers
 [`build_vm_compilers`]: https://pub.dartlang.org/packages/build_vm_compilers
 
-## Module algorithms
+## Module creation
 
 The overall process for emitting modules follows these steps:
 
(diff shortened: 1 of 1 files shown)


build_modules/README.md
```diff
@@ -32,3 +32,70 @@ building your Dart code.
...
+1. Emit a `.module.library` asset for every `.dart` library containing a
+   description of the directives (imports, exports, and parts), a list of the
+   `dart:` imports used, and a determination of whether the library is an
+   "entrypoint". Entrypoint libraries are either those which are likely to be
+   directly imported (they are under `lib/` but not `lib/src/`) or which have a
+   `main`. Only the libraries that can be imported (they aren't part files) will
+   have an output. This step is mainly separated out for better invalidation
+   behavior since the output will change less frequently than the code. The
+   outputs from this step are non differentiated by platform.
+1. Emit package level `.meta_module` assets. This is an aggregation of the
```
jakemac53: 2. (etc)

build_modules/README.md
```diff
@@ -32,3 +32,70 @@ building your Dart code.
...
+   `.module.library` information with a first run at creating a module
+   structure. The output depends on the algorithm, described below. The outputs
+   from this step are specific to a platform and will exclude any libraries
```
jakemac53: I don't think we exclude anything? We just mark them as not supported.

build_modules/README.md
```diff
@@ -32,3 +32,70 @@ building your Dart code.
...
+1. Emit package level `.meta_module.clean` assets. This step performs an extra
+   run of the strongly connected components algorithm so that any libraries
+   which are in an import cycle are grouped into a single module. This is done
+   as a separate step so that it can be sure it may read all the `.meta_module`
+   results across packages in a dependency cycle so that it may resolve import
+   cycles across packages. As modules are merged due to import cycles the new
+   module becomes the union of their sources and dependencies. Dependencies are
+   resolved to the "primary source" of the module containing the imported
+   library.
+1. Emit library level `.module` assets which contain a filtered view of the
```
jakemac53: The `library level` comment here is a bit confusing, I am not sure what that means in this context? Maybe it can just be removed.

build_modules/README.md
natebosch: Discussed offline.

Markdown will render this with the correct numbers, using `1.` for all of them means if we add new entries we don't need to manually renumber them ourselves and keeps the diff cleaner.

Given that the number of steps is unlikely to change here, and is pretty small, I'm fine doing a manual numbering since it does look nicer in the source.

build_modules/README.md
```diff
@@ -32,3 +32,70 @@ building your Dart code.
 
 [`build_web_compilers`]: https://pub.dartlang.org/packages/build_web_compilers
 [`build_vm_compilers`]: https://pub.dartlang.org/packages/build_vm_compilers
+
+## Module algorithms
```
jakemac53: I think this should be moved to right above the actual module algorithms, and this title should be renamed `Module discovery strategy`, `Module creation`, or something like that?

build_modules/README.md
natebosch: Good catch. Fixed.

build_modules/README.md
natebosch: Done

build_modules/README.md
natebosch: It was meant to be distinct from "package level" above. I think it does read clearly with this part removed so I dropped it.

natebosch:
> We should also add an example of configuring the module strategy?

I think that might be a separate section. Along with maybe some advice on when to use each?

In my mind this documentation was mostly intended for authors of the package. I could move it out of the README to a docs directory or something if you think it would be better.

build_modules/README.md
```diff
@@ -32,3 +32,71 @@ building your Dart code.
 
 [`build_web_compilers`]: https://pub.dartlang.org/packages/build_web_compilers
 [`build_vm_compilers`]: https://pub.dartlang.org/packages/build_vm_compilers
+
+## Module creation
+
+The overall process for emitting modules follows these steps:
+
+1. Emit a `.module.library` asset for every `.dart` library containing a
+   description of the directives (imports, exports, and parts), a list of the
```
jakemac53: Should we also mention the config specific dependency information here? Or does that fall under `directives`?

natebosch: I think that falls under `directives`. I mention config specific imports in one of the later steps.

jakemac53:
> In my mind this documentation was mostly intended for authors of the package. I could move it out of the README to a docs directory or something if you think it would be better.

I think that it is fine where it is, we could maybe link from here to something under `docs` for configuring it at some point.

PR #2244, opened 2019-05-02
Refactor resolver crawling

natebosch:
Fixes https://github.com/dart-lang/sdk/issues/36773

Previously the crawling was intermingled with handling changes and we
had a single use implementation of `FutureGroup` to capture all the work
happening. Refactor to use `crawlAsync` from `package:graphs` for the
crawling portion, and separate out the crawling from notifying the driver
that the contents have been changed. By the time `driver.changeFile` is
called the `resourceProvider` should have updated content for imported
libraries.

Since the `_cache*` maps are used to know what specifically needs to be
changed in the `resourceProvider`, the updates to both places needs to
happen synchronously. 

Update the invalidation of dependency cache to store asset digests
instead of contents.

### DIFF
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -21,7 +21,7 @@ const _ignoredSchemes = ['dart', 'dart-ext'];
 class BuildAssetUriResolver extends UriResolver {
   /// A cache of the directives for each Dart library.
   ///
-  /// This stored across builds and is only invalidated if we read a file and
+  /// This is stored across builds and is only invalidated if we read a file and
   /// see that it's content is different from what it was last time it was read.
   final _cachedAssetDependencies = <AssetId, Set<AssetId>>{};
 
@@ -42,6 +42,8 @@ class BuildAssetUriResolver extends UriResolver {
   /// later). If this happens we don't want to hide the asset from the analyzer.
   final seenAssets = HashSet<AssetId>();
 
+  /// Crawl the transitive imports from [entryPoints] and ensure that the
+  /// content of each asset is updated in [resourceProvider] and [driver].
   Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
       AnalysisDriver driver) async {
     final changes =
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -46,8 +46,9 @@ class BuildAssetUriResolver extends UriResolver {
   /// content of each asset is updated in [resourceProvider] and [driver].
   Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
       AnalysisDriver driver) async {
-    final changes =
+    final changedPaths =
         await crawlAsync<AssetId, _AssetState>(entryPoints, (id) async {
+      final path = assetPath(id);
       if (!await buildStep.canRead(id)) {
         if (seenAssets.contains(id)) {
           // ignore from this graph, some later build step may still be using it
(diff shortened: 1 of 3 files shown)


natebosch:
With this change I'm not able to see a failure in >100 test runs, previously on my machine it always took less than 10 runs to see the failure.

It's not clear to me the exact bug this solves, but I think either way it's easier to follow this logic.

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -27,54 +39,47 @@ class BuildAssetUriResolver extends UriResolver {
...
-              .where((id) => id != null)
-              .toSet();
-          _cachedAssetDependencies[assetId] = dependencies;
+  Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
+      AnalysisDriver driver) async {
+    final states =
+        await crawlAsync<AssetId, _AssetState>(entryPoints, (id) async {
+      try {
+        return _AssetState(id, await buildStep.readAsString(id));
+      } catch (e) {
```
jakemac53: We should probably only catch AssetNotFoundException here

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -27,54 +39,47 @@ class BuildAssetUriResolver extends UriResolver {
...
+        return _AssetState.removed(id);
+      }
+    }, (id, state) {
+      if (state.isRemoved) return const [];
+      if (_cachedAssetContents[id] == state.content) {
+        return _cachedAssetDependencies[id];
+      }
+      _cachedAssetContents[id] = state.content;
+      return _cachedAssetDependencies[id] = _parseDirectives(state);
+    }).toList();
```
jakemac53: Can we leave this as a stream and use await for, or does that bring back the bug? I am mostly concerned about the potential memory consumption of this list (it could get large and we could have a lot of them).

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: This was translating what we had before which caught anything, I agree that it would be better to be more specific.

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: That seems to restore the problem.

natebosch: I noticed an optimization to bring us closer to previous behavior - I was doing updates here even for assets that hadn't changed. If I add another `isChanged` boolean I can filter the list down to just the changes which should help keep it smaller.

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -11,70 +12,81 @@ import 'package:analyzer/file_system/memory_file_system.dart';
...
+import 'package:crypto/crypto.dart';
+import 'package:graphs/graphs.dart';
 import 'package:path/path.dart' as p;
 
 const _ignoredSchemes = ['dart', 'dart-ext'];
 
 class BuildAssetUriResolver extends UriResolver {
+  /// A cache of the directives for each Dart library.
+  ///
+  /// This stored across builds and is only invalidated if we read a file and
```
grouma: [nit] This is* stored

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -11,70 +12,81 @@ import 'package:analyzer/file_system/memory_file_system.dart';
...
-                var uri = Uri.parse(directive.uri.stringValue);
-                return !_ignoredSchemes.any(uri.isScheme);
-              })
-              .map((d) => AssetId.resolve(d.uri.stringValue, from: assetId))
-              .where((id) => id != null)
-              .toSet();
-          _cachedAssetDependencies[assetId] = dependencies;
+  final seenAssets = HashSet<AssetId>();
+
+  Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
```
grouma: Can we get a doc comment on this function?

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: Done

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: Done

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -11,70 +12,83 @@ import 'package:analyzer/file_system/memory_file_system.dart';
...
+  /// see that it's content is different from what it was last time it was read.
   final _cachedAssetDependencies = <AssetId, Set<AssetId>>{};
-  final _cachedAssetContents = <AssetId, String>{};
+
+  /// A cache of the digest for each Dart asset.
+  ///
+  /// This is stored across builds and used to invalidate the values in
+  /// [_cachedAssetDependencies] only when the actual content of the library
+  /// changed.
+  final _cachedAssetDigests = <AssetId, Digest>{};
```
jakemac53: Comparing digests might actually be slower than strings? Digests that are equal aren't identical, but Strings I think are canonicalized?

In practice I believe this actually won't matter because we keep the old digest instances around instead of bashing over with the new one if it doesn't change.

jakemac53:
In general I think this will be ok but I do think we need to verify performance on at least angular components example to make sure it doesn't unexpectedly cause a regression.

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: for reasons I don't understand it turns out I _must_ use `await for` and collect up the changed paths to update the `driver` afterwards. I have no idea why updating the `resourceProvider` must happen during the async crawling...

I also don't know why the problem doesn't surface in this repo but does show up when I try to build a larger project...

jakemac53: :/

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: > Strings I think are canonicalized?

Not always. If you read a file into a string twice then `first == second` will be true but `identical(first, second)` will not be true. I think the `identical` case would happen for either `Digest` or `String` for us anyway, since we cache things and should always give back the same instance, but I don't know if the `operator ==` for `Digest` takes advantage of that.


Either way, I don't think I can find any performance difference but I didn't add a stopwatch... I wouldn't be surprised to find that when files are changed the performance improvement we get from comparing digests more the offsets the potential optimizations in the equals case.

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: Hmm, we've got a much worse bug here than I thought.

Somehow all the way back in https://github.com/dart-lang/build/pull/1983 I was never writing to `seenAssets`, so all of our out-of-order build step handling was completely broken. I don't know how this hasn't been manifesting in even worse ways...

natebosch: OK, there are some really subtle async things happening. I think what I'm seeing is that we can't have any async steps between updating `_cachedAsset*` and the `resourceProvider`.

build_resolvers/lib/src/build_asset_uri_resolver.dart
jakemac53: The `==` operator for Digest uses `ListEquality` from package:collection which does do an instance check first.

I think this should be fine since we cache the instances.

PR #2248, opened 2019-05-03
Fix exceptions in PackageAssetReader

natebosch:
- Don't throw for missing files from `canRead` - the correct behavior is
  to return `false`.
- Change `ArgumentError` to `AssetNotFoundException` in read methods.
- Shorten the implementations to use `??` fallbacks.

### DIFF
--- a/build_test/lib/src/package_reader.dart
+++ b/build_test/lib/src/package_reader.dart
@@ -83,7 +83,7 @@ class PackageAssetReader extends AssetReader
     if (id.package == _rootPackage) {
       return File(p.canonicalize(p.join(_rootPackagePath, id.path)));
     }
-    throw UnsupportedError('Unable to resolve $id');
+    return null;
   }
 
   String get _rootPackagePath {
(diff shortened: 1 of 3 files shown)


natebosch:
I discovered this bug when I tried to refactor a `try/catch` into a `canRead` in https://github.com/dart-lang/build/pull/2244

build_test/lib/src/package_reader.dart
```diff
@@ -125,29 +125,15 @@ class PackageAssetReader extends AssetReader
...
-  Future<String> readAsString(AssetId id, {Encoding encoding = utf8}) {
-    final file = _resolve(id);
-    if (file == null) {
-      throw ArgumentError('Could not read $id.');
-    }
-    return file.readAsString(encoding: encoding);
-  }
+  Future<String> readAsString(AssetId id, {Encoding encoding = utf8}) =>
+      _resolve(id)?.readAsString(encoding: encoding) ??
+      (throw AssetNotFoundException(id));
```
grouma: Curious why this needs parenthesis.

natebosch: it has to do with how tightly the different syntax binds. When you have a `throw` on the right side of `??` you need to wrap it in parens.

There has been some discussion and the google3 community isn't completely sold on the `?? (throw)` pattern... The requirement to use parens is just part of the reason.

Personally, I like it more than the alternative...

PR #2249, opened 2019-05-04
Handle fake package roots in PackageAssetReader

natebosch:
Fixes another case where `canRead` would throw instead of return false.

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.10.7+3
+
+- Handle the case where the root package in a `PackageAssetReader` is a fake
+  package.
+
 ## 0.10.7+2
 
 - Avoid throwing for missing files from `PackageAssetReader.canRead`.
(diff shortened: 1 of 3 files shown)


natebosch:
This time I actually tested it with the changes in https://github.com/dart-lang/build/pull/2244 to confirm the tests pass 😝

build_test/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 0.10.7+3
+
+- Handle the case where the root package in a `PackageAssetReader` is a fake
```
kevmoo: When does that happen? Just curious...

natebosch: `PackageAssetReader` is designed to let you read assets from real packages imported by the current isolate. We combine it with asset readers for in-memory assets that have fake packages. With things like [`resolveSource`](https://github.com/dart-lang/build/blob/b3c63cc5ed1b6993b557f6f2faf4d153b6979cc2/build_test/lib/src/resolve_source.dart#L27) we pass the [`rootPackage`](https://github.com/dart-lang/build/blob/b3c63cc5ed1b6993b557f6f2faf4d153b6979cc2/build_test/lib/src/resolve_source.dart#L177) as one of those fake ones.

kevmoo: Ah! Okay. Still zero idea. That's fine. I trust you. 😁

Issue #2208, opened 2019-04-22
build_daemon ServerLog should have fields for Level and Logger Name


grouma:
Replacing https://github.com/dart-lang/webdev/issues/137

jakemac53:
We should also add `error` and `stackTrace` (nullable)

PR #2250, opened 2019-05-06
update the ServerLog to more closely resemble a LogRecord

jakemac53:
Closes https://github.com/dart-lang/build/issues/2208


jakemac53:
I have updates for build_runner and webdev pending also, but they are blocked on this update (trying to keep things small)

build_daemon/lib/data/server_log.dart
```diff
@@ -4,14 +4,47 @@
...
+            'Only pre-defined log levels are supported, see [Level.LEVELS] '
+                'from package:logging.'));
+    return LogRecord(logLevel, message, loggerName ?? '', error,
+        stackTrace == null ? null : StackTrace.fromString(stackTrace));
+  }
+
   ServerLog._();
-  String get log;
+
+  int get level;
```
grouma: Should we make this an enum? This would likely prevent the potential exception you see above.

jakemac53: Hmm, that actually might be nice. Then we can add a simple helper here for converting between levels also?

We would still have to throw the exception when converting from a `LogRecord` though, but that is less common.

jakemac53: Done

jakemac53:
PTAL, I had to add in the build_runner changes as well to get the integration tests running

build_runner/lib/src/entrypoint/daemon.dart
```diff
@@ -65,9 +66,17 @@ class DaemonCommand extends BuildRunnerCommand {
       );
       await startupLogSub.cancel();
 
-      // Forward server logs to daemon command STDIO.
-      var logSub =
-          builder.logs.listen((serverLog) => stdout.writeln(serverLog.log));
+      // Serialize server logs to daemon command STDIO.
```
grouma: Please update the comment as this is not accurate.

build_daemon/lib/data/server_log.dart
```diff
@@ -4,14 +4,102 @@
 
 import 'package:built_value/built_value.dart';
 import 'package:built_value/serializer.dart';
+import 'package:logging/logging.dart' as logging;
 
 part 'server_log.g.dart';
 
+enum Level {
+  finest,
```
natebosch: I'm not sure if we want to spread the choices of `package:logging` here... I would be in favor of shortening this to `fine, info, warning, severe`

build_daemon/lib/data/server_log.dart
```diff
@@ -4,14 +4,102 @@
...
+  severe,
+  shout,
+}
+
+/// Converts a [logging.Level] to a [Level].
+///
+/// Throws an [ArgumentError] if the [level] is not one of
+/// [logging.Level.LEVELS].
+Level fromLoggingLevel(logging.Level level) {
+  if (level == logging.Level.FINEST) {
```
natebosch: switch?

jakemac53: Can't unfortunately

build_daemon/lib/data/server_log.dart
jakemac53: We expose all these levels to builder authors though, so we should retain them imo.

These aren't just the opinions of `package:logging` either, they are at least standardized across Java/Dart :D.

jakemac53:
Gah, built value doesn't support regular enums. Will update tomorrow to use their custom bloated thing...

build_daemon/lib/data/server_log.dart
```diff
@@ -2,16 +2,94 @@
...
+import 'package:built_collection/built_collection.dart';
 import 'package:built_value/built_value.dart';
 import 'package:built_value/serializer.dart';
+import 'package:logging/logging.dart' as logging;
 
 part 'server_log.g.dart';
 
+/// Logging levels, these have a 1:1 mapping with the levels from
+/// `package:logging`.
+class Level extends EnumClass with Comparable<Level> {
```
grouma: This is better IMO but man... that's a lot of boilerplate :(

jakemac53: Ya, it does have some advantages, like allowing more robust comparison and simplifying the translations between the types since it has an actual name field. Also the name won't ever get mangled or anything. :man_shrugging:

build_daemon/lib/data/server_log.dart
```diff
@@ -2,16 +2,94 @@
...
+  static Serializer<Level> get serializer => _$levelSerializer;
+
+  static const Level FINEST = _$finest;
```
natebosch: why all caps?

jakemac53: The name is derived from this, and we want it to line up with the old names (which were upper case). The alternative would be upper casing it during log output etc, which we would end up doing all the time.

natebosch: now that we have the build daemon for a programmatic API we don't need to worry as much about keeping identical log output. We can pick different labels.

I think it would be good to try something like https://github.com/dart-lang/build/pull/1165 again

PR #2251, opened 2019-05-06
Add new pkg:pedantic lints

kevmoo:


kevmoo:
> Should we be including the `package:pedantic` analysis_options.yaml directly?

ideally, yes, but it'll break because we're relative to package dir...one place where mono_repos are painful

jakemac53:
This lint will be nice, that one case you had to fix was actually caused by a `dartfmt` (at some point it got an extra level indentation which forced it down a line). This lint should catch those now so we can avoid this pattern more robustly.

PR #2253, opened 2019-05-07
Accept lists of file changes rather than individual changes.

davidmorgan:


build_daemon/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_daemon
-version: 0.6.0
+version: 0.7.0
```
jakemac53: We have some more breaking changes coming so it would be good to roll these all together in one release (which will be the 1.0.0 release)

build_daemon/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 0.7.0
+
+- Accept file change notifications as `Stream<List<WatchEvent>>` instead of
+  `Stream<WatchEvent>`. This allows file change notifications to be sent as
+  batches of simultaneous changes, preventing over-triggering of builds.
```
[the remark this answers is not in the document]
davidmorgan: For file systems where you get batches of changes simultaneously--which we have internally--debouncing is less reliable and slower than just using the batches as provided by the file system.

So this allows us to do the right thing internally--while also allowing debouncing externally. We should probably do that, too :)

build_daemon/pubspec.yaml
davidmorgan: SGTM. Not too worried about how this gets released. Please just point me in the right direction :)

build_daemon/CHANGELOG.md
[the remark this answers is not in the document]
jakemac53: Ah ok that makes sense, this allows us to deal with the debouncing where we actually do the file listening as well (and batch up the changes), so I think that is a decent abstraction.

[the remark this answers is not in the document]
davidmorgan: Yup, that's the idea. Thanks.

jakemac53:
it looks like this needs a dartfmt for travis

jakemac53:
Oh, because of the mono_repo setup you would also have to fix up build_runner in the same cl here.

davidmorgan:
dartfmt done.

How does the monorepo setup use local versions when testing? i.e. how can I trigger that locally? Thanks.

jakemac53:
> How does the monorepo setup use local versions when testing? i.e. how can I trigger that locally? Thanks.

We create whole package fixtures with dependency overrides in their pubspec, see here https://github.com/dart-lang/build/blob/master/build_runner/test/integration_tests/utils/build_descriptor.dart#L93. Essentially we make sure the fixture matches the same versions as what the root package is using.

We also have the `_test` directory which has some additional e2e tests and has path dependency overrides for all packages in the mono repo.

davidmorgan:
Thanks; took a shot at it, let's see what travis thinks...

build_runner/pubspec.yaml
```diff
@@ -12,7 +12,7 @@ dependencies:
   async: ">=1.13.3 <3.0.0"
   build: ">=1.0.0 <1.2.0"
   build_config: ">=0.4.0 <0.4.1"
-  build_daemon: ^0.6.0
+  build_daemon: ^0.7.0-dev
```
jakemac53: You will want to add a dependency_override for this as well

davidmorgan: Done. (I think this is what you meant...)

grouma:
LGTM assuming @jakemac53 has already published his bug fix. We won't be able to publish the `build_runner` changes until after Google IO.

davidmorgan:
Travis looks happy after the rebase. AppVeyor also looks happy but github says it's pending--is there a way to nudge it?

Thanks :)

jakemac53:
Restarted appveyor - it looks like it never terminated which was hopefully just a flake

davidmorgan:
Thanks, looks good to go, I'll merge it.

What does appveyor do, BTW?

jakemac53:
Appveyor is our windows testing solution right now - but travis supports windows so we will hopefully switch to that soon, we only get 1 bot at a time with appveyor.

PR #2256, opened 2019-05-08
add build_daemon dep override to build_config

jakemac53:
this should fix travis


jakemac53:
> Would it work to instead remove the other overrides? We should have published packages now that allow the latest `build_config`...

It doesn't work because we have a tight version constraint (`>=0.4.0 <0.4.1`).

PR #2280, opened 2019-05-21
update build_web_compilers for org-dartlang-app uris

jakemac53:
This gets hot restart working with the previous fix in https://github.com/dart-lang/sdk/issues/36736.

We should probably discuss if we need to backport a 1.x fix as well.


natebosch:
> We should probably discuss if we need to backport a 1.x fix as well.

We don't need to. 1.x was on the analyzer ddc which isn't changing behavior.

PR #2283, opened 2019-05-22
Upstream changes required to land ddc + flutter web

jonahwilliams:
These changes mostly involve exposing more configuration to the constructor, except in the case of bootstrap ddc where I'm still using a src member.

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.1.4
+## 2.2.0
 
 - Make the `librariesPath` in the `KernelBuilder` configurable.
 - Fixed bug where the configured dart SDK was ignored.
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.1.4
+
+- Make the `librariesPath` in the `KernelBuilder` configurable.
+- Fixed bug where the configured dart SDK was ignored.
+
 ## 2.1.3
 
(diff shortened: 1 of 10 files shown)


build_modules/lib/src/kernel_builder.dart
```diff
@@ -58,11 +58,17 @@ class KernelBuilder implements Builder {
...
+  ///
+  /// If not provided, defaults to "lib/libraries.json".
+  final String librariesPath;
+
   KernelBuilder(
       {@required this.platform,
       @required this.summaryOnly,
       @required this.sdkKernelPath,
       @required this.outputExtension,
+      this.librariesPath,
```
jakemac53: nit: instead of allowing this to be null, assign it to the default in the initializer (same pattern as `platformSdk`)

build_modules/lib/src/kernel_builder.dart
```diff
@@ -58,11 +58,17 @@ class KernelBuilder implements Builder {
   /// directory, which contains the platform kernel files.
   final String platformSdk;
 
+  /// The libraries file for the current platform.
```
jakemac53: nit: clarify this should be an absolute path

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.1.4
```
jakemac53: nit: technically these are new features (plus a bug fix) so we should increment the minor version number

jakemac53:
Also the unit test needs to be updated to pass the platform https://travis-ci.org/dart-lang/build/jobs/535990115#L1332

PR #2286, opened 2019-05-23
update build_daemon to support serializing logs over stdout

jakemac53:
Also updated build_runner daemon command to use that for pre-connection logs

### DIFF
--- a/build_daemon/lib/client.dart
+++ b/build_daemon/lib/client.dart
@@ -43,10 +43,27 @@ Future<void> _handleDaemonStartup(
 
   // The daemon may log critical information prior to it successfully
   // starting. Capture this data and forward to the logHandler.
+  //
+  // Whenever we see a `logStartMarker` we will parse everything between that
+  // and the `logEndMarker` as a `ServerLog`. Everything else is considered a
+  // normal INFO level log.
+  StringBuffer nextLogRecord;
   var sub = stdout.where((line) => !_isActionMessage(line)).listen((line) {
-    logHandler(ServerLog((b) => b
-      ..level = Level.INFO
-      ..message = line));
+    if (nextLogRecord != null) {
+      if (line == logEndMarker) {
+        logHandler(serializers.deserialize(jsonDecode(nextLogRecord.toString()))
+            as ServerLog);
+        nextLogRecord = null;
+      } else {
+        nextLogRecord.writeln(line);
+      }
+    } else if (line == logStartMarker) {
+      nextLogRecord = StringBuffer();
+    } else {
+      logHandler(ServerLog((b) => b
+        ..level = Level.INFO
+        ..message = line));
+    }
   });
 
   var daemonAction =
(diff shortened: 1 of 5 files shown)


build_daemon/lib/client.dart
```diff
@@ -43,10 +43,27 @@ Future<void> _handleDaemonStartup(
...
+  // and the `logEndMarker` as a `ServerLog`. Everything else is considered a
+  // normal INFO level log.
+  StringBuffer nextLogRecord;
   var sub = stdout.where((line) => !_isActionMessage(line)).listen((line) {
-    logHandler(ServerLog((b) => b
-      ..level = Level.INFO
-      ..message = line));
+    if (nextLogRecord != null) {
+      if (line == logEndMarker) {
+        logHandler(serializers.deserialize(jsonDecode(nextLogRecord.toString()))
```
grouma: We probably should put this deserialization in a try catch just in case.

build_daemon/lib/constants.dart
```diff
@@ -10,6 +10,16 @@ const readyToConnectLog = 'READY TO CONNECT';
 const versionSkew = 'DIFFERENT RUNNING VERSION';
 const optionsSkew = 'DIFFERENT OPTIONS';
 
+/// These are used when serializing logs messages over stdout.
```
grouma: [nit] serializing log* messages

build_daemon/lib/client.dart
jakemac53: done

build_daemon/lib/constants.dart
jakemac53: done

build_daemon/lib/client.dart
```diff
@@ -43,10 +43,35 @@ Future<void> _handleDaemonStartup(
 
   // The daemon may log critical information prior to it successfully
   // starting. Capture this data and forward to the logHandler.
+  //
+  // Whenever we see a `logStartMarker` we will parse everything between that
+  // and the `logEndMarker` as a `ServerLog`. Everything else is considered a
+  // normal INFO level log.
+  StringBuffer nextLogRecord;
```
natebosch: Would it be sensible to write a `StreamTransformer<String, String>` that collects between the start and end markers so that we separate out that logic from how we are handling log records?

build_daemon/lib/constants.dart
```diff
@@ -10,6 +10,16 @@ const readyToConnectLog = 'READY TO CONNECT';
...
 const optionsSkew = 'DIFFERENT OPTIONS';
 
+/// These are used when serializing log messages over stdout.
+///
+/// Serialized logs must be preceded with the [logStartMarker] on its own line,
+/// and terminated with a [logEndMarker] also on its own line.
+///
+/// This allows multi-line logs to be sanely serialized via stdout, and mixed
+/// with other generic messages.
+const logStartMarker = 'BUILD DAEMON LOG START';
```
natebosch: [nit] do we want something shorter?

jakemac53: Only a few logs actually are  handled this way, we went for uniqueness over efficiency here for now.

build_daemon/lib/client.dart
jakemac53: We could do that, it would also allow us to unit test it...

PR #2287, opened 2019-05-23
Docs for using Resolver in an aggregate builder

natebosch:


docs/writing_an_aggregate_builder.md
```diff
@@ -122,3 +122,27 @@ class ListAllFilesBuilder implements Builder {
...
+a build step is a `.dart` library, and then only for the code transitively
+imported by that library. If an aggregate builder needs to resolve Dart code
+form the inputs it globs then it needs to be split into two steps. Each of these
+steps can be implemented as a separate `builder_factory` (and `Builder` class)
+but roll into the same builder definition in `build.yaml` if they can both
+output to cache - otherwise for the final output to source they need to be
+separate builder definitions in `build.yaml` and use `runs_before` or
+`required_inputs` to ensure that they happen in the correct order.
+
+1. A `Builder` with `buildExtensions` of `{'.dart': ['.some_name.info']}`. Use
```
grouma: These cases seem unrelated to the above information. Maybe add a context sentence to better separate the details? e.g. "Usage scenarios:"

natebosch: These are the 2 builders that achieve the above goal.

jakemac53: This was confusing to me as well, maybe add a note on line 131 like `(see below)` to tie these together more.

Or maybe split the above paragraph in two, and move the second half (about build.yaml) below.

docs/writing_an_aggregate_builder.md
```diff
@@ -122,3 +122,27 @@ class ListAllFilesBuilder implements Builder {
...
+## Using a `Resolver`
+
+The `Resolver` provided by the build system only works when the primary input to
+a build step is a `.dart` library, and then only for the code transitively
+imported by that library. If an aggregate builder needs to resolve Dart code
+form the inputs it globs then it needs to be split into two steps. Each of these
+steps can be implemented as a separate `builder_factory` (and `Builder` class)
+but roll into the same builder definition in `build.yaml` if they can both
+output to cache - otherwise for the final output to source they need to be
+separate builder definitions in `build.yaml` and use `runs_before` or
```
jakemac53: We need to provide an example here I think, otherwise this introduces a lot of things the reader probably hasn't heard of. We could also link to docs for these options to help.

docs/writing_an_aggregate_builder.md
natebosch: Good call on splitting the paragraph in 2. I think it's more clear that way.

docs/writing_an_aggregate_builder.md
natebosch: I dropped the bit about `runs_before` and added an example for `required_inputs`. I think it's a bit more clear that way.

PR #2291, opened 2019-05-29
Refactor argument handling for DDC builder

natebosch:
- Use conditional and spread collection elements to replace some of the
  code that looked more iterative.
- Replace `transitiveDeps` with `transitiveJsModules` since we were
  otherwise always changing their extensions repeatedly.
- Drop `allDeps` variable which was otherwise an identical collection as
  `allAssetIds`.
- Make more arguments use the `--name=value` pattern so that it is clear
  how they are related as opposed to subsequent arguments.

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -109,11 +109,13 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   // These can be consumed for hot reloads.
   var moduleDigests = <String, String>{
     for (var jsId in transitiveJsModules)
-      jsId.path.replaceFirst('lib/', 'packages/${jsId.package}/'):
-          '${await buildStep.digest(jsId)}',
+      _moduleDigestKey(jsId): '${await buildStep.digest(jsId)}',
   };
   await buildStep.writeAsString(appDigestsOutput, jsonEncode(moduleDigests));
 }
+
+String _moduleDigestKey(AssetId jsId) =>
+    '${ddcModuleName(jsId)}$jsModuleExtension';
 
 final _lazyBuildPool = Pool(16);
 
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -29,9 +29,9 @@ Future<void> bootstrapDdc(BuildStep buildStep, {DartPlatform platform}) async {
       .decode(await buildStep.readAsString(moduleId)) as Map<String, dynamic>);
 
   // First, ensure all transitive modules are built.
-  List<Module> transitiveDeps;
+  List<AssetId> transitiveJsModules;
   try {
-    transitiveDeps = await _ensureTransitiveModules(module, buildStep);
+    transitiveJsModules = await _ensureTransitiveJsModules(module, buildStep);
   } on UnsupportedModules catch (e) {
     var librariesString = (await e.exactLibraries(buildStep).toList())
         .map((lib) => AssetId(lib.id.package,
@@ -68,8 +68,6 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   // Map from module name to module path for custom modules.
   var modulePaths = SplayTreeMap.of(
       {'dart_sdk': r'packages/build_web_compilers/src/dev_compiler/dart_sdk'});
-  var transitiveJsModules = [jsId]..addAll(transitiveDeps
-      .map((dep) => dep.primarySource.changeExtension(jsModuleExtension)));
   for (var jsId in transitiveJsModules) {
     // Strip out the top level dir from the path for any module, and set it to
     // `packages/` for lib modules. We set baseUrl to `/` to simplify things,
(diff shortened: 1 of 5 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -109,14 +107,11 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
...
-  for (var dep in transitiveDeps.followedBy([module])) {
-    var assetId = dep.primarySource.changeExtension(jsModuleExtension);
-    moduleDigests[
-            assetId.path.replaceFirst('lib/', 'packages/${assetId.package}/')] =
-        (await buildStep.digest(assetId)).toString();
-  }
-
+  var moduleDigests = <String, String>{
+    for (var jsId in transitiveJsModules)
+      jsId.path.replaceFirst('lib/', 'packages/${jsId.package}/'):
```
jakemac53: The wrapping here makes this a bit weird to follow, consider extracting the key into a function so it doesn't wrap?

Maybe at the same time make this more robust (check the path actually starts with lib before blindly replacing).

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -92,20 +92,21 @@ Future<void> _createDevCompilerModule(Module module, BuildStep buildStep,
...
+  var transitiveKernelDeps = [
+    for (var dep in transitiveDeps)
+      dep.primarySource.changeExtension(ddcKernelExtension)
+  ];
   var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
 
-  var allAssetIds = Set<AssetId>()
-    ..addAll(module.sources)
-    ..addAll(transitiveKernelDeps);
+  var allAssetIds = {...module.sources, ...transitiveKernelDeps};
```
jakemac53: consider an explicit `<AssetId>` type or something here? At a glance its not clear if this is a set or a map.

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -114,56 +115,33 @@ Future<void> _createDevCompilerModule(Module module, BuildStep buildStep,
       '-o',
       jsOutputFile.path,
       debugMode ? '--source-map' : '--no-source-map',
+      for (var dep in transitiveDeps) _summaryArg(dep),
```
jakemac53: :champagne: cc @munificent this might be one of the best ones I have seen yet...

build_web_compilers/lib/src/dev_compiler_builder.dart
natebosch: Sure.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
natebosch: Refactored to call `ddcModuleName` and extracted a method.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -109,12 +109,14 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
...
     for (var jsId in transitiveJsModules)
-      jsId.path.replaceFirst('lib/', 'packages/${jsId.package}/'):
-          '${await buildStep.digest(jsId)}',
+      _moduleDigestKey(jsId): '${await buildStep.digest(jsId)}',
   };
   await buildStep.writeAsString(appDigestsOutput, jsonEncode(moduleDigests));
 }
 
+String _moduleDigestKey(AssetId jsId) =>
+    '${ddcModuleName(jsId)}$jsModuleExtension';
```
jakemac53: `ddcModuleName` will also strip the extension, which I assume would break things here?

natebosch: that's why I'm putting it back... did I introduce a bug with this?

jakemac53: oh, didn't catch you were adding it back, sgtm

build_web_compilers/lib/src/dev_compiler_builder.dart
munificent: Yeah, it works so well for command line args that it almost seems tailor-made for them. :)

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -114,56 +115,33 @@ Future<void> _createDevCompilerModule(Module module, BuildStep buildStep,
...
+      '--multi-root=.',
+      '--track-widget-creation',
+      '--inline-source-map',
+      if (useIncrementalCompiler) ...[
+        '--reuse-compiler-result',
+        '--use-incremental-compiler',
+      ],
+      // The urls to compile, using the package: path for files under lib and the
+      // full absolute path for other files.
+      ...module.sources.map((id) {
```
jakemac53: nit: `for (var id in module.sources) _sourceArg(id)` (and extract out the logic), for consistency?

build_web_compilers/pubspec.yaml
```diff
@@ -1,11 +1,11 @@
 name: build_web_compilers
-version: 2.1.0
+version: 2.1.1-dev
 description: Builder implementations wrapping Dart compilers.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
 
 environment:
-  sdk: ">=2.3.0-dev.0.1 <3.0.0"
+  sdk: ">=2.3.0 <3.0.0"
```
kevmoo: This will block the current stable Flutter SDK from picking this up. 😢

jakemac53: we don't need to release anyways any time soon :man_shrugging: I don't think we should focus to much on it, this is the correct constraint.

build_web_compilers/lib/src/dev_compiler_builder.dart
natebosch: Good call

PR #2299, opened 2019-06-04
fix command line arg for build_modules

jakemac53:
This also fixes the failing test by adding a `hasMain` boolean to `ModuleLibrary` and using that instead of `isEntrypoint` when copying over modules.

### DIFF
--- a/build_modules/lib/src/module_library.dart
+++ b/build_modules/lib/src/module_library.dart
@@ -51,12 +51,16 @@ class ModuleLibrary {
   /// The `dart:` libraries that this library directly depends on.
   final Set<String> sdkDeps;
 
+  /// Whether this library has a `main` function.
+  final bool hasMain;
+
   ModuleLibrary._(this.id,
       {@required this.isEntryPoint,
       @required Set<AssetId> deps,
       @required this.parts,
       @required this.conditionalDeps,
-      @required this.sdkDeps})
+      @required this.sdkDeps,
+      @required this.hasMain})
       : _deps = deps,
         isImportable = true;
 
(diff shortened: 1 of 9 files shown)


grouma:
How do we make sure that failures are recognized as such?

jakemac53:
> How do we make sure that failures are recognized as such?

I can file a separate issue for that, but we need to have a non-zero exit code when there is an unrecognized command line arg, which must not be happening today?

grouma:
Is that not the case today:
https://github.com/dart-lang/build/blob/master/build_runner/bin/build_runner.dart#L37

jakemac53:
The issue with the exit code is a bug in the generated script, it assumes it is being ran as an isolate and never sets the exit code. The code you pointed to never gets ran if you run the script directly.

grouma:
What is the additional main logic for?

jakemac53:
> What is the additional main logic for?

Fixing the test so it actually passes - this has always been broken (we would see extra modules for some of these files that we don't expect otherwise, since they are entrypoints)

jakemac53:
Updated the description to highlight that change as well

build_modules/lib/src/module_library.dart
```diff
@@ -116,21 +121,25 @@ class ModuleLibrary {
         deps.add(linkedId);
       }
     }
+    // Allow two or fewer arguments so that entrypoints intended for use with
```
natebosch: Is this comment in the right place? Should it be on `_hasMainMethod` instead?

jakemac53: whoops ya, refactor screwup

Issue #2270, opened 2019-05-14
Cannot create link, path... A required privilege is not held by the client


cooler-king:
When a test build completes it crashes.

Dart SDK 2.3.0
Windows x64
WebStorm
Large project.

![image](https://user-images.githubusercontent.com/11008367/57666084-b4e20400-75cc-11e9-9ee7-0d48a41b1e16.png)

Other stuff gets created in the same place:
![image](https://user-images.githubusercontent.com/11008367/57666375-b65ffc00-75cd-11e9-9017-40e3e82de252.png)

Looks like `packages` folders under `AppData` get special treatment in Windows... or maybe just `test\packages`.

natebosch:
cc @grouma - did we somehow change the defaults for `useSymlinks` when we switched to including that in `target.outputLocation`?

https://github.com/dart-lang/build/blob/184f1f5479898731dede72c6286835d4b8b7aa3f/build_runner_core/lib/src/environment/io_environment.dart#L42-L46

jakemac53:
I think the problem is that we always have done a _directory_ symlink for the packages dir on windows, because anything else is really costly. That worked fine when we had a junctions implementation. Now it crashes if you aren't an admin.

natebosch:
I think the best we can do here is print a better error linking to an FAQ explaining that it needs to be run as admin.

We discussed the possibility of falling back to file copies, but that's going to be inefficient. We'd want a warning in that case.

cooler-king:
What about falling back to the junction implementation that used to work if it fails?  Being unable to run tests unless you run as admin is somewhere between inconvenient and a quality risk.  It breaks tools and messes with dev team worflows.

natebosch:
> What about falling back to the junction implementation that used to work if it fails?

The SDK doesn't give us the option. We could request that the SDK falls back, but it's unlikely they'll want to do that. We could also investigate shelling out to create a junction, but it's a fair bit of complexity and could be brittle.

We haven't had many complaints so far and the SDK migrated off junction around 2.1 I think, so I'm not sure if that's a signal that most users are running with admin anyway.

jakemac53:
It looks like enabling developer mode solves the issue (see https://docs.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development).

I am going to add some explicit error handling for this pointing to those docs, and then close this issue.

kam-ox:
[...]

PR #2303, opened 2019-06-05
handle symlink creation failures, link to dev mode docs

jakemac53:
Closes https://github.com/dart-lang/build/issues/2270

Below is some example output:

```
> pub run build_runner build -o build
[INFO] Generating build script completed, took 451ms
[INFO] Reading cached asset graph completed, took 260ms
[INFO] Checking for updates since last build completed, took 919ms
[INFO] Running build completed, took 264ms
[INFO] Caching finalized dependency graph completed, took 166ms
[SEVERE] Unable to create symlink build\web\packages. Note that to create symlinks on windows you need to either run in a console with admin priviledges or enable developer mode (see https://docs.microsoft.com/en-us/windows/uwp/get-started/enable-your-device-for-development).
[SEVERE] Unable to create merged directory for build.
[SEVERE] Failed after 2.1s
```

### DIFF
--- a/build_runner_core/lib/src/environment/create_merged_dir.dart
+++ b/build_runner_core/lib/src/environment/create_merged_dir.dart
@@ -141,7 +141,7 @@ Future<bool> _createMergedOutputDir(
 
     return true;
   } on FileSystemException catch (e) {
-    if (e.osError.errorCode != 1314) rethrow;
+    if (e.osError?.errorCode != 1314) rethrow;
     var devModeLink =
         'https://docs.microsoft.com/en-us/windows/uwp/get-started/'
         'enable-your-device-for-development';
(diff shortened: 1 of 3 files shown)


build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -88,57 +88,68 @@ Future<bool> _createMergedOutputDir(
...
+      var paths = outputAssets.map((id) => id.path).toList()..sort();
+      var content = paths.join(_manifestSeparator);
+      await _writeAsString(
+          outputDir, AssetId(packageGraph.root.name, _manifestName), content);
+    });
 
-  return true;
+    return true;
+  } on FileSystemException catch (e) {
+    if (e.osError.errorCode != 1314) rethrow;
```
grouma: None of these members can be null correct?

jakemac53: good catch, https://api.dartlang.org/stable/2.3.1/dart-io/FileSystemException/osError.html

if only the language had a way to express that at the type system level... :D

PR #2306, opened 2019-06-06
Pass target name to kernel worker

### LINKED ISSUE #2174: Support passing `--target` to the kernel worker
Today the kernel worker will make an assumption about the target depending on whether summaries are used. We want to make this explicit so that the two settings are not tied together.

It would be handy if this matched the platform name as we already use in the module builder - but there is an incompatibility. The on the build side we call it `ddc`, while on the kernel worker side we call it `devcompiler`

Some options:
- Make `ddc` an allowed alias for `devcompiler` in the kernel worker. Not breaking.
- Add `kernelTarget` to the `DartPlatform`. Breaking in `build_modules`.
- Add `kernelTarget` to the constructor for `KernelBuilder`. Breaking in `build_modules`.
- Hack together a translation between the two in `KernelBuilder`.
- Rename `ddc` to `devcompiler` in `build_web_compilers`. Breaking in `build_web_compilers` (I think).

natebosch:
Closes #2174

This will be more reliable and surface errors more quickly than relying
on the defaulting depending on other arguments like `--summary-only`.

### DIFF
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -303,6 +303,7 @@ Future<void> _addRequestArguments(
     '--multi-root-scheme=$multiRootScheme',
     '--exclude-non-sources',
     summaryOnly ? '--summary-only' : '--no-summary-only',
+    '--target=${platform.name}',
     '--libraries-file=${p.toUri(librariesPath)}',
     if (useIncrementalCompiler) ...[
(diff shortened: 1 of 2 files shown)


natebosch:
We should be good with `2.3.0`. The name was changed in the SDK in https://github.com/dart-lang/sdk/commit/d886c85d3d804a6aee1e2535960402ff18bd5b12 and that should be in `2.3.0`

Issue #2161, opened 2019-04-10
build_runner watch stops detecting changes on Windows


dergemkr:
When I run the `pub run build_runner watch -o build` on Windows for an Angular project, it will stop detecting changes after one or two successful builds. This is not an issue on Linux - I've tested with the same project and Dart 2.2 on Linux and it continues to detect changes.

The command will usually get through one or two successful builds before it stops detecting changes, but interestingly, it will continue to detect an indefinite number of changes while the build continues to fail. Maybe it has something to do with writing the output directory? 

* Dart SDK Version: Dart VM version: 2.2.0 (Tue Feb 26 15:04:32 2019 +0100) on "windows_x64"
* Package Versions:
  * build: 1.1.0
  * build_runner: 1.2.3
  * build_web_compilers: 1.2.0

To ensure it wasn't specific to my project, I downloaded the [TOH-6 example](https://github.com/angular-examples/toh-6), upgraded build_web_compilers to 1.2.0and ran the above command on the project. I tested making changes to `web/main.dart` and `lib/app_component.dart` to see if changes were detected, and they were detected so long as the build failed.

jakemac53:
On windows it uses a polling file watcher I believe (we just use package:watcher). It wouldn't surprise me if that combined with `-o` is really slow - but I wouldn't expect it to stop working entirely :/

jakemac53:
Ok so actually it looks like we are explicitly choosing the polling watcher on windows https://github.com/dart-lang/build/blob/2fb938207512a4355a5d2e0af66a6bc46488dc08/build_runner/lib/src/generate/directory_watcher_factory.dart#L13 

I don't know if the linked issues are relevant still or not.

dergemkr:
I suppose I should have clarified: By stops, I mean that it doesn't expose a crash message. There's just no subsequent builds after the build succeeded message. I just tested, and so long as "really slow" is not > 5 minutes, I don't think it's a delay.

I'm really liking this customizable build logic in Dart 2, though! I had always wanted this while working on Angular Dart < 4.

jakemac53:
Interestingly it looks like we switched to a PollingDirectoryWatcher for exactly this reason (see https://github.com/dart-lang/build/issues/1031)?

I wonder if somehow you aren't picking up the polling directory watcher?

dergemkr:
That issue does seem related - it continues to watch if I run with output directory `../build`. How do I determine if I'm using the polling watcher or not?

jakemac53:
[...]

PR #2307, opened 2019-06-07
use a polling watcher on windows again

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2161

I am not sure exactly when/why we stopped overriding this, only the daemon command was using the polling watcher on windows and the other commands were not.


jakemac53:
@dergemkr can you confirm this resolves your issue?

dergemkr:
I've tested this fix and it resolves #2161. Thank you!

PR #2308, opened 2019-06-12
Fix source map rewriting to ignore raw paths with no scheme

jakemac53:
In preparation for https://dart-review.googlesource.com/c/sdk/+/105703, today the logic is very naive and would overwrite uris that it shouldn't touch.

Eventually we could delete this logic entirely, as once that change has landed and been released we will no longer need to touch the source maps at all.


jakemac53:
cc @Markzipan

PR #2312, opened 2019-06-13
[WIP] Example of providing invalidated files through daemon client

jonahwilliams:
TBD corresponding changes to build_runner_core to disable watching.

### DIFF
--- a/build_daemon/lib/src/server.dart
+++ b/build_daemon/lib/src/server.dart
@@ -74,7 +74,13 @@ class Server {
         if (request is BuildTargetRequest) {
           _buildTargetManager.addBuildTarget(request.target, channel);
         } else if (request is BuildRequest) {
-          await _build(_buildTargetManager.targets, <WatchEvent>[]);
+          var events = <WatchEvent>[];
+          if (request.invalidatedFiles != null) {
+            for (var filePath in request.invalidatedFiles) {
+              events.add(WatchEvent(ChangeType.ADD, filePath));
+            }
+          }
+          await _build(_buildTargetManager.targets, events);
         }
       }, onDone: () {
         _removeChannel(channel);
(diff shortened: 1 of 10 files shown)


build_daemon/lib/src/server.dart
```diff
@@ -74,7 +74,13 @@ class Server {
         if (request is BuildTargetRequest) {
           _buildTargetManager.addBuildTarget(request.target, channel);
         } else if (request is BuildRequest) {
-          await _build(_buildTargetManager.targets, <WatchEvent>[]);
+          var events = <WatchEvent>[];
+          if (request.invalidatedFiles != null) {
+            for (var filePath in request.invalidatedFiles) {
+              events.add(WatchEvent(ChangeType.ADD, filePath));
```
jakemac53: We discussed a bit earlier but we will need to figure out what type of change event to provide here.

We might actually need to go investigate all these files and compare them to our last state to create a proper change type?

jonahwilliams: We could also provide the change type ourselves, the flutter tool will know if something is modified/added, though we don't explicitly track that currently

jakemac53: I suppose then the protocol would change to add a change type? That would probably be idea from our perspective, as it matches what we need to pass to build runner.

jakemac53:
After some further thoughts and in person discussion, I think we don't actually want to provide the list of invalidated sources directly. Instead, we can add the "build on demand" mode. In that mode we will not set up file watchers, and would do our own file system scan for changes on demand whenever asked to do a build. We already have this logic since its what we do for the first build.

We should also expose this as a general mode for webdev/build_runner. This is likely to give a better experience for windows users who are seeing their cpu get utilized to 100% by the polling file watcher.

jonahwilliams:
Closing this since it is not the approach we agreed upon

PR #2317, opened 2019-06-19
Shutdown notification

grouma:
- Add new `failureType` field to `ShutdownNotification`
- Watch for build script updates and send corresponding `ShutdownNotification`

Closes https://github.com/dart-lang/build/issues/2094


grouma:
Discussed offline. It would be nice to expose static `FailureType`s. However these values do not make sense to reside in `package:build_daemon` and `package:webdev` does not directly depend on `package:build_runner` where they belong. Leaving it as is.

PR #2328, opened 2019-06-25
Ensure the `pub run test` process exits.

### LINKED ISSUE #2327: Test command can leave the child `pub run test` process running in background
* Dart SDK Version: 2.3.2 and 2.5.0-dev.0.0
* Build package versions:
  * build_runner: 1.6.0
  * build_runner_core: 3.0.6
  * build_test: 0.10.8
* MacOSX

---

The `pub run test` command consumes the first `sigint` instance in order to give the test runner a chance to finish in-progress work (new test suites are prevented from starting, in-progress suites are allowed to finish). However, the `pub run build_runner test` command will exit on the first instance of `sigint`. The result is that interrupting a `build_runner test` process can leave the `pub run test` process running in the background.

A repro can be found here: https://github.com/evanweible-wf/build_runner_test_hanging_test_process

---

Locally, I was able to update the `TestCommand` at `build_runner/lib/src/entrypoint/test.dart` like so:

```diff
   Future<int> _runTests(String precompiledPath) async {
     stdout.writeln('Running tests...\n');
     var extraTestArgs = argResults.rest;
     var testProcess = await Process.start(
         pubBinary,
         [
           'run',
           'test',
           '--precompiled',
           precompiledPath,
         ]..addAll(extraTestArgs),
         mode: ProcessStartMode.inheritStdio);
-    return testProcess.exitCode;
+    StreamSubscription sigintSub;
+    sigintSub = ProcessSignal.sigint.watch().listen((signal) async {
+      testProcess.kill(signal);
+      await testProcess.exitCode;
+      await sigintSub?.cancel();
+      sigintSub = null;
+    });
+    final testExitCode = await testProcess.exitCode;
+    unawaited(sigintSub?.cancel());
+    return testExitCode;
   }
 }
```

and this resolved the issue. The sigint received by the build_runner process gets explicitly forwarded to the test process, and the handler doesn't return until the test process exits, which in turn keeps the build_runner process open until that happens. As a result, you can interrupt the build_runner process with `ctrl+c` and if the test process consumes that event, the build_runner process will transitively do the same.

evanweible-wf:
The `build_runner test` command starts a child `pub run test` process which in some cases consumes the first SIGINT signal without exiting. If the build_runner process doesn't wait for this child process to exit, it may remain running in the background.

The `build_runner test` command now watches for the exit signals (SIGINT and SIGTERM) and prevents its own process from exiting until the child `pub run test` process exits, which should happen after the second SIGINT signal.

@jakemac53 I'm not very familiar with the testing setup in this repo and I'm not sure if there's a clear way to verify that the child process exits via a unit/integration test. Any ideas?

Fixes #2327.

### DIFF
--- a/build_runner/lib/src/entrypoint/test.dart
+++ b/build_runner/lib/src/entrypoint/test.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:args/command_runner.dart';
+import 'package:async/async.dart';
 import 'package:build_runner_core/build_runner_core.dart';
 import 'package:io/io.dart';
 import 'package:path/path.dart' as p;
@@ -135,7 +136,9 @@ class TestCommand extends BuildRunnerCommand {
           precompiledPath,
         ]..addAll(extraTestArgs),
         mode: ProcessStartMode.inheritStdio);
-    return testProcess.exitCode;
+    _ensureProcessExit(testProcess);
+    final testExitCode = await testProcess.exitCode;
+    return testExitCode;
   }
 }
 
@@ -153,6 +156,21 @@ void _ensureBuildTestDependency(PackageGraph packageGraph) {
     throw _BuildTestDependencyError();
   }
 }
+
+void _ensureProcessExit(Process process) {
+  var signalsSub = _exitProcessSignals.listen((signal) async {
+    await process.exitCode;
+  });
+  process.exitCode.then((_) {
+    signalsSub?.cancel();
+    signalsSub = null;
+  });
+}
+
+Stream<ProcessSignal> get _exitProcessSignals => Platform.isWindows
+    ? ProcessSignal.sigint.watch()
+    : StreamGroup.merge(
+        [ProcessSignal.sigterm.watch(), ProcessSignal.sigint.watch()]);
 
 class _BuildTestDependencyError extends StateError {
   _BuildTestDependencyError() : super('''
--- a/build_runner/lib/src/entrypoint/test.dart
+++ b/build_runner/lib/src/entrypoint/test.dart
@@ -137,8 +137,7 @@ class TestCommand extends BuildRunnerCommand {
         ]..addAll(extraTestArgs),
         mode: ProcessStartMode.inheritStdio);
     _ensureProcessExit(testProcess);
-    final testExitCode = await testProcess.exitCode;
-    return testExitCode;
+    return testProcess.exitCode;
   }
 }
 
@@ -159,6 +158,7 @@ void _ensureBuildTestDependency(PackageGraph packageGraph) {
 
 void _ensureProcessExit(Process process) {
   var signalsSub = _exitProcessSignals.listen((signal) async {
+    stdout.writeln('waiting for subprocess to exit...');
     await process.exitCode;
   });
   process.exitCode.then((_) {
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/entrypoint/test.dart
```diff
@@ -135,7 +136,9 @@ class TestCommand extends BuildRunnerCommand {
           precompiledPath,
         ]..addAll(extraTestArgs),
         mode: ProcessStartMode.inheritStdio);
-    return testProcess.exitCode;
+    _ensureProcessExit(testProcess);
+    final testExitCode = await testProcess.exitCode;
+    return testExitCode;
```
jakemac53: nit: you can just `return testProcess.exitCode;` here.

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -154,6 +157,21 @@ void _ensureBuildTestDependency(PackageGraph packageGraph) {
   }
 }
 
+void _ensureProcessExit(Process process) {
+  var signalsSub = _exitProcessSignals.listen((signal) async {
+    await process.exitCode;
```
jakemac53: nit: consider logging something to stdout here `stdout.writeln('waiting for subprocess to exit...');`

build_runner/lib/src/entrypoint/test.dart
evanweible-wf: whoops, removed some code I had here previously. Thanks :)

evanweible-wf:
@jakemac53 testing this locally, yes multiple signals are sent to the child process (presumably because of `inheritStdio`). If I run `pub run build_runner test` in that repro package, wait for it to start running tests, and then ctrl+c twice very quickly, both processes exit immediately and the test process does _not_ log the "Press ctrl+c again to exit immediately" line, which makes sense because they delay that print statement by 1 second and cancel it if a second SIGINT is received first.

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -154,6 +156,22 @@ void _ensureBuildTestDependency(PackageGraph packageGraph) {
   }
 }
 
+void _ensureProcessExit(Process process) {
+  var signalsSub = _exitProcessSignals.listen((signal) async {
+    stdout.writeln('waiting for subprocess to exit...');
+    await process.exitCode;
```
jakemac53: @natebosch brought up that we actually don't need this await here either :)

natebosch: Yeah I think this can be:

```dart
var signalsSub = _exitProcessSignals.listen((_) {
    stdout.writeln('waiting for subprocess to exit...');
});
```

evanweible-wf: Fixed in 9c1707f

evanweible-wf:
I'm not sure what is causing the CI failure. `dartanalyzer --fatal-infos --fatal-warnings .` in the `build_runner/` directory on 2.5.0-dev.0.0 passes for me locally

natebosch:
@evanweible-wf - Travis looks like it is red before your change. Should be fixed by https://github.com/dart-lang/build/pull/2329

evanweible-wf:
@jakemac53 @natebosch fyi this has a passing CI now that #2329 merged

PR #2329, opened 2019-06-25
Add ignores for deprecated analyzer APIs

natebosch:
We will need to migrate these to newer APIs or rewrite. For now unblock
submits by adding ignores.


jakemac53:
cc @stereotype441

PR #2331, opened 2019-06-27
Cache modules while writing them

natebosch:
Today every module which is read must be deserialized at least once from
the asset. If the module was also written during the same build we can
skip this step by writing to the cache directly in addition to the
asset.

- Update the interface to operate on `List<int>` instead of a `Map` and
  let the client deal with all aspects of serialization. This will be
  useful if we want to promote this class to `package:build`.
- Add a `toBytes` argument and a `write` method.
- Write through the cache in `ModuleBuilder` and `MetaModuleBuilder`.

### DIFF
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build
-version: 1.1.4
+version: 1.1.5-dev
 description: A build system for Dart.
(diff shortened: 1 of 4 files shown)


natebosch:
Yeah I'd like to explore digest based invalidation as well...

PR #2335, opened 2019-06-28
Fix per breaking change to HttpRequest and HttpClientResponse

tvolkert:
A recent change to the Dart SDK updated `HttpRequest` and
`HttpClientResponse` from implementing `Stream<List<int>>` to
implementing `Stream<Uint8List>`.

This forwards-compatible change prepares for that SDK breaking
change by casting the Stream to `List<int>` before transforming
it.

dart-lang/sdk#36900


tvolkert:
@jakemac53 found another one.

jakemac53:
Thanks for proactively finding that! I guess we only run our integration tests against the bleeding edge, seems like we should maybe change that.

Issue #2314, opened 2019-06-14
Add a flag to disable auto building


jonahwilliams:
Instead of watching files and eagerly building, we should expose a flag that a) disables file watching and debouncing and b) only starts a build when specifically requested via the daemon. This is to integrate with the flutter_tools resident runner.

grouma:
There is a concern that tools provide an incomplete list of sources. The alternative approach is to allow the build_daemon to continue watching for changed files but instead support a mode of manually triggering the builds.

jonahwilliams:
Yes, i think that approach would be perfectly acceptable as well

jakemac53:
On windows for now at least the watcher comes at a very large cost because of the polling watcher. At least on windows we should probably do a regular file scan, and I would prefer to only deal with one code path so that makes me lean towards the file scan on demand approach.

jonahwilliams:
That's a good point, i would also be concerned that with a polling watcher we might "miss" events unless there was a sufficient debounce period, but that would probably slow the whole thing down again

PR #2336, opened 2019-06-28
Manual build mode support

grouma:
Closes https://github.com/dart-lang/build/issues/2314

- Create a public entrypoint for backend implementations of the daemon protocol.
  - Refer to `lib/daemon.dart`.
- Update client `connect` method to now take an optional `buildMode`.
  The default mode is auto in which builds will automatically occur on changes.
  The alternative mode is manual in which builds will only occur when triggered
  with the client `startBuild` method.
- Add enum of build modes to `constants.dart`.
- Update `build_runner` to support the latest `build_daemon`.
- Add tests for each of the modes


build_daemon/lib/change_provider.dart
```diff
@@ -0,0 +1,21 @@
...
+abstract class ChangeProvider {}
+
+/// A provider of changes which are manually collected right before a build.
+abstract class ManualChangeProvider implements ChangeProvider {
+  Future<List<WatchEvent>> collectChanges();
+}
+
+/// A provider of changes which automatically cause builds to occur.
+abstract class AutoChangeProvider implements ChangeProvider {
+  /// If multiple files change together then they should be sent in the same
```
natebosch: [nit] this isn't phrased as a doc comment.

https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs
https://dart.dev/guides/language/effective-dart/documentation#do-start-doc-comments-with-a-single-sentence-summary

build_daemon/lib/change_provider.dart
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2019, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+
+import 'package:watcher/src/watch_event.dart';
+
+abstract class ChangeProvider {}
```
natebosch: I'd be more comfortable if both methods were defined here and we have a doc comment that says it's fine for one method to always return "empty" if the other method encompasses all changes.

In the long term we can also consider going through the `collectChanges` path even if the build was triggered by something on the `changes` stream with the intent of hypothetical use cases like "we auto rebuild if you edit in `lib/`, but if you edit somewhere else you need to manually trigger a build"

build_daemon/lib/change_provider.dart
grouma: Done.

build_daemon/lib/change_provider.dart
grouma: Done.

Issue #2257, opened 2019-05-08
Use new BuildConfig.parse({configYamlPath}) when build_config 0.4.1 is released


kevmoo:
https://github.com/dart-lang/build/blob/93b9f3bdbbe7e94f13cce15a8e5909b72fe422b6/build_runner/lib/src/package_graph/build_config_overrides.dart#L37

Will give more helpful errors when parsing fails.

kevmoo:
CC @natebosch RE https://github.com/dart-lang/build/pull/2247

natebosch:
Was fixed in https://github.com/dart-lang/build/pull/2386

PR #2341, opened 2019-07-03
Prepare to publish build_config

natebosch:
Towards #2257

Will publish and then bump versions from other packages.


natebosch:
Will comment out dependency overrides before publishing.

PR #2344, opened 2019-07-04
Invalidate cache by digest instead of pessimistic

natebosch:


build_modules/lib/src/module_cache.dart
```diff
@@ -53,9 +60,26 @@ class DecodingCache<T> {
...
+      final entry = _cached[id];
+      if (entry.needsCheck) {
+        final digest = await reader.digest(id);
+        if (digest != entry.digest) {
+          entry
+            ..needsCheck = false
+            ..digest = digest
+            ..value = Result.capture(reader.readAsBytes(id).then(_fromBytes));
+        } else {
+          entry.needsCheck = false;
```
jakemac53: nit: move this out of the else and don't assign it in the if (it should unconditionally be done)

build_modules/lib/src/module_cache.dart
```diff
@@ -53,9 +60,24 @@ class DecodingCache<T> {
   /// content dependencies will be tracked through [reader].
   Future<T> find(AssetId id, AssetReader reader) async {
     if (!await reader.canRead(id)) return null;
-    var result = _cached.putIfAbsent(
-        id, () => Result.capture(reader.readAsBytes(id).then(_fromBytes)));
-    return Result.release(result);
+    if (_cached.containsKey(id)) {
+      final entry = _cached[id];
+      if (entry.needsCheck) {
+        final digest = await reader.digest(id);
```
natebosch: We might want to put the digest into a `Future<Result>` as well...

I think we could end up doing some duplicate work here if two builders ask for the same content that needs to be checked - both will try to get the digest and both will read and do the serialization...

I'll need to give this some thought.

kevmoo:
> This needs a changelog entry and ideally an update to the doc comment on the function!

Ugh. Wrong issue. Sorry!

build_modules/lib/src/module_cache.dart
```diff
@@ -64,6 +86,15 @@ class DecodingCache<T> {
...
     await writer.writeAsBytes(id, _toBytes(instance));
-    _cached[id] = Result.capture(Future.value(instance));
+    _cached[id] = _Entry()
+      ..needsCheck = false
+      ..value = Result.capture(Future.value(instance));
   }
 }
+
+/// A cached [T] and it's digest that may not need to be checked.
+class _Entry<T> {
```
kevmoo: DBR: Add a ctor, make fields final, avoid silly bugs later 😄

natebosch:
I'm going to try this again on a fresh branch.

PR #2348, opened 2019-07-10
WIP: use platform specific paths for the sdk dill file always

jakemac53:


natebosch:
@jakemac53 - where are we at with this change? Still need it?

jakemac53:
This is a nice to have from the perspective of removing hacks, I am going to re-run and if things are happy we might as well submit I think.

jakemac53:
We resolved this in other ways and I don't want to risk further breakage so I am just going to close this

PR #2350, opened 2019-07-10
Write a dill manifest instead of concatenating

natebosch:
The resulting file will be smaller. The SDK supports relative paths to
kernel files as of https://github.com/dart-lang/sdk/commit/b0a11c9d255229759bbb0c7a50fd342f29a639f2

- Migrate to for loop elements instead of iteratively building up a
  list.
- Use `canRead` to ensure that all the transitive modules are available
  and otherwise get the same errors as we would get with the old
  approach.
- Write a text file with the relative paths of the transitive modules.


natebosch:
This would mean that it's no longer possible to copy a _single_ .dill file and run with the VM...

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
```diff
@@ -63,16 +64,19 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
...
-        module.primarySource.changeExtension(vmKernelModuleExtension)
-      ].followedBy(transitiveModules.map(
-          (m) => m.primarySource.changeExtension(vmKernelModuleExtension)));
-      var appContents = <int>[];
-      for (var dependencyId in transitiveKernelModules) {
-        appContents.addAll(await buildStep.readAsBytes(dependencyId));
-      }
-      await buildStep.writeAsBytes(
+        module.primarySource.changeExtension(vmKernelModuleExtension),
+        for (var dep in transitiveModules)
```
jakemac53: :tada:

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
```diff
@@ -63,16 +64,19 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
...
+          dep.primarySource.changeExtension(vmKernelModuleExtension),
+      ];
+      await Future.wait(transitiveKernelModules.map(buildStep.canRead));
```
jakemac53: This won't actually throw it just returns a `List<bool>` - we would need to do a `.any` on the result and throw.

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
```diff
@@ -63,16 +64,19 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
...
+      var appContents = [
+        '#@dill',
+        for (var dependencyId in transitiveKernelModules)
+          p.relative(dependencyId.path, from: buildStep.inputId.path),
```
jakemac53: We should use `p.url` here so that the output is hermetic. Hopefully that works in the vm on windows..............

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
natebosch: I added a `log.severe`. I'm not sure the best way to surface this...

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
natebosch: We'll need to verify before submitting - I'm not sure if we have a test for that that runs on appveyor...

natebosch: No test running on appveyor 😞

natebosch:
I don't know what happened here, I can't get this to work with the Dart vm anymore...

natebosch:
Blocked by https://github.com/dart-lang/sdk/issues/37610

build_vm_compilers/lib/src/vm_entrypoint_builder.dart
kevmoo: There is no end to the delight in this feature CC @munificent

jakemac53:
Did we want to push forward with this or just close?

natebosch:
I think it's worth revisiting. I haven't tried this since the fix for https://github.com/dart-lang/sdk/issues/37610 went in.

natebosch:
I'm not sure if we are running tests in the SDK that verify the VM supports running these files, and at some point (potentially with the Dart CLI work? cc @devoncarew ) it looks like that's been broken. Trying to run a file that starts with exactly the characters `#@dill\n` is not being handled correctly and it's trying to parse the file as Dart code instead.

Since we've deprioritized `build_vm_compilers` and the VM doesn't seem to have a high priority on dill manifest support in general I think we can close this out.

Issue #2356, opened 2019-07-12
BuildRunnerDaemonBuilder should respect --skip-build-script-checks 


jonahwilliams:
In https://github.com/dart-lang/build/pull/2317/files#diff-d93edda0ea4440c7871bf77eb8bd3017R72 a check was added for a build script check. This will cause a daemon spawned by the flutter_tool to never successfully start, since we don't require the project to have a build_runner/daemon dependency.

instead we'd need something like
```
    if (!_buildOptions.skipBuildScriptCheck && _builder.buildScriptUpdates
        .hasBeenUpdated(changes.map<AssetId>((change) => change.id).toSet())) {
      _buildScriptUpdateCompleter.complete();
      return;
    }
```

grouma:
On it.

PR #2359, opened 2019-07-12
Respect skip build script configuration

grouma:
Closes https://github.com/dart-lang/build/issues/2356

### DIFF
--- a/build_runner/test/daemon/daemon_test.dart
+++ b/build_runner/test/daemon/daemon_test.dart
@@ -13,6 +13,7 @@ import 'package:build_daemon/client.dart';
 import 'package:build_daemon/constants.dart';
 import 'package:build_daemon/data/build_status.dart';
 import 'package:build_daemon/data/build_target.dart';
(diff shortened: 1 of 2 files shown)


build_runner/test/daemon/daemon_test.dart
```diff
@@ -66,9 +67,11 @@ main() {
     await runPub('a', 'run', args: ['build_runner', 'clean']);
   });
 
-  Future<BuildDaemonClient> _startClient({BuildMode buildMode}) {
+  Future<BuildDaemonClient> _startClient(
+      {BuildMode buildMode, List<String> options}) {
+    options ??= [];
     buildMode ??= BuildMode.Auto;
-    var args = ['run', 'build_runner', 'daemon'];
+    var args = ['run', 'build_runner', 'daemon', ...options];
```
jakemac53: This would have the downside it seems of causing options skew if some other build daemon client wants to run as well, but I guess that would be by design in this case?

grouma: Yes.

Issue #2313, opened 2019-06-14
Expose daemonPort as getter/method on Daemon client.


jonahwilliams:
...This is required for the flutter_tool to pass through to webdev eventually. Since I believe the client should know the port already we could expose it as a getter/setter.

jonahwilliams:
Err, not as a setter - that would make no sense :)

jakemac53:
Ah so the asset server port (which I think is the specific thing you need here) is not a general build daemon concept (its build_runner specific). That complicates this a little bit.

grouma:
Do we need the asset port:
https://github.com/dart-lang/build/blob/master/build_runner/lib/src/entrypoint/daemon.dart#L91

Or the build daemon port:
https://github.com/dart-lang/build/blob/master/build_daemon/lib/client.dart#L22

grouma:
If we need to expose the asset port we will need to do that through `package:build_runner` as this is not a `build_daemon` concept.

If we need to expose the `daemon` port then we can easily modify the client API for `package:build_daemon`. 

@jonahwilliams can you confirm what is required?

jonahwilliams:
Sorry for the delay!

What is needed is the `daemonPort` that web dev requires here: https://github.com/dart-lang/webdev/pull/440/files#diff-b84db82f5fe57b165c3d044fa5916e5bL24

This can be grabbed via the utility method in https://github.com/flutter/flutter/pull/34252/files#diff-01a7653ade6cc1a5182fdebf929ad64bR496

(I could also keep using the above utility method)

jakemac53:
That is a poorly named parameter, its really the asset server port :).

PR #2360, opened 2019-07-12
Expose assetServerPort

grouma:
Closes https://github.com/dart-lang/build/issues/2313

### DIFF
--- a/build_runner/lib/src/daemon/constants.dart
+++ b/build_runner/lib/src/daemon/constants.dart
@@ -2,8 +2,19 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:io';
+
 import 'package:build_daemon/constants.dart';
 import 'package:path/path.dart' as p;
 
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/daemon/constants.dart
```diff
@@ -2,8 +2,19 @@
...
 import 'package:path/path.dart' as p;
 
 String assetServerPortFilePath(String workingDirectory) =>
     p.join(daemonWorkspace(workingDirectory), '.asset_server_port');
+
+/// Returns the port of the daemon asset server.
+int assetServerPort(String workingDirectory) {
+  var portFile = File(assetServerPortFilePath(workingDirectory));
+  if (!portFile.existsSync()) {
+    throw Exception('Unable to read daemon asset port file.');
```
natebosch: Is there something more specific we can use than `Exception`?

grouma: Not anything defined currently. But we should do something similar to:
https://github.com/dart-lang/build/blob/master/build_daemon/lib/client.dart#L191

Issue #2346, opened 2019-07-08
Support latest pkg:analyzer


kevmoo:
Need at least

- [ ] build
- [ ] build_modules
- [ ] build_resolvers
- [ ] build_web_compilers
- [ ] build_vm_compilers

jakemac53:
All of these are now published

PR #2361, opened 2019-07-12
Allow analyzer 0.37.0 in all packages

jakemac53:
Overriding the versions for now so we can run tests with the new version, will push another commit that removes the overrides if everything is happy.

Closes https://github.com/dart-lang/build/issues/2346

### DIFF
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -12,8 +12,6 @@ dev_dependencies:
     path: pkgs/provides_builder/
 
 dependency_overrides:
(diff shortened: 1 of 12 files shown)


_test/pubspec.yaml
```diff
@@ -34,3 +34,4 @@ dependency_overrides:
     path: ../build_web_compilers
   scratch_space:
     path: ../scratch_space
+
```
natebosch: but why? https://github.com/dart-lang/build/pull/2366

PR #2366, opened 2019-07-18
Drop trailing newline

natebosch:
No need for a blank line at the end of any text file, and this triggers
a warning in my git prechecks.

### DIFF
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -34,4 +34,3 @@ dependency_overrides:
     path: ../build_web_compilers
   scratch_space:
     path: ../scratch_space
-


jakemac53:
This is a common practice for editors to always insert newlines and I don't
know all the arguments but they generally do more good than harm. GitHub
also complains about files with no newline with a visual indication in the
file.

natebosch:
> always insert newlines

The problem isn't the trailing newline, it's the _blank_ line followed by a trailing newline. That is the file previously ended with `something\n\n` and now it ends with `something\n`

jakemac53:
Oh ok I only read the title (on the phone) :D.

natebosch:
I probably should have said "blank line" instead of "newline"

jakemac53:
I thought vscode would actually auto-delete these too (basically it would enforce exactly one newline) but I guess that must have been a setting I had on before that I need to turn back on.

PR #2372, opened 2019-07-23
DDC entrypoint update

Markzipan:
An upcoming SDK release changes DDC's entrypoints to be package-relative, which will break compatibility with older build_web_compilers versions. This change makes build_web_compilers compatible with both the old and new entrypoints. 

See: https://dart-review.googlesource.com/c/sdk/+/108942

Fixes https://github.com/dart-lang/build/issues/2395

### DIFF
--- a/build_web_compilers/lib/src/ddc_names.dart
+++ b/build_web_compilers/lib/src/ddc_names.dart
@@ -2,10 +2,22 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// Escape [name] to make it into a valid identifier.
+import 'package:path/path.dart' as _p;
+
+/// Transforms a path to a valid JS identifier.
 ///
-/// This is copied from
-/// https://github.com/dart-lang/sdk/blob/300fd66f5c5e7497674ea4dd9fd21be91ee3c71e/pkg/dev_compiler/lib/src/analyzer/code_generator.dart#L6390
+/// This logic must be synchronized with [pathToJSIdentifier] in DDC at:
+/// pkg/dev_compiler/lib/src/compiler/module_builder.dart
+String pathToJSIdentifier(String path) {
+  path = _p.normalize(path);
+  if (path.startsWith('/') || path.startsWith('\\')) {
+    path = path.substring(1, path.length);
+  }
+  return toJSIdentifier(
+      path.replaceAll('\\', '__').replaceAll('/', '__').replaceAll('..', '__'));
+}
+
+/// Escape [name] to make it into a valid identifier.
 String toJSIdentifier(String name) {
   if (name.isEmpty) return r'$';
 
--- a/build_web_compilers/lib/src/ddc_names.dart
+++ b/build_web_compilers/lib/src/ddc_names.dart
@@ -8,6 +8,10 @@ import 'package:path/path.dart' as p;
 ///
 /// This logic must be synchronized with [pathToJSIdentifier] in DDC at:
 /// pkg/dev_compiler/lib/src/compiler/module_builder.dart
+/// 
+/// For backwards compatibility, if this pattern is changed, 
+/// [dev_compiler_bootstrap.dart] must be updated to accept both old and new
+/// patterns.
 String pathToJSIdentifier(String path) {
   path = p.normalize(path);
   if (path.startsWith('/') || path.startsWith('\\')) {
--- a/build_web_compilers/lib/src/ddc_names.dart
+++ b/build_web_compilers/lib/src/ddc_names.dart
@@ -8,8 +8,8 @@ import 'package:path/path.dart' as p;
 ///
 /// This logic must be synchronized with [pathToJSIdentifier] in DDC at:
 /// pkg/dev_compiler/lib/src/compiler/module_builder.dart
-/// 
-/// For backwards compatibility, if this pattern is changed, 
+///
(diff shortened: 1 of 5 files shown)


build_web_compilers/lib/src/ddc_names.dart
```diff
@@ -2,10 +2,25 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-/// Escape [name] to make it into a valid identifier.
+import 'package:path/path.dart' as _p;
```
natebosch: Why `_p` instead of `p`?

build_web_compilers/lib/src/ddc_names.dart
```diff
@@ -2,10 +2,25 @@
...
+
+/// Transforms a path to a valid JS identifier.
 ///
-/// This is copied from
-/// https://github.com/dart-lang/sdk/blob/300fd66f5c5e7497674ea4dd9fd21be91ee3c71e/pkg/dev_compiler/lib/src/analyzer/code_generator.dart#L6390
+/// This logic must be synchronized with [pathToJSIdentifier] in DDC at:
+/// pkg/dev_compiler/lib/src/compiler/module_builder.dart
+String pathToJSIdentifier(String path) {
+  path = _p.basenameWithoutExtension(_p.normalize(path));
+  if (path.startsWith('/') || path.startsWith('\\')) {
```
natebosch: When would we end up with `\` at the start?

Markzipan: I think this is more a side effect of stuff happening on DDC's side. I thought it better to keep the code consistent (even if there's some extraneous logic).

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -62,9 +62,13 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   // See https://github.com/dart-lang/sdk/issues/27262 for the root issue
   // which will allow us to not rely on the naming schemes that dartdevc uses
   // internally, but instead specify our own.
-  var appModuleScope = toJSIdentifier(
+  var simpleAppModuleScope = toJSIdentifier(
       _context.withoutExtension(_context.basename(buildStep.inputId.path)));
 
+  // Like above but with a more modern module-relative entrypoint.
```
jakemac53: module-relative doesn't seem to be accurately describing this - it looks like it is package-relative?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -62,9 +62,13 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   // See https://github.com/dart-lang/sdk/issues/27262 for the root issue
   // which will allow us to not rely on the naming schemes that dartdevc uses
   // internally, but instead specify our own.
-  var appModuleScope = toJSIdentifier(
+  var simpleAppModuleScope = toJSIdentifier(
```
jakemac53: nit: rename to `oldAppModuleScope` to match the param and a bit more accurately describe it

build_web_compilers/lib/src/ddc_names.dart
```diff
@@ -2,10 +2,25 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-/// Escape [name] to make it into a valid identifier.
+import 'package:path/path.dart' as p;
+
+/// Transforms a path to a valid JS identifier.
 ///
-/// This is copied from
-/// https://github.com/dart-lang/sdk/blob/300fd66f5c5e7497674ea4dd9fd21be91ee3c71e/pkg/dev_compiler/lib/src/analyzer/code_generator.dart#L6390
+/// This logic must be synchronized with [pathToJSIdentifier] in DDC at:
```
natebosch: How synchronized exactly? How much backwards and forwards compatibility do we have?

Markzipan: It should be logically identical (so copy-paste). Luckily, pathToJSIdentifier won't change much.

natebosch: We need to document in the commit message exactly what the rollout looks like. Something like.

"An upcoming release of the SDK will break compatibility with older versions of `build_web_compilers` that use the existing pattern. This change makes the compilers forwards compatible with upcoming SDK and remains compatible with old SDKs."

Then this sentence in the doc should be made more clear. Something like

"If the SDK changes patterns this logic should be updated to accept the new pattern in addition to the old pattern."

We can't be "synchronized" with the SDK because we don't have 1:1 releases with the SDK.

Markzipan: An ideal case would be to use some standard Uri to JS-compliant identifier scheme that could be shared across our codebase. Synchronicity has been an implicit requirement since DDC decided to export names in a special way. However, this is a pretty unique case, as the scheme usually isn't volatile.

jakemac53: As Nate mentioned synchronicity just isn't possible though. There is an issue filed somewhere about getting a better api from ddc that would allow us to not have to hardcode so much boostrapping logic so we wouldn't own this type of thing at all, which is really the only way forward.

jakemac53:
@Markzipan looks like dartfmt needs to be ran then travis should be happy

PR #2379, opened 2019-07-25
Update README for style

natebosch:
- Remove configurable build extension since it's confusing given the
  capabilities of the build system. Technically it is a valid thing to
  do as long as the builder factory hardcodes the extension and it's not
  coming from builder options, however it's more clear and
  representative of real world usage to hardcode it in the builder.
- Drop optional `new`.
- Update Builder method signatures to match current code.
- Switch some block comments to use 2 slashes and remove square braces.
- Add an `await` and remove comment about how it is optional - this
  isn't info that most users care about since it's almost never going to
  be an optimization to omit it.
- Drop references to barback and transformers - they've been gone long
  enough that referring to them isn't useful anymore.

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -19,28 +19,21 @@ and allows reading/writing files and resolving Dart source code.
 An interface into the dart [analyzer][pub:analyzer] to allow resolution of code
 that needs static analysis and/or code generation.
 
-## Differences between the `build` package and `pub` + `barback`.
-
-If you currently implement transformers with `package:barback` for use with
-Dart v1 `pub build` and `pub serve`, see [Upgrading from barback][].
-
 ## Implementing your own Builders
 
-If you have written a `barback` `Transformer` in the past, then the
-[`Builder`][dartdoc:Builder] API should be familiar to you. The main difference
-is that `Builders` must always configure outputs based on input extensions.
+A `Builder` gets invoked one by one on it's inputs, and may output new files.
 
 The basic API looks like this:
 
 ```dart
 abstract class Builder {
-  /// You can only output files in `build` that are configured here. You are not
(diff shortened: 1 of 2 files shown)


build/README.md
```diff
@@ -19,60 +19,49 @@ and allows reading/writing files and resolving Dart source code.
...
-
-If you currently implement transformers with `package:barback` for use with
-Dart v1 `pub build` and `pub serve`, see [Upgrading from barback][].
-
 ## Implementing your own Builders
 
-If you have written a `barback` `Transformer` in the past, then the
-[`Builder`][dartdoc:Builder] API should be familiar to you. The main difference
-is that `Builders` must always configure outputs based on input extensions.
+A `Builder` gets invoked one by one on it's inputs, and may output new files.
```
jakemac53: ... based on those inputs

build/README.md
```diff
@@ -19,60 +19,49 @@ and allows reading/writing files and resolving Dart source code.
...
 The basic API looks like this:
 
 ```dart
 abstract class Builder {
-  /// You can only output files in `build` that are configured here. You are not
-  /// required to output all of these files, but no other [Builder] is allowed
-  /// to produce the same outputs.
+  /// You can only output files that are configured here. You are not required
+  /// to output all of these files, but no other builder is allowed to produce
+  /// the same outputs.
```
jakemac53: Should we expand this comment to better describe what this actually means (suffix substitution)?

build/README.md
natebosch: Done

build/README.md
natebosch: I expanded it a bit - I don't want to go into too much detail here.

PR #2383, opened 2019-07-26
Add notice to hot reload docs

jakemac53:
This functionality will likely be deleted since we did an investigation and nobody is using them.


docs/hot_module_reloading.md
```diff
@@ -1,3 +1,5 @@
+# NOTICE: THESE HOOKS WILL LIKELY SOON BE DEPRECATED/DELETED
```
natebosch: I might drop the "likely"

We could also add a note below with something like "if this breaks your use case please file an issue at ..."

jakemac53: done

PR #2387, opened 2019-07-30
Add a check for empty kernel files

natebosch:
Towards #2362

We don't know where the zero byte files are being introduced to the
system - if they are coming from the kernel worker try to catch them
earlier by checking for when we copy a file with no content. This won't
solve the issue but if errors surface through this they should be more
clear.

- Add a `requireContent` argument to `copyOutput`. In the future we
  might consider making the default either with or without an output -
  we don't know of any use cases where we want to allow copying a zero
  byte file.
- Use the new argument from the kernel builder to check earlier whether
  a kernel file is empty.

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.4.1
+
+- Require non-empty output from kernel build steps.
+
 ## 2.4.0
 
 - Allow overriding the target name passed to the kernel worker.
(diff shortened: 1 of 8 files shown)


build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.4.1
```
jakemac53: nit: leave `-dev` until we are actually ready to publish (dev dependencies removed) - ideally in the pubspec too

natebosch: Will keep that in mind for next time - I plan to publish right away so I won't bother waiting for another round of travis this time.

jakemac53: sgtm

Issue #2389, opened 2019-08-05
Generated Dart code is not visible to other builders if library is resolved twice


simolus3:
## Problem
We have two builders (one custom defined below, one from `built_value`) generating different Dart files that should be `part of` a user defined library:
```dart
// this is the input, called example.dart
import 'package:built_value/built_value.dart';

part 'example.sample.dart'; // generated by a custom builder, source below
part 'example.g.dart'; // generated by built_value with source_gen

abstract class Example implements Built<Example, ExampleBuilder> {
  SomeClass get entry; // defined in "example.sample.dart"

  Example._();
  factory Example([void Function(ExampleBuilder) updates]) = _$Example;
}
```

The custom builder looks like this, it basically writes `SomeClass` into a part file for the `example.dart` file above:
```dart
import 'dart:async';
import 'package:build/build.dart';
Builder createSomeBuilder([BuilderOptions options]) => SimpleBuilder();

class SimpleBuilder extends Builder {
  @override
  FutureOr<void> build(BuildStep buildStep) async {
    final input = buildStep.inputId;
    log.fine('Running build-use-generated-dart on ${input.path}');
    if (input.path.endsWith('example.dart')) {
      final output = input.changeExtension('.sample.dart');

      // the issue only occurs when the next line is present!!
      await buildStep.inputLibrary;

      await buildStep.writeAsString(output, r'''
part of 'example.dart';
class SomeClass {
  final String field;
  const SomeClass(this.field);
}''');
    }
  }
  @override
  final Map<String, List<String>> buildExtensions = const {
    'dart': ['sample.dart']
  };
}
```

When the `await buildStep.inputLibrary` is commented out, the build runs without a problem and the generated code from `built_value` contains the `SomeClass` field.   
However, when `await buildStep.inputLibrary` is uncommented, it looks like `built_value` is unable to resolve the generated class:
```
[SEVERE] built_value_generator:built_value on lib/example.dart:
Error in BuiltValueGenerator for abstract class Example implements Built<Example, dynamic>.
Please make the following changes to use BuiltValue:

1. Make field entry have non-dynamic type. If you are already specifying a type, please make sure the type is correctly imported.
package:built_value_generator/src/value_source_class.dart 575:28        ValueSourceClass.generateCode
package:built_value_generator/built_value_generator.dart 53:52          BuiltValueGenerator.generate
package:source_gen/src/builder.dart 280:35                              _generate
[...]

PR #2392, opened 2019-08-06
Call AnalysisDriver.changeFile for new files

jakemac53:
Related to https://github.com/dart-lang/build/issues/2389 but this doesn't appear to solve the problem.

In isolated reproduction cases this does appear necessary to call though.

### DIFF
--- a/build_resolvers/test/resolver_test.dart
+++ b/build_resolvers/test/resolver_test.dart
@@ -74,6 +74,39 @@ void main() {
       }, resolvers: AnalyzerResolvers());
     });
 
+    test('handles discovering previously missing parts', () async {
+      var resolvers = AnalyzerResolvers();
+      await resolveSources({
+        'a|web/main.dart': '''
+              part 'main.g.dart';
+
+              class A implements B {}
+              ''',
+      }, (resolver) async {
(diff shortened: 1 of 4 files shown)


natebosch:
According to https://github.com/dart-lang/build/pull/2244#discussion_r280909427 I skipped the `changeFile` as an attempt at an (unmeasured) optimization. I did it because the previous code path only called `changeFile` for files we had previously seen.

Given that context this seems safe to me.

jakemac53:
I was able to create a test that I think captures the bug as well, left it in as a skipped test but it should help to validate things and with debugging as well.

build_resolvers/test/resolver_test.dart
```diff
@@ -74,6 +74,39 @@ void main() {
...
+              class A implements B {}
+              ''',
+        'a|web/main.g.dart': '''
+              class B {}
+              ''',
+      }, (resolver) async {
+        var lib = await resolver.libraryFor(entryPoint);
+        var clazz = lib.getType('A');
+        expect(clazz, isNotNull);
+        expect(clazz.interfaces.length, 1);
```
natebosch: [hit] `hasLength(1)`

Issue #2393, opened 2019-08-07
--log-performance crashes on Windows


Fox32:
I'm using build_runner on Windows and just found the `--log-performance` command line option and wanted to try it. However the application crashes when I specify it:

```
>pub run build_runner build --output out --log-performance out
[INFO] Generating build script completed, took 651ms
[INFO] Reading cached asset graph completed, took 246ms
[INFO] Checking for updates since last build completed, took 1.2s
[INFO] Running build completed, took 58ms
[INFO] Caching finalized dependency graph completed, took 159ms
[SEVERE] Unhandled build failure!
FileSystemException: Cannot create file, path = '...\out/2019-08-07T11:29:25.601099' (OS Error: Die Syntax für den Dateinamen, Verzeichnisnamen oder die Datenträgerbezeichnung ist falsch.
, errno = 123)
[SEVERE] Failed after 253ms
```

The exception is in German, but more or less means that the file path contains invalid characters. In this case it's the colon `:` in the date format which isn't allowed on Windows. For better Windows compatibility you could use a different separator in the date format.

However as a workaround, you can use `--track-performance` together with `serve` to view the performance at the `/$perf` endpoint. That you UI is really nice!

*Additional info:*
* Dart SDK Version: Dart VM version: 2.4.0 (Wed Jun 19 11:53:45 2019 +0200) on "windows_x64"
* build_runner version: 1.6.5

jakemac53:
This should be fixed now (the invalid characters part at least, the path still seems a bit wonky so feel free to file a new issue or comment here if you still have problems).

Also note that while this adds logging it is in a somewhat incomplete state - we don't provide any way of viewing those performance logs etc. If you wanted to contribute here that would likely be welcome.

If you want to dig into and parse these logs they are produced by the `timing` package, so you can deserialize them using those classes to create your own UI or whatever you like. If you come up with anything cool let us know!

jakemac53:
Actually sorry I was slightly wrong, to deserialize the logs you would want to use our `BuildPeformance` class here https://github.com/dart-lang/build/blob/master/build_runner_core/lib/src/generate/performance_tracker.dart#L21. I forgot we wrap up the timings objects in a few more layers.

Fox32:
Thanks for fixing this that fast.
[...]

PR #2397, opened 2019-08-07
create windows compatible directory names when doing performance logs

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2393

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -280,9 +280,9 @@ class _SingleBuild {
         var now = DateTime.now();
         var logPath = p.join(
             _logPerformanceDir,
-            '${_fourDigits(now.year)}-${_twoDigits(now.month)}-'
-            '${_twoDigits(now.day)}_${_twoDigits(now.hour)}-'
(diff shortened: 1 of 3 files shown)


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -855,3 +859,15 @@ String _actionLoggerName(
       : primaryInput.uri;
   return '${phase.builderLabel} on $asset';
 }
+
+String _twoDigits(int n) {
```
natebosch: Are we re-implementing `DateFormatter` to avoid a dependency on `intl`?

How about using `padLeft` for this?

String _twoDigits(int n) => '$n'.padLeft(2, '0');

jakemac53: > Are we re-implementing `DateFormatter` to avoid a dependency on `intl`

Yes, it seems crazy basic date formatting doesn't come with DateTime but :man_shrugging: 


> How about using `padLeft` for this?

Sure, I just copied the internal impl from DateTime but I can change it. I will also remove `_fourDigits` since I realized its totally unecessary for this case (we don't have to deal with old dates).

PR #2399, opened 2019-08-08
build_config: support the latest pkg:json_annotation, prepare release

kevmoo:
Bumped the min version of pubspec_parse to remove a class so a
`hide` can be removed, too


grouma:
Looks like some stuff needs to be published?

kevmoo:
Yup. I'm on it!

PR #2410, opened 2019-08-13
Reduce flakiness

grouma:
This should solve the flakiness. I believe there were a couple issues going on:

- We were killing the daemon process mid build and then attempting to clean up the build directory
  - Although the daemon is guaranteed to exit it isn't guaranteed to exit immediately which can cause errors with the test
- The daemon has logic to automatically shut down if a client doesn't connect in a certain amount of time
  - Some tests are slow so the client wouldn't connect in time and therefore get a hold of the lock causing issues
  - To resolve this we simply extend the default timeout

Closes: https://github.com/dart-lang/build/issues/2413


build_runner/test/daemon/daemon_test.dart
```diff
@@ -172,8 +176,8 @@ main() {
         ..registerBuildTarget(webTarget)
         ..startBuild();
       clients.add(client);
-      var buildResults = await client.buildResults.first;
-      expect(buildResults.results.first.status, BuildStatus.started);
+      await client.buildResults
+          .firstWhere((b) => b.results.first.status == BuildStatus.succeeded);
```
jakemac53: We should probably use `expectLater` and `emitsThrough` here instead of this to get better failure messages (same below)

build_runner/test/daemon/daemon_test.dart
```diff
@@ -28,6 +27,20 @@ main() {
   final webTarget = DefaultBuildTarget((b) => b..target = 'web');
   final testTarget = DefaultBuildTarget((b) => b..target = 'test');
   var clients = <BuildDaemonClient>[];
+
+  // This needs to be run before setUp since a call to `d.create` will add
```
natebosch: This doesn't reflect my understanding of how the test package works. I would not expect the order of called between this and setup to matter. Instead I think we should be calling addTearDown after creating the client.

grouma: `d.sandbox` implicitly calls `addTearDown`. Since `package:test` will execute tear downs in order they are added the theory is that this could cause issues as we are mucking with the file system while the daemon is running.

natebosch: We should discuss on Monday. The call to `setUp` is not when this happens, it's when the runner invokes that callback. The order of calls to give the runner these callbacks shouldn't matter, and the comment didn't make sense because the runner will always invoke callbacks from `addTeardown` before the ones from `teardown`

grouma: Ahh. I understand what you are getting at. Yeah that makes sense and this change wouldn't be required unless we were doing a call to `d.sandbox` directly in the main function. Nonetheless, there is still flakiness I need to track down. I'm convinced that the file lock isn't working as intended on Travis for some reason. I'll add some logging to confirm the theory.

PR #2415, opened 2019-08-15
WIP: Add --build-filter option to only build matching outputs

jakemac53:
Progress towards https://github.com/flutter/flutter/issues/37810.

Example usage: `pub run build_runner build --build-filter=test/some_test.*.js` (builds all js outputs of a test)

I will do some further cleanup and add tests, but sending this out for preliminary feedback on the approach.

- Adds a `BuildFilter` class which is a package + glob pair
- Allows you to pass an iterable of filters to the build impl
- If any filter is provided, only non-optional outputs matching a filter and their inputs will be built.

Future cleanup work:

- Remove buildDirs as an internal concept and translate that into `<dir>/**` build filters.
  - Will be slightly less performant due to glob matching instead of prefix matching.
- Update daemon builder to expect build targets to be these build filters instead of directories 

Pros of this approach:

- Simple to implement, no crazy assumptions built in
- Simple concept to understand from the outside
- Allows outside tools to decide what they need to be built

Cons:

- syntax is kind of ugly (but webdev or other products can paper over this)
  - we could possibly just use rest args instead of an explicit option, but serve/build mode would have to be a bit smarter
- outside tools have to understand what needs to be built for a given task they want to accomplish
  - they have to know to build *.js files for web tests, or *.dill files for vm tests etc
  - any extra data files for the test that are generated they would have to know about (this is also a pro that the build system becomes more agnostic though).


jakemac53:
cc @jonahwilliams I couldn't add you as a reviewer but PTAL, this would affect how you use build daemon ultimately (what targets you ask to be built)

build_runner_core/lib/src/util/build_dirs.dart
```diff
@@ -4,18 +4,27 @@
...
+/// Returns whether or not [id] should be built based upon [buildDirs],
+/// [phase], and optional [buildFilters].
+bool shouldBuildForDirs(AssetId id, List<String> buildDirs, BuildPhase phase,
+    {Iterable<BuildFilter> buildFilters}) {
   // Always build non-hidden outputs, since they ship with the package.
   // Otherwise it is too easy to ship not up to date files on accident.
   if (!phase.hideOutput) return true;
 
   // Always build `lib/` dirs, anything in them might be available to any app.
   if (id.path.startsWith('lib/')) return true;
```
jonahwilliams: Does this imply that the main entrypoints of a flutter application will always be built? (due to existing under lib)

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -399,3 +419,36 @@ Set<BuildDirectory> _parseBuildDirs(ArgResults argResults) {
...
 }
+
+/// Returns build filters parsed from [buildFilterOption] arguments.
+///
+/// These support `package:` uri syntax as well as regular path syntax,
+/// with glob support.
+List<BuildFilter> _parseBuildFilters(
+    ArgResults argResults, String rootPackage) {
+  var filterArgs = argResults[buildFilterOption] as List<String>;
+  if (filterArgs?.isNotEmpty != true) return null;
```
natebosch: [nit] https://dart.dev/guides/language/effective-dart/usage#do-use--to-convert-null-to-a-boolean-value

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -399,3 +419,36 @@ Set<BuildDirectory> _parseBuildDirs(ArgResults argResults) {
...
+  var filters = <BuildFilter>[];
+  for (var arg in filterArgs) {
```
natebosch: What if we add a `BuildFilter.fromArg(String rootPackage, String arg)` constructor, then we could do something like this here:

```dart
try {
  return [
    for(var arg in filterArgs)
      BuildFilter.forArg(rootPackage, arg),
  ];
} on FormatException catch(e) {
  throw ArgumentError.value(...);
}
```

build_runner_core/lib/src/util/build_dirs.dart
```diff
@@ -4,18 +4,27 @@
...
 
+  // Apply the build filters, if present.
+  if (buildFilters?.isNotEmpty == true) {
```
natebosch: [nit] `??`

build_runner_core/lib/src/util/build_dirs.dart
```diff
@@ -4,18 +4,27 @@
...
+    if (!buildFilters.any((f) => f.matches(id))) {
```
natebosch: [nit] Can we use `&&` instead of nested conditions?

build_runner/lib/src/entrypoint/base_command.dart
```diff
@@ -88,7 +88,11 @@ abstract class BuildRunnerCommand extends Command<int> {
       ..addFlag(symlinkOption,
           defaultsTo: symlinksDefault,
           negatable: true,
-          help: 'Symlink files in the output directories, instead of copying.');
+          help: 'Symlink files in the output directories, instead of copying.')
+      ..addMultiOption(buildFilterOption,
+          help: 'An explicit filter of files to build. Relative paths and '
```
natebosch: We need to make it clear that this is an `OR` not an `AND` when you give multiple filters.

build_runner_core/lib/src/util/build_dirs.dart
jakemac53: It does, yes :sob: .

I don't think we can change this unfortunately. Sass is a really concrete example - although this entire feature could have other issues with sass too, but this at least helps paper over some of the problems.

Note that this is only for `required` builders still (but web entrypoints are one of those). I really don't think we can avoid building required builders from dependencies because we have no concept of a RUNFILES equivalent (ie: what files are required at runtime for a given app).

build_runner/lib/src/entrypoint/base_command.dart
```diff
@@ -88,7 +88,11 @@ abstract class BuildRunnerCommand extends Command<int> {
...
+              '`package:` uris are supported, including glob syntax for paths '
+              'portions (but not package names).');
```
grouma: Optionally include an example?

jakemac53:
> I'm a little concerned how this will impact `webdev` but overall this approach works for me. You'll want to keep @liuming0 tied into this change as this will impact Bolt. It actually may solve some problems for him.

I talked with @natebosch a bit this morning about this, specifically regarding the hack that @jonahwilliams brought up where we always build everything under `lib`.

I think we can remove that by basically just doing the following, and it will allow webdev to retain its current behavior without negatively impacting flutter web:

- Allow glob support for package names
- Have webdev translate `webdev serve <directory>` into two build filters:
  - package:*/** (equivalent of the lib hack)
  - <directory>/**

In general the auto-building of lib does feel like it shouldn't be hardcoded - this gives tools the flexibility to keep the current behavior but also do something better later (we could some way of configuring runfiles for entrypoints, and maybe the above would just be the default?).

In the meantime build_daemon could support either top level directory names (which it would translate into the above filters), or explicit build filters. We would have to add a bit of a hack to detect directory names and treat them different but that shouldn't be terrible.

Eventually we do a breaking change in build_daemon and remove directory building support entirely, and force the explicit `<dir>/**` glob instead. I want to stop doing any build_daemon breaking changes until we fix our release process though.

build_runner_core/lib/src/util/build_dirs.dart
jonahwilliams: Well, I guess they will get cached at least. Ideally we'd have a way to avoid multiple dart2js invocations for an app with more than one entrypoint, but as long as we can skip compiling extra tests this seems like a win.


Of course, dart2js itself could be more modular...

jakemac53: See my recent comment https://github.com/dart-lang/build/pull/2415#issuecomment-522083547.

For flutter web since it doesn't support custom builders anyways it is well known that you don't have to actually build these lib files, and that proposal would allow us to delete that code and have webdev explicitly opt into it instead.

jakemac53:
> Do we want to try to match `InputSet` semantics or anything like that? I guess we can do that in a backwards compatible way in the future since these semantics effectively match the `include` field for an `InputSet`.

This is a bit different because it supports package names as well. The `InputSet` class is always used within the context of a package and doesn't have a package name itself.

build_runner/lib/src/entrypoint/options.dart
jakemac53: Sure, done

build_runner_core/lib/src/util/build_dirs.dart
jakemac53: sure, done

jakemac53:
Note that I am going to be OOO through labor day and then I can pick this back up. I don't think there is a huge sense of urgency but I do think we should move forward with this.

yjbanov:
Is it reasonable to expect that users will know the names of the output files? Other than that, LGTM.

jakemac53:
> Is it reasonable to expect that users will know the names of the output files? Other than that, LGTM.

Probably not for all scenarios, no. However this is more about the underlying functionality - and it enables the clients (webdev, flutter web, build_runner) to provide clean abstractions over this.

Also, allowing glob support helps here.

jonahwilliams:
I would imagine the flutter tool would handle mapping the test command/run command to the correct output fies

jakemac53:
> I would imagine the flutter tool would handle mapping the test command/run command to the correct output fies

Yes, exactly. To be clear this isn't a perfect solution (flutter tool has to hardcode some assumptions about output files), but I think the general functionality is useful regardless.

One other solution we had discussed a while ago is allowing you to build all the outputs of some input file (for instance a test dart file) - but I realized that has some problems of its own:

- it would build for all platforms that the app supports (so you might build for vm and web even though you only cared about web for instance).
- in release mode it would compile both the original test and the bootstrapped version of the test (*.browser_test.dart).
- it doesn't really make sense for anything but entrypoints - where this approach is more general (you can build any generated file in isolation).

yjbanov:
If it's mostly API for tools, then it makes sense to me. It doesn't have to be super user-friendly.

yjbanov:
Is this going in? We could use this in a couple of places as soon as it lands.

jakemac53:
I am planning on picking this back up but after I finish up working on the devtools integration tests which I understand as being higher priority.

It was my understanding previously that there is a workaround in the flutter code today so this isn't as urgent. If there are new use cases or I misunderstood then I could pick this back up instead. It needs a fair bit more work in terms of tests and cleanup though.

evanweible-wf:
@jakemac53 I pulled this branch down and tried it out in a few of our projects and it is a huge help to one of our major paint points in transitioning devs to dart2: iterating on tests in large projects. Using a build filter to target the outputs for a single test file and then only running that one test file results in incremental build times in the seconds rather than minutes (often 4+ mins).

We would love to use this in our test-running abstraction by automatically mapping test files that should be run to build filters so that devs can iterate on individual tests or a subset of all tests without paying the cost of building everything.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -399,3 +418,24 @@ Set<BuildDirectory> _parseBuildDirs(ArgResults argResults) {
...
 }
+
+/// Returns build filters parsed from [buildFilterOption] arguments.
+///
+/// These support `package:` uri syntax as well as regular path syntax,
+/// with glob support for both package names and paths.
+List<BuildFilter> _parseBuildFilters(
+    ArgResults argResults, String rootPackage) {
+  var filterArgs = argResults[buildFilterOption] as List<String>;
+  if (filterArgs?.isNotEmpty ?? false) return null;
```
evanweible-wf: I think this is supposed to be:
```suggestion
  if (filterArgs?.isEmpty ?? true) return null;
```
Making that change locally got me up and running.

jakemac53:
@evanweible-wf yep makes sense - any project with a large number of tests will see major benefit here (especially when iterating on core libs but only to fix a single test, etc).

I think it would be worth making general solutions here too - potentially `build_runner test` should understand what tests it is running. Or maybe now with the build_daemon we just remove build_runner test and add built in support to package:test. We could then add annotations like `@BuiltData(['my_generated_data_file.txt'])` or something as well so that the test package could request not just the JS but also any other generated files based on that. Lots of options :).

jakemac53:
Or actually maybe we should add a general annotation to like the build_config package and the entrypoint builders would understand that and make sure to request those files when they build in addition to whatever other files are required to compile...

build_runner/lib/src/entrypoint/options.dart
jakemac53: Heh, nice catch.

evanweible-wf:
> Or maybe now with the build_daemon we just remove build_runner test and add built in support to package:test.

This option sounds awesome to me! It's been a bit of an odd paradigm shift for devs to have to know that some tests can be run via `pub run test` directly, but others require going through `pub run build_runner test`. If we could get back to a scenario where `pub run test` is the primary touch point and it knows how to integrate with the build system automatically, I think that would go a long way to simplify that mental model.

jakemac53:
> If we could get back to a scenario where `pub run test` is the primary touch point and it knows how to integrate with the build system automatically, I think that would go a long way to simplify that mental model.

Definitely agreed on that front. There are a lot of details to be worked out but the major hurdle has already been completed (allowing multiple tools to coordinate on a single build).

yjbanov:
@jacob314 we have a workaround for framework tests where we use the API directly. We do not have a workaround for engine tests where we use the command line.

jakemac53:
PTAL: I added tests and changed the behavior a little bit.

- Now if any build filters are provided we will not do the pessimistic building of all files under `lib` from all packages.
  - Note that in order to run a basic web app you will always have to provide `--build-filter "package:build_web_compilers/**/*"` to get the required js files from the SDK.
- I also updated the `OptionalOutputTracker` to respect build filters. This means build filters are respected for both output directories and in serve mode it will limit the files you can see.

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -36,6 +37,12 @@ final _defaultWebDirs = const ['web', 'test', 'example', 'benchmark'];
 
 /// Base options that are shared among all commands.
 class SharedOptions {
+  /// A set of explicit filters for files to build.
```
natebosch: Should we make it a set? If we keep it as an Iterable should we change the comment to `/// The explicit filters for files to build.`

build_runner_core/lib/src/generate/options.dart
```diff
@@ -54,22 +55,58 @@ class LogSubscription {
   final StreamSubscription<LogRecord> logListener;
 }
 
+/// Describes a set files that should be built.
```
natebosch: [nit] "a set of files"

jakemac53:
Ok this is now published as `build_runner` version `1.7.0`, which will require the correct versions of the other packages (`build_runner_core` `0.4.0`  and `build_daemon` `2.1.0`)

PR #2419, opened 2019-08-21
Set charset for Dart content types

grouma:
This should address https://github.com/dart-lang/webdev/issues/527

### DIFF
--- a/build_runner/pubspec.yaml
+++ b/build_runner/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_runner
-version: 1.6.6
+version: 1.6.7
 description: Tools to write binaries that run builders.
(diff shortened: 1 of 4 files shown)


build_runner/pubspec.yaml
```diff
@@ -26,7 +26,7 @@ dependencies:
   js: ^0.6.1+1
   logging: ^0.11.2
   meta: ^1.1.0
-  mime: ^0.9.3
+  mime: ^0.9.3+3
```
natebosch: why do this?

kevmoo: guessing https://pub.dev/packages/mime#0963

PR #2420, opened 2019-08-21
Add an option to skip the platform check for specific packages

natebosch:
Flutter web must allow some `dart:io` import in `package:flutter`, but
adding `dart:io` to the `DartPlatform` causes the module crawling to
pick the imports in some cases - we want the `DartPlatform` to match the
`libraries.json` that will be used by DDC and kernel in terms of which
libraries are supported, but to avoid the failures for bad imports.

The kernel and DDC workers have always allowed imports to missing
`dart:` libraries with stubbed implementations that will throw if they are
used at runtime. The restrictions on imports are implemented in the
build systems instead of the compilers. This argument allows loosening
the restriction in specific flutter packages while keeping the handrails for
user code.

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.5.0
+
+- Add an option to skip the unsupported module check for modules in specified
+  packages.
+
 ## 2.4.3
 
 - Allow analyzer version 0.38.0.
--- a/build_modules/lib/src/modules.dart
+++ b/build_modules/lib/src/modules.dart
@@ -138,8 +138,10 @@ class Module {
   /// If [throwIfUnsupported] is `true`, then an [UnsupportedModules]
   /// will be thrown if there are any modules that are not supported.
   Future<List<Module>> computeTransitiveDependencies(BuildStep buildStep,
-      {bool throwIfUnsupported = false}) async {
+      {bool throwIfUnsupported = false,
+      Set<String> skipPlatformCheckPackages = const {}}) async {
     throwIfUnsupported ??= false;
+    skipPlatformCheckPackages ??= const {};
     final modules = await buildStep.fetchResource(moduleCache);
     var transitiveDeps = <AssetId, Module>{};
     var modulesToCrawl = {primarySource};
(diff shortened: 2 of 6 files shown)


jonahwilliams:
Friendly bump :)

jakemac53:
> Flutter web must allow some `dart:io` import in `package:flutter`

This seems like the core issue? Can we not remove those imports (or hide them behind conditional imports?)

jonahwilliams:
It's not a trivial change: https://github.com/flutter/flutter/pull/38773

natebosch:
> can you add some more context?

I updated the first comment on the PR and will make the commit message match.

> Can we not remove those imports

In the long term I hope we will so that we can restore the build system restrictions on imports since it's friendlier to users. In the short term I think this is the right change.

jakemac53:
We could potentially make it take a whitelist of packages to allow these imports in instead of disabling the check entirely?

I am concerned people will be really confused when their app starts throwing at runtime due to some transitive dep that is trying to use dart:io....

natebosch:
@jonahwilliams - I'll take a stab at adding a `Set<String> ignorePlatformCheckPackages` argument where you could pass `{'flutter'}` to ease the concerns about UX - would that work for you?

jonahwilliams:
As a short term solution, either should be fine.

As for the long-term behavior of build_runner/ddc in the presence of dart:io/unsupported libraries, I'm not convinced that the current build_runner behavior is any less confusing than the resulting (potential) runtime exceptions. I think this behavior should be re-evaluated and potentially include some UXR work from Flutter.

jakemac53:
While it is true that we would refuse to compile an app that might actually work, the same could be said for all static errors. Skipping these import checks is basically equivalent to allowing imports to a non-existent uri and then only failing at runtime if you tried to actually use that library.

Improving the ergonomics of platform specific code would certainly be good, but I think that would have to happen at the language level. We could also work on the error messages etc to be more intuitive.

jonahwilliams:
What is the path forward here?

I understand that I'm asking for a quick hack, but this shouldn't effect anyone _besides_ the flutter tool. We should not be putting the dart:io problem on the critical path for flutter web

jakemac53:
I think the short term fix we should shoot for is allowing a (hopefully temporary) whitelist of packages that are allowed to ignore the import checks.

natebosch:
@jonahwilliams - can you try with the latest packages from this branch? You'll need to override both `build_web_compilers` and `build_modules` and change the boolean argument into `skipPlatformCheckPackages: {'flutter'}`

jonahwilliams:
Will do

PR #2432, opened 2019-09-05
Update ManualChangeProvider to do a lazy file system scan

jakemac53:
Should solve https://github.com/flutter/flutter/issues/39696

- Extracted existing file system scanning logic into an `AssetManager` class (in build_runner_core)
- Updated the ManualChangeProvider to use an `AssetManager` to lazily scan the file system instead of using a file watcher and collecting changes.
- Updated the existing test to verify changes were actually picked up by the new build.

### DIFF
--- a/build_runner/lib/src/daemon/change_providers.dart
+++ b/build_runner/lib/src/daemon/change_providers.dart
@@ -25,13 +25,13 @@ class AutoChangeProvider implements ChangeProvider {
 /// Computes changes with a file scan when requested by a call to
 /// [collectChanges].
 class ManualChangeProvider implements ChangeProvider {
-  final AssetManager _assetManager;
+  final AssetTracker _assetTracker;
 
-  ManualChangeProvider(this._assetManager);
+  ManualChangeProvider(this._assetTracker);
 
   @override
   Future<List<WatchEvent>> collectChanges() async {
-    var updates = await _assetManager.collectChanges();
+    var updates = await _assetTracker.collectChanges();
     return List.of(updates.entries
         .map((entry) => WatchEvent(entry.value, '${entry.key}')));
   }
--- a/build_runner/lib/src/daemon/change_providers.dart
+++ b/build_runner/lib/src/daemon/change_providers.dart
@@ -3,8 +3,13 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:build_daemon/change_provider.dart';
+import 'package:build_runner_core/src/generate/build_definition.dart';
 import 'package:watcher/src/watch_event.dart';
 
+/// Continually updates the [changes] stream as watch events are seen on the
+/// input stream.
+///
+/// The [collectChanges] method is a no-op for this implementation.
 class AutoChangeProvider implements ChangeProvider {
   final Stream<List<WatchEvent>> _changes;
 
@@ -17,21 +22,18 @@ class AutoChangeProvider implements ChangeProvider {
   Future<List<WatchEvent>> collectChanges() async => [];
 }
 
-// TODO(grouma) - collect changes through a one time file scan instead of
-// buffering changes from a change stream. This will have better performance
-// on Windows.
+/// Computes changes with a file scan when requested by a call to
+/// [collectChanges].
 class ManualChangeProvider implements ChangeProvider {
-  final _changes = <WatchEvent>[];
+  final AssetManager _assetManager;
 
-  ManualChangeProvider(Stream<List<WatchEvent>> changes) {
-    changes.listen(_changes.addAll);
-  }
+  ManualChangeProvider(this._assetManager);
 
   @override
   Future<List<WatchEvent>> collectChanges() async {
-    var result = _changes.toList();
-    _changes.clear();
-    return result;
+    var updates = await _assetManager.collectChanges();
+    return List.of(updates.entries
+        .map((entry) => WatchEvent(entry.value, '${entry.key}')));
   }
 
   @override
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -24,6 +24,7 @@ import 'package:build_runner_core/build_runner_core.dart'
     hide BuildResult, BuildStatus;
 import 'package:build_runner_core/build_runner_core.dart' as core
     show BuildStatus;
+import 'package:build_runner_core/src/generate/build_definition.dart';
 import 'package:build_runner_core/src/generate/build_impl.dart';
 import 'package:stream_transform/stream_transform.dart' show debounceBuffer;
 import 'package:watcher/watcher.dart';
@@ -193,7 +194,8 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
         builders, daemonOptions.builderConfigOverrides,
         isReleaseBuild: daemonOptions.isReleaseBuild);
 
-    var graphEvents = PackageGraphWatcher(packageGraph,
+    // Only actually used for the AutoChangeProvider.
+    Stream<List<WatchEvent>> graphEvents() => PackageGraphWatcher(packageGraph,
             watch: (node) =>
                 PackageNodeWatcher(node, watch: defaultDirectoryWatcherFactory))
         .watch()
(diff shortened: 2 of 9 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -204,17 +206,16 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
               true,
               expectedDeletes,
-            ));
-
-    var changes = graphEvents
+            ))
         .map((data) => WatchEvent(data.type, '${data.id}'))
         .map((change) => [change]);
 
     var changeProvider = daemonOptions.buildMode == BuildMode.Auto
```
natebosch: Should we pull out all this into a `_createChangeProvider` function? Or maybe move it into the constructor for `AutoChangeProvider`?

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -204,17 +206,16 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
               // Assume we will create an outputDir.
               true,
               expectedDeletes,
-            ));
-
-    var changes = graphEvents
+            ))
         .map((data) => WatchEvent(data.type, '${data.id}'))
         .map((change) => [change]);
```
natebosch: If we skip this can we also skip the `.map((changeLists) => changeLists.expand((l) => l).toList()`?

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -204,17 +206,16 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
 
     var changeProvider = daemonOptions.buildMode == BuildMode.Auto
-        ? AutoChangeProvider(changes
+        ? AutoChangeProvider(graphEvents()
             .transform(debounceBuffer(buildOptions.debounceDelay))
             .map((changeLists) => changeLists.expand((l) => l).toList()))
```
natebosch: Should this get moved to the previous chain of function calls?

build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -64,6 +67,120 @@ class BuildDefinition {
       _Loader(environment, options, buildPhases).prepareWorkspace();
 }
 
+/// Understands how to find all assets relevant to a build as well as compute
+/// updates to those assets.
+class AssetManager {
```
natebosch: How about `AssetTracker`? Or maybe `AssetDiffer`?

build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -64,6 +67,120 @@ class BuildDefinition {
...
+  final AssetGraph _assetGraph;
+  final RunnerAssetReader _reader;
+  final TargetGraph _targetGraph;
+
+  AssetManager(this._assetGraph, this._reader, this._targetGraph);
+
+  Future<Map<AssetId, ChangeType>> collectChanges() async {
```
natebosch: Can we add a doc comment?

build_runner/lib/src/daemon/daemon_builder.dart
jakemac53: Oh true it does seem like that, will give it a shot

build_runner/lib/src/daemon/daemon_builder.dart
jakemac53: hopefully this can just go away now as you mentioned above, will play with it

build_runner/lib/src/daemon/daemon_builder.dart
jakemac53: I don't want to move it all into the constructor I don't think (it is a good thing imo that it just accepts a stream of changes right now). I think it is probably fine as is - I could move the inline function to an actual instance method but I don't know that there is really any value there (and I would have more boilerplate for passing in params etc even though its only used locally in this function).

build_runner_core/lib/src/generate/build_definition.dart
jakemac53: Renamed AssetTracker

build_runner/lib/src/daemon/change_providers.dart
```diff
@@ -17,21 +22,18 @@ class AutoChangeProvider implements ChangeProvider {
   Future<List<WatchEvent>> collectChanges() async => [];
 }
 
-// TODO(grouma) - collect changes through a one time file scan instead of
-// buffering changes from a change stream. This will have better performance
-// on Windows.
```
grouma: I think there is a tracking issue that should be linked.

build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -193,7 +194,8 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
         builders, daemonOptions.builderConfigOverrides,
         isReleaseBuild: daemonOptions.isReleaseBuild);
 
-    var graphEvents = PackageGraphWatcher(packageGraph,
+    // Only actually used for the AutoChangeProvider.
+    Stream<List<WatchEvent>> graphEvents() => PackageGraphWatcher(packageGraph,
```
grouma: Why even create it then? Should we move the conditional above?

build_runner/test/daemon/daemon_test.dart
```diff
@@ -191,9 +191,9 @@ main() {
       await d.dir('a', [
         d.dir('web', [
           d.file('main.dart', '''
-                main() {
-                  print('hello world');
-                }'''),
+main() {
+  print('goodbye world');
```
grouma: Dark...

build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -64,6 +67,122 @@ class BuildDefinition {
...
+class AssetTracker {
+  final AssetGraph _assetGraph;
+  final RunnerAssetReader _reader;
+  final TargetGraph _targetGraph;
+
+  AssetTracker(this._assetGraph, this._reader, this._targetGraph);
+
+  /// Checks for and returns any file system changes compared to the current
+  /// state of the asset graph.
+  Future<Map<AssetId, ChangeType>> collectChanges() async {
```
grouma: Would be nice to get some unit tests in this package as well.

build_runner/lib/src/daemon/daemon_builder.dart
natebosch: Could also consider skipping the local function and moving the whole chunk of code inline where we call the constructor - I'm not sure how ugly that would make things with a large chunk like that in a ternary conditional...

build_runner/lib/src/daemon/daemon_builder.dart
jakemac53: I moved it into a function here so it only gets lazily created if its invoked

build_runner/test/daemon/daemon_test.dart
jakemac53: lol I could go for the exclamation point instead

build_runner/lib/src/daemon/daemon_builder.dart
jakemac53: Ya that gets pretty ugly :/

I also did try extracting it to a separate method but that ends up needing 6 parameters so its pretty boilerplaty compared to what I have now.

build_runner_core/lib/src/generate/build_definition.dart
jakemac53: done

build_runner/lib/src/daemon/change_providers.dart
jakemac53: I can't find one, if you do please close and link it

jakemac53:
Should solve https://github.com/flutter/flutter/issues/39696

- Extracted existing file system scanning logic into an `AssetManager` class (in build_runner_core)
- Updated the ManualChangeProvider to use an `AssetManager` to lazily scan the file system instead of using a file watcher and collecting changes.
- Updated the existing test to verify changes were actually picked up by the new build.

PR #2435, opened 2019-09-06
Don't track the build script if the check is skipped

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/621.

When passing --skip-build-script-check we will now stop attempting to track the build script entirely, and thus won't warn if there are sources that don't exist in the asset graph.

### DIFF
--- a/build_runner_core/lib/src/changes/build_script_updates.dart
+++ b/build_runner_core/lib/src/changes/build_script_updates.dart
@@ -13,14 +13,26 @@ import '../asset/reader.dart';
 import '../asset_graph/graph.dart';
 import '../package_graph/package_graph.dart';
 
+/// Functionality for detecting if the build script itself or any of its
+/// transitive imports have changed.
 class BuildScriptUpdates {
   final Set<AssetId> _allSources;
   final bool _supportsIncrementalRebuilds;
 
   BuildScriptUpdates._(this._supportsIncrementalRebuilds, this._allSources);
 
-  static Future<BuildScriptUpdates> create(RunnerAssetReader reader,
-      PackageGraph packageGraph, AssetGraph graph) async {
+  /// Creates a [BuildScriptUpdates] object, using [reader] to ensure that
+  /// the [graph] is tracking digests for all transitive sources.
+  ///
+  /// If [skipUpdatesCheck] is `true` then all checks are skipped and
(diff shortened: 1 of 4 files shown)


build_runner_core/lib/src/changes/build_script_updates.dart
```diff
@@ -13,14 +13,26 @@ import '../asset/reader.dart';
...
+  /// Creates a [BuildScriptUpdates] object, using [reader] to ensure that
+  /// the [graph] is tracking digests for all transitive sources.
+  ///
+  /// If [skipUpdatesCheck] is `true` then all checks are skipped and
+  /// [hasBeenUpdated] will always return `false`.
+  static Future<BuildScriptUpdates> create(
+      RunnerAssetReader reader,
+      PackageGraph packageGraph,
+      AssetGraph graph,
+      bool skipUpdatesCheck) async {
```
natebosch: how about a named argument?

Also, maybe call it `{bool disabled = false}`?

build_runner_core/lib/src/changes/build_script_updates.dart
```diff
@@ -94,3 +106,16 @@ class BuildScriptUpdates {
     return null;
   }
 }
+
+/// Always returns false for [hasBeenUpdated], used when we want to skip
+/// the build script checks.
+class _NoopBuildScriptUpdates implements BuildScriptUpdates {
```
natebosch: [optional] Could avoid the `UnimplementedError()` required methods if we make both implementations different classes:

```dart
abstract class BuildScriptUpdates {
  static Future<BuildScriptUpdates> create(...) {}

  bool hasBeenUpdated();
}

class _MirrorBuildScriptUpdates implements BuildScriptUpdates {
  ...
}

 class _NoopBuildScriptUpdates implements BuildScriptUpates {
  @override
  bool hasBeenUpdated(_) => false;
}
```

build_runner_core/lib/src/changes/build_script_updates.dart
jakemac53: done

build_runner_core/lib/src/changes/build_script_updates.dart
jakemac53: Done

PR #2447, opened 2019-09-18
Add support for dependency pruning to build/build_runner_core

jakemac53:
~Sending out for early design review - this is my current prototype for dependency pruning.~ Ready for primetime now!

cc @evanweible-wf @jonahwilliams 

## Current api overview

Added ~two~ one new api to the `BuildStep` abstract class:

- ~`void removeDependency(AssetId id);`~
- ~`Iterable<AssetId> get removedDependencies;`~
- `void reportUnusedAssets(Iterable<AssetId> ids);`

~I am not super happy with the exact api here, definitely looking for suggestions for improvement.~ I am much  happier now with the api here, based on suggestions.

## General concept

Gives `Builder` implementations the power to remove some of their input dependencies, even though they did actually check for the existence of or read the contents of some input.

This is a port of an experimental feature in `bazel`, which has been proven to dramatically reduce build times when making edits to heavily shared libraries.

Specifically, the Dart compilers can take advantage of this feature to not depend on all transitively imported packages when compiling a given module - they can "prune" their dependencies to only the transitive inputs that the module actually uses. In practice it is common to see an order of magnitude fewer dependencies for any given module as a result.

## Trying out the full proof of concept

You can experiment with this feature by depending on the `unused-inputs-kernel` branch of this package, by adding the following to your pubspec:

```yaml
dependency_overrides:
  build:
    git:
      ref: unused-inputs-kernel
      path: build
  build_runner_core:
    git:
      ref: unused-inputs-kernel
      path: build_runner_core
  build_modules:
    git:
      ref: unused-inputs-kernel
      path: build_modules
  build_web_compilers:
    git:
      ref: unused-inputs-kernel
      path: build_web_compilers
```

## Isn't this non-hermetic?

Yes, this allows for potential non-hermetic builds for badly behaved builders. It is up to the builder to make sure they really didn't depend on the contents of that file, and that no change to that file (or the existence thereof) could affect the output of the builder.

This is an explicit tradeoff being made because the performance gains are such that it is worth the risk.

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.2.0
+
+- Add `BuildStep.removeDependency(AssetId id)` method to `BuildStep` which
+  allows you to delete a dependency on some input after the fact.
+  - This should be used very carefully as it can easily lead to correctness
+    issues if abused.
+- Add the `BuildStep.removedDependencies` iterable which gives access to all
+  the removed deps.
+
 ## 1.1.6
 
 - Allow analyzer version 0.38.0.
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,11 +1,12 @@
 ## 1.2.0
 
-- Add `BuildStep.removeDependency(AssetId id)` method to `BuildStep` which
-  allows you to delete a dependency on some input after the fact.
-  - This should be used very carefully as it can easily lead to correctness
-    issues if abused.
-- Add the `BuildStep.removedDependencies` iterable which gives access to all
-  the removed deps.
+- Add the `void reportUnusedAssets(Iterable<AssetId> ids)` method to the
+  `BuildStep` class.
+  - **WARNING**: Using this introduces serious risk of non-hermetic builds.
+  - Indicates to the build system that `ids` were read but their content has
+    no impact on the outputs of the build.
+  - Build system implementations can choose to support this feature or not,
+    and it should be assumed to be a no-op by default.
 
 ## 1.1.6
 
--- a/build/lib/src/builder/build_step.dart
+++ b/build/lib/src/builder/build_step.dart
@@ -81,12 +81,10 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
 
-  /// **WARNING**: Expert level feature only, use at your own risk.
+  /// Indicates that [ids] were read but their content has no impact on the
+  /// outputs of this step.
   ///
-  /// Marks a set of ids as unused even if they were read by this build step.
-  ///
-  /// This should only be used if an asset was read, but none of it's content,
-  /// or the fact that it exists, actually has an impact on the output.
+  /// **WARNING**: Using this introduces serious risk of non-hermetic builds.
   ///
   /// If this file changes or becomes unreadable on the next build this build
   /// step may not run.
@@ -94,7 +92,7 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// **Note**: This is not guaranteed to have any effect and it should be
   /// assumed to be a no-op by default. Implementations must explicitly
   /// choose to support this feature.
-  void reportUnusedAssets(Iterable<AssetId> id);
+  void reportUnusedAssets(Iterable<AssetId> ids);
 }
 
 abstract class StageTracker {
--- a/build/lib/src/builder/build_step.dart
+++ b/build/lib/src/builder/build_step.dart
@@ -81,9 +81,20 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
 
-  void removeDependency(AssetId id);
-
-  Iterable<AssetId> get removedDependencies;
+  /// **WARNING**: Expert level feature only, use at your own risk.
+  ///
+  /// Marks a set of ids as unused even if they were read by this build step.
+  ///
+  /// This should only be used if an asset was read, but none of it's content,
+  /// or the fact that it exists, actually has an impact on the output.
+  ///
+  /// If this file changes or becomes unreadable on the next build this build
+  /// step may not run.
+  ///
+  /// **Note**: This is not guaranteed to have any effect and it should be
+  /// assumed to be a no-op by default. Implementations must explicitly
+  /// choose to support this feature.
+  void reportUnusedAssets(Iterable<AssetId> id);
 }
 
 abstract class StageTracker {
--- a/build/lib/src/builder/build_step.dart
+++ b/build/lib/src/builder/build_step.dart
@@ -80,6 +80,10 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// Returns value returned by [action].
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
+
+  void removeDependency(AssetId id);
+
+  Iterable<AssetId> get removedDependencies;
 }
 
 abstract class StageTracker {
--- a/build/lib/src/builder/build_step.dart
+++ b/build/lib/src/builder/build_step.dart
@@ -86,7 +86,7 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   ///
   /// **WARNING**: Using this introduces serious risk of non-hermetic builds.
   ///
-  /// If this file changes or becomes unreadable on the next build this build
+  /// If these files change or become unreadable on the next build this build
   /// step may not run.
   ///
   /// **Note**: This is not guaranteed to have any effect and it should be
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -57,17 +57,15 @@ class BuildStepImpl implements BuildStep {
 
   bool _isComplete = false;
 
-  @override
-  final removedDependencies = <AssetId>{};
+  final void Function(Iterable<AssetId>) _reportUnusedAssets;
 
   BuildStepImpl(this.inputId, Iterable<AssetId> expectedOutputs, this._reader,
       this._writer, this._rootPackage, this._resolvers, this._resourceManager,
-      [this._stageTracker = NoOpStageTracker.instance])
-      : _expectedOutputs = expectedOutputs.toSet();
-
-  @override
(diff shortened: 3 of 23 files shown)


build/lib/src/builder/build_step_impl.dart
```diff
@@ -57,11 +57,18 @@ class BuildStepImpl implements BuildStep {
...
+  final removedDependencies = <AssetId>{};
+
   BuildStepImpl(this.inputId, Iterable<AssetId> expectedOutputs, this._reader,
       this._writer, this._rootPackage, this._resolvers, this._resourceManager,
       [this._stageTracker = NoOpStageTracker.instance])
       : _expectedOutputs = expectedOutputs.toSet();
 
+  @override
+  void removeDependency(AssetId id) =>
+      (removedDependencies as Set<AssetId>).add(id);
```
jakemac53: These casts are unfortunate, and I end up doing it in build_runner_core right now also - so this is a place where the current api is not ideal for sure.

build/lib/src/generate/run_builder.dart
```diff
@@ -26,12 +26,16 @@ import 'expected_outputs.dart';
...
+/// If [removedDependencies] is provided an entry will be added for each input
+/// which gives all the removed dependencies for that input.
 Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
     AssetReader reader, AssetWriter writer, Resolvers resolvers,
     {Logger logger,
     ResourceManager resourceManager,
     String rootPackage,
-    StageTracker stageTracker = NoOpStageTracker.instance}) async {
+    StageTracker stageTracker = NoOpStageTracker.instance,
+    Map<AssetId, Iterable<AssetId>> removedDependencies}) async {
```
jakemac53: This is also a weird api that is a side effect of putting this on the `BuildStep` class - this method can create multiple `BuildStep` instances and doesn't return them. Because of that I end up allowing you to pass in this map to get the information back out :(.

grouma:
This approach looks good to me. Just a quick thought:

> it is up to the builder to make sure they really didn't depend on the contents of that file, and that no change to that file (or the existence thereof) could affect the output of the builder.

Do we not have enough information in the build system to validate that the file was not read by the builder? Can we not provide some safe guards? What about validating that the hash of the non-dependent files stays the same after the build step.

jakemac53:
> Do we not have enough information in the build system to validate that the file was not read by the builder? Can we not provide some safe guards?

Usually we would, but when it comes to external compilers we always do "read" all transitive deps because we run the compiler in a temp directory and copy the deps eagerly over there.

Also in practice the compilers do actually read all the transitive deps - kernel especially requires a fully valid and resolved program. It can however track which libraries were actually used to generate the output kernel file to allow trimming the deps after the fact.

>  What about validating that the hash of the non-dependent files stays the same after the build step.

Not sure exactly what you mean here, can you elaborate?

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -781,11 +789,15 @@ class _SingleBuild {
       SingleStepReader reader,
       AssetWriterSpy writer,
       String actionDescription,
-      Iterable<ErrorReport> errors) async {
+      Iterable<ErrorReport> errors,
+      {Iterable<AssetId> removedDependencies}) async {
     if (outputs.isEmpty) return;
+    var actualInputs = removedDependencies != null
```
grouma: [optional] Maybe `usedInputs` instead?

grouma:
> > What about validating that the hash of the non-dependent files stays the same after the build step.
> 
> Not sure exactly what you mean here, can you elaborate?

Ignore that thought actually. It would only work if all the build steps ran sequentially. Since build steps can happen in parallel, and it's possible for other build steps to modify non-dependent files, we can't guarantee that they won't change.

build/lib/src/generate/run_builder.dart
natebosch: I'm not a fan of "output collections" like this. Can we pass a callback? If we pass it in to the constructor of `BuildStepImpl` it would also obviate the need for adding an API on `buildStep` that is only useful to the build system.

```
...
void Function(AssetId, Set<AssetId>) reportUnusedAssets

...

BuildStepImpl(input, outputs,...
  reportUnusedAssets: (ids) => reportUnusedAssets(input, ids),
);
```

build/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 1.2.0
+
+- Add `BuildStep.removeDependency(AssetId id)` method to `BuildStep` which
+  allows you to delete a dependency on some input after the fact.
+  - This should be used very carefully as it can easily lead to correctness
```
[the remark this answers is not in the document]
davidmorgan: You probably want to always rerun on existence changes; in bazel these change the actual command line of the action, guaranteeing a rerun. The action can do something with that data before it decides to read anything, e.g. figuring out which files to read based on what exists.

davidmorgan:
Thanks for taking this on :) ... as I'm not familiar with the details of this code, I just dropped one high-level comment :)

build/CHANGELOG.md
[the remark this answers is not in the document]
jakemac53: Unfortunately right now we don't have a way of tracking existence of a file explicitly - it is either an input or not and we incorporate existence into the content hash.

It is true I suppose though that a kernel action would fail if some transitive dep no longer existed even if it wasn't used?

build/lib/src/builder/build_step_impl.dart
jakemac53: I can guard the cast with an is check and then do `toSet` if it fails.

build/lib/src/generate/run_builder.dart
jakemac53: Ya I like that more - and it also removes the need for the casts to `Set`.

Actually I think I will remove the internal set entirely and move that out to the client code as well, by making this `void Function reportUnusedAsset(AssetId, AssetId)`?

build/lib/src/builder/build_step.dart
```diff
@@ -80,6 +80,10 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// Returns value returned by [action].
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
+
+  void removeDependency(AssetId id);
+
+  Iterable<AssetId> get removedDependencies;
```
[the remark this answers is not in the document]
jakemac53: This is removed now

build/lib/src/builder/build_step.dart
```diff
@@ -80,6 +80,10 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// Returns value returned by [action].
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
+
+  void removeDependency(AssetId id);
```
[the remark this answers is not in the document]
jakemac53: This is a set now, I had the same thought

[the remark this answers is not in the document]
jakemac53: done

build_runner_core/lib/src/generate/build_impl.dart

build/lib/src/builder/build_step_impl.dart
jakemac53: nvm this no longer exists

build/lib/src/generate/run_builder.dart
jakemac53: this is done now

build/CHANGELOG.md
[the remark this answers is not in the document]
jakemac53: I added one mitigation to the actual kernel/ddk worker implementations to mitigate the risk by not reporting unused inputs if the action failed based on offline conversations.

The issue around asset existence is real but I don't believe it will actually cause any meaningful problems - it can cause non-hermetic _failed_ builds but not non-hermetic _successful_ builds, if that makes sense.

build/lib/src/builder/build_step.dart
```diff
@@ -80,6 +80,21 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// Returns value returned by [action].
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
+
+  /// **WARNING**: Expert level feature only, use at your own risk.
```
natebosch: [nit] Needs a header sentence.

```
/// Indicates that [ids] were read but their content has no impact on the outputs of this step.
///
/// ***WARNING***: Using this introduces serious risk of non-hermetic builds.
///
/// ...
```

build/lib/src/generate/run_builder.dart
```diff
@@ -26,12 +26,18 @@ import 'expected_outputs.dart';
 /// automatically disposed of (its up to the caller to dispose of it later). If
 /// one is not provided then one will be created and disposed at the end of
 /// this function call.
+///
+/// If [reportUnusedAssetsForInput] is provided then all calls to
+/// `BuildStep.reportUnusedAsests` in [builder] will be forwarded to this
```
natebosch: [nit] typo - `Asests`

build/CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+    issues if abused.
+- Add the `BuildStep.removedDependencies` iterable which gives access to all
```
natebosch: Changelog still needs updating.

jakemac53:
It looks like we got broken by a new release of front_end or analyzer :( investigating separately in https://github.com/dart-lang/build/issues/2450

build/lib/src/builder/build_step.dart
```diff
@@ -80,6 +80,19 @@ abstract class BuildStep implements AssetReader, AssetWriter {
   /// Returns value returned by [action].
   /// [action] can be async function returning [Future].
   T trackStage<T>(String label, T Function() action, {bool isExternal = false});
+
+  /// Indicates that [ids] were read but their content has no impact on the
+  /// outputs of this step.
+  ///
+  /// **WARNING**: Using this introduces serious risk of non-hermetic builds.
+  ///
+  /// If this file changes or becomes unreadable on the next build this build
```
evanweible-wf: Should this be "If any of these files.."?

jakemac53:
> I'm assuming the `--used-input-files` option is already implemented by the dev compilers, is that right?

Correct - the common frontend implements it and then our kernel/ddc workers use that flag to enable it.

Once this PR lands ill be sending out the next one to start using those flags by default.

build/lib/src/builder/build_step.dart
jakemac53: good point, done

PR #2449, opened 2019-09-19
Add node/** to default sources files

### LINKED ISSUE #2448: Add node/** to the default sources
See https://github.com/pulyaevskiy/node-interop/issues/25

natebosch:
Closes #2448

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.0.1
+
+- Include `node/**` in the default set of sources (when there is no target
+  defined) for the root package.
+
 ## 4.0.0
 
 ### New Feature: Build Filters
--- a/build_runner_core/lib/src/generate/options.dart
+++ b/build_runner_core/lib/src/generate/options.dart
@@ -28,6 +28,7 @@ const List<String> defaultRootPackageWhitelist = [
   'test/**',
   'tool/**',
   'web/**',
+  'node/**',
   'pubspec.yaml',
   'pubspec.lock',
 ];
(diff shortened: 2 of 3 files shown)


natebosch:
Will need a changelog and pubspec update.

I _think_ this is the only place we need to handle this. @pulyaevskiy could you confirm? You can run with this change with a dependency override in your pubspec.

```yaml
dependency_overrides:
  build_runner_core:
    git:
      ref: node-includes
```

pulyaevskiy:
Thanks, this is great!

I tried removing `build.yaml` in e2e test project and in one of the libraries and run the build with this version (also needed `path: build_runner_core`). And it's working! 🎉

I think we should be good.

build_runner_core/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.0.1
```
jakemac53:
```suggestion
## 4.0.1-dev
```

natebosch: It's moot now since I rebased to `4.1.0-dev`

This has come up in the past - I thought we had landed on putting `-dev` in the pubspec but not bothering with the changelog...

jakemac53: I like it in both myself, I would actually lean towards having it in the changelog and not the pubspec if anything?

jakemac53: I mostly just wanted an excuse to use the suggestion feature since @evanweible-wf enlightened me to its presence.

natebosch: > I would actually lean towards having it in the changelog and not the pubspec if anything?

It's easier to forget to touch it if it's only in the changelog.

jakemac53: I disagree, you can't publish a version that is already published so you are physically incapable of not updating the pubspec. The changelog on the other hand is easy to forget.

jakemac53: I still prefer both though, I like them being in sync.

PR #2453, opened 2019-09-20
generalize the required resources feature when bootstrapping ddc apps

jakemac53:
I noticed that when rolling the latest build_web_compilers into flutter_web this was throwing since they handle sdk resources differently.

cc @jonahwilliams


jakemac53:
Fixing the test failures here as a part of landing https://github.com/dart-lang/build/pull/2447, and then I will re-run tests here and merge.

PR #2458, opened 2019-09-23
Support unused inputs reporting for Kernel/DDK

jakemac53:
Uses the new `build` api to report unused kernel dependencies.

The kernel and ddk workers support a flag to write the used dependencies to a file on disk, so we take the difference between all the transitive inputs and this list to compute the unused kernel inputs.

There are a couple defensive mechanisms built into this as well:

- We never report unused deps if there is a failure.
- We never report unused deps if it claims there are zero unused deps (this is assumed to be a failure of some sort, and should only happen if there are no deps to begin with).

### DIFF
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -211,11 +211,8 @@ Future<void> _createKernel(
           .map((line) => kernelInputPathToId[line])
           .where((id) => id != null)
           .toSet();
-      for (var dep in kernelDeps) {
-        if (!usedInputs.contains(dep)) {
-          buildStep.removeDependency(dep);
-        }
-      }
+      buildStep.reportUnusedAssets(
+          kernelDeps.where((id) => !usedInputs.contains(id)));
       await usedInputsFile.parent.delete(recursive: true);
     }
   }
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -165,7 +165,7 @@ Future<void> _createKernel(
     if (trackUnusedInputs) {
       usedInputsFile = await File(p.join(
               (await Directory.systemTemp.createTemp('kernel_builder_')).path,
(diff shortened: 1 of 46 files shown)


jakemac53:
Also apologies for the commit history - I forked the original cl for this work and merging in from master was a lot easier than a rebase.

build_modules/lib/src/kernel_builder.dart
```diff
@@ -131,16 +136,19 @@ Future<void> _createKernel(
...
+    @required bool useIncrementalCompiler,
+    @required bool trackUnusedInputs}) async {
   var request = WorkRequest();
   var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
   var outputId = module.primarySource.changeExtension(outputExtension);
   var outputFile = scratchSpace.fileFor(outputId);
 
   File packagesFile;
+  File usedInputsFile;
+  Map<String, AssetId> kernelInputPathToId;
```
natebosch: can we add a doc comment?

build_modules/lib/src/kernel_builder.dart
```diff
@@ -187,11 +204,57 @@ Future<void> _createKernel(
...
+
+    // Note that we only want to do this on success, we can't trust the unused
+    // inputs if there is a failure.
+    if (usedInputsFile != null) {
+      await reportUnusedKernelInputs(
+          usedInputsFile, kernelDeps, kernelInputPathToId, buildStep);
+    }
   } finally {
     await packagesFile.parent.delete(recursive: true);
+    await usedInputsFile.parent.delete(recursive: true);
```
natebosch: what if this is null?

jakemac53: it throws! :D

build_modules/lib/src/kernel_builder.dart
jakemac53: I added some doc comments to these both and moved them into their own block (same for the ddc builder)

Issue #2464, opened 2019-09-25
build_web_compilers 2.5.0 does not require Dart 2.5.0 in pubspec.yaml, but should


evanweible-wf:
build_web_compilers 2.5.0 was just released and has started causing build failures in projects that are still using a Dart SDK version <2.5.0:

An example of the build failures:
```
[SEVERE] build_web_compilers:ddc on package:fluri/fluri.ddc.module:
Error creating fluri|lib/fluri.ddc.dill
Error creating kernel summary for module:fluri|lib/fluri.ddc.dill

Response:--dart-sdk-summary=file:///usr/local/Cellar/dart/2.4.1/libexec/lib/_internal/ddc_sdk.dill --output=/private/var/folders/vb/k8ccjw095q16jrwktw31ctmm0000gn/T/scratch_spacezDfcoe/packages/fluri/fluri.ddc.dill --packages-file=file:///var/folders/vb/k8ccjw095q16jrwktw31ctmm0000gn/T/kernel_builder_TNt2dG/.packages --multi-root-scheme=org-dartlang-app --exclude-non-sources --summary-only --target=ddc --libraries-file=file:///usr/local/Cellar/dart/2.4.1/libexec/lib/libraries.json --reuse-compiler-result --use-incremental-compiler --used-inputs=/var/folders/vb/k8ccjw095q16jrwktw31ctmm0000gn/T/kernel_builder_ezioKq/used_inputs.txt --source=package:fluri/fluri.dart
FormatException: Could not find an option named "used-inputs".
#0      Parser.validate (package:args/src/parser.dart:261:21)
#1      Parser.parseLongOption (package:args/src/parser.dart:250:7)
#2      Parser.parse (package:args/src/parser.dart:90:11)
#3      ArgParser.parse (package:args/src/arg_parser.dart:307:45)
#4      computeKernel (file:///b/s/w/ir/cache/builder/sdk/utils/bazel/kernel_worker.dart:157:38)
<asynchronous suspension>
#5      KernelWorker.performRequest (file:///b/s/w/ir/cache/builder/sdk/utils/bazel/kernel_worker.dart:58:26)
<asynchronous suspension>
#6      AsyncWorkerLoop.run.<anonymous closure> (package:bazel_worker/src/worker/async_worker_loop.dart:33:41)
#7      _rootRun (dart:async/zone.dart:1124:13)
#8      _CustomZone.run (dart:async/zone.dart:1021:19)
[...]

PR #2465, opened 2019-09-25
ignore the trackInputs option to fix issue #2464

jakemac53:
Adding both versions in this pr - will merge as two separate commits and publish separately.


evanweible-wf:
@jakemac53 @natebosch good to merge?

Issue #2451, opened 2019-09-20
build_runner throwing error while using json_serializable


boylenssen:
* Dart SDK Version (2.5.0)
* build_runner 1.6.7

I'm on MacOS. 

Reproduce steps:
create project `flutter create <some_project>`. 
Add de dependencies to use `json_serializable` => `json_serializable: ^3.2.2`, dev_dependency: `build_runner: ^1.6.7`

Run `flutter packages pub run build_runner build -v`

Pubspec and command output attached.
[build_runner_output.txt](https://github.com/dart-lang/build/files/3634550/build_runner_output.txt)

[pubspec.lock.txt](https://github.com/dart-lang/build/files/3634551/pubspec.lock.txt)
[pubspec.yaml.txt](https://github.com/dart-lang/build/files/3634552/pubspec.yaml.txt)

boylenssen:
I thought I tried it before, but now I added `json_serializable: ^3.2.2` as below `dev_dependencies`, instead of `dependencies`. So documentation of that library might be wrong.

Now I don't get an error.

boylenssen:
Attached are 2 projects, they look identical too me, but one of them gives no issues, the other throws the error when running these commands:

`flutter clean`
`flutter packages pub run build_runner build --delete-conflicting-outputs`


[NOT_working_version.zip](https://github.com/dart-lang/build/files/3634780/NOT_working_version.zip)
[working_version.zip](https://github.com/dart-lang/build/files/3634781/working_version.zip)

boylenssen:
Workaround for now: setting the version of the analyzer in the `pubspec.lock` file from `0.38.3` to `0.38.2`

jakemac53:
The analyzer version is selected by pub based on the constraints in your package and your transitive deps. For flutter projects this will largely be determined by whatever version flutter has pinned for a package, because they pin very specific versions that they have actually tested.

jakemac53:
cc @stereotype441 do you know if this was an intended change or what we should be doing here?

conghaonet:
downgrade analyzer version from 0.38.3 to 0.38.2
dependencies:
  //analyzer: ^0.38.2
  analyzer: 0.38.2

d-jiang:
after downgrade analyszer, run `flutter pub get`

scheglov:
I found the reason why it does not work with the stable version of Flutter SDK.
[...]

PR #2467, opened 2019-09-26
Create a custom summary from the users sdk

jakemac53:
Fixes https://github.com/dart-lang/sdk/issues/38499.
Fixes https://github.com/dart-lang/build/issues/2451.

Lazily generates an sdk summary from the users SDK files and caches it under `.dart_tool/build_resolvers`. On my machine this adds ~3 seconds on first use.

The summary is invalidated based on the `Platform.version` which is cached in a file next to the summary.

A custom summary can also be provided (for non-pub package scenarios). I am going to validate this can be rolled successfully internally before landing.

Note that this has one nice property of allowing build_runner and build_test to share the summary file, which is one advantage over implementing this with a Builder.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -126,8 +126,6 @@ class AnalyzerResolver implements ReleasableResolver {
   }
 }
 
-typedef ResolverFactory = Future<Resolver> Function();
-
 class AnalyzerResolvers implements Resolvers {
   /// Nullable, the default analysis options are used if not provided.
   final AnalysisOptions _analysisOptions;
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -135,7 +135,7 @@ class AnalyzerResolvers implements Resolvers {
   /// Defaults to [_defaultSdkSummaryGenerator].
   final Future<String> Function() _sdkSummaryGenerator;
 
-  /// Lazy, all access must be preceded by a call to [_ensureInitialized].
+  // Lazy, all access must be preceded by a call to `_ensureInitialized`.
   AnalyzerResolver _resolver;
   BuildAssetUriResolver _uriResolver;
 
@@ -191,7 +191,7 @@ Future<String> _defaultSdkSummaryGenerator() async {
 
   var cacheDir = p.join(dartToolPath, 'build_resolvers');
   var summaryPath = p.join(cacheDir, 'sdk.sum');
-  var versionFile = File(summaryPath + '.version');
+  var versionFile = File('$summaryPath.version');
   var summaryFile = File(summaryPath);
 
   // Invalidate existing summary/version files if present.
@@ -206,7 +206,7 @@ Future<String> _defaultSdkSummaryGenerator() async {
     await summaryFile.delete();
   }
 
-  /// Generate the summary and version files if necessary.
+  // Generate the summary and version files if necessary.
   if (!await summaryFile.exists()) {
     var watch = Stopwatch()..start();
     _logger.info('Generating SDK summary...');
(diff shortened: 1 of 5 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -123,31 +126,102 @@ class AnalyzerResolver implements ReleasableResolver {
   }
 }
 
+typedef ResolverFactory = Future<Resolver> Function();
```
natebosch: Unused? Remove it?

build_resolvers/lib/src/resolver.dart
```diff
@@ -123,31 +126,102 @@ class AnalyzerResolver implements ReleasableResolver {
...
-  final BuildAssetUriResolver _uriResolver;
+  /// Nullable, the default analysis options are used if not provided.
+  final AnalysisOptions _analysisOptions;
+
+  /// A function that returns the path to the SDK summary when invoked.
+  ///
+  /// Defaults to [_defaultSdkSummaryGenerator].
+  final Future<String> Function() _sdkSummaryGenerator;
+
+  /// Lazy, all access must be preceded by a call to [_ensureInitialized].
```
natebosch: [nit] 2 slashes, or add a header sentence to format like a doc comment.

build_resolvers/lib/src/resolver.dart
```diff
@@ -123,31 +126,102 @@ class AnalyzerResolver implements ReleasableResolver {
...
+}
+
+/// Lazily creates a summary of the users SDK and caches it under
+/// `.dart_tool/build_resolvers`.
+///
+/// This is only intended for use in typical dart packages, which must
+/// have an already existing `.dart_tool` directory (this is how we
+/// validate we are running under a typical dart package and not a custom
+/// environment).
+Future<String> _defaultSdkSummaryGenerator() async {
```
natebosch: should this be top level?

build_resolvers/lib/src/resolver.dart
```diff
@@ -123,31 +126,102 @@ class AnalyzerResolver implements ReleasableResolver {
...
+    if (lastVersion != Platform.version) {
+      await versionFile.delete();
+      if (await summaryFile.exists()) await summaryFile.delete();
+    }
+  } else if (await summaryFile.exists()) {
+    // If there is no version file we can't validate the version here.
+    await summaryFile.delete();
+  }
+
+  /// Generate the summary and version files if necessary.
```
natebosch: [nit] 2 slashes

build_resolvers/lib/src/resolver.dart
```diff
@@ -123,31 +126,102 @@ class AnalyzerResolver implements ReleasableResolver {
...
+  var dartToolPath = '.dart_tool';
+  if (!await Directory(dartToolPath).exists()) {
+    throw StateError(
+        'The default analyzer resolver can only be used when the current '
+        'working directory is a standard pub package.');
+  }
+
+  var cacheDir = p.join(dartToolPath, 'build_resolvers');
+  var summaryPath = p.join(cacheDir, 'sdk.sum');
+  var versionFile = File(summaryPath + '.version');
```
natebosch: [nit] `'$summaryPath.version'`

https://dart.dev/guides/language/effective-dart/usage#prefer-using-interpolation-to-compose-strings-and-values

build_resolvers/lib/src/resolver.dart
jakemac53: it is?

natebosch: Diff reading fail. Please ignore

jakemac53:
This is going to force a patch internally but there isn't any way to avoid that. I have confirmed a fairly minimal patch can get things passing though.

PR #2470, opened 2019-09-26
check the analyzer path in addition to sdk version before reusing the summary

jakemac53:
At lunch I realized that we also need to check the analyzer version and invalidate if that changes :sob: .


jakemac53:
> This would be fine, would it be sensible to write them both to the same file?

I considered that but I think its cleaner this way, and its more extensible (adding additional lines/info to the one file would be more breaking than adding more checks).

PR #2480, opened 2019-10-03
Read assets from same build step

simolus3:
I've made two changes
1. allow builders to resolve assets that aren't a transitive dependency of their main input. 
2. allow builders to read generated assets written in the same step. There's a special case in `SingleStepReader._isReadable` so that these nodes don't have themselves as primary input in the graph.

### DIFF
--- a/build/lib/src/analyzer/resolver.dart
+++ b/build/lib/src/analyzer/resolver.dart
@@ -27,7 +27,9 @@ abstract class Resolver {
   /// * Throws [NonLibraryAssetException] if [assetId] is not a Dart library.
   Future<LibraryElement> libraryFor(AssetId assetId);
 
-  /// Returns the first library identified by [libraryName].
+  /// Returns the first resolved library identified by [libraryName]. A library
+  /// is resolved if it's recursively accessible from the entry point or
+  /// subsequent calls to [libraryFor] and [isLibrary].
   ///
   /// If no library can be found, returns `null`.
   ///
--- a/build/lib/src/analyzer/resolver.dart
+++ b/build/lib/src/analyzer/resolver.dart
@@ -27,11 +27,11 @@ abstract class Resolver {
   /// * Throws [NonLibraryAssetException] if [assetId] is not a Dart library.
   Future<LibraryElement> libraryFor(AssetId assetId);
 
-  /// Returns the first resolved library identified by [libraryName]. A library
-  /// is resolved if it's recursively accessible from the entry point or
-  /// subsequent calls to [libraryFor] and [isLibrary].
+  /// Returns the first resolved library identified by [libraryName].
   ///
-  /// If no library can be found, returns `null`.
+  /// A library is resolved if it's recursively accessible from the entry point
+  /// or subsequent calls to [libraryFor] and [isLibrary]. If no library can be
+  /// found, returns `null`.
   ///
   /// **NOTE**: In general, its recommended to use [libraryFor] with an absolute
   /// asset id instead of a named identifier that has the possibility of not
--- a/build/lib/src/analyzer/resolver.dart
+++ b/build/lib/src/analyzer/resolver.dart
@@ -16,7 +16,8 @@ abstract class Resolver {
   /// or is a `part of` file (not a standalone Dart library).
   Future<bool> isLibrary(AssetId assetId);
 
-  /// All libraries accessible from the entry point, recursively.
+  /// All libraries recursively accessible from the entry point or subsequent
+  /// calls to [libraryFor] and [isLibrary].
   ///
   /// **NOTE**: This includes all Dart SDK libraries as well.
   Stream<LibraryElement> get libraries;
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -40,17 +40,26 @@ class BuildAssetUriResolver extends UriResolver {
   /// When actions can run out of order an asset can move from being readable
   /// (in the later phase) to being unreadable (in the earlier phase which ran
   /// later). If this happens we don't want to hide the asset from the analyzer.
-  final seenAssets = HashSet<AssetId>();
+  final globallySeenAssets = HashSet<AssetId>();
+
+  /// The assets which have been resolved from a [BuildStep], either as an
+  /// input, subsequent calls to a resolver, or a transitive import thereof.
+  final _buildStepAssets = <BuildStep, HashSet<AssetId>>{};
 
   /// Crawl the transitive imports from [entryPoints] and ensure that the
   /// content of each asset is updated in [resourceProvider] and [driver].
   Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
       AnalysisDriver driver) async {
+    final seenInBuildStep =
+        _buildStepAssets.putIfAbsent(buildStep, () => HashSet());
+    final relevantEntries =
+        entryPoints.where((asset) => !seenInBuildStep.contains(asset));
+
     final changedPaths =
-        await crawlAsync<AssetId, _AssetState>(entryPoints, (id) async {
+        await crawlAsync<AssetId, _AssetState>(relevantEntries, (id) async {
       final path = assetPath(id);
       if (!await buildStep.canRead(id)) {
-        if (seenAssets.contains(id)) {
+        if (globallySeenAssets.contains(id)) {
           // ignore from this graph, some later build step may still be using it
           // so it shouldn't be removed from [resourceProvider], but we also
           // don't care about it's transitive imports.
@@ -63,7 +72,8 @@ class BuildAssetUriResolver extends UriResolver {
         }
         return _AssetState.removed(path);
       }
-      seenAssets.add(id);
+      globallySeenAssets.add(id);
+      seenInBuildStep.add(id);
       final digest = await buildStep.digest(id);
       if (_cachedAssetDigests[id] == digest) {
         return _AssetState.unchanged(path, _cachedAssetDependencies[id]);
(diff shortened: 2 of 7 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -60,11 +61,20 @@ class PerActionResolver implements ReleasableResolver {
...
+  Future<bool> isLibrary(AssetId assetId) =>
+      _resolveIfNecesssary(assetId).then(_delegate.isLibrary);
 
   @override
   Future<LibraryElement> libraryFor(AssetId assetId) =>
-      _delegate.libraryFor(assetId);
+      _resolveIfNecesssary(assetId).then(_delegate.libraryFor);
+
+  Future<AssetId> _resolveIfNecesssary(AssetId id) async {
+    if (!_delegate._uriResolver.seenAssets.contains(id)) {
```
natebosch: `seenAssets` is not specific to a `buildStep` so this could cause use to expose information without properly calling `canRead` for this step. I think we need to somehow have a per-build step understanding of what we have already crawled...

build_resolvers/lib/src/resolver.dart
```diff
@@ -60,11 +61,20 @@ class PerActionResolver implements ReleasableResolver {
...
 
   @override
-  Future<bool> isLibrary(AssetId assetId) => _delegate.isLibrary(assetId);
+  Future<bool> isLibrary(AssetId assetId) =>
+      _resolveIfNecesssary(assetId).then(_delegate.isLibrary);
 
   @override
   Future<LibraryElement> libraryFor(AssetId assetId) =>
-      _delegate.libraryFor(assetId);
+      _resolveIfNecesssary(assetId).then(_delegate.libraryFor);
```
natebosch: We already had a potential for non-hermetic information transfer here. We need to carefully consider whether this makes things any worse.

build_resolvers/lib/src/resolver.dart
jakemac53: Ya, I didn't realize this wasn't per build step - not guarding it behind this check would also mean a regression for _all_ calls to `libraryFor` so we can't just omit this either.

If we do track this per build step we can potentially make `performResolve` more efficient too, so it doesn't re-crawl things its already crawled in the same step.

build_resolvers/lib/src/resolver.dart
simolus3: Am I correct to assume that this is the potential problem (with or without the `_resolveIfNecessary` step)?

1. Builder foo is run on `a|lib/foo.dart`, which doesn't import anything
2. Builder bar is run on `a|lib/bar.dart`, but calls `libraryFor(a|lib/foo.dart)`. That works, because the library has already been resolved, but doesn't recognize `foo.dart` as an input.

In that case, keeping a per-buildstep list of assets that have been crawled __and__ wrapping the resolver in `_resolveIfNecessary` would improve the situation, right?

simolus3: Just say the comment that mentions crawling the element model manually. I don't think that gets worse now as the `libraries` stream only crawls from the build step input. Should we include libraries found via subsequent calls to `libraryFor` in that stream?

jakemac53: We were talking in person about this exact same question haha. It isn't clear what should happen. It is weird for it to change after you ask for a new library.

I think probably the best situation is if it was a method that took a root AssetId to crawl from, but that would be a breaking change now.

I would probably be fine with leaving it as is and we could add a method like `transitiveLibrariesFor(AssetId)`  or something in the future if people need to crawl libraries from the non-primary input.

simolus3: > The thing that sucks about that though is that `findLibraryByName` won't work for things that aren't imported by the primary input.

What do you think is the best way to solve that? Keep a per-buildstep map of `<String, LibraryElement>` based on crawled assets? We could populate that map in the `BuildAssetUriResolver`.

simolus3: I think including them in the stream if the `libraries` getter has been called _after_ a previous call to `libraryFor` makes sense - after all, the library has been found at that point and is accessible for the resolver. That also means that it will be included in `findLibraryByName`.

simolus3:
@natebosch Is there any update on this? I think I addressed your last review by constructing a per-buildstep set of crawled assets. Sorry for the nagging, but it's been some time and I would love to work on something that requires this.

jakemac53:
Aplogies for the delay here. One of us will get back to this soon (I will try and at least took by EOW).

build/lib/src/analyzer/resolver.dart
```diff
@@ -16,7 +16,8 @@ abstract class Resolver {
   /// or is a `part of` file (not a standalone Dart library).
   Future<bool> isLibrary(AssetId assetId);
 
-  /// All libraries accessible from the entry point, recursively.
+  /// All libraries recursively accessible from the entry point or subsequent
+  /// calls to [libraryFor] and [isLibrary].
```
jakemac53: We should also add a similar note to `findLibraryByName` that it can only find libraries that have been previously resolved (or are transitive imports of one)

build_resolvers/lib/src/resolver.dart
```diff
@@ -25,16 +25,24 @@ final _logger = Logger('build_resolvers');
...
 
   @override
   Stream<LibraryElement> get libraries async* {
     final seen = Set<LibraryElement>();
     final toVisit = Queue<LibraryElement>();
-    for (final entryPoint in _entryPoints) {
+
+    // keep a copy of entry points in case [_resolveIfNecessary] is called
+    // before this stream is done.
+    final entryPoints = Set.of(_entryPoints);
```
jakemac53: nit: we can just create a list at this point which would be a bit cheaper (we already know there are no duplicates)

You can just use `.toList()` to get that.

build_resolvers/lib/src/resolver.dart
```diff
@@ -60,14 +68,31 @@ class PerActionResolver implements ReleasableResolver {
...
-  Future<bool> isLibrary(AssetId assetId) => _delegate.isLibrary(assetId);
+  Future<bool> isLibrary(AssetId assetId) =>
+      _resolveIfNecesssary(assetId).then(_delegate.isLibrary);
 
   @override
   Future<LibraryElement> libraryFor(AssetId assetId) =>
-      _delegate.libraryFor(assetId);
+      _resolveIfNecesssary(assetId).then(_delegate.libraryFor);
+
+  Future<AssetId> _resolveIfNecesssary(AssetId id) async {
```
jakemac53: In general we avoid returning things just for the purpose of creating a fluid interface (https://dart.dev/guides/language/effective-dart/design#avoid-returning-this-from-methods-just-to-enable-a-fluent-interface). This doesn't fit perfectly with that guideline but I think it fits the spirit of it.

I would probably just return a `Future<void>` here and then use async await at the call sites:

```dart
Future<bool> isLibrary(AssetId assetId) async {
  await _resolveIfNecesssary(assetId);
  return _delegate.isLibrary(assetId);
}
```

I am curious what @natebosch thinks... I don't feel super strongly given this is a private api.

natebosch: I agree with @jakemac53 - let's make this return `Future<void>`

srikrsna:
Will this fix #2389 ?

jakemac53:
Travis had a flake downloading the sdk - I restarted it so as long as that is green I think we are pretty much ready to land this.

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -40,17 +40,25 @@ class BuildAssetUriResolver extends UriResolver {
...
 
   /// Crawl the transitive imports from [entryPoints] and ensure that the
   /// content of each asset is updated in [resourceProvider] and [driver].
   Future<void> performResolve(BuildStep buildStep, List<AssetId> entryPoints,
       AnalysisDriver driver) async {
-    final changedPaths =
-        await crawlAsync<AssetId, _AssetState>(entryPoints, (id) async {
+    final seenInBuildStep =
+        _buildStepAssets.putIfAbsent(buildStep, () => HashSet());
+    final notCrawled = (AssetId asset) => !seenInBuildStep.contains(asset);
```
natebosch: https://dart.dev/guides/language/effective-dart/usage#do-use-a-function-declaration-to-bind-a-function-to-a-name

```suggestion
bool notCrawled(AssetId asset) => !seenInBuildStep.contains(asset);
```

build/lib/src/analyzer/resolver.dart
```diff
@@ -26,7 +27,9 @@ abstract class Resolver {
   /// * Throws [NonLibraryAssetException] if [assetId] is not a Dart library.
   Future<LibraryElement> libraryFor(AssetId assetId);
 
-  /// Returns the first library identified by [libraryName].
+  /// Returns the first resolved library identified by [libraryName]. A library
```
natebosch: [nit] Blank line after the first sentence.

https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

jakemac53:
I will work on updating pubpsecs/changelogs across packages today and get things merged :tada:.

Thanks @srikrsna for your patience!

jakemac53:
This has all been published, let me know if you see any issues!

simolus3:
Thanks @jakemac53, I managed to parse and resolve Dart expressions in a builder using the latest resolver version.
A small gotcha I ran into was using `testBuilder` from `build_test`. By default, the asset reader used can't read written outputs, so it behaves different to a regular build. I've written [a workaround](https://github.com/simolus3/moor/blob/261fbe6ba0dee0b339fa9bd3458229d382dc8af2/moor_generator/test/test_utils.dart#L7-L40) for my tests, but I'm wondering if this should be the default behavior of `testBuilder`. It's simple to fix and I can open another PR for that if you agree on changing the behavior.

jakemac53:
Yes changing the behavior of `testBuilder` sounds like the correct approach - it is sort of a mini-implementation of the build package for tests and thus will need to conform to the new behavior.

PR #2481, opened 2019-10-04
Use the kernel version of dart_sdk.js

natebosch:
We ship with two copies of the SDk, one compiled with kernel and the
other with analyzer. There may be subtle behavior differences between
the two.


natebosch:
Yet another reason it's good this is no longer hardcoded in `build_runner`...

Issue #2484, opened 2019-10-04
skipPlatformCheck logic is negated


jonahwilliams:
https://github.com/dart-lang/build/blob/68b3d29a624754959d65aefcc509a918cca11746/build_web_compilers/lib/src/dev_compiler_bootstrap.dart#L167

This should be `throwIfUnsupported: !skipPlatformCheck`

jakemac53:
doh! :man_facepalming:

PR #2485, opened 2019-10-04
fix skipPlatformCheck args

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2484


Hixie:
FYI: After this commit, Flutter for Web no longer builds. I'm going to pin us to 2.6.1 for now.

cc @yjbanov

Hixie:
never mind, it broke us because we were working around the bug this pr fixed

PR #2489, opened 2019-10-08
Give a more consistent ordering for builders

### LINKED ISSUE #2487: Builder ordering appears to be different on windows
When adding windows bots for all our unit tests I had to skip the `_test/test-generated_script_integration_test.dart`, it looks like it is ordering build_web_compilers before build_vm_compilers when the opposite is expected.

natebosch:
Fixes #2487 

The strongly connected components algorithm doesn't make any guarantees
about the ordering of values that don't depend on eachother, when there
are multiple valid topological orderings it doesn't make any guarantees
about which it will use. We can count on it being consistent though, it
is based on the order of the children it sees from `edges` for each
node.

The ordering of `builders` when we look for the topological sort may
depend on both the ordering of the topological sort of the package
graph, as well as the order it iterates keys within the builder
definitions from the parsed `build.yaml`. Since these aren't guaranteed
we also don't have a stable ordering of the builders to start with. The
order of the edges is based on the order we see them here.

Since we don't care about _which_ arbitrary ordering we get, it should
be sufficient to pick some to start with. We can sort by builder keys
which are guaranteed to be unique.


natebosch:
This makes us less likely to do an "expected" order between builders that aren't _required_ to be in any given order, but are defined by packages that have an ordering relationship via their dependencies. This shouldn't matter, since it isn't required by their inputs or `runs_before` etc, but it could also cause subtle behavior changes in poorly configured builders that were relying on it using working.

We could pick an order that also takes package dependency relationships into account, but it would be more invasive and the code that focuses on ordering would be scattered in a few more places. It would also run counter to the idea that this ordering is meant to be arbitrary unless explicitly expressed in our supported config.

_test/test/goldens/generated_build_script.dart
```diff
@@ -108,23 +101,30 @@ final _builders = <_i1.BuilderApplication>[
       }),
       defaultReleaseOptions: _i8.BuilderOptions({'compiler': 'dart2js'}),
       appliesBuilders: ['build_web_compilers:dart2js_archive_extractor']),
+  _i1.apply('build_web_compilers:sdk_js_copy', [_i7.sdkJsCopyBuilder],
```
jakemac53: We might want to add a `runs_before: [build_web_compilers|ddc]` to this one...

jakemac53: Sorry that would be `build_web_compilers|entrypoint`.

jakemac53:
Btw pinning analyzer to 0.38.4 fixes the current failure. I am checking that in as a part of my windows tests pr.

_test/test/goldens/generated_build_script.dart
natebosch: https://github.com/dart-lang/build/pull/2490

PR #2496, opened 2019-10-11
Adding logic to post-process extension method names from DDC stack traces.

Markzipan:
State of our current stack trace post-processing:

* Running with webdev loads the mapper in build/build_web_compilers/web/stack_trace_mapper.dart.

* Flutter web depends on dev_compiler/web/stack_trace_mapper.dart (hard-coded here: https://github.com/flutter/flutter/blob/master/packages/flutter_tools/lib/src/build_runner/web_fs.dart#L389) to process the stringified stack trace. I suspect google3 does the same thing.

* Flutter devtools depends on webdev/dwds/debugging/debugger.dart to process the stack frame (but the stringified stack message is still processed by the mapper in build_web_compilers).

So whenever the scheme used for renaming JS identifiers/names changes (either in Kernel or DDC), we need to update build/build_web_compilers/web/stack_trace_mapper.dart (mapStackTrace) + copy it over to dev_compiler/web AND update webdev/dwds/debugging/debugger.dart (dartFramesFor).

See: https://github.com/dart-lang/webdev/pull/723

### DIFF
--- a/build_web_compilers/pubspec.yaml
+++ b/build_web_compilers/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 2.6.4-dev
+version: 2.6.4
 description: Builder implementations wrapping Dart compilers.
 author: Dart Team <misc@dartlang.org>
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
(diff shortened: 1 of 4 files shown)


Markzipan:
cc: @sigmundch

build_web_compilers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 2.6.3
+version: 2.6.4-dev
```
jakemac53:
```suggestion
version: 2.6.4
```

PR #2500, opened 2019-10-12
Ensure all help is handled in outer binary

### LINKED ISSUE #2498: Help commands require generating build script
(On the latest of all build_* packages.)

When running a command with the `-h|--help` flag, the `build_runner` executable has to first go through the process of generating the build script and then snapshotting it if the asset graph has changed or is missing, e.g.:

```bash
$ pub run build_runner serve -h
[INFO] Generating build script completed, took 322ms
[WARNING] Deleted previous snapshot due to missing asset graph.
[INFO] Creating build script snapshot... completed, took 11.0s
Runs a development server that serves the specified targets and runs builds based on file system updates.

Usage: build_runner serve [arguments] [directories] [<directory>[:<port>]]...
```

Could the script generation and snapshot code paths be short-circuited if the help flag is provided?

### LINKED ISSUE #2499: build_runner help <command> prints usage info for root exe
(On the latest of all build_* packages.)

The help output for the root `build_runner` exe states at the end that you can run `build_runner help <command>` in order to get help info specific to that command (afaik this is provided by the `CommandRunner` class):

```
$ pub run build_runner -h
...
Run "build_runner help <command>" for more information about a command.
```

However, running `build_runner help <command>` just prints out the same usage info and not the info specific to the command:

```
$ pub run build_runner help serve
Unified interface for running Dart builds.

Usage: build_runner <command> [arguments]

Global options:
-h, --help    Print this usage information.

Available commands:
  build   Performs a single build on the specified targets and then exits.
  clean   Cleans up output from previous builds. Does not clean up --output directories.
  help    Display help information for build_runner.
  run     Performs a single build, and executes a Dart script with the given arguments.
  serve   Runs a development server that serves the specified targets and runs builds based on file system updates.
  test    Performs a single build of the test directory only and then runs tests using the compiled assets.
  watch   Builds the specified targets, watching the file system for updates and rebuilding as appropriate.

Run "build_runner help <command>" for more information about a command.
```

natebosch:
Fixes #2498, fixes #2499

Previously we'd only handle help in the outer binary if it was invoked
specifically as the `help` command, or if no command was given. If we
passed `-h` as a sub argument to another command we wouldn't recognize
it and would still generate the build script. We also always printed the
top level usage rather than the usage that would be printed by the real
help command, which may be specific to a sub command.

- Dig 2 levels deep for the `-h` or `--help` flags. This way it can be
  passed as either a top level flag, or a flag for the specific sub
  command.
- When printing usage for help specifically, use the command runner
  since it already implements figuring out whether the print outer usage
  or usage for a sub command.
- Handle the missing command as it's own case after we know we didn't
  handle help. Set the exit code to `ExitCode.usage` for this case, and
  print the top level usage.

This pattern won't work if we start adding nested commands more deeply
within other commands since we only pick up the `-h` from the top level
or 1 command deep.

### DIFF
--- a/_test/test/help_test.dart
+++ b/_test/test/help_test.dart
@@ -4,7 +4,6 @@
 
 @TestOn('vm')
 import 'dart:async';
-import 'dart:io';
 
 import 'package:test/test.dart';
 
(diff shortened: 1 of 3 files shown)


jakemac53:
It might be worth a quick test too for this to ensure we don't regress it

natebosch:
Updated tests for help output so that it looks for specific strings for each command.

Issue #733, opened 2017-12-18
How should we deal with flutter and "dart:ui" in generators?


davidmorgan:
Running the built_value generator on flutter code currently spits out a severe error:

```[SEVERE] Instance of 'PartBuilder' on dartsugar|lib/models/module_data.dart: Unable to resolve asset ID for "dart:ui"```

It still works, but I wonder if 1) there is potential for things to break down the line, in which case we should make sure this doesn't happen; and 2) there's a way to do it without getting the warning. Any ideas please?

See: https://github.com/google/built_value.dart/issues/282

jakemac53:
This seems like a bug, it looks like you are hitting this line https://github.com/dart-lang/code_transformers/blob/718b2b206d0ed0bac7214c5678d4d58c28f73f67/lib/src/resolver_impl.dart#L458, which has an assert above it that the [scheme != dart](https://github.com/dart-lang/code_transformers/blob/718b2b206d0ed0bac7214c5678d4d58c28f73f67/lib/src/resolver_impl.dart#L450), so in checked mode this would throw.

natebosch:
This could be tricky. The SDK we use while generating the code doesn't have `dart:ui` so it probably can't resolve any of it. We'll probably have to add the ability to point to a different SDK or something for the Resolver?

matanlurey:
Yeah, we'll need a way to use Flutter's Dart SDK, not the same as `which dart`.

I'm making this M1 because it doesn't prevent the initial MVP for the web.

davidmorgan:
Thanks. Yes, I don't think this is blocking right now.

matanlurey:
I thought about this a little bit, it might be simple if Flutter exposes some help here:

https://github.com/flutter/flutter/issues/13360

matanlurey:
I thought about this a bit, this is tricky, because the Flutter Dart _tool_ SDK isn't the same as the Flutter Dart _engine_ SDK; we need the second one if you want to be able to resolve `dart:ui`.

matanlurey:
I guess `flutter script` (if added) could provide the Flutter engine location as `DART_SDK=..`

s-bauer:
Any update on this? It's a problem because the reflectable package is also unable to correctly handle types like `Color` from the dart:ui package

eernstg:
@matanlurey, concerning

> the Flutter Dart tool SDK isn't the same as the Flutter Dart engine SDK;
> we need the second one if you want to be able to resolve dart:ui.

[...]

PR #2501, opened 2019-10-14
add flutter sdk support to build_resolvers

jakemac53:
Fixes https://github.com/dart-lang/build/issues/733

We always start with the Dart sdk, but now if we can find the `_embedder.yaml` file for the flutter sdk then we add any libraries found in that to the Dart sdk before summarization.

This has the effect of making any flutter libraries available to builds that are using the flutter sdk.

~WIP: Still thinking the best way to test this, we probably have to download the whole flutter sdk and run a test using that.~ Test has been added now.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -280,10 +280,8 @@ void _addFlutterLibraries(AbstractDartSdk sdk,
   var embedderYamlFile =
       resourceProvider.getFile(p.join(embedderSdkPath, '_embedder.yaml'));
   if (!embedderYamlFile.exists) {
-    _logger.warning(
-        'Unable to find flutter _embedder.yaml file, flutter types will not '
-        'be resolvable.');
-    return;
+    throw StateError('Unable to find flutter libraries, please run '
+        '`flutter precache` and try again.');
   }
 
   var embedderYaml = loadYaml(embedderYamlFile.readAsStringSync()) as YamlMap;
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -6,15 +6,20 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:io';
 
-import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/src/dart/analysis/driver.dart' show AnalysisDriver;
-import 'package:analyzer/src/generated/engine.dart' hide Logger;
-import 'package:analyzer/src/generated/source.dart';
 import 'package:analyzer/src/summary/summary_file_builder.dart';
+
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/file_system/physical_file_system.dart';
+import 'package:analyzer/src/generated/sdk.dart';
+import 'package:analyzer/src/generated/source.dart';
+import 'package:analyzer/src/dart/analysis/driver.dart' show AnalysisDriver;
+import 'package:analyzer/src/dart/sdk/sdk.dart';
+import 'package:analyzer/src/generated/engine.dart' hide Logger;
 import 'package:build/build.dart';
 import 'package:logging/logging.dart';
 import 'package:package_resolver/package_resolver.dart';
 import 'package:path/path.dart' as p;
+import 'package:yaml/yaml.dart';
 
 import 'analysis_driver.dart';
 import 'build_asset_uri_resolver.dart';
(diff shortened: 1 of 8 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -6,15 +6,20 @@ import 'dart:async';
 import 'dart:collection';
 import 'dart:io';
 
+import 'package:analyzer/src/summary/summary_file_builder.dart';
+
```
natebosch: [nit] remove extra blank line

build_resolvers/lib/src/resolver.dart
```diff
@@ -249,3 +254,38 @@ Future<void> _createSdkVersionFile(File sdkVersionFile) async {
   await sdkVersionFile.create(recursive: true);
   await sdkVersionFile.writeAsString(Platform.version);
 }
+
+List<int> buildSdkSummary(String dartSdkPath) {
```
natebosch: private?

build_resolvers/lib/src/resolver.dart
```diff
@@ -249,3 +254,38 @@ Future<void> _createSdkVersionFile(File sdkVersionFile) async {
...
+  // Prepare the Dart SDK.
+  var resourceProvider = PhysicalResourceProvider.INSTANCE;
+  var dartSdkFolder = resourceProvider.getFolder(dartSdkPath);
+  var dartSdk = FolderBasedDartSdk(resourceProvider, dartSdkFolder, true)
+    ..useSummary = false
+    ..analysisOptions = AnalysisOptionsImpl();
+
+  // Try to load the flutter engine _embedder.yaml file and add any missing
```
natebosch: Consider extracting some methods with clear names over commenting blocks of code.

build_resolvers/lib/src/resolver.dart
```diff
@@ -249,3 +254,38 @@ Future<void> _createSdkVersionFile(File sdkVersionFile) async {
...
+  // libraries to the dart sdk.
+  var embedderSdkPath =
+      p.join(p.dirname(dartSdkPath), 'pkg', 'sky_engine', 'lib');
+  var embedderYamlFile =
+      resourceProvider.getFile(p.join(embedderSdkPath, '_embedder.yaml'));
```
natebosch: Does `resourceProvider` expect platform specific directory separators or should these be normalized with `p.url`?

jakemac53: Once I figure out how to write a test we will find out ;)

jakemac53:
> What should our testing strategy be for this? I suspect we don't want to bring in a Flutter SDK on travis?

I think we do essentially have to do that. I want to try and make that only happen for this test suite though.

build_resolvers/test/flutter_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Fast fail the script on failures.
+set -ex
```
natebosch: There are no `select`, `case`, or `for` here. I'm not sure why we'd want the extra verbosity if there were... Why do we need the `x` option?

build_resolvers/test/flutter_test.sh
```diff
@@ -0,0 +1,10 @@
...
+
+# Get Flutter.
+echo "Cloning master Flutter branch"
+git clone https://github.com/flutter/flutter.git ./flutter
+
+./flutter/bin/cache/dart-sdk/bin/dart test/resolver_test
```
natebosch: I thought `flutter/bin/cache` didn't exist until after some setup step - is this working for you?

build_resolvers/test/flutter_test.sh
jakemac53: was just copy/pasting the header from other scripts, I can remove it though. The `-e` is the only part that really matters certainly.

build_resolvers/test/flutter_test.sh
jakemac53: Nope - still working through this.

I grabbed these travis setup steps from the devtools repo but they must be doing some additional stuff later on.

build_resolvers/lib/src/resolver.dart
jakemac53: Looks like this works!

build_resolvers/test/resolver_test.dart
```diff
@@ -195,4 +197,36 @@ void main() {
...
+
+                MyClass(this.color);
+              } ''',
+      }, (resolver) async {
+        var entry = await resolver.libraryFor(AssetId('a', 'lib/a.dart'));
+        var classDefinition = entry.getType('MyClass');
+        var color = classDefinition.getField('color');
+
+        if (_isFlutter) {
+          expect(color.type.name, equals('Color'));
```
natebosch: [nit] I think we typically haven't used `equals` in this repo

jakemac53: :man_shrugging: not worth rerunning the ci imo

PR #2505, opened 2019-10-16
Use the native windows directory watcher by default

jakemac53:
Fixes https://github.com/dart-lang/webdev/issues/436

Updates the `daemon`, `serve` and `watch` commands to use the native watcher on windows by default, as well as adding a `--use-polling-watcher` flag which reverts to the old watcher.

If we see an exception coming from the directory watcher we log a severe warning pointing users at that flag (as of https://github.com/dart-lang/sdk/commit/8866cdb8be58f9228548436a3bfee939bc1887e4 it will throw if the native watcher stops working).

Updates the min sdk to the not-yet-released 2.6.0-dev.8.0 which has the new exception. I will wait for that to actually be released before merging/publishing this.

cc @supermuka @mnordine you can try this out in the meantime with this dependency override:

```yaml
dependency_overrides:
  build_runner:
    git:
      ref: windows-file-watcher
      path: build_runner
```

### DIFF
--- a/build_runner/lib/src/entrypoint/options.dart
+++ b/build_runner/lib/src/entrypoint/options.dart
@@ -12,24 +12,28 @@ import 'package:build_daemon/constants.dart';
 import 'package:build_runner_core/build_runner_core.dart';
 import 'package:meta/meta.dart';
 import 'package:path/path.dart' as p;
+import 'package:watcher/watcher.dart';
+
+import '../generate/directory_watcher_factory.dart';
 
 const buildFilterOption = 'build-filter';
+const configOption = 'config';
 const defineOption = 'define';
 const deleteFilesByDefaultOption = 'delete-conflicting-outputs';
-const liveReloadOption = 'live-reload';
+const failOnSevereOption = 'fail-on-severe';
+const hostnameOption = 'hostname';
 const hotReloadOption = 'hot-reload';
+const liveReloadOption = 'live-reload';
 const logPerformanceOption = 'log-performance';
 const logRequestsOption = 'log-requests';
 const lowResourcesModeOption = 'low-resources-mode';
-const failOnSevereOption = 'fail-on-severe';
-const hostnameOption = 'hostname';
 const outputOption = 'output';
-const configOption = 'config';
-const verboseOption = 'verbose';
 const releaseOption = 'release';
 const trackPerformanceOption = 'track-performance';
 const skipBuildScriptCheckOption = 'skip-build-script-check';
 const symlinkOption = 'symlink';
+const usePollingWatcherOption = 'use-polling-watcher';
+const verboseOption = 'verbose';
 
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/entrypoint/options.dart
```diff
@@ -98,18 +102,9 @@ class SharedOptions {
 
   factory SharedOptions.fromParsedArgs(ArgResults argResults,
       Iterable<String> positionalArgs, String rootPackage, Command command) {
-    var buildDirs = _parseBuildDirs(argResults);
+    var buildDirs = _parseBuildDirs(argResults)
+      ..addAll(_parsePositionalBuildDirs(positionalArgs, command));
```
natebosch: [optional] How about a list literal?

```dart
var buildDirs = [
  ..._parseBuildDirs(argResults),
  ..._parsePositionalBuildDirs(positionalArgs, command),
];
```

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -419,6 +473,26 @@ Set<BuildDirectory> _parseBuildDirs(ArgResults argResults) {
...
 }
 
+/// Parses positional arguments as plain build directories.
+///
+/// Only allows top level directories as arguments, otherwise a
+/// [UsageException] is thrown.
+Set<BuildDirectory> _parsePositionalBuildDirs(
+    Iterable<String> positionalArgs, Command command) {
+  var buildDirs = <BuildDirectory>{};
+  for (var arg in positionalArgs) {
```
natebosch: [optional] How about

```dart
String _checkTopLevel(String arg) {
  var parts = p.split(arg);
  if (parts.length > 1 || arg == '.') {
    throw UsageException(...);
  }
  return arg;
}

Set<BuildDirectory> _parsePositionalBuildDirs(...) =>
    {for(var arg in positionalArgs) BuildDirectory(_checkTopLevel(arg))};
```

build_runner/lib/src/entrypoint/options.dart
jakemac53: well, in this case it would actually be less efficient and longer?

natebosch: IMO it would be more clear. In this case I didn't even get it correct, I should have used `{}` instead of `[]`. If we us a Set literal that would be explicit. We could also change both `_parse` to return an Iterable and not lose any efficiency from this version.

jakemac53: now you will just wonder if its a map or set :D

jakemac53:
Note that `2.6.0-dev.8.0` is now released so I will go ahead and merge this and publish once its green.

jakemac53:
released/tagged

mnordine:
Working!!! :)

supermuka:
Wonderful, it's working for me too! :)

PR #2506, opened 2019-10-16
enable source span tests on windows by forcing ascii rendering

jakemac53:


build_config/test/errors_test.dart
```diff
@@ -2,11 +2,14 @@
...
 // BSD-style license that can be found in the LICENSE file.
 
-@OnPlatform({'windows': Skip('yaml errors look different on windows')})
 import 'package:build_config/build_config.dart';
+import 'package:term_glyph/term_glyph.dart' as glyph;
 import 'package:test/test.dart';
 
 void main() {
+  // Ensures consistent rendering on windows/linux.
+  glyph.ascii = false;
```
natebosch: I had the impression that we'd also have to change our characters on lines 22-24...

jakemac53: This is forcing it into unicode mode which is what those are - ascii is the fallback for windows which would normally be on by default.

natebosch: Reading comprehension fail on my part. I think I'm more used to see this forced to `true` for tests and assumed this was doing the same.

PR #2507, opened 2019-10-16
add environment option to the build_web_compilers|ddc builder

jakemac53:
- Added `environment` option to the `KernelBuilder`, configuration is up to other packages that use this builder.
- Added an `environment` option to the `DevCompilerBuilder`.
  - This can be configured using the `environment` option of the
    `build_web_compilers|ddc` builder.
  - This is forwarded to both kernel and dev compiler workers.
  - The expected value is a `Map<String, String>` and is equivalent to
    providing `-D<key>=<value>` command line arguments.
  - This option should only be set globally, and will throw if it ever recieves
    two different values. This is to ensure all modules are compiled with the
    same environment.

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_builder.dart
+++ b/build_web_compilers/lib/src/dev_compiler_builder.dart
@@ -47,7 +47,7 @@ class DevCompilerBuilder implements Builder {
   final String librariesPath;
 
   /// Any extra args to pass to ddc.
-  final List<String> ddcArgs;
+  final Map<String, String> environment;
 
   DevCompilerBuilder(
       {bool useIncrementalCompiler,
@@ -56,7 +56,7 @@ class DevCompilerBuilder implements Builder {
       this.sdkKernelPath,
       String librariesPath,
       String platformSdk,
-      List<String> ddcArgs})
+      Map<String, String> environment})
       : useIncrementalCompiler = useIncrementalCompiler ?? true,
         platformSdk = platformSdk ?? sdkDir,
         librariesPath = librariesPath ??
@@ -69,7 +69,7 @@ class DevCompilerBuilder implements Builder {
             jsSourceMapExtension
           ],
         },
-        ddcArgs = ddcArgs ?? <String>[];
+        environment = environment ?? {};
 
   @override
   final Map<String, List<String>> buildExtensions;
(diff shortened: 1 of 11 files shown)


jakemac53:
I like having this more generic than supporting environment explicitly.

We already support `dart2js_args` for the entrypoint builder - dart2js has to be configured in a separate builder like that unfortunately.

natebosch:
One concern I have about exposing a `ddc_args` is that is expands the surface area that would be breaking to change in DDC. Today we lean on the fact that we only really "support" the arguments and argument combinations that we use in practice here and internally. Once we open this up free-form then any user could start relying on any arguments...

jakemac53:
cc @Markzipan @nshahan @vsmenon for thoughts on exposing arbitrary args.

Imo it greatly simplifies the logic here and is more usable from a user perspective (they don't have to understand as many concepts/custom options), but Nate does have some potentially valid concerns.

jakemac53:
Note that I realized this also has to be plumbed through the kernel worker - which isn't supported in the current SDK. I am working on an upstream change for that.

This does also swing the balance pretty clearly in favor of a custom `environment` option since its no longer ddc specific.

jakemac53:
Ok, I pushed an update here but note that the tests will fail until I can upstream changes to support the kernel argument.

Also today it appears that kernel is not in fact inlining these constants, so we may not need to pass this along after all. It is unclear so I will try to confirm with @johnniwinther what the expected behavior is here.

johnniwinther:
Constants are not evaluated in outlines. This was not a deliberate choice, but we like it, so we'll keep it that way.

jakemac53:
Ok @natebosch ptal I removed all kernel updates here so we will no longer provide an environment there given that it isn't required (and isn't supported in the current sdk).

jakemac53:
@johnniwinther would you be willing to add a test to the sdk that constants are not inlined in outlines just to help ensure we don't regress here? (or is there one already)

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -46,13 +46,17 @@ class DevCompilerBuilder implements Builder {
   /// If not provided, defaults to "lib/libraries.json" in the sdk directory.
   final String librariesPath;
 
+  /// Any extra args to pass to ddc.
```
natebosch: [nit] Update the comment since it isn't arbitrary args anymore.

Issue #1625, opened 2018-06-29
Support clean JSON output of test result


eredo:
Previously it was possible to redirect the `pub run test -r json` output to a file and than parse the json file accordingly (ex. `pub run test -r json > test_results.json`). Now `pub run build_runner test -- -r json` cannot be redirected to a file because of the `build` output. A possible solution would be to provide an `--output` argument for `build_runner test` which pipes the `test` output to the file.

jakemac53:
Makes sense, we already have a `--output` flag but we could add a `--test-output` flag or similar.

Note that there is a workaround today, you can do `pub run build_runner build test -o <output-dir>`, and then `pub run test  -r json --precompiled <output-dir>`.

Some other options could be:

- Adding a flag to silence the build output entirely
- Adding a json reporter mode to the build process as well

matanlurey:
Long-er term I imagine you'll want a JSON reporter for other reasons (IDE integration, etc).

jakemac53:
We have some other ideas coming down the line which would be relevant as well when it comes to tool integration, so we might want to hold off on a full json reporter. Possibly something like this https://jvican.github.io/slides/BSP.pdf (or maybe just implement that with some language extensions).

evanweible-wf:
I took a stab at fixing this issue here: https://github.com/dart-lang/build/pull/2508

I'm not sure if there are still plans for a full json reporter as mentioned above, but this would at least make it possible to get a usable json test report with a single build_runner command.

jakemac53:
The test package supports --file-reporter now, which I think can be used to accomplish this

PR #2508, opened 2019-10-19
Add option to write test process stdout to file

evanweible-wf:
We've been using the workaround for #1625 (run build and tests separately) to collect the JSON output from the test runner, but it'd be nice for that to be supported in build_runner directly so that CI test configs are more straightforward.

This is an attempt at fixing #1625 by adding a `--test-stdout` option that, when set, results in the test command running the test process in normal mode and piping the stdout to the given file path. Stderr is still forwarded to the main build process so that non-test errors remain visible.

If this approach looks alright, I can follow up with a couple tests and changelog entry.

@jakemac53 @natebosch

### DIFF
--- a/build_runner/lib/src/entrypoint/test.dart
+++ b/build_runner/lib/src/entrypoint/test.dart
@@ -17,6 +17,8 @@ import '../generate/build.dart';
 import 'base_command.dart';
 import 'options.dart';
 
+const _testStdout = 'test-stdout';
+
 /// A command that does a single build and then runs tests using the compiled
 /// assets.
 class TestCommand extends BuildRunnerCommand {
@@ -25,7 +27,10 @@ class TestCommand extends BuildRunnerCommand {
           // Use symlinks by default, if package:test supports it.
           symlinksDefault:
               _packageTestSupportsSymlinks(packageGraph) && !Platform.isWindows,
-        );
+        ) {
+    argParser.addOption(_testStdout,
+        help: 'Write the test process stdout to this file path.');
+  }
 
   @override
   String get invocation =>
@@ -114,7 +119,11 @@ class TestCommand extends BuildRunnerCommand {
         return result.failureType.exitCode;
       }
 
-      return await _runTests(tempPath);
+      File testStdout;
+      if (argResults[_testStdout] != null) {
+        testStdout = File(argResults[_testStdout] as String);
+      }
+      return await _runTests(tempPath, testStdout: testStdout);
     } on _BuildTestDependencyError catch (e) {
       stdout.writeln(e);
       return ExitCode.config.code;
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/entrypoint/test.dart
```diff
@@ -17,6 +17,8 @@ import '../generate/build.dart';
 import 'base_command.dart';
 import 'options.dart';
 
+const _testStdout = 'test-stdout';
```
evanweible-wf: Open to better name suggestions here

build_runner/lib/src/entrypoint/test.dart
```diff
@@ -136,8 +146,30 @@ class TestCommand extends BuildRunnerCommand {
...
+        if (testStdout != null) {
+          var sink = testStdout.openWrite();
+          futures.add(sink
+              .addStream(testProcess.stdout)
+              .then((_) => sink.flush)
+              .then((_) => sink.close));
+        }
+        await Future.wait(futures);
+        stdout.writeln('Test results written to ${testStdout.path}');
+      } catch (e) {
```
natebosch: What types of exceptions might we catch here?

evanweible-wf:
@natebosch I like that approach better, too. Looking at the reporter implementations, I don't think it would be too difficult to add that in.

The `JsonReporter` funnels all of its output through an `_emit()` method that calls `print()`, so all of the logic in that reporter should be able to be shared with a file-based version that just writes to a file instead of printing.

The `CompactReporter` and `ExpandedReporter` don't centralize their printing anywhere (each have at least 7 calls to `print()`, but that should be easy enough to DRY up for the purposes of sharing with a file reporter. The `CompactReporter` also uses `stdout.write` since it rewrites lines, but those could be ignored when writing to a file – presumably if someone wants to write the compact output to file, all they care about is the final result in a single line.

The runner does have some special casing when in debug mode based on the reporter – it will only output the debug info to the terminal if it's not the json reporter so as to not break its contract, but that wouldn't apply to the file reporters.

Since the `Runner` constructor is private, I think we could update it to accept a file reporter as an optional param without it being a breaking change, and then any place where the reporter is interacted with would just need to be updated to also interact with the file reporter.

Alternatively, the runner and `debug()` fn could be updated to accept a list of reporters and then file-based reporters could be added to the existing reporter registry, but that seems a bit odd.. you would never enable multiple stdout reporters and I can't think of a scenario where multiple file reporters would be useful.

build_runner/lib/src/entrypoint/test.dart
evanweible-wf: Probably just `FileSystemException`s, potentially when creating the file. I ended up having to update line 160 locally to support writing to a directory that doesn't yet exist:

```
var sink = (testStdout..createSync(recursive: true)).openWrite();
```

natebosch:
I got started with https://github.com/dart-lang/test/pull/1101 to experiment.
As far as I can tell it should be safe to have multiple reporters and to allow a reporter to write to an arbitrary `Sink<List<int>>`

We'll need to figure out how to handle option parsing for this, but I think it's promising.

evanweible-wf:
I'm going to close this since it seems to make a lot more sense in pkg:test.

PR #2510, opened 2019-10-21
Invalidate cached modules by digest

natebosch:
Within the build that writes a new value we don't have the digest so at
the end of the build the value will be cleared away and must be read
from the file at least once.
As soon as a value is read from a file we also track the digest and only
invalidate the deserialized value when the digest has changed.

### DIFF
--- a/build_modules/lib/src/module_cache.dart
+++ b/build_modules/lib/src/module_cache.dart
@@ -6,6 +6,7 @@ import 'dart:convert';
 
 import 'package:async/async.dart';
 import 'package:build/build.dart';
+import 'package:crypto/crypto.dart';
 
 import 'meta_module.dart';
 import 'modules.dart';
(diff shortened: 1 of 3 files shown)


natebosch:
This is the replacement for #2344 which handles not trying to perform the same digest step multiple times using an `onGoingCheck` field.

build_modules/lib/src/module_cache.dart
```diff
@@ -53,9 +59,30 @@ class DecodingCache<T> {
...
+          entry.digest = Result.capture(reader.digest(id));
+          if (await Result.release(entry.digest) != previousDigest) {
+            entry.value =
+                Result.capture(reader.readAsBytes(id).then(_fromBytes));
+          }
+          entry
+            ..needsCheck = false
+            ..onGoingCheck = null;
+        }();
+        await entry.onGoingCheck;
```
jakemac53: [nit]: You could move this await up to the assignment

natebosch: Done

natebosch:
Added a test and changelog.

robbecker-wf:
@natebosch Can we get a build_modues release with this in it?

natebosch:
I'm planning on getting https://github.com/dart-lang/build/pull/2512 merged and then I'll start publishing. If I end up needing to use a tighter constraint on analyzer from this package I'll instead publish this first and then do the analyzer bump in another publish so that there isn't any trouble picking this version up.

PR #2518, opened 2019-10-28
Update golden for build entrypoint

natebosch:
The `built_value` package added a non-dev dependency on
`built_value_generator`, so we now pick up that package for any
transitive dep of `build_daemon`.


natebosch:
See https://github.com/google/built_value.dart/commit/ddf656d45886e3f1f9e45f6ae27c679c292054c4#commitcomment-35698906

PR #2524, opened 2019-10-30
s/hide Logger/show AnalysisOptions

MichaelRFairhurst:
This will unblock us from removing Logger as a part of the analysis engine namespace. (hiding a non-exist term is a build breaking error in g3)

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.2.3-dev
+
+Change a `hide` declaration to a `show` declaration to support a
+`package:analyzer` change.
+
 ## 1.2.2
 
 - Update to `package:analyzer` version `0.39.0`.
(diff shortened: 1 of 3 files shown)


jakemac53:
Do we need to publish this soon or will this be in the next analyzer release and we can wait until then?

MichaelRFairhurst:
Technically this is not a breaking change. It breaks google3 just like unused imports do. But it won't need to be published. Obviously better/safer to though!

jakemac53:
Ok - can you update the pubspec/changelog to `1.2.3-dev` (changelog can just have a very brief entry but something should exist for each pending or released version)

MichaelRFairhurst:
OK, done!

MichaelRFairhurst:
@jakemac do you prefer merge, squash, or rebase?

jakemac53:
We generally prefer squash & merge

jakemac53:
The presubmit error does seem related to this - https://travis-ci.org/dart-lang/build/jobs/605197109#L661

jakemac53:
(or possibly its a bug in the lint?)

MichaelRFairhurst:
I changed `[AnalysisContext]` to `\`AnalysisContext\``, but if you prefer, can `show AnalysisContext`.

jakemac53:
What you did is fine - its weird to add something to the scope just for a doc comment imo :D

PR #2533, opened 2019-11-05
Use extensions from package:stream_transform

natebosch:
- Bump the minimum SDK constraint on the packages that use
  `package:stream_transform` to 2.6.0 with support for extension
  methods.
- Migrate usages of `.transform(something())` to `.something()`.
- Migrate some `.where((e) => e is File)` to `.whereType<File>()`.


natebosch:
I added `-dev` to the version in the 2 pubspecs which didn't already have it. I'm not sure if this is worth a changelog entry...

jakemac53:
> I added `-dev` to the version in the 2 pubspecs which didn't already have it. I'm not sure if this is worth a changelog entry...

I prefer to have the pubspec and changelog always agree on the version - and its probably worth mentioning we now require the latest dart sdk? We could have it in there with no entries for now as well.

natebosch:
Added a changelog entry for each

PR #2534, opened 2019-11-06
Fix strict-inference hints in build_runner

natebosch:
Add return types on all methods, including `main`. Mostly the missing
types were `void`.

Add argument types on all arguments where they can't be inferred. This
is a clear improvement in a few places where we otherwise had dynamic
calls in the body. In some places where we are ignoring the argument
change the argument type to `void` to make it clear it won't be used.
Remove one unnecessary optional argument since it wasn't required to fit
the interface we were implementing.

Add a `void` return type to function types where we don't use the
return. In some places switch to the newer function type syntax since it
is more clear.


natebosch:
I'm not 100% on the `void` argument types, I'd also be fine making those `dynamic`.

For the most part I feel like this is an improvement. If we all agree I can clean up the other packages as well.

grouma:
I think I prefer `void` over `dynamic` so LGTM.

Issue #2543, opened 2019-11-06
support passing VM flags to subprocesses


sigmundch:
Users would like to be able to plumb VM flags to subprocesses. The main motivating example at this time is to specify a different `--old-gen-heap-size` to the VM when invoking dart2js to prevent OOM errors. See https://github.com/dart-lang/sdk/issues/27883 for an example

@jakemac53 - not sure if this is the right repo to track this?

jakemac53:
It should be done with an environment variable, so only `build` will have to know about it.

jakemac53:
How do we want to support passing multiple arguments here? separate by commas or...?

jakemac53:
Actually I will just go with space separated args

jakemac53:
planned name for the environment variable is `BUILD_DART2JS_VM_ARGS`

jakemac53:
This has been publish in build_modules 2.7.0. Multiple arguments can be provided by separating them with spaces.

robbecker-wf:
Thanks! Can confirm this does in fact get `--old-gen-heap-size` through to dart2js when running webdev.

PR #2544, opened 2019-11-13
add BUILD_DART2JS_VM_ARGS environment variable support

jakemac53:
Closes https://github.com/dart-lang/build/issues/2543.

I haven't actually tested it yet, nor do I have any idea of how to write a test for it :sob: .

### DIFF
--- a/build_modules/lib/src/workers.dart
+++ b/build_modules/lib/src/workers.dart
@@ -107,6 +107,13 @@ final frontendDriverResource =
   __frontendDriver = null;
 });
 
+const _dart2jsVmArgsEnvVar = 'BUILD_DART2JS_VM_ARGS';
+final _dart2jsVmArgs = () {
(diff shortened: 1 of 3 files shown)


build_modules/lib/src/workers.dart
```diff
@@ -107,13 +107,21 @@ final frontendDriverResource =
   __frontendDriver = null;
 });
 
+const _dart2jsVmArgsEnvVar = 'BUILD_DART2JS_VM_ARGS';
+final _dart2jsVmArgs = () {
+  var env = Platform.environment[_dart2jsVmArgsEnvVar];
+  if (env == null) return <String>[];
```
natebosch: [nit] `return env?.split(' ') ?? <String>[];`

jakemac53:
I was able to manually test this and it works - adding an actual test would require bumping the min constraint for build_modules in build_web_compilers which is probably not worth doing.

PR #2549, opened 2019-11-18
Read written assets in build_test

simolus3:
Now that the build runner supports reading assets that were written in the same build step, the behavior of `testBuilder` can diverge from a real build.

To fix this, I've added the `WrittenAssetReader`, which can read assets that were written to a `RecordingAssetWriter`. The `testBuilder` then uses that reader to match the behavior of `build_runner_core`.

### DIFF
--- a/build_test/lib/src/written_asset_reader.dart
+++ b/build_test/lib/src/written_asset_reader.dart
@@ -9,21 +9,36 @@ import 'package:glob/glob.dart';
 
 import 'in_memory_writer.dart';
 
-/// A [MultiPackageAssetReader] reading outputs written into a
-/// [RecordingAssetWriter].
+/// A [MultiPackageAssetReader] which supports reads from previous outputs.
 class WrittenAssetReader extends MultiPackageAssetReader {
   final RecordingAssetWriter source;
 
-  WrittenAssetReader(this.source);
+  /// An optional [AssetWriterSpy] to limit what's readable through this reader.
+  ///
+  /// Only assets reported as written trough this [AssetWriterSpy] can be read
+  /// from this reader. When null, all assets from [source] are available.
+  final AssetWriterSpy filterSpy;
+
+  WrittenAssetReader(this.source, [this.filterSpy]);
 
   @override
   Future<bool> canRead(AssetId id) {
-    return Future.value(source.assets.containsKey(id));
+    var canRead = source.assets.containsKey(id);
+    if (filterSpy != null) {
+      canRead = canRead && filterSpy.assetsWritten.contains(id);
+    }
+
+    return Future.value(canRead);
   }
 
   @override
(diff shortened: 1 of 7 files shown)


build_test/lib/src/written_asset_reader.dart
```diff
@@ -0,0 +1,50 @@
...
+  @override
+  Future<List<int>> readAsBytes(AssetId id) {
+    if (!source.assets.containsKey(id)) {
+      throw AssetNotFoundException(id);
+    }
+    return Future.value(source.assets[id]);
+  }
+
+  @override
+  Future<String> readAsString(AssetId id, {Encoding encoding}) {
```
jakemac53: Nit: rewrite this as:

```dart
Future<String> readAsString(AssetId id, {Encoding encoding}) async {
  encoding ??= utf8;
  return encoding.decode(await readAsBytes(id));
}
```

Bashing over the optional parameter in the null case is generally the safest way to provide a default for it - it prevents accidental usage of the possibly null original value.

And then we generally use async/await unless there is a strong reason not to :)

build_test/test/written_asset_reader_test.dart
```diff
@@ -0,0 +1,43 @@
...
+    await writer.writeAsString(AssetId.parse('foo|a.txt'), 'a');
+    await writer.writeAsString(AssetId.parse('foo|lib/b.dart'), 'b');
+    await writer.writeAsString(AssetId.parse('bar|a.txt'), 'b_a');
+
+    reader = WrittenAssetReader(writer);
+  });
+
+  test("doesn't read assets that weren't written", () async {
+    final doesntExist = AssetId.parse('foo|c.txt');
+    expect(reader.canRead(doesntExist), completion(isFalse));
```
jakemac53: nit: you can just use async/await here instead of relying on the async completers which are sort of weird (the rest of the test still executes before the expect is validated since there is no await).

`expect(await reader.canRead(doesntExist), false);`

Same for tests below (except the stream one)

jakemac53:
I just realized one issue with this - `testBuilder` can run the builder on multiple primary inputs. As it is written right now this will allow any build step to read any asset written by any other build step which it shouldn't do (and could cause the test to pass when a real build would fail).

This can likely be resolved by changing the logic here https://github.com/dart-lang/build/pull/2549/files#diff-c0865ae8abb605df886af14abe861af7R155 to run each primary input on its own instead of all at once, and using the AssetReaderSpy to expose the assets that were written for just that input.

jakemac53:
@simolus3 ready for merge?

simolus3:
@jakemac53 Yes :+1:

jakemac53:
Thanks for the followup contribution!

jakemac53:
tagged/released as build_test 0.10.10

PR #2560, opened 2019-12-04
Stop plumbing unused `verbose` through options

natebosch:
This used to have other meaning, however now it is only used to set a
default log level.

Fixes a failing travis since the analyzer recently started reporting
this as a hint.


build_runner_core/lib/src/generate/options.dart
```diff
@@ -39,7 +39,6 @@ class LogSubscription {
   factory LogSubscription(BuildEnvironment environment,
       {bool verbose, Level logLevel}) {
     // Set up logging
-    verbose ??= false;
```
jakemac53: this looks like it should still be here

natebosch: Oops! yes it should

PR #2578, opened 2019-12-16
Make resolver throw an exception on non-libraries

simolus3:
From the docs of [`Resolver.libraryFor`](https://pub.dev/documentation/build/latest/build/Resolver/libraryFor.html):
> Throws `NonLibraryAssetException` if assetId is not a Dart library.

Contrary to the docs, the most common resolver from `build_resolvers` actually returns null instead of throwing said exception.
I think the actual behavior should match the docs here, so I've changed the behavior of `AnalyzerResolver`. I think this behavior can be useful, but I'm not sure if some users might rely on that implementation now (e.g. doing `!= null` checks instead of wrapping it in a try block). Personally I'd also be fine with the alternative of changing the documentation, if that's what you prefer.

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,6 +1,8 @@
-## 1.3.1-dev
+## 1.3.1
 
-- Internal changes.
+- Add an exception on trying to resolve an `AssetId` that is not a Dart
+  library with `libraryFor` to fit the contract expressed by the doc comment on
+  `Resolver`.
 
 ## 1.3.0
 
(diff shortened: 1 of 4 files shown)


jakemac53:
This seems like the correct behavior to me.

cc @natebosch wdyt about releasing this without a breaking change?

natebosch:
This is either a regression or a difference from bazel behavior. We should add a changelog entry as a bug fix release.

natebosch:
@jakemac53 - I added a changelog. Can you take another look? Once it's got a second approval I can merge and publish.

natebosch:
Thanks @simolus3 for the fix!

PR #2588, opened 2020-01-06
Fix scratch space race condition with pending writes

jakemac53:
Fixes the potential root cause of https://github.com/dart-lang/sdk/issues/38102

If we failed to write one file we would previously eagerly return that error, but still have a bunch of other pending writes which are cached.

If we then scheduled another build while those writes were still pending they could end up with the cached result.

The fix is to not eagerly fail and make sure we wait for all pending writes before returning.


jakemac53:
Investigating https://github.com/dart-lang/sdk/issues/39921 is what made me realize this race condition but I don't think it this would fix that particular error.

PR #2597, opened 2020-01-14
Eagerly report writes to build step

natebosch:
Solves a potential bug if a builder does not `await` the future returned
by `copyOutput` then the build step may finish before the asset is read
from the temp directory and the `writeAsString` call may be rejected.

### DIFF
--- a/scratch_space/CHANGELOG.md
+++ b/scratch_space/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.1.0-dev
+
+- Change `copyOutput` to take a `BuildStep` instead of an `AssetWriter`. This
+  shouldn't break most uses which are likely already passing in a `BuildStep`.
+
 ## 0.0.4+2
 
 - Fix a race condition bug where `ensureAssets` would complete before all
--- a/scratch_space/lib/src/scratch_space.dart
+++ b/scratch_space/lib/src/scratch_space.dart
@@ -43,10 +43,7 @@ class ScratchSpace {
     return ScratchSpace._(tempDir);
   }
 
-  /// Copies [id] from the tmp dir and writes it back using the [writer].
-  ///
-  /// Note that [BuildStep] implements [AssetWriter] and that is typically
-  /// what you will want to pass in.
+  /// Copies [id] from the tmp dir and writes it back using the [buildStep].
   ///
   /// This must be called for all outputs which you want to be included as a
   /// part of the actual build (any other outputs will be deleted with the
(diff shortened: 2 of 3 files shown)


natebosch:
I'm not entirely convinced this is worth the breaking change to solve, AFAIK this isn't causing any problems today.

If we decide to pursue this change we'll need a regression test. Opening for discussion.

jakemac53:
I don't think its worth trying to support the unawaited future use case. In general unawaited futures in builders can cause us all kinds of problems.  If we have instances where we aren't awaiting this we should update those though.

natebosch:
I couldn't find anywhere that we weren't awaiting this, just noticed this as a possibility while reading the code.

Do you think we should apply this change on our next breaking release for safety?

jakemac53:
I am not convinced it would make things better. It would actually allow you to write assets after the build step has completed which I think is worse?

natebosch:
>  It would actually allow you to write assets after the build step has completed

The build step would be held open by the fact that the writes hadn't finished yet. This change can't introduce any new bugs as far as I can tell, it only means that it's not possible to write certain other bugs.

Buggy today but works with this change:

```dart
Future<void> build(BuildStep buildStep) async {
  // a bunch of work.
  for (var id in outputs) {
    scratchSpace.copyOutput(id, buildStep);
  }
}
```

The missing `await` for `copyOutput` means that some writes will end up failing.

jakemac53:
Oh right because we wait for pending writes.

I am still on the fence TBH because by making code like this "work" we could be making it more likely for it to be written in the first place. Having it give an error (it would today error all the time right?) in some sense keeps people on the rails a bit.

natebosch:
I'm also on the fence.

The thing that's pushing me towards wanting to do this anyway, is that we specifically made the same design decisions about `BuildStep.writeAsBytes` already. We intentionally made it so you are allowed to not await that return (we even wanted to hide the Future at one point), and we added the `FutureOr` argument type to make it even less likely for an author to shoot themselves in the foot around futures. It feels sensible to have `ScratchSpace.copyOutput` keep as much of the ergonomics of `BuildStep.writeAsBytes` as possible because they are both things that happen around the "end" of the build step and are closely related.

A builder isn't supposed to _do_ anything in reaction to the write being done, so the `await` _only_ accomplished correct usage of the build system.

> Having it give an error (it would today error all the time right?) in some sense keeps people on the rails a bit.

I think there is a small possibility of having an error get masked by subtle differences in async file read timing if some other work might be holding open the build step through at least a loop or 2 around the even queue, but yeah I'd expect that if an author is missing the `await` it should fail pretty much always.

jakemac53:
We don't really encourage not awaiting `writeAs*` any more - we have already changed the docs around this to actually await it I believe. If I recall correctly this was originally done for misguided performance reasons rather than trying to help out users.

natebosch:
Yeah it was mostly because you have to use things like `Future.wait` to avoid accidentally serializing writes and it feels awkward to _force_ the author to wait for things that they don't care about just to appease the build system that does want to wait.

Issue #2596, opened 2020-01-14
build_test fails with an out-of-call-stack exception with build_resolvers 1.3.1


kevmoo:
build_resolvers: 1.3.1
build_test: 0.10.11

See commit https://github.com/dart-lang/build/commit/b6fdcb0508c6786113ceda73ba17506dcb51718a

This is the failing test:

https://github.com/kevmoo/source_gen_test/blob/c1aff801cba9f46c27a1da854fbeed75fb7b2a1e/test/init_library_reader_test.dart#L54-L62

jakemac53:
https://github.com/dart-lang/build/blob/master/build_test/lib/src/resolve_source.dart#L197 😢

kevmoo:
> https://github.com/dart-lang/build/blob/master/build_test/lib/src/resolve_source.dart#L197 😢

I tried just awaiting that – still has the problem!

jakemac53:
Awaiting that fixes the issue but then the test has to be updated to expect the new exception type (which is the documented one, previously it threw a different one). I think it might be breaking for other use cases though to await there, I need to look into it a bit more.

cc @matanlurey

jakemac53:
Ya if we awaited there we would block on the `tearDown` future being resolved so this would just hang tests that use that.

I don't really know what the correct solution is going to be here.

natebosch:
Reopening for more discussion. This change can cause some legitimate failures to get lost.

When I drop the `.catchError` here I can't repro any failure with the `source_gen_test` tests. @kevmoo - do you expect that something changed in those tests?

I'm not sure how much retaining these specific errors would help though. If the async error comes after the builder completes normally we'd still ignore those errors.

https://github.com/dart-lang/build/blob/2d36ee33d138edd35542b3f9633f7032a54f19fc/build/lib/src/builder/logging.dart#L28

jakemac53:
@natebosch should we close this at this point?

natebosch:
I think #3253 and #3245 are examples of things that we can miss because of this design.

I found both of those examples while I was experimenting with migrating those tests to `package:checks`.

I think this is still a likely issue when using `expect`, but with `checks` we'd more likely get an `unawaited_future`, and the `await` resolves the issue.

I think that the complexity to solve this is not justified for continued safety with a library that has a foot-gun around async expectations. If we invest in async safety for `package:matcher` I'd first look at adding lints to use `expectLater` with `AsyncMatcher` which I think also would add enough safety for this problem.

PR #2599, opened 2020-01-14
Fix a bug in build_test where the resolve* apis would leak unhandled async errors

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2596

Specifically if the `action` callback throws any exceptions then we would leak those as unhandled async errors.

We can't await the call to `runBuilder` because that would block on the `tearDown` future which would hang tests.

The solution is to just swallow all errors from the unawaited future, and rely on the `onDone` Future to report most errors. We try/catch the action callback and forward the errors to the `onDone` completer.

### DIFF
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -228,10 +228,8 @@ class _ResolveSourceBuilder<T> implements Builder {
   @override
   Future<void> build(BuildStep buildStep) async {
     if (_resolverFor != buildStep.inputId) return;
-    T result;
     try {
-      result = await _action(buildStep.resolver);
-      onDone.complete(result);
+      onDone.complete(await _action(buildStep.resolver));
     } catch (e, s) {
       onDone.completeError(e, s);
     }
--- a/build_test/lib/src/resolve_source.dart
+++ b/build_test/lib/src/resolve_source.dart
@@ -193,14 +193,22 @@ Future<T> _resolveAssets<T>(
     sourceAssets: inputAssets,
     rootPackage: rootPackage,
   );
-  // We don't care about the results of this build.
-  unawaited(runBuilder(
-    resolveBuilder,
-    inputAssets.keys,
-    MultiAssetReader([inMemory, assetReader]),
-    InMemoryAssetWriter(),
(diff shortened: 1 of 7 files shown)


build_test/lib/src/resolve_source.dart
```diff
@@ -193,14 +193,22 @@ Future<T> _resolveAssets<T>(
...
-    resolveBuilder,
-    inputAssets.keys,
-    MultiAssetReader([inMemory, assetReader]),
-    InMemoryAssetWriter(),
-    resolvers ?? defaultResolvers,
-  ));
+  // We don't care about the results of this build, but we also can't await
+  // it because that would block on the `tearDown` of the `resolveBuilder`.
+  //
+  // We also dont want to leak unhandled async errors so we swallow them.
```
natebosch: I don't understand this part. If we swallow unhandled async errors I think we mask bugs in the test? Does it work if we do `unawaited(runBuilders(...).catchError((_) {}));`

jakemac53: Oh right - I think that would work and be better :). The builder itself doesn't have unawaited futures in its build method (well unless the action does).

jakemac53:
Looks like something else in the latest dev broke us - investigating.

build_test/lib/src/resolve_source.dart
```diff
@@ -220,8 +227,11 @@ class _ResolveSourceBuilder<T> implements Builder {
   @override
   Future<void> build(BuildStep buildStep) async {
     if (_resolverFor != buildStep.inputId) return;
-    var result = await _action(buildStep.resolver);
-    onDone.complete(result);
+    try {
+      onDone.complete(await _action(buildStep.resolver));
```
natebosch: [optional] If we aren't worried about `_action` throwing synchronously we can skip the `try/catch`. `Completer.complete` takes a `FutureOr<T>` and forwards async errors correctly.

```dart
onDone.complete(_action(buildStep.resolver));
```

jakemac53: We do need to catch synchronous failures as well

jakemac53:
bad dev release is causing the integration test failures - merging on red.

Issue #2528, opened 2019-10-31
`package:flutter` resolution in `built_value`


davidmorgan:
I think the latest release was intended to help with

https://github.com/google/built_value.dart/issues/513

--did I understand correctly?

Thanks!

jakemac53:
Correct, that should work now.

This is based on the `build_resolvers` version however not the `build` version, you will need version `1.2.1`.

jakemac53:
Re-opening this, after following up in the other issue I was able to reproduce with flutter version 1.9.1+hotfix6 and [this](https://gist.github.com/MrCreeper1008/40209abce5265e37d5cff5153988a3f7) pubspec.lock file which has the expected versions.

jakemac53:
OK, I have tracked down the root issue here. The problem is that `Platform.version` for Flutter 1.9.1 (and probably previous versions) looks just like a normal Dart SDK version (specifically, `2.5.0 (Fri Sep 6 20:10:36 2019 +0200) on "linux_x64"`).

We detect flutter by searching for the word `flutter` in this version string, which apparently was not introduced until flutter version `1.10.0`.

I don't know of a better way to detect whether we are running in flutter or not, @jonahwilliams are you aware of anything more robust on older flutter versions?

I am reluctant to add anything to hacky given that this does work on the latest stable flutter releases.

MrCreeper1008:
Not sure if I am adding anything here, but is it possible to detect Flutter with `where flutter` or something like that? Or am I missing anything?

jakemac53:
There isn't an obvious built in way to detect flutter other than the version string.

However, I spoke with some colleagues and we came up with a solution that will work (and be more robust than the current version string check going forward as well). We can use conditional imports to create our own boolean which is set based on the presence of `dart:ui` or not.

jakemac53:
It looks like we should also do a patch release for 1.2.1+1 since the newest build_resolvers requires the latest analyzer which isn't possible to get yet.

jakemac53:
Actually nevermind that fix doesn't work because the resolver is actually running in the Dart VM so `dart:ui` does not exist :sob: .

jakemac53:
@MrCreeper1008 is updating to flutter 1.10.0 or later an option for you?

MrCreeper1008:
Yes; should i upgrade?

jakemac53:
[...]

PR #2603, opened 2020-01-23
Add dart:ui directory existence check to help determine environment

shyndman:
Prior to this PR, the inclusion of `dart:ui` types were conditional on the string `"flutter"` appearing in the dart executable's version string. Unfortunately, several recent versions of Flutter `dart` omitted this string, causing type resolution failures in downstream packages. 

This PR resolves the situation by introducing a fallback check, which considers the existence of an SDK-relative `../pkg/sky_engine/lib` directory evidence of running via a Flutter-provided Dart binary.

Fixes https://github.com/dart-lang/build/issues/2528

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -250,8 +250,7 @@ Future<String> _defaultSdkSummaryGenerator() async {
     var watch = Stopwatch()..start();
     _logger.info('Generating SDK summary...');
     await summaryFile.create(recursive: true);
-    var sdkPath = p.dirname(p.dirname(Platform.resolvedExecutable));
-    await summaryFile.writeAsBytes(_buildSdkSummary(sdkPath));
+    await summaryFile.writeAsBytes(_buildSdkSummary());
 
     await _createDepsFile(depsFile, currentDeps);
     watch.stop();
@@ -284,14 +283,16 @@ Future<void> _createDepsFile(
   await depsFile.writeAsString(jsonEncode(currentDeps));
 }
 
-List<int> _buildSdkSummary(String dartSdkPath) {
+List<int> _buildSdkSummary() {
   var resourceProvider = PhysicalResourceProvider.INSTANCE;
-  var dartSdkFolder = resourceProvider.getFolder(dartSdkPath);
+  var dartSdkFolder = resourceProvider.getFolder(_runningDartSdkPath);
   var sdk = FolderBasedDartSdk(resourceProvider, dartSdkFolder)
     ..useSummary = false
     ..analysisOptions = AnalysisOptionsImpl();
 
-  if (_isFlutter) _addFlutterLibraries(sdk, resourceProvider, dartSdkPath);
+  if (isFlutter) {
+    _addFlutterLibraries(sdk, resourceProvider);
+  }
 
   var sdkSources = {
     for (var library in sdk.sdkLibraries) sdk.mapDartUri(library.shortName),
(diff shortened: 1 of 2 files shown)


shyndman:
Here you go Jake. Hope this works out.

Tests are passing on Flutter Dart and standalone.

build_resolvers/lib/src/resolver.dart
```diff
@@ -284,14 +283,16 @@ Future<void> _createDepsFile(
   await depsFile.writeAsString(jsonEncode(currentDeps));
 }
 
-List<int> _buildSdkSummary(String dartSdkPath) {
+Future<List<int>> _buildSdkSummary(String dartSdkPath) async {
```
jakemac53: Lets just just the new private `_runningDartSdkPath` variable since that is all that we pass in here anyways - then we cal also remove the argument from `isFlutterSdk` and the null handling there as well.

shyndman: sg. give me a few.

nit picking, but how do you feel about getting rid of the asynchrony on `isFlutterSdk`, if we're willing to use an `existsSync` in place of the `exists`?

For what it's worth, I've noticed there's a lint that discourages use of `exists`, but that might have baked in assumptions about where the code is running (and whether the files are truly local).

shyndman: Also, happy to keep the version string check. We can just `||` it.

jakemac53: I would be fine with using `existsSync` in `isFlutterSdk`. We should also convert that method to a final variable in that case, so we only do it once. Final variables are lazily evaluated so it will only happen on first access.

> Also, happy to keep the version string check. We can just || it.

Ya, lets go ahead and do this as it shouldn't hurt and might be more future proof :).

shyndman: Alright, done, and updated the commit/PR description. PTAL.

build_resolvers/lib/src/resolver.dart
```diff
@@ -304,23 +305,31 @@ List<int> _buildSdkSummary(String dartSdkPath) {
 /// [sdk].
 void _addFlutterLibraries(AbstractDartSdk sdk,
     ResourceProvider resourceProvider, String dartSdkPath) {
```
jakemac53: I think `dartSdkPath` is unused here now :)

build_resolvers/lib/src/resolver.dart
```diff
@@ -304,23 +305,31 @@ List<int> _buildSdkSummary(String dartSdkPath) {
...
   }
 }
 
-final _isFlutter = Platform.version.contains('flutter');
+/// Path to the running dart's SDK root.
+final _runningDartSdkPath = p.dirname(p.dirname(Platform.resolvedExecutable));
+
+/// Path where the dart:ui package will be found, if executing via the dart
+/// binary provided by the Flutter SDK.
+final _dartUiPath = p.canonicalize(
```
jakemac53: nit: use `p.normalize` instead of `p.canonicalize` - we don't need the canonical (absolute) path just a normalized one :).

build_resolvers/lib/src/resolver.dart
shyndman: Oops. Done.

build_resolvers/lib/src/resolver.dart
shyndman: Done.

shyndman:
> No need to force push for the final cleanup here - we will squash and merge

Missed this. Noted for next time. Habit from Gerrit.

jakemac53:
> Missed this. Noted for next time. Habit from Gerrit.

No worries, for github its just easier to review without the force pushes

Issue #2605, opened 2020-01-24
Add the ability to enable experiments for DDC builds


jakemac53:
We should add and `experiments` configuration option to the `build_web_compilers|ddc` builder, which will be a `List<String>`.

Each listed experiment will translate to an `--enable-experiment=<name>` command line argument for DDC and Kernel.

We also need to ensure this option is applied globally similar to the `environment` option.

cc @nshahan @leafpetersen

PR #2607, opened 2020-01-27
Allow enabling of experiments for ddc

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2605

cc @nshahan @leafpetersen


build_web_compilers/lib/builders.dart
```diff
@@ -96,12 +98,19 @@ Map<String, String> _readEnvironmentOption(BuilderOptions options) {
   return Map.from((options.config[_environmentOption] as Map) ?? {});
 }
 
+List<String> _readExperimentOption(BuilderOptions options) {
+  return List<String>.from(
```
natebosch: [nit] the `<String>` isn't necessary here since it gets inferred from the return type.

jakemac53: for some reason I was thinking this was an implicit cast and not type inference :P

PR #2609, opened 2020-01-29
Make patching easier for build_resolvers

jakemac53:


jakemac53:
nevermind this didn't help like I thought it would I forgot we do a lot more than just this (basically replace the entire sdk summary generator function)

Issue #2601, opened 2020-01-17
Config option for disabling all auto-applied builders for a target


evanweible-wf:
We have a few projects that are large enough to warrant some not-so-pretty `build.yaml` configuration in order to improve build times. This requires specifying some custom targets, which in turn requires disabling of some auto-applied builders in each target that would otherwise produce conflicting outputs. The config for this looks like so:

```yaml
targets:
  $default:
    sources:
      include:
        - lib/**
    dependencies:
      - :module_bootstrap
    builders:
      # disable all auto-applied builders except compiler builders (bwc and bvc)
      build_modules|module_library:
        enabled: false
      build_test|test_bootstrap:
        enabled: false
      # etc.

  module_bootstrap:
    sources:
      include:
        - example/**
        - lib/**
        - test/**
        - web/**
    builders:
      # disable all the compiler builders, but allow the rest of the auto-applied
      # builders to run in this target so that all dart codegen is done here.
      build_web_compilers|ddc_modules:
        enabled: false
      build_web_compilers|dart2js_modules:
        enabled: false
      build_web_compilers|ddc:
        enabled: false
      build_web_compilers|entrypoint:
        enabled: false
      # also disable build_vm_compilers builders here, too, if used

# The following targets are defined for the common top-level directories
# and must have all auto-applied builders disabled except the compiler builders

  example:
    sources:
      include:
        - example/**
    dependencies:
      - $default
    builders:
      build_modules|module_library:
        enabled: false
      build_test|test_bootstrap:
        enabled: false
      # etc.

      build_web_compilers|entrypoint:
        generate_for:
          include:
            - "example/**.dart"

  test:
    sources:
      include:
        - test/**
    dependencies:
      - $default
    builders:
      build_modules|module_library:
        enabled: false
      build_test|test_bootstrap:
        enabled: false
      # etc.

      build_web_compilers|entrypoint:
        generate_for:
          include:
            - "test/**.browser_test.dart"

  web:
    sources:
      include:
        - web/**
    dependencies:
      - $default
    builders:
      build_modules|module_library:
        enabled: false
      build_test|test_bootstrap:
        enabled: false
      # etc.

      build_web_compilers|entrypoint:
        generate_for:
          include:
            - "web/**.dart"
```

There are two main disadvantages of this approach:

- It's verbose
- It will break as soon as a new builder is added (either directly or transitively through a dependency)

---

[...]

PR #2612, opened 2020-02-03
Add an option to disable builder auto applications on a target

jakemac53:
Adds the `auto_apply_builders` option to `target` configuration. Disables normal automatic builder application for all builders for that target when set to `false`.

Any builder which is explicitly configured is also implicitly enabled. 

Closes https://github.com/dart-lang/build/issues/2601.

### DIFF
--- a/build_runner_core/lib/src/package_graph/apply_builders.dart
+++ b/build_runner_core/lib/src/package_graph/apply_builders.dart
@@ -366,10 +366,13 @@ bool _shouldApply(
   if (builderConfig?.isEnabled != null) {
     return builderConfig.isEnabled;
   }
-  return builderApplication.filter(node.package) ||
-      (applyWith[builderApplication.builderKey] ?? const []).any(
-          (anchorBuilder) =>
-              _shouldApply(anchorBuilder, node, applyWith, allBuilders));
+  if (node.target.autoApplyBuilders) {
+    return builderApplication.filter(node.package) ||
+        (applyWith[builderApplication.builderKey] ?? const []).any(
+            (anchorBuilder) =>
+                _shouldApply(anchorBuilder, node, applyWith, allBuilders));
+  }
+  return false;
 }
 
 /// Inverts the dependency map from 'applies builders' to 'applied with
--- a/build_runner_core/lib/src/package_graph/apply_builders.dart
+++ b/build_runner_core/lib/src/package_graph/apply_builders.dart
@@ -366,13 +366,12 @@ bool _shouldApply(
   if (builderConfig?.isEnabled != null) {
     return builderConfig.isEnabled;
   }
-  if (node.target.autoApplyBuilders) {
-    return builderApplication.filter(node.package) ||
-        (applyWith[builderApplication.builderKey] ?? const []).any(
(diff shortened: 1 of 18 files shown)


build_runner_core/lib/src/package_graph/apply_builders.dart
```diff
@@ -366,10 +366,13 @@ bool _shouldApply(
   if (builderConfig?.isEnabled != null) {
     return builderConfig.isEnabled;
   }
-  return builderApplication.filter(node.package) ||
-      (applyWith[builderApplication.builderKey] ?? const []).any(
-          (anchorBuilder) =>
-              _shouldApply(anchorBuilder, node, applyWith, allBuilders));
+  if (node.target.autoApplyBuilders) {
```
jakemac53: I actually wonder if the logic should be a bit different here.

If a builder is explicitly enabled and applies other builders, should those also be enabled?

jakemac53: cc @evanweible-wf what would you expect here?

For instance a source_gen `SharedPartBuilder` needs to apply the `source_gen:combining_builder`.

However enabling that automatically could cause issues if you have multiple source_gen builders you want to enable  on different targets... (each target would try to enable the same builder and then you would get duplicate outputs and have to explicitly disable it again...)

evanweible-wf: It would be nice to only have to to enable/configure the known builders and allow them to apply their own transitive builders.. just to make sure I'm understanding this, the problem you mentioned would only arise if we were splitting source gen builders across targets like so:

```
    module_bootstrap_json:
      builders:
        # disable all builders except json_serializable
    module_bootstrap_built:
      builders:
        # disable all builders except built_value_generator

    # etc..
```

Is that right?

jakemac53: Correct, for the source_gen example it would only matter if splitting builders across targets.

It might be reasonable though to say that you just shouldn't do that as it explicitly means you are duplicating some sources across targets which we don't really want people to do (but we don't enforce).

natebosch: +1 We should still enable the builders that get applied for reasons other than their `auto_apply:` config.

jakemac53: Seems like we are on the same page here I will update the logic.

I think making this only affect builders that are applied via `auto_apply` is a good way to think about it.

jakemac53:
Updated and added a test around `applies_builders`, PTAL

jakemac53:
> Should we consider making this an enum so that we can expand behavior in the future? We could start with `allow_any` (corresponds to `true` in this implementation) and `none` (corresponds to `false` in this implementation) and then we have the option of expanding to, for instance, allow auto apply that happened based on dependencies, but disallow `all_packages` or something? Not sure if that ends up getting too complex...

I think it is better to just keep it simple - how something choose to apply itself is supposed to be an implementation detail and not something a user relies on as a public part of its api.

evanweible-wf:
Thanks for adding this!!

PR #2617, opened 2020-02-05
add a skipped test that can be used to validate fixes to Issue #2389

jakemac53:
This should help with trying to fix https://github.com/dart-lang/build/issues/2389 as it is a pretty simple test case to validate and doesn't use the build system at all.


jakemac53:
Closing in favor of https://github.com/dart-lang/build/pull/2618 which has a similar test

PR #2618, opened 2020-02-05
Fix resolver behavior when encountering non-existing asset ids

### LINKED ISSUE #2389: Generated Dart code is not visible to other builders if library is resolved twice

simolus3:
I follow the advice from https://github.com/dart-lang/build/issues/2389#issuecomment-582595934. Basically, we're not supposed to return a null source if the uri could refer to a valid asset id. However, we used to return null if we haven't seen the asset id yet (e.g. it doesn't exist). Now, we just return the source pointing to the file where a non-existing asset id would be. If it gets created in a later build step, the analyzer will recognize this and handle it correctly.

With manual testing, I can verify that this indeed fixes #2389. Also contains, and thus closes #2392.


jakemac53:
woooo! :tada:

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -130,16 +143,18 @@ class BuildAssetUriResolver extends UriResolver {
 
   @override
   Source resolveAbsolute(Uri uri, [Uri actualUri]) {
-    final cachedId = lookupAsset(uri);
-    if (cachedId == null) return null;
-    return resourceProvider
-        .getFile(assetPath(cachedId))
-        .createSource(cachedId.uri);
+    final assetId = parseAsset(uri);
+    if (assetId == null) {
```
jakemac53: nit: for this one I would just do a one liner `if (assetId == null) return null;`

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -130,16 +143,18 @@ class BuildAssetUriResolver extends UriResolver {
...
+      return null;
+    }
+    final source =
```
jakemac53: nit: just return right away we don't need the local variable

build_resolvers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_resolvers
-version: 1.3.2
+version: 1.3.3-dev
```
jakemac53: Go ahead and remove the `-dev` - I can publish this after the merge

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -130,16 +143,18 @@ class BuildAssetUriResolver extends UriResolver {
...
+        resourceProvider.getFile(assetPath(assetId)).createSource(assetId.uri);
+    return source;
   }
 
   @override
   Uri restoreAbsolute(Source source) =>
-      lookupAsset(source.uri)?.uri ?? source.uri;
+      lookupCachedAsset(source.uri)?.uri ?? source.uri;
```
simolus3: I'm not sure if we should enforce the asset being cached here, or if we can just remove `lookupCachedAsset` entirely. The documentation from `UriResolver` says that `restoreAbsolute` "should be based solely on [source.fullName].", to me that sounds like further checks unnecessary. What do you think, @jakemac53?

jakemac53: Tbh I would just go for the minimal change (what we have here). This code has always been very fragile and I am not sure what it might break.

@scheglov would know the most though.

jakemac53: I am going to go ahead and merge/publish as this isn't changing the behavior. If we decide to change it further we can always do a followup :).

jakemac53:
published!

PR #2620, opened 2020-02-07
add isLibrary test for generated part files

jakemac53:
Adds a test that validates this issue https://github.com/dart-lang/source_gen/issues/447 - skipped until we get a resolution.

### DIFF
--- a/build_resolvers/test/resolver_test.dart
+++ b/build_resolvers/test/resolver_test.dart
@@ -2,6 +2,8 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:convert';
+
 import 'package:build/build.dart';
 import 'package:build_test/build_test.dart';
 import 'package:test/test.dart';
@@ -138,6 +140,7 @@ void main() {
               class A implements B {}
               ''',
         'a|web/main.g.dart': '''
+              part of 'main.dart';
               class B {}
               ''',
       }, (resolver) async {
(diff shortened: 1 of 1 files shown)


simolus3:
I believe this not working is an analyzer bug, I've opened https://github.com/dart-lang/sdk/issues/40546 to clarify.

@jakemac53 As a workaround, we could expose the `AnalysisDriverScheduler` to `BuildAssetUriResolver`. Then, we need to add this at the end of `performResolve`:
```dart
changedPaths.forEach(driver.changeFile);
// added:
if (changedPaths.isNotEmpty) {
  await scheduler.waitForIdle();
}
```

jakemac53:
Interesting - so that implies `changeFile` is doing something async but doesn't return a `Future` and we end up with a race condition here?

Lets give @scheglov a chance to look - it might be that your proposed solution is reasonable even long term if that is the only mechanism to wait for everything to complete.

scheglov:
Yes, `AnalysisDriver.changeFile()` does not do any work itself, it puts the file path into the queue of changed files, and schedules analysis work to do. Actual work is done later, asynchronously. If the are files for which the clients subscribed using `AnalysisDriver.addFile()`, and for which analysis results are reported via the `results` stream, affected files will be re-analyzed.

Is this related to https://github.com/dart-lang/sdk/issues/40546?

jakemac53:
Yes - this is related to dart-lang/sdk#40546

jakemac53:
> Actual work is done later, asynchronously.

Should this happen as a part of the `driver.getSourceKind` call itself (is it intended to recognize outdated results and wait for the new results) - or should we use `waitForIdle` or something else to wait for the results to propagate?

scheglov:
> > Actual work is done later, asynchronously.
> 
> Should this happen as a part of the `driver.getSourceKind` call itself (is it intended to recognize outdated results and wait for the new results) - or should we use `waitForIdle` or something else to wait for the results to propagate?

Yes, every `AnalysisDriver` API should return results that are consistent with previously reported changes via `changeFile`. The fact that `driver.getSourceKind` does not, is a bug, and I'm fixing it.

PR #2636, opened 2020-03-02
Support the package_config.json file

jakemac53:
Updates create_merged_dir and scratch_space to copy a modified version of the package_config.json file into their directories.

This file is modified by changing the `rootUri` and the `packageUri`. The `packageUri` is always the empty string, and the `rootUri` is either `<multiRootScheme>:///packages/<package-name>` (for the scratch_space) or `packages/<package-name>` (for merged output directories).

Note that for the shared scratch_space resource used by `build_modules` and the `build_*_compilers` packages, we now create a single package_config.json file and use that instead of a unique one per action.

## Invalidation

The package_config.json file is added as an internal source so it always exists as a part of the build.

Because a package_config.json change can arbitrarily change the build by changing the language version some package should use, we throw a `BuildScriptChanged` exception and delete the entire previous build whenever that file changes.

### DIFF
--- a/build_runner/test/generate/watch_test.dart
+++ b/build_runner/test/generate/watch_test.dart
@@ -379,7 +379,7 @@ a:file://different/fake/pkg/path
 ''');
 
         expect(await results.hasNext, isFalse);
-        expect(logs.length, 1);
+        expect(logs, hasLength(1));
         expect(
             logs.first.message,
             contains('Terminating builds due to package graph update, '
@@ -405,7 +405,7 @@ a:file://different/fake/pkg/path
             jsonEncode(newConfig));
 
         expect(await results.hasNext, isFalse);
-        expect(logs.length, 1);
+        expect(logs, hasLength(1));
         expect(
             logs.first.message,
             contains('Terminating builds due to package graph update, '
@@ -439,7 +439,7 @@ a:file://different/fake/pkg/path
             var next = await results.next;
             expect(next.status, BuildStatus.failure);
             expect(next.failureType, FailureType.buildConfigChanged);
-            expect(logs.length, 1);
+            expect(logs, hasLength(1));
             expect(logs.first.message,
                 contains('Terminating builds due to a:build.yaml update'));
           });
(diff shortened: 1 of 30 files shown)


build_runner/test/generate/watch_test.dart
```diff
@@ -383,6 +386,32 @@ a:file://different/fake/pkg/path
...
+        checkBuild(result, outputs: {'a|web/a.txt.copy': 'a'}, writer: writer);
+
+        var newConfig = Map.of(_packageConfig);
+        newConfig['extra'] = 'stuff';
+        await writer.writeAsString(
+            AssetId('a', '.dart_tool/package_config.json'),
+            jsonEncode(newConfig));
+
+        expect(await results.hasNext, isFalse);
+        expect(logs.length, 1);
```
natebosch:
```suggestion
        expect(logs, hasLength(1));
```

build_runner/test/generate/watch_test.dart
```diff
@@ -799,3 +828,10 @@ Future terminateWatch() async {
   await _terminateWatchController.close();
   _terminateWatchController = null;
 }
+
+final _packageConfig = {
```
natebosch: [nit] 
```suggestion
const _packageConfig = {
```

build_runner_core/lib/src/environment/create_merged_dir.dart
```diff
@@ -112,15 +113,12 @@ Future<bool> _createMergedOutputDir(
...
-          .join('\r\n');
-      var packagesAsset = AssetId(packageGraph.root.name, '.packages');
-      await _writeAsString(outputDir, packagesAsset, packagesFileContent);
-      outputAssets.add(packagesAsset);
+      outputAssets
+        ..addAll(await Future.wait(builtAssets.map((id) => _writeAsset(
+            id, outputDir, root, packageGraph, reader, symlinkOnly, hoist))))
+        ..add(await _writeCustomPackagesFile(packageGraph, outputDir))
+        ..add(await _writeModifiedPackageConfig(
+            packageGraph.root.name, reader, outputDir));
```
natebosch: [optional] Should we go for more in the `Future.wait`?

```suggestion
      outputAssets.addAll(await Future.wait([
          for(var id in builtAssets)
             _writeAsset(id, outputDir, root, packageGraph, reader, symlinkOnly, hoist),
           _writeCustomPackagesFile(packageGraph, outputDir),
           _writeModifiedPackageConfig(packageGraph.root.name, reader, outputDir),
       ]);
```

PR #2637, opened 2020-03-02
Avoid treating AssetId paths as URIs

### LINKED ISSUE #2635: Watcher error for emacs temp file
Seems to be similar to this issue: https://github.com/dart-lang/build/issues/1134.

If I use `pub run build_runner watch`, and edit a file with Emacs, 

```
[SEVERE] built_value_generator:built_value on lib/src/.#types.dart:

NoSuchMethodError: The getter 'element' was called on null.
Receiver: null
Tried calling: element
[SEVERE] Unhandled build failure!
FileSystemException: Cannot create file, path = '.dart_tool/build/098d3ee73e6cc294616d3a2e2c3c81ad/error_cache/cone_lib/0/lib/src/.' (OS Error: Is a directory, errno = 21)
[SEVERE] Failed after 7ms
```

Environment information:
```
$ dart --version
Dart VM version: 2.7.1 (Thu Jan 23 13:02:26 2020 +0100) on "macos_x64"
```

```yaml
# pubspec.lock
...
  build_runner:
    dependency: "direct dev"
    description:
      name: build_runner
      url: "https://pub.dartlang.org"
    source: hosted
    version: "1.7.4"
...
  built_value:
    dependency: "direct main"
    description:
      name: built_value
      url: "https://pub.dartlang.org"
    source: hosted
    version: "7.0.9"
  built_value_generator:
    dependency: "direct dev"
    description:
      name: built_value_generator
      url: "https://pub.dartlang.org"
    source: hosted
    version: "7.0.9"
...
```

natebosch:
Fixes #2635

We normalize to URI path separators, and have considered them to be
valid URIs in some places. This isn't safe since do don't do any
encoding so special characters like `#` are not escaped. We can't do
actual URI encoding for AssetIds since `$lib$` is used unencoded in too
many places. We will consider AssetIds to be, as much as possible posix
style instead of URI style. Remove some cases of treating them as URIs.

- Use the current OS style for `errorCacheDirectory` since it is only
  ever used for OS file operations and does not get stored anywhere.
- Use `p.posix.split` over `p.fromUri` when normalizing separators in
  asset ID paths just before an OS file operation.

### DIFF
--- a/build_runner_core/lib/src/util/constants.dart
+++ b/build_runner_core/lib/src/util/constants.dart
@@ -18,7 +18,7 @@ String assetGraphPathFor(String path) =>
 /// Relative path to the directory which holds serialized versions of errors
 /// reported during previous builds.
 final errorCachePath =
-    p.url.join(cacheDir, _scriptHashFor(_scriptPath), 'error_cache');
+    p.join(cacheDir, _scriptHashFor(_scriptPath), 'error_cache');
 
 final String _scriptPath = Platform.script.scheme == 'file'
     ? p.url.joinAll(
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.4.1-dev
+
+- Avoid treating `AssetId` paths as URIs.
+
 ## 4.4.0
 
 - Support the `auto_apply_builders` target configuration added in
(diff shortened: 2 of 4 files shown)


build_runner_core/lib/src/util/constants.dart
```diff
@@ -18,7 +18,7 @@ String assetGraphPathFor(String path) =>
 /// Relative path to the directory which holds serialized versions of errors
 /// reported during previous builds.
 final errorCachePath =
-    p.url.join(cacheDir, _scriptHashFor(_scriptPath), 'error_cache');
+    p.join(cacheDir, _scriptHashFor(_scriptPath), 'error_cache');
```
jakemac53: `p.posix`?

natebosch: This one can be the host OS style since it's only going to flow through to another OS style `p.join`. The only existing usage was round tripping between `p.url.join` then `p.fromUrl`, so I'm keeping it OS style the whole way instead.

PR #2639, opened 2020-03-04
Release build_runner_core and build_runner

jakemac53:
Going to do these two first, then build_modules and build_web_compilers in a followup.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.8.0-dev
+## 1.8.0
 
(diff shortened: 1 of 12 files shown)


jakemac53:
Somehow the latest build_runner_core is breaking the multiple client test for build_daemon here. No idea how yet though :sob: .

jakemac53:
The issue was we were getting false update notifications for files like `.packages`, `pubspec.lock`, and of course `.dart_tool/package_config.json`.

I updated the watcher to actually check if the digest has changed before accepting a change at face value.

PR #2647, opened 2020-03-10
Fix new error in latest SDK

kevmoo:


build_runner_core/lib/src/logging/failure_reporter.dart
```diff
@@ -85,7 +85,7 @@ class FailureReporter {
         final errorReports = jsonDecode(await errorFile.readAsString());
         final actionDescription = '${(errorReports as List).first} (cached)';
         final logger = Logger(actionDescription);
-        for (final List error in errorReports.skip(1)) {
+        for (final error in errorReports.skip(1)) {
```
jakemac53: We would ideally get the `List` type for error here, otherwise its just falling back on dynamic. I think we should likely file an SDK issue.

kevmoo: https://github.com/dart-lang/sdk/issues/40962

PR #2654, opened 2020-03-25
migrate build_test to package_config

jakemac53:
Note that we were actually using package_resolver under `lib`, and were actually missing the dep before.


kevmoo:
@jakemac53 – I took the liberty to do the follow-up with the SDK bump. Please take a look!

Issue #2640, opened 2020-03-04
Stop using package_resolver


kevmoo:
https://github.com/dart-lang/build/search?l=YAML&q=package_resolver&type=Code

- [x] _test_common
- [x] build_daemon
- [x] build_resolvers
- [x] build_test
- [x] build_runner

jakemac53:
done!

PR #2657, opened 2020-03-26
replace package_resolver use in _test_common with package_config

jakemac53:
Part of https://github.com/dart-lang/build/issues/2640


jakemac53:
I can't figure out how this is causing the build_vm_compilers failure... will keep looking into it

Issue #2645, opened 2020-03-06
Sharing .dart_tool in CI


todbachman-wf:
We currently share the `.dart_tool` directory between CI systems in order to speed up overall CI times as well as reduce the resource load on those systems. Sharing the `.dart_tool` directory is easier for us to do in a generic way than compiling to one or more build directories and sharing them. Sharing the `.dart_tool` directory has also typically sped up `pub get`s for us, although to be honest I'm not sure how much we benefit from that anymore.

This is a sketch of how we've been doing this:
- CI Build System:
  - pub get
  - pub run build_runner build
  - pub run build_runner build --release
  - tar czf dart_tool.tgz --exclude='*.snapshot' .dart_tool pubspec.lock
- CI Test System (executes the runs below in parallel):
  - DDC unit test run:
    - tar xzf dart_tool.tgz
    - pub get
    - pub run build_runner test
  - Dart2Js unit test run:
    - tar xzf dart_tool.tgz
    - pub get
    - pub run build_runner test --release

By reusing the `.dart_tool` directory from our Build system in our unit test runs in our Test system, those unit test runs don't have to recompile the code again. For us the compilation step of running unit tests can be fairly resource intensive and so avoiding that means we can reduce the memory and CPU requirements for our test system instances. Since those configs don't compile the code they also run faster and free up test instances sooner and mean we don't have to have as many test instances available in our pool.

Unfortunately for us, the `4.5.0` release of `build_runner_core` resulted in this approach breaking down. Specifically, if the `package_config.json` file changes it invalidates the builds. That file contains absolute paths which are not the same in our Build and Test systems, so we wouldn't be able to reuse that file between those systems. I believe this is causing all of our builds to now be invalidated and we're doing full builds in our unit test runs.

So, I'm wondering if it might be possible for you all to provide a way to more easily share the `.dart_tool` directory across systems. Or perhaps you can suggest a work around for us for the specific issue we've now hit with the `package_config.json` invalidating our builds. Or do you have a better approach you would suggest that might be equally easy to implement in a generic way?

#### Environment:
* Dart SDK Version: 2.7.1
* What package(s) from this repo you are using:
  - build_runner: 1.8.0
  - build_runner_core: 4.5.1

jakemac53:
Hmmm, ya the absolute paths in there are unfortunate. I have been seeing similar issues on our travis runs as well.

[...]

PR #2659, opened 2020-03-30
Depend on package language versions instead of entire package_graph.json

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2645

This adds a map of package name to language version to the asset graph, and checks that during deserialization. If it changes we throw away the graph and restart.

Note that this is a breaking change in build_runner_core, as some factory methods are now static methods that return a Future, and `PackageNode` now also has a language version.

This also reverts some of the previous logic and moves the check from build_script_updates.dart to build_definition.dart.


jakemac53:
A few windows things to work through, ill do that tomorrow

PR #2661, opened 2020-03-31
release build_runner

jakemac53:


build_runner/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## 1.8.1-dev
+## 1.8.1
 
 - Update to `build_runner_core` version `^5.0.0`.
```
natebosch: Is it worth copying anything from that changelog here since this is the one more users see?

jakemac53: I don't think so, its mostly implementation detail and a bug fix.

Issue #2531, opened 2019-11-04
build_web_compilers fails without lib/ in sources


lexaknyazev:
**pubspec.yaml**
```yaml
dev_dependencies:
    build_runner: '^1.7.2'
    build_web_compilers: '^2.7.1'
```

**build.yaml**
```yaml
targets:
  $default:
    sources:
      - "web/**"
```

```
> dart --version
Dart VM version: 2.6.0-dev.8.2 (Thu Oct 24 09:35:58 2019 +0200) on "windows_x64"

> pub run build_runner build
[INFO] Generating build script completed, took 290ms
[INFO] Creating build script snapshot... completed, took 11.3s
[INFO] Building new asset graph completed, took 1.0s
[INFO] Checking for unexpected pre-existing outputs. completed, took 1ms
[SEVERE] build_web_compilers:ddc_modules on web/main.dart:

NoSuchMethodError: The getter 'modules' was called on null.
Receiver: null
Tried calling: modules
[SEVERE] build_web_compilers:entrypoint on web/main.dart:

AssetNotFoundException: demo|web/main.ddc.module
[INFO] Running build completed, took 572ms
[INFO] Caching finalized dependency graph completed, took 122ms
[SEVERE] Failed after 705ms
```

Adding `"lib/**"` to `sources` in `build.yaml` fixes the build, although the folder does not exist.

jakemac53:
This is because of the synthetic `lib/$lib$` source file. The module builders require this to be available.

lexaknyazev:
Could `build_config` auto-insert it at runtime?

jakemac53:
We could hack it so those synthetic nodes always exists but it would mean users wouldn't be able to exclude those sources and it would cause worse problems. For instance you would essentially never be able to create more than one target without having duplicate sources between them which will cause overlapping output errors.

lexaknyazev:
Thanks for the clarification. I guess the best short-term solution here is to improve the error message then.

kuhnroyal:
@jakemac53 After updating to build_runner 1.9.0 I get the warning that the `$package$` source is missing. How do I fix this? I added `$package$` to the sources which silences the warning but I doubt this is the correct way.

jakemac53:
That is actually the correct way - its just a bit weird. This makes it be treated as an input for your package which allows builders to run using it. That is just a special file, what we call a "placeholder", which builders that run on an entire package can use as their primary input.

kuhnroyal:
[...]

PR #2664, opened 2020-04-01
check for placeholders when building a target graph from a package graph

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2531

Logs a warning if any required placeholders are not included in any target for a package.

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,4 +1,10 @@
+## 5.1.0-dev
+
+- Add a warning if a package is missing some required placholder files,
+  including `$package$` and `lib/$lib$`. 
+
 ## 5.0.0
+
 ### Breaking changes
 
 - `PackageGraph.forPath` and `PackageGraph.forThisPackage` are now static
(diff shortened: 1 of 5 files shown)


jakemac53:
I don't really like the idea of adding any magic here personally, I would prefer to just log the warning and let people fix it on their own. Otherwise they won't fix it, and then later on when they introduce some ambiguity (with another target for instance) they will run into problems.

natebosch:
Can this surface in a dependency that has an incorrect `build.yaml`? Maybe they defined a target in order to get some codegen and never tried with `build_web_compilers` or `build_vm_compilers`. If a user sees this warning in a dependency, would they need to add a `<package>.build.yaml` to get around it?

jakemac53:
It does warn about any build.yaml it finds, yes. 

That is why I only made this a warning - any build which previously worked will still work, but you may get a warning if some dependency is not set up properly.

They could also add a `<package>.build.yaml` to hide the warning.

jakemac53:
Note also that this isn't specific at all to build_vm_compilers or build_web_compilers - its a general requirement now.

PR #2666, opened 2020-04-06
Pass featureSet to SummaryBuilder.build

jakemac53:
I am unsure what I should actually be passing here.

What we really want is the feature set that is supported for the current running SDK version I think? But there is no api for that. As far as I can tell analyzer hardcodes the default feature set but its SDK constraint is not sufficient to make that valid.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,8 @@ jobs:
       env: PKGS="build build_config build_daemon build_modules build_resolvers build_runner build_runner_core build_test build_vm_compilers build_web_compilers example scratch_space"
       script: ./tool/travis.sh dartfmt dartanalyzer_0
     - stage: analyze_and_format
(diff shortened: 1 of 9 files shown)


kevmoo:
QQ: does this change the lower bound of the SDK we use?

scheglov:
You can get the set of features supported by a language version by creating `FeatureSet.fromEnableFlags([])` and then `restrictToVersion(languageVersion)`. Determining the running version of SDK and mapping it to the language version is up to the client.

jakemac53:
It looks like we also need to bump our min analyzer version, any idea what version this was added in  @scheglov ?

jakemac53:
also PTAL at the updated code

scheglov:
The feature landed in https://dart-review.googlesource.com/c/sdk/+/132800, in 2020-01-22.
The [latest version](https://pub.dev/packages/analyzer/versions) of analyzer (`0.39.5`) should include it.

jakemac53:
@kevmoo ya it looks like it does since we need the latest analyzer, updated everything to >=2.6.0

Issue #2658, opened 2020-03-27
Support language versioning for builders


jakemac53:
- [x] pass the correct `Packages` to for the analysis driver
- [x] expose language version on LibraryElement
- [ ] write docs for builder authors with guidance

jakemac53:
cc @scheglov do you know when the next analyzer version will be released with the `packages` argument to AnalysisDriver? I see it in the sdk but it hasn't been published yet.

scheglov:
We plan publishing a new version of analyzer soon.
https://github.com/dart-lang/sdk/issues/41222

https://dart-review.googlesource.com/c/sdk/+/141464 will include expose language version from `LibraryElement`.

jakemac53:
PR pending here https://github.com/dart-lang/build/pull/2669. This contains everything but the docs.

PR #2669, opened 2020-04-15
Support getting the language version for libraries

jakemac53:
Part of https://github.com/dart-lang/build/issues/2658

Adds a required `PackageConfig` argument to the internal `analysisDriver` method and convert that into the analyzers `Packages` object.

- Note that we have to use our internal custom paths for the rootDir and packageDir when we do the conversion, since those are the paths the analyzer sees.

Also supports passing a custom `PackageConfig` to `AnalyzerResolvers` (defaults to the current isolates package config).

### DIFF
--- a/build_resolvers/lib/src/analysis_driver.dart
+++ b/build_resolvers/lib/src/analysis_driver.dart
@@ -6,6 +6,7 @@ import 'package:analyzer/src/dart/analysis/byte_store.dart'
     show MemoryByteStore;
 import 'package:analyzer/src/dart/analysis/driver.dart';
 import 'package:analyzer/src/dart/analysis/file_state.dart';
+import 'package:analyzer/src/context/packages.dart' show Packages, Package;
 import 'package:analyzer/src/dart/analysis/performance_logger.dart'
     show PerformanceLog;
 import 'package:analyzer/src/generated/engine.dart'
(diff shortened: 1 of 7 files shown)


build_resolvers/lib/src/analysis_driver.dart
```diff
@@ -33,17 +42,32 @@ AnalysisDriver analysisDriver(BuildAssetUriResolver buildAssetUriResolver,
 
   var logger = PerformanceLog(null);
   var scheduler = AnalysisDriverScheduler(logger);
+  var packages = Packages({
```
natebosch, line 45: [optional] This is taking up a really big chunk of the method - consider pulling it out to a utility method

build_resolvers/lib/src/resolver.dart
```diff
@@ -183,18 +184,39 @@ class AnalyzerResolvers implements Resolvers {
...
+  ///
+  /// If no [_analysisOptions] is provided, then an empty one is used.
+  ///
+  /// If no [sdkSummaryGenerator] is provided, a default one is used that only
+  /// works for typical `pub` packages.
+  ///
+  /// If no [_packageConfig] is provided, then one is created from the current
+  /// [Isolate.packageConfig].
+  ///
+  /// **NOTE**: The [_packageConfig] is used only for language versions, we
```
natebosch, line 200: Given our previous experience allowing these types to leak into our interfaces and then later migrating - should we consider some sort of `Map` here instead? It could also reduce the confusion leading to this note.

jakemac53: talked offline, clarified the comment but left the type as its really only us who use this api and it allows for the most long term flexibility if new required fields are added.

jakemac53:
> But I don't have much better suggestions.

Make them public :)

PR #2671, opened 2020-04-19
Fix null error in _buildAnalyzerPackages when package in packageConfig has a null languageVersion so accessing languageVersion.major leads to a NoSuchMethodError

acoutts:
I originally noticed this when mobx_codegen was failing in my project to generate my store class code. Digging deeper, the exception was this:

```
NoSuchMethodError: The getter 'major' was called on null.
Receiver: null
Tried calling: major
dart:core                                                                                     Object.noSuchMethod
package:build_resolvers/src/analysis_driver.dart 71:39                                        _buildAnalyzerPackages
package:build_resolvers/src/analysis_driver.dart 46:18                                        analysisDriver
package:build_resolvers/src/resolver.dart 218:26                                              AnalyzerResolvers._ensureInitialized.<fn>
package:build_resolvers/src/resolver.dart                                                     AnalyzerResolvers._ensureInitialized.<fn>
package:build_resolvers/src/resolver.dart 221:6                                               AnalyzerResolvers._ensureInitialized
package:build_resolvers/src/resolver.dart 226:11                                              AnalyzerResolvers.get
package:build_runner_core/src/performance_tracking/performance_tracking_resolvers.dart 19:58  PerformanceTrackingResolvers.get.<fn>
package:build_runner_core/src/generate/performance_tracker.dart 302:15                        _NoOpBuilderActionTracker.trackStage
package:build_runner_core/src/performance_tracking/performance_tracking_resolvers.dart 19:16  PerformanceTrackingResolvers.get
package:build                                                                                 BuildStepImpl.resolver
package:source_gen/src/builder.dart 68:32                                                     _Builder.build
package:build                                                                                 runBuilder
package:build_runner_core/src/generate/build_impl.dart 486:19                                 _SingleBuild._runForInput.<fn>.<fn>.<fn>
package:build_runner_core/src/generate/performance_tracker.dart 302:15                        _NoOpBuilderActionTracker.trackStage
package:build_runner_core/src/generate/build_impl.dart 484:23                                 _SingleBuild._runForInput.<fn>.<fn>
package:build_runner_core/src/generate/build_impl.dart                                        _SingleBuild._runForInput.<fn>.<fn>
package:timing/src/timing.dart 222:44                                                         NoOpTimeTracker.track
package:build_runner_core/src/generate/build_impl.dart 441:22                                 _SingleBuild._runForInput.<fn>
package:pool/pool.dart 127:28                                                                 Pool.withResource
package:build_runner_core/src/generate/build_impl.dart 437:17                                 _SingleBuild._runForInput
package:build_runner_core/src/generate/build_impl.dart 375:38                                 _SingleBuild._runBuilder.<fn>
dart:async                                                                                    Future.wait
package:build_runner_core/src/generate/build_impl.dart 374:36                                 _SingleBuild._runBuilder
package:build_runner_core/src/generate/build_impl.dart 320:20                                 _SingleBuild._runPhases.<fn>.<fn>
package:build_runner_core/src/generate/build_impl.dart                                        _SingleBuild._runPhases.<fn>.<fn>
package:build_runner_core/src/generate/performance_tracker.dart 184:15                        _NoOpBuildPerformanceTracker.trackBuildPhase
package:build_runner_core/src/generate/build_impl.dart 316:47                                 _SingleBuild._runPhases.<fn>
package:timing/src/timing.dart 222:44                                                         NoOpTimeTracker.track
package:build_runner_core/src/generate/build_impl.dart 310:32                                 _SingleBuild._runPhases
package:build_runner_core/src/logging/logging.dart 25:30                                      logTimedAsync
package:build_runner_core/src/generate/build_impl.dart 267:26                                 _SingleBuild._safeBuild.<fn>
dart:async                                                                                    runZoned
package:build_runner_core/src/generate/build_impl.dart 262:5                                  _SingleBuild._safeBuild
package:build_runner_core/src/generate/build_impl.dart 209:24                                 _SingleBuild.run
package:build_runner_core/src/generate/build_impl.dart 94:56                                  BuildImpl.run
package:build_runner_core/src/generate/build_runner.dart 25:14                                BuildRunner.run
package:build_runner                                                                          BuildCommand.run
package:args/command_runner.dart 197:27                                                       CommandRunner.runCommand
package:args/command_runner.dart 112:25                                                       CommandRunner.run.<fn>
dart:async                                                                                    new Future.sync
package:args/command_runner.dart 112:14                                                       CommandRunner.run
package:build_runner                                                                          run
.dart_tool/build/entrypoint/build.dart 22:22                                                  main
```

In `build_resolvers/lib/src/analysis_driver.dart` in method `_buildAnalyzerPackages()`, the unsafe access to `package.languageVersion.major` and `package.languageVersion.minor` led to a `NoSuchMethodError` when `languageVersion` was null, which was the case a number of packages in my dependency tree.

Simple fix here to give some default values and use the null-safe `.?` operator, and now my code generators are running fine again.

```
Flutter 1.17.0 • channel beta • https://github.com/flutter/flutter.git
Framework • revision d3ed9ec945 (13 days ago) • 2020-04-06 14:07:34 -0700
Engine • revision c9506cb8e9
Tools • Dart 2.8.0 (build 2.8.0-dev.18.0 eea9717938)
```
```
 mobx_codegen: ^1.0.3
```

### DIFF
--- a/build_resolvers/lib/src/analysis_driver.dart
+++ b/build_resolvers/lib/src/analysis_driver.dart
@@ -67,8 +67,10 @@ Packages _buildAnalyzerPackages(
       for (var package in packageConfig.packages)
         package.name: Package(
           name: package.name,
-          languageVersion: Version(package.languageVersion?.major ?? 0,
-              package.languageVersion?.minor ?? 0, 0),
+          languageVersion: package.languageVersion == null
+              ? null
+              : Version(package.languageVersion.major,
+                  package.languageVersion.minor, 0),
           // Analyzer does not see the original file paths at all, we need to
           // make them match the paths that we give it, so we use the `assetPath`
           // function to create those.
--- a/build_resolvers/lib/src/analysis_driver.dart
+++ b/build_resolvers/lib/src/analysis_driver.dart
@@ -67,8 +67,8 @@ Packages _buildAnalyzerPackages(
       for (var package in packageConfig.packages)
         package.name: Package(
           name: package.name,
-          languageVersion: Version(
-              package.languageVersion.major, package.languageVersion.minor, 0),
+          languageVersion: Version(package.languageVersion?.major ?? 0,
+              package.languageVersion?.minor ?? 0, 0),
           // Analyzer does not see the original file paths at all, we need to
           // make them match the paths that we give it, so we use the `assetPath`
           // function to create those.


acoutts:
@googlebot I signed it!

mit-mit:
cc @natebosch

jakemac53:
The default language version is afaik supposed to be the language version of the current SDK - I thought that the `package_config` package already resolved this for us :/.

I am curious how you are getting a null language version also... 🤔 (I thought this would only happen if you are a standalone dart file which isn't supported anyways in package:build).

cc @lrhn for confirmation about what the default should be.

cc @scheglov can pass a null version here for analyzer and will it handle the defaulting?

jakemac53:
cc @lrhn would it make sense for package_config to do this in order to centralize the defaulting logic and make sure its done consistently?

jakemac53:
Ok I confirmed this happens when you are missing an SDK constraint - I didn't think we allowed publishing to pub without an SDK constraint, but if you have all path dependencies that could be quite common.

So one workaround in the meantime is to add sdk constraints to your pubspecs, something like this:

```yaml
environment:
  sdk: ">=2.7.0 <3.0.0"
```

build_resolvers/lib/src/analysis_driver.dart
```diff
@@ -67,8 +67,8 @@ Packages _buildAnalyzerPackages(
       for (var package in packageConfig.packages)
         package.name: Package(
           name: package.name,
-          languageVersion: Version(
-              package.languageVersion.major, package.languageVersion.minor, 0),
+          languageVersion: Version(package.languageVersion?.major ?? 0,
+              package.languageVersion?.minor ?? 0, 0),
```
jakemac53, line 71: It looks like analyzer allows a null version, which should handle the defaulting correctly, so this is I think the more semantically correct solution: 

```suggestion
          languageVersion: package.languageVersion == null
              ? null
              : Version(package.languageVersion.major,
                  package.languageVersion.minor, 0),
```

jakemac53:
I went ahead and committed my suggestion and will merge and follow up with https://github.com/dart-lang/build/pull/2672 for the release, should get it out in an hour or so.

lrhn:
The `package_config` package does not know the current language version. I merely serializes and deserializes a configuration. If a package entry has no language version entry in the JSON configuration, then it also has no language version in the Dart object model of that configuration.

The code using the `package_config` package should interpret "no language version" as meaning "current language version". That's what you get if you run a script file that you just wrote, with no language marker and no `pubspec.yaml`. It should be rare for packages configured by Pub, but the JSON format is not restricted to what Pub generates.

So, when you read `languageVersion` (which is nullable), you should probably be doing `packageConfig.languageVersion ?? currentVersion`.

jakemac53:
> So, when you read `languageVersion` (which is nullable), you should probably be doing `packageConfig.languageVersion ?? currentVersion`.

Right, we don't have any notion of current version either though. So the fix I have for now is to pass a null version to the analyzer. This may end up leaking all the way back to the code generators as null and not the "current version" (unless analyzer computes some current version and uses that), but at least it won't throw.

jakemac53:
Actually I think the best case scenario here is in fact that we get a `null` back for the code generator. Especially if it is trying to generate code that will be published on pub, where the language version would be different depending on who is running the code (since its whatever their current language version is).

jakemac53:
The fix for build_resolvers has been released as build_resolvers `1.3.6`

PR #2672, opened 2020-04-20
Release build_resolvers with null language version fix

jakemac53:
Also adds a test which validates the behavior

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.3.6
+
+- Fix bug when a package has no language version (as a result of having no sdk
+  constraint).
+
 ## 1.3.5
 
 - Create and pass a correct `Packages` argument to analysis driver, this
--- a/build_resolvers/pubspec.yaml
+++ b/build_resolvers/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_resolvers
-version: 1.3.5
+version: 1.3.6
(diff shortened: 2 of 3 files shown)


jakemac53:
Note that it does appear that analyzer gives us the _default_ language version here, and not `null` which would actually likely be what we want for codegen.

It _probably_ doesn't matter because I don't think we allow publishing without an SDK constraint any more, and published codegen files in packages is where this would really cause a problem.

jakemac53:
@scheglov I added a test here that the language version matches the current SDK version if not specified, but it looks like we are getting `2.8` even on a `2.9` sdk?

Is that a bug in analyzer or is there something else I need to do?

jakemac53:
It looks like this is a hardcoded value in the analyzer, https://github.com/dart-lang/sdk/blob/b774160615a741ce40ff6fd8f3bf7bf5f78556c4/pkg/analyzer/lib/src/dart/analysis/experiments.g.dart#L11.

But the SDK constraints in the analyzer here don't align with that version.... :/

jakemac53:
Updating the expect for now to expect `ExperimentStatus.currentVersion` from the analyzer (which is what it will give us internally) - I will file an issue on analyzer about this though.

scheglov:
This works as intended - from the element model we should return the actual language version.
And analyzer returns the the version that it understands as the current version by default.

jakemac53:
Filed https://github.com/dart-lang/sdk/issues/41583

PR #2674, opened 2020-04-21
Reduce the chances of different build phase order

### LINKED ISSUE #2645: Sharing .dart_tool in CI

natebosch:
Fixes #2645

Sort the list of packages we get from `PackageConfig`. This won't fully
guarantee consistency since we later have behavior that doesn't make
ordering guarantees, but in practice with a consistent package order to
start with we _should_ get a consistent build phase order.

### DIFF
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -111,9 +111,6 @@ final _builders = <_i1.BuilderApplication>[
   _i1.apply('provides_builder:throwing_builder', [_i4.throwingBuilder],
       _i1.toDependentsOf('provides_builder'),
       hideOutput: true),
-  _i1.applyPostProcess(
-      'provides_builder:some_post_process_builder', _i4.somePostProcessBuilder,
-      defaultGenerateFor: const _i3.InputSet()),
   _i1.applyPostProcess('build_modules:module_cleanup', _i5.moduleCleanup,
       defaultGenerateFor: const _i3.InputSet()),
   _i1.applyPostProcess('build_web_compilers:dart2js_archive_extractor',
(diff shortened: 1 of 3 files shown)


natebosch:
I suspect we could also get inconsistencies if things like `pubspec.yaml` files change their dependency orders but I haven't dug deeply. I think we've had potential bugs lurking here for a long time and in practice things have always worked out until now. This (should) fix the immediate issue, but doesn't address the deeper and longer standing problem.

natebosch:
> please add a changelog entry as well

I have one on `build_runner_core` - did you want me to copy it to `build_runner`?

jakemac53:
I must have just missed it, lgtm

Issue #2675, opened 2020-04-21
Allow enabling of experiments for the AnalyzerResolver. 


jakemac53:
We need to do this so that users of build_runner will be able to enable any experiments in their code.

We should probably just follow the standard set by the SDK here and add an `--enable-experiment` flag. We will need to throw away the world if it changes from the previous run - or at least that would be the easiest option.

natebosch:
@jakemac53 - is this done?

PR #2682, opened 2020-04-24
Enable experiment flag

jakemac53:
First part of https://github.com/dart-lang/build/issues/2675

Before publishing I am going to send a followup to update build_web_compilers/build_vm_compilers to use this flag, and get some tests written as well (although I haven't decided how to test it actually... since experiments come and go).


build/lib/experiments.dart
```diff
@@ -0,0 +1,5 @@
+// Copyright (c) 2020, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+export 'src/experiments.dart' show enabledExperiments, withEnabledExperiments;
```
natebosch, line 5: [optional] Would `dart_experiments.dart` or `language_experiments.dart` reduce confusion against something related to build/builder experiments?

jakemac53: We could do `language_experiments.dart`, `languageExperiments`, and `withLanguageExperiments`. I don't feel strongly either way, wdyt?

natebosch: After looking again, I'm not sure that "language experiments" is very widely used. We don't use that term at all on https://github.com/dart-lang/sdk/blob/master/docs/process/experimental-flags.md#flag-format-for-cli-based-tools

cc @leafpetersen - do you think we should use the general term "experiments" to refer to Dart language feature experiment flags, or should we use something more specific?

@jakemac53 - I noticed that our [docs are wrong](https://github.com/dart-lang/sdk/issues/41676) when they show how to use the flag. The VM requires the `=`, not a space separated option. Would it be sensible to manually require the same here to reduce confusion on slightly different UX depending on how you run? `package:args` allows either, but we could special case it to check that it's used, or maybe issue a warning?

leafpetersen: Experiment is fine by me.

jakemac53: > Would it be sensible to manually require the same here to reduce confusion on slightly different UX depending on how you run? `package:args` allows either, but we could special case it to check that it's used, or maybe issue a warning?

I think its fine if we are a bit more permissive here than the vm or other tools, it isn't worth sacrificing usability just for the sake of 100% consistency.

PR #2683, opened 2020-04-28
[WIP] Use experiment zone to plumb flags through to kernel/analyzer

jakemac53:
Deprecates the old ways of doing things, but leaves support in for now with warnings.

I also refactored build_resolvers a bit to pass the current SDK language version if it is `null` for the package, since the analyzer won't choose the correct default based on the current sdk (its based on analyzer version).

@stereotype441 @scheglov The test I added is failing here for a couple reasons:

* https://github.com/dart-lang/sdk/issues/41691
* Unknown issue where I can't get the library to be opted in even when the experiment definitely is enabled and the language version is equal to the current SDK language version (via the override comment).

Closes https://github.com/dart-lang/build/issues/2685


jakemac53:
We came to a resolution in the linked issue, this now throws a `StateError` if you attempt to enable experiments but the sdk/analyzer do not agree on the current language version.

I also added in the fix for https://github.com/dart-lang/sdk/issues/41820.

jakemac53:
@stereotype441 is there any chance you can help debug the test added in this PR? It is failing saying that the experiment is not enabled still, but I have debugged it and the experiment definitely is enabled on the analysis context. (I added this line to the test `print(lib.context.analysisOptions.contextFeatures.isEnabled(ExperimentalFeatures.non_nullable));`), which prints `true`.

build_resolvers/test/resolver_test.dart
```diff
@@ -355,6 +355,25 @@ void main() {
...
       }, resolvers: AnalyzerResolvers());
     });
+
+    test('Respects withEnabledExperiments', () async {
+      Logger.root.level = Level.ALL;
+      Logger.root.onRecord.listen(print);
+      await withEnabledExperiments(
+          () => resolveSources({
+                'a|web/main.dart': '''
+// @dart=${sdkLanguageVersion.major}.${sdkLanguageVersion.minor}
```
scheglov, line 365: 1. What version do you get here?
2. What version is in `ExperimentStatus.currentVersion` in analyzer?
3. Do you have a version of analyzer that includes https://dart-review.googlesource.com/c/sdk/+/145220?

jakemac53: > * What version do you get here?

2.9

> * What version is in `ExperimentStatus.currentVersion` in analyzer?

2.9

> * Do you have a version of analyzer that includes https://dart-review.googlesource.com/c/sdk/+/145220?

I have reverted to not call `restrictToVersion` at all if any experiments are enabled, so it shouldn't matter. Instead it now throws if you attempt to use an experiment and the analyzer and SDK don't agree on a language version.

jakemac53: The test also has a check that the language version is correct, (it expects the library language version to equal the sdk language version)

scheglov: I'm not sure where you reverted the call to `restrictToVersion`, but analyzer [calls it itself](https://github.com/dart-lang/sdk/blob/2ccfc88f25e378ac82e2548bd8f5117d8ebea6ab/pkg/analyzer/lib/src/dart/analysis/feature_set_provider.dart#L33) to determine the set of enabled features for a file.

jakemac53: My analyzer version though fwiw is the latest published version, 0.39.8. I don't see any tags in the SDK for this release so I don't know if its included or not.

scheglov: Try to set breakpoint in `FeatureSetProvider` and check which versions and features you get.

jakemac53: ok, thanks for that code pointer @scheglov I noticed there that for non-packages there is a totally different feature set that has to be set, the `nonPackageFeatureSet`. If I set that in addition to the `contextFeatures` then the test passes 🎉

jakemac53: Actually, it looks like it also requires a dependency override on analyzer/_fe_analyzer_shared to the latest version in the SDK (to pick up that fix).

When can we expect a new version of analyzer to be published?

jakemac53:
ok, PTAL @natebosch 

I have confirmed the test will pass as soon as analyzer published a new version, and manually did some e2e testing for the whole workflow and it works.

build_resolvers/lib/src/resolver.dart
```diff
@@ -385,20 +385,36 @@ existing issues if present to avoid duplicates.
 final _dartUiPath =
     p.normalize(p.join(_runningDartSdkPath, '..', 'pkg', 'sky_engine', 'lib'));
 
-/// The current feature set based on the current sdk version
-FeatureSet _featureSet({List<String> enableExperiments}) =>
-    FeatureSet.fromEnableFlags(enableExperiments ?? [])
-        .restrictToVersion(_sdkLanguageVersion);
+/// The current feature set based on the current sdk version and enabled
+/// experiments
```
natebosch, line 389: [nit] End with a period.

PR #2696, opened 2020-05-14
Fix AnalyzerResolver methods isLibrary and libraryFor 

jakemac53:
These now do explicit `canRead` checks to try and keep people on the rails.

Also added `.exists()` checks to sources since we can get an empty source instead of a null source now.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -91,6 +91,7 @@ class PerActionResolver implements ReleasableResolver {
 
   @override
   Future<LibraryElement> libraryFor(AssetId assetId) async {
+    if (!await _step.canRead(assetId)) throw AssetNotFoundException(assetId);
     await _resolveIfNecesssary(assetId);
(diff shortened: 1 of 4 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -134,7 +137,9 @@ class AnalyzerResolver implements ReleasableResolver {
     var path = assetPath(assetId);
     var uri = assetId.uri;
     var source = _driver.sourceFactory.forUri2(uri);
-    if (source == null) throw ArgumentError('missing source for $uri');
+    if (source == null || !source.exists()) {
+      throw ArgumentError('missing source for $uri');
```
natebosch, line 141: [out of scope?] I wonder if this should have been `AssetNotFoundException`... Is there any risk in changing it?

jakemac53: I will go ahead and change it, there is some small risk but it should be very minor.

Issue #2343, opened 2019-07-03
testBuilder should provide a way to specify language versions


srawlins:
See the accepted [Dart Language Versioning](https://github.com/dart-lang/language/blob/master/accepted/future-releases/language-versioning/feature-specification.md) specification.

In summary, the "Dart language version" of a package can be specified in the `pubspec.yaml` of a pub package, and forwarded to the `.packages` file of a local pub checkout (e.g. by `pub get`). Additionally, an individual library can override the Dart language version that its package specifies, via a comment at the top, e.g. `// @dart = 2.0`. Only overrides to _earlier_ versions are allowed.

So, `testBuilder` (and similar methods?) should allow the test author to specify what Dart language version the source assets are using. Presumably one language version per package declared in the source assets. Builder authors will want to test, e.g., that they can build correct code when one asset uses Dart language version 2.5, and that asset imports another asset, which uses Dart language version 2.0.

#2229 is related: at a time when an unreleased Dart language version requires an experiment flag (such as now, with the NNBD experiment flag), then `testBuilder` will also need to "pass  on" that flag.

natebosch:
Can this be put in to the source asset content?

srawlins:
Alas. The only Dart language versioning that can be specified in a library content is _overriding_ the Dart language version specified in a package, to an earlier version.

natebosch:
Ah, in that case we need to start by figuring out whether we need to communicate package version information to the analyze in the non-test case.

@stereotype441 - can you weight in? Will we need to do something around language versioning with the `AnalysisDriver`?

jakemac53:
This seems like a larger problem. Imo it isn't reasonable to expect codegen authors to generate different code based on the supported language version of the library they are generating into. Ideally we wouldn't even surface that information to them.

One option would be for code generators to simply put in the `// @dart = x.x` comment based on _their_ minimum sdk constraint, but that has several issues as well (especially if you can't upgrade the version). It also makes combining builders harder but that should be resolvable.

jakemac53:
cc @lrhn for your thoughts since I think you own the feature

Given NNBD it seems like code generators may not be able to avoid knowing about these language version constraints...

jakemac53:
[...]

PR #2702, opened 2020-05-21
support language versioning and experiments in build_test apis

jakemac53:
Closes https://github.com/dart-lang/build/issues/2343

Adds support to the `resolve*` and `testBuilder` apis for passing a `PackageConfig` object which is used for setting language versions.

Also respects the `withEnabledExperiments` zone if running in one to allow enabling of experiments. Note that the experiment support is untested as there is no way to write a test that won't break in the future.

cc @srawlins


srawlins:
🤩 Thanks @jakemac53

PR #2703, opened 2020-05-21
update build_resolvers constraint in build_test

jakemac53:
forgot to bump this in the last pr

### DIFF
--- a/build_test/pubspec.yaml
+++ b/build_test/pubspec.yaml
@@ -10,7 +10,7 @@ dependencies:
   async: ">=1.2.0 <3.0.0"
   build: ">=1.3.0 <2.0.0"
   build_config: ">=0.2.0 <0.5.0"
-  build_resolvers: ">=0.2.0 <2.0.0"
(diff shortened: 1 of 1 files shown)


build_test/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
   async: ">=1.2.0 <3.0.0"
   build: ">=1.3.0 <2.0.0"
   build_config: ">=0.2.0 <0.5.0"
-  build_resolvers: ">=0.2.0 <2.0.0"
+  build_resolvers: ">=1.3.5 <2.0.0"
```
natebosch, line 13:
```suggestion
  build_resolvers: ^1.3.0
```

Or did we publish a feature change in a patch release? `^1.3.5` would be fine in that case.

jakemac53: Yes - this is the release that adds the PackageConfig parameter.

natebosch: Oops!

PR #2706, opened 2020-05-24
Use iterable.expand.take over iterable.fold.take

natebosch:
Previously we would iterate over _all_ of `pendingActions.values` and
collect them to a list, only to throw all but 5 away. Now we only
iterate as far in `values` as it takes to reach `actionsToLogMax` since
there is no `List`, it's all iterables.


natebosch:
I noticed when I was playing around with the analysis server that we weren't getting inference for the `combined` argument and it was treated as `dynamic`. When I started to correct that I noticed there is an overall better approach.

Issue #2684, opened 2020-05-04
Tests run under build_runner don't allow user-defined platform


alanknight-wk:
* Dart SDK Version 
2.7.2
* What package(s) from this repo you are using, and the version (i.e. `build_runner 0.7.12`)
build_test - 0.10.12+1
* What builder(s) you are using (or writing yourself). Try to give a short summary of what they do.
* Whether you are using Windows, MacOSX, or Linux (if applicable)
Mac
* Whether you are using Chrome, Safari, Firefox, Edge (if applicable)
Chrome
* Any other packages or constraints we should know about

I have a test suite where I define a new platform. It runs fine under pub test, but pub run build_runner test fails with the error 

Undefined variable ...  @TestOn('chromeWithExtension')

It feels like this should be relatively simple. If build didn't complain, test would run this successfully. It seems like it just needs build to invoke test in a way that it sees dart_test.yaml and can answer about those platforms. 

The motivation is that I have a Chrome Extension and want to be able to test both with the extension loaded and not loaded.

natebosch:
Can you run with `--verbose` to get a stack trace?

natebosch:
It looks like we hardcode known platforms.

https://github.com/dart-lang/build/blob/65e54a4aa5ebb1a7d17040f3a592cfb2f215f357/build_test/lib/src/test_bootstrap_builder.dart#L52-L54

@jakemac53 - do you know if user defined platforms have enough information to know if we should consider it a browser test?

alanknight-wk:
/me waves

User-defined platforms can only inherit from existing platforms, so any check (other than a hard-coded list) that tells you if a regular platform is a browser test should also work on user-defined tests. For example, my motivating use case is to define a sub-platform of Chrome with additional command-line options.

alanknight-wk:
```

Error on line 2, column 10 of test/clipboard_extension_test.dart: Undefined variable.
  ╷
2 │ @TestOn('chromeWithExtension')
  │          ^^^^^^^^^^^^^^^^^^^
  ╵
package:test_core/src/runner/parse_metadata.dart 516:7                  _Parser._contextualize
package:test_core/src/runner/parse_metadata.dart 123:12                 _Parser._parsePlatformSelector
package:test_core/src/runner/parse_metadata.dart 117:7                  _Parser._parseTestOn
package:test_core/src/runner/parse_metadata.dart 83:18                  _Parser.parse
[...]

PR #2708, opened 2020-05-28
Support custom platforms in build_test

jakemac53:
Adds support for recognizing custom platforms in TestOn annotations during bootstrapping.

Fixes https://github.com/dart-lang/build/issues/2684

### DIFF
--- a/build_test/lib/src/test_bootstrap_builder.dart
+++ b/build_test/lib/src/test_bootstrap_builder.dart
@@ -1,6 +1,7 @@
 // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+
 import 'dart:async';
 
 import 'package:build/build.dart';
--- a/build_test/lib/src/test_bootstrap_builder.dart
+++ b/build_test/lib/src/test_bootstrap_builder.dart
@@ -1,14 +1,13 @@
 // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
-
 import 'dart:async';
 
 import 'package:build/build.dart';
+import 'package:crypto/crypto.dart';
 import 'package:path/path.dart' as p;
 // ignore: deprecated_member_use
 import 'package:test_core/backend.dart';
-import 'package:test_core/src/runner/configuration/load.dart';
 
 /// A [Builder] that injects bootstrapping code used by the test runner to run
 /// tests in --precompiled mode.
@@ -38,8 +37,8 @@ class TestBootstrapBuilder extends Builder {
     var browserRuntimes =
         Runtime.builtIn.where((r) => r.isBrowser == true).toList();
     var nodeRuntimes = [Runtime.nodeJS];
-    if (await buildStep.canRead(AssetId(id.package, 'dart_test.yaml'))) {
-      var config = load('dart_test.yaml');
+    var config = await _ConfigLoader.instance.load(id.package, buildStep);
+    if (config != null) {
       for (var customRuntime in config.defineRuntimes.values) {
         var parent = customRuntime.parent;
         if (vmRuntimes.any((r) => r.identifier == parent)) {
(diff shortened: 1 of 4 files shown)


build_test/lib/src/test_bootstrap_builder.dart
```diff
@@ -32,10 +33,41 @@ class TestBootstrapBuilder extends Builder {
     var assetPath = id.pathSegments.first == 'lib'
         ? p.url.join('packages', id.package, id.path)
         : id.path;
+
+    var vmRuntimes = [Runtime.vm];
+    var browserRuntimes =
+        Runtime.builtIn.where((r) => r.isBrowser == true).toList();
+    var nodeRuntimes = [Runtime.nodeJS];
+    if (await buildStep.canRead(AssetId(id.package, 'dart_test.yaml'))) {
+      var config = load('dart_test.yaml');
```
natebosch, line 42: Can we find a way to do this only once, instead of once per test file? Maybe it can be a `Resource`?

natebosch:
I think exposing this from `test_core` should be OK.

build_test/lib/src/test_bootstrap_builder.dart
jakemac53: Sure, a Resource would make sense.

jakemac53: Ended up not needing a resource as there is nothing worth bothering cleaning up explicitly like extra processes, temp dirs, etc.

jakemac53:
this is ready for review @natebosch

build_test/lib/src/test_bootstrap_builder.dart
```diff
@@ -76,3 +107,34 @@ class TestBootstrapBuilder extends Builder {
     }
   }
 }
+
+/// Manages a cache of [Configuration] per package, to avoid duplicating work
+/// across build steps.
+///
+/// Can safely be used across the build as configuration is invalidated by its
```
natebosch, line 114: I don't understand this line. Are you saying that test config changing causes us to throw away the asset graph?

build_test/lib/src/test_bootstrap_builder.dart
```diff
@@ -76,3 +107,34 @@ class TestBootstrapBuilder extends Builder {
...
+/// digest.
+class _ConfigLoader {
+  _ConfigLoader._();
+
+  static final instance = _ConfigLoader._();
+
+  final _configByPackage = <String, Future<Configuration>>{};
```
natebosch, line 121: When would we have a different package? I don't think we'd let you read this for anything but the root package since it's not in `lib/` right?

build_test/lib/src/test_bootstrap_builder.dart
jakemac53: I am saying that this class uses the digests of the configuration files it reads to always give you an accurate result.

build_test/lib/src/test_bootstrap_builder.dart
jakemac53: In theory you could have tests under lib or something, or in some other build system scenario maybe you have more than one. This just future proofs things and doesn't have any assumptions about that.

jakemac53: The package comes from the primary input ID, so it works for any test under any package.

natebosch: Will `reader.canRead` always return false for any `AssetId(package, 'dart_test.yaml')` where `package` is not the root package, because we don't include other sources in the asset graph?

https://github.com/dart-lang/build/blob/93de086d54c5e70c7c504fc5ab1130df81d5d741/build_runner_core/lib/src/generate/build_definition.dart#L170-L173

jakemac53: Yes it will - in build_runner. These are theoretically general builders though and that could change. We could also want to plug this into some other build system some day which does not have that same restriction.

natebosch: I don't think we should be writing this for a theoretical build system, but on the other hand it doesn't seem like it makes the code any more complicated since we don't know the root package anyway... 🤷

jakemac53: right, I would otherwise be writing some assert that I only ever see the same package - which wouldn't end up being any more complicated and this way it just isn't an issue 🤷

PR #2709, opened 2020-05-29
Prep test for NNBD

grouma:
Ensure the right VM args `--enable-experiment=non-nullable` is passed to child process.

### DIFF
--- a/build_daemon/test/daemon_test.dart
+++ b/build_daemon/test/daemon_test.dart
@@ -196,9 +196,8 @@ Future<Process> _runDaemon(var workspace, {int timeout = 30}) async {
 
   var args = [
     ...Platform.executableArguments,
-    if (Platform.executableArguments
-            .indexWhere((arg) => arg.startsWith('--packages=')) ==
-        -1)
+    if (!Platform.executableArguments
+        .any((arg) => arg.startsWith('--packages')))
       '--packages=${(await Isolate.packageConfig).path}',
     'test.dart'
   ];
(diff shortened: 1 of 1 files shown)


build_daemon/test/daemon_test.dart
```diff
@@ -194,10 +194,15 @@ Future<Process> _runDaemon(var workspace, {int timeout = 30}) async {
...
       ''').create();
 
-  var packageArg = '--packages=${(await Isolate.packageConfig).path}';
-
-  var process = await Process.start(
-      Platform.resolvedExecutable, [packageArg, 'test.dart'],
+  var args = [
+    ...Platform.executableArguments,
+    if (Platform.executableArguments
+            .indexWhere((arg) => arg.startsWith('--packages=')) ==
```
jakemac53, line 200: This should just check for `--packages`, you can pass `--packages <path>` as well afaik.

build_daemon/test/daemon_test.dart
```diff
@@ -194,10 +194,15 @@ Future<Process> _runDaemon(var workspace, {int timeout = 30}) async {
...
+        -1)
```
natebosch, line 201:
```suggestion
    if (!Platform.executableArguments
            .any((arg) => arg.startsWith('--packages=')))
```

build_daemon/test/daemon_test.dart
natebosch: `Setting VM flags failed: Unrecognized flags: packages`

PR #2715, opened 2020-06-03
Add FAQ section on using multiple configurations

natebosch:


docs/faq.md
```diff
@@ -38,6 +38,54 @@ targets:
...
+## How can I build with multiple configurations
+
+The build system supports two types of builds, "dev" and "release". By default
+with `build_runner` the "dev" version is built by default regardless of the
+command used, build in release mode by passing the `--release` flag. With
+`webdev` the default mode for the `serve` command is dev, and the default mode
+for the `build` command is release. The `build` command can use dev mode with
+the `--no-release` flag.
+
+Options can be configured per mode, and they are merged by key. The `options`
```
filiph, line 50: I don't know what it means that configurations are merged by key. Not necessarily a show stopper - I'm a beginner at this, and this is probably focused on advanced users. Just wanted to point it out.

docs/faq.md
```diff
@@ -38,6 +38,54 @@ targets:
...
+          - -O3
+```
+
+If you need other configurations in addition to dev and release, you can define
+multiple `build.yaml` files. For instance if you have a `build.debug.yaml` file
+you can build with `--config debug` and this file will be used instead of the
+default `build.yaml`. The dev and release flavors still apply. `pub run
+build_runner serve --config debug` will use the `dev_options` in
+`build.debug.yaml`, while `pub run build_runner build --config debug --release`
+will use the `release_options` in `build.debug.yaml`.
```
filiph, line 81: Does this `--config` option work also with `webdev`? If so, it would be great to mention it. If not, same.

docs/faq.md
jakemac53: This is intended for anybody to use/read, it is supposed to be the most searchable/easy spot to get common questions answered.

So, if its not meaningful to you we should consider rewording it :)

docs/faq.md
jakemac53: I think  you have to put it after a `--`, so `webdev <args> -- --config debug` for instance. Webdev doesn't support all the build_runner arguments directly but you can forward additional args to build_runner that way.

docs/faq.md
natebosch: Yeah, I'm trying to decide between expanding on this here, or writing another FAQ or document about option merging and linking it here.

jakemac53: We could add a tiny entry like "What is a builder key" and we could link references to "key" to that.

jakemac53: Also ya the merging seems like it might be deserving of its own section as it is somewhat complicated to explain.

docs/faq.md
```diff
@@ -38,6 +38,113 @@ targets:
...
+
+The build system supports two types of builds, "dev" and "release". By default
+with `build_runner` the "dev" version is built by regardless of the command
+used, build in release mode by passing the `--release` flag. With `webdev` the
+default mode for the `serve` command is dev, and the default mode for the
+`build` command is release. The `build` command can use dev mode with the
+`--no-release` flag.
+
+Options can be configured per mode, and they are [merged by
+key](#how-is-the-configuration-for-a-builder-resolved) with the defaults
```
natebosch, line 51: @kevmoo - do you know if there is any way to make this shorter and less likely to break if we reword the header?

kevmoo: I don't sadly – is there a standard way to put anchors in markdown?

natebosch:
@jakemac53 - can you take another look with the new section on merging?

docs/faq.md
natebosch: Some other markdown renderers support things like `# Really long header that might not be stable {#stable-short-anchor}`

Github doesn't seem to and I can't find any alternative so far. Oh well.

PR #2718, opened 2020-06-04
Allow the latest build_runner_core

natebosch:


natebosch:
No relevant changes for the changelog, there were some minor technically breaking changes but no impact to `build_runner`.

PR #2727, opened 2020-06-12
Refactor factories to constructor initializers

natebosch:
When a factor is trivially forwarding to another generative constructor
with some defaulted arguments it can be written as an initializer list
instead, similar to a `super` constructor. This makes it clear that the
factory is not used for other purposes, such as returning a potentially
different concrete type.


natebosch:
Opening for discussion. I just today learned that this is even a thing. I'm curious what other people think.

jakemac53:
Huh, ya I hadn't heard of this either. I am not opposed to the change, its a bit less verbose and I can't think of a downside really? Even if you have never heard of this feature I think its fairly obvious what it is doing.

eernstg:
Drive-by comment, just in case anyone is curious about the precise rules&mdash;the phrase to look for in the spec is [redirecting generative constructor](https://github.com/dart-lang/language/blob/123ba363819b4be0af80b8a94ef6bd0b36db51ba/specification/dartLangSpec.tex#L3044). :smile:

natebosch:
I think it makes sense to go forward with this - it's pretty minor but I do think in these cases it reads a bit nicer and it may help us to use this feature more in the future.

grouma:
Is there a lint to prevent this pattern popping up in the future?

natebosch:
> Is there a lint to prevent this pattern popping up in the future?

No and I don't think it would make sense to add one. These changes do technically show up in the public API and there are valid reasons to pick one over the other. For instance not having any public generative constructors makes it impossible to use `extends`, and a `factory` can't be used as a `super` constructor even if some other generative constructor exists. Using a generative constructor is always the most flexible when possible, but flexibility may not be the goal so a lint could have false positives.

natebosch:
My goal here is mainly to get some more use of this pattern into our code so we're used to seeing it and that hopefully will lead us to use it more (I also plan on trying to keep an eye for places to use it in the future) since I do think it reads nicely.

@grouma @jakemac53 - do you think it's worth merging this? Or should I drop it?

grouma:
Feel free to merge imo.

natebosch:
> Feel free to merge imo.

I need an approval first 😉

PR #2733, opened 2020-06-18
Support new step build mode

grouma:
Add a new `Step` build mode. In this mode, the build daemon automatically watches for files and does corresponding builds. However, after each build, the daemon will collect file changes and wait to build until there is a continue request.

The goal of this change is so that other tools that integrate with the build daemon can have more control over when the builds occur while still maintaining the watch behavior.

Note that the build daemon can have multiple clients so the step behavior may be problematic in that scenario. I have some ideas for solutions but do not want to introduce the added complexity until necessary.


natebosch:
What is going to be the client that starts using this now?

build_daemon/README.md
```diff
@@ -1 +1,6 @@
 This package provides support for running builds in a daemon.
+
+Note since `package:build_daemon` is a dependency of `package:build_runner`,
+you should build this package with the following command:
```
natebosch, line 4:
```suggestion
if this is not a clean git checkout you should build this package with the following command:
```

grouma:
> What is going to be the client that starts using this now?

The idea is that the client within `package:webdev` will use this. After a build result, it will run the Frontend Server. Once that completes it will issue the continue command. This should resolve the race condition issue.

> I'm really confused why there wasn't more difference in build_runner...

The complexity is handled in `package:build_daemon` so really it just needed to support the new argument. Also you forget how amazing I am...

build_daemon/README.md
```diff
@@ -1 +1,6 @@
 This package provides support for running builds in a daemon.
+
+Note since `package:build_daemon` is a dependency of `package:build_runner`,
+if this is not a clean git checkout you should build this package with the following command:
+
+dart .dart_tool/build/entrypoint/build.dart.snapshot build --skip-build-script-check
```
jakemac53, line 6: This file won't exist for everybody - they may need to at least do a `pub run build_runner generate-build-script` command and then I would recommend dropping the `.snapshot` here since that also won't be generated.

build_daemon/lib/daemon.dart
```diff
@@ -89,6 +91,24 @@ class Daemon {
     }));
   }
 
+  BuildMode _buildModeFor(Set<String> options) {
+    for (var option in options) {
+      if (option.startsWith('--$buildModeFlag')) {
+        var modeValue = option.split('=').last;
```
jakemac53, line 97: There are other valid ways of providing this arg, it is probably worth just using the arg parser here imo? But we should at least support the `--build-mode <mode>` style.

build_daemon/lib/src/server.dart
```diff
@@ -83,6 +89,10 @@ class Server {
               ? _buildTargetManager.targets
               : _buildTargetManager.targetsForChanges(changes);
           await _build(targets, changes);
+        } else if (request is ContinueRequest) {
+          if (!_stepCompleter.isCompleted) {
```
jakemac53, line 93: I do think we need to do one of two things here:

- Wait for all connected clients to issue a continue request
- Only allow one client to connect in this mode

build_runner/test/daemon/daemon_test.dart
```diff
@@ -240,6 +252,53 @@ main() {
...
     });
 
+    test('step build mode does not automatically build without continuing',
+        () async {
+      await _startDaemon(buildMode: BuildMode.Step);
+      var client = await _startClient(buildMode: BuildMode.Step)
+        ..registerBuildTarget(webTarget);
+      clients.add(client);
+      // Let the target request propagate.
+      await Future<void>.delayed(Duration(seconds: 2));
```
jakemac53, line 262: 😭 does `_startClient` need to be updated? Maybe we need a `registerSuccess` response to let us know the build daemon is ready? These types of artifical delays need to be a last resort.

That doesn't have to happen in this PR but it would be a good general change I think and help eliminate real world race conditions as well.

build_daemon/lib/daemon.dart
grouma: There are no other valid ways since the client actually embeds this argument and it does it exactly as parsed.

build_daemon/README.md
grouma: I committed Nate's suggestion which addresses this.

build_runner/test/daemon/daemon_test.dart
grouma: That could help although this only shows up in testing. I can actually rework this test to not require the delay.

build_daemon/lib/src/server.dart
grouma: I referenced this in the PR description. Long term I do think we'll leverage this feature with Blaze so this will have to be addressed but I'm not sure it's worth concerning ourselves with at the moment. If you feel strongly I'll probably just add in the logic to wait for all connected clients.

natebosch:
> The idea is that the client within `package:webdev` will use this. After a build result, it will run the Frontend Server. Once that completes it will issue the continue command. This should resolve the race condition issue.

Are we starting that work today? I would rather not put this support in unless we are sure we will need it...

natebosch:
> The complexity is handled in `package:build_daemon` so really it just needed to support the new argument. Also you forget how amazing I am...

Ah I see now the change for `_stepCompleter`. Looks like someone brilliant must have helped with that whole `asyncMapBuffer` structure, and the idea of this step mode that fits so nicely with our existing code 😉

build_daemon/lib/daemon.dart
jakemac53: That seems like a pretty strong assumption to make still?

build_daemon/lib/src/server.dart
jakemac53: I do think we need to implement one of these options. Otherwise this is incurring significant technical debt.

build_runner/test/daemon/daemon_test.dart
jakemac53: If you can rework it to not require the delay please do.

PR #2739, opened 2020-06-26
Merged metadata

grouma:
- Update `build_web_compilers|ddc` builder to produce a `.metadata` file.
- Update `build_web_compilers|entrypoint` builder to produce a `.merged_metadata` file
  which consists of new line separated `.metadata` content produced by DDC.
- Bump the minimum SDK to one in which DDC supports `--experimental-emit-debug-metadata`

### DIFF
--- a/build_web_compilers/pubspec.yaml
+++ b/build_web_compilers/pubspec.yaml
@@ -1,10 +1,10 @@
 name: build_web_compilers
-version: 2.11.0
+version: 2.12.0
 description: Builder implementations wrapping Dart compilers.
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
 
 environment:
-  sdk: ">=2.8.1 <3.0.0"
+  sdk: ">=2.9.0-10.0.dev <3.0.0"
 
 dependencies:
   analyzer: ">=0.30.0 <0.40.0"
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,16 @@ jobs:
       env: PKGS="build build_config build_daemon build_resolvers build_runner_core build_test build_vm_compilers scratch_space"
       script: ./tool/travis.sh dartanalyzer_1
     - stage: analyze_and_format
-      name: "SDK: 2.8.1; PKGS: build_modules, build_web_compilers; TASKS: `dartanalyzer --fatal-warnings .`"
+      name: "SDK: 2.8.1; PKG: build_modules; TASKS: `dartanalyzer --fatal-warnings .`"
       dart: "2.8.1"
       os: linux
-      env: PKGS="build_modules build_web_compilers"
+      env: PKGS="build_modules"
+      script: ./tool/travis.sh dartanalyzer_1
+    - stage: analyze_and_format
+      name: "SDK: dev; PKG: build_web_compilers; TASKS: `dartanalyzer --fatal-warnings .`"
+      dart: dev
+      os: linux
+      env: PKGS="build_web_compilers"
       script: ./tool/travis.sh dartanalyzer_1
     - stage: unit_test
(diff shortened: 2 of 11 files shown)


build_web_compilers/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: build_web_compilers
-version: 2.11.0
+version: 2.12.0
 description: Builder implementations wrapping Dart compilers.
 homepage: https://github.com/dart-lang/build/tree/master/build_web_compilers
 
 environment:
-  sdk: ">=2.8.1 <3.0.0"
+  sdk: ">=2.9.0-10.0.dev <3.0.0"
```
jakemac53, line 7: cc @natebosch @leafpetersen 

Note that this will potentially cause problems with the NNBD tech preview...this package will now be opted in if the experiment is enabled. Concretely that would affect builder migrations, no builder can run `pub run --enable-experiment=non-nullable build_runner <command>` with that stable sdk (or the current dev), they would have to update to the later `2.10.0-x.x.dev` release.

This highlights exactly the scenario we were concerned about last week. Not having an explicit language version field separate from the sdk version is problematic for cases like these. Imagine NNBD was already released, we would be blocked on our NNBD migration in order to depend on this functionality, but this package will never migrate to NNBD most likely, or certainly not for a very long time (the priority is extremely low to do that for this package, and it has a lot of transitive deps it would anyways be blocked on). At the same time it regularly depends on new compiler functionality. The workaround is language version comments in _every single file_ but gross.

grouma:
> It is probably worth adding some basic tests here (you should be able to just update the existing ones to expect additional outputs). If we don't care about the exactly content it would be fine to just expect some file.

Ah yes. I was intending on doing that but obviously forgot. Should be good now.

build_web_compilers/pubspec.yaml
natebosch: FWIW I don't mind language version comments in every file.

grouma:
> These new files need to be added to the source cleanup builder.
> 
> https://github.com/dart-lang/build/blob/026cefa21ee3814712c5505c8fdafee6ec033553/build_web_compilers/lib/builders.dart#L70-L73
> 

These files are not written to source. Do we still need to clean them up?

> Given that we need to clean them up for production builds, can we also look for a more specific file extension? I'm worried that `.metadata` might already be used by some other builder that wouldn't want their outputs deleted. Can we use `.ddc.metatada`? Or is there already a `.ddc.` extension in there?

The extension is actually `.ddc.js.metadata`. We don't have control of the extension here as it is decided by DDC / the SDK.

natebosch:
> These files are not written to source. Do we still need to clean them up?

Yes, otherwise they will be included in the output directory if you use the `--output` flag (which is always used by `webdev` for `webdev build`). Technically we don't claim to support "production" builds with DDC, but for consistency we should clean up up anyway.

> The extension is actually `.ddc.js.metadata`. We don't have control of the extension here as it is decided by DDC / the SDK.

Perfect. We can make sure that the file cleanup targets it that extension specifically.

build_web_compilers/pubspec.yaml
jakemac53: That is a pretty large technical debt burden and not something we are likely to remember to continue doing if we add new files though? We won't be broken or notice since we aren't enabling the experiment...

leafpetersen: Noted.  The --enable-experiment has become more of a public facing thing than I at least had originally anticipated.  Short term, I don't think much we can do about it. Not sure what the right resolution is long term.  I guess one possibility is to require something extra in the pubspec to opt in (at least during the experiment phase) or equivalently use something besides the min sdk version to opt in.  The latter ship feels like it has sailed though.

PR #2741, opened 2020-06-29
Clean up metadata files

grouma:


build_web_compilers/lib/builders.dart
```diff
@@ -69,8 +69,11 @@ PostProcessBuilder dart2jsArchiveExtractor(BuilderOptions options) =>
 // General purpose builders
 PostProcessBuilder dartSourceCleanup(BuilderOptions options) =>
     (options.config['enabled'] as bool ?? false)
-        ? const FileDeletingBuilder(['.dart', '.js.map'])
-        : const FileDeletingBuilder(['.dart', '.js.map'], isEnabled: false);
+        ? const FileDeletingBuilder(
+            ['.dart', '.js.map', '.ddc.js.metadata', '.merged_metadata'])
```
natebosch, line 73: Can this be `.ddc.js.merged_metadata`?

grouma: No. We can change the behavior however (which will require internal and `package:dwds` changes).

natebosch: I'm not quite as concerned as I would be about `.metadata` when it comes to conflicts.

Is this file 1 per package, or is it 1 per entry point?

If we do think it should be changed, now would be the time to do it. Do you have a rough estimate on cost? @jakemac53  do you think that it's OK for us to use such a general name?

grouma: `.metadata` is one per package while `.merged_metadata` is one per entrypoint.

jakemac53: Hmm ya that isn't ideal. I don't know how much work it would actually require to fix though. I don't think it is likely to be a huge issue or anything

grouma: It's minimal. I'll just fix it now.

PR #2743, opened 2020-07-02
remove deprecated analyzer api usage

jakemac53:


jakemac53:
> Do we need to bump the dep?

I decided not to since its only used in tests. We could, but we can't distinguish the version between regular deps and dev deps I don't think?

build_resolvers/test/resolver_test.dart
```diff
@@ -367,8 +367,10 @@ int? get x => 1;
                 ''',
               }, (resolver) async {
                 var lib = await resolver.libraryFor(entryPoint);
-                expect(lib.languageVersionMajor, sdkLanguageVersion.major);
-                expect(lib.languageVersionMinor, sdkLanguageVersion.minor);
+                expect(lib.languageVersion.effective.major,
+                    sdkLanguageVersion.major);
```
kevmoo, line 371: nit: I'm starting to add trailing commas when I force a wrap in an arg list – do what you like, thoug!

jakemac53: we don't do that, generally

jakemac53:
merged for now, this isn't published so we could still increase the dep in a followup if we think that is necessary

PR #2753, opened 2020-07-13
Change build_web_compilers to dev version

grouma:
Since it now depends on a dev SDK we must release it as a dev version.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.12.0
+## 2.12.0-dev.1
 
 - Update `build_web_compilers|ddc` builder to produce a `.metadata` file.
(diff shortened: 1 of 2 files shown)


jakemac53:
> Since it now depends on a dev SDK we must release it as a dev version.

Is this a warning or error? If it is a warning we can bypass it. If it is an error we should file an issue to turn it into a warning.

We don't want to be blocked from doing any stable releases until the next stable sdk.

grouma:
This is an error: 

`Packages with an SDK constraint on a pre-release of the Dart SDK should themselves be published as a pre-release version. `

grouma:
Looks like the check is actually server side. We'll move forward with this change and publish a non pre-release version when `2.9.0` lands. This should unblock consuming the package in `package:dwds`.

PR #2754, opened 2020-07-13
Format markdown files

natebosch:


natebosch:
I've started getting in the habit of autoformatting my markdown files when I edit them, but I keep having to undo it for these files since it adds a bunch of unrelated noise.

Unfortunately the only markdown formatter I'm aware of (and the one used here) is internal only so I don't think we can enforce this on Travis or anything like that 😢 

I _personally_ think it's nice to get these into a consistent state now (maybe I should do the rest of the repo too, include README files?) even if we can't enforce it, but I'm open to the idea that without enforcement it's an uphill battle that isn't worth fighting.

natebosch:
@grouma thought it might be better not to submit this if we can't enforce it.

I might also look again to see if I can find external tools that work well enough. It looks like prettier might be an option that I missed before.

jakemac53:
I don't feel strongly either way. I would push back if on one of my prs you tried to get me to conform to some formatter I can't reasonably use, but I don't think it hurts to bring things to a consistent state.

natebosch:
@grouma - are you OK with a 1 time format for these?

PR #2756, opened 2020-07-16
Stop using deprecated package:analyzer/analyzer.dart

scheglov:


jakemac53:
We will need to update the pubspec/changelog versions, and possibly analyzer constraints as well.

If your branch allows me to push to it I can do that, although I am not sure what the required analyzer versions are.

jakemac53:
Ok I believe I updated to the correct version, I set it to 0.36.4 based on the changelog entries.

scheglov:
Oh, I see you have already found that `parseString` existed since `0.36.4`.

Issue #2772, opened 2020-08-03
Add back `overrideGeneratedOutputDirectory`


jakemac53:
This was removed in https://github.com/dart-lang/build/pull/2716, but it is used by flutter web (see these failures https://cirrus-ci.com/task/6070672998793216?command=main#L86).

PR #2776, opened 2020-08-03
Remove the export of constants from _test_common

natebosch:
This export is overly broad and masks the places where a test _should_
be getting an import from the public library instead of `src/`. This
makes it more clear which tests are using the internal details.

If this had been the pattern before the issue in #2772 _may_ have been
more obvious. Its a better pattern to avoid this in either case.


jakemac53:
@natebosch ready to merge?

Issue #2774, opened 2020-08-03
Scratch space package config doesn't work for non-package uris


jakemac53:
See the code here https://github.com/dart-lang/build/blob/master/build_modules/lib/src/scratch_space.dart#L94

We need to set up a different root uri for the main package, and the package uri should point at the specific packages/<name> dir.

This is somewhat complicated because we don't actually know the root package I don't think.

PR #2777, opened 2020-08-03
fix null safety support to allow non-package uris, add integration test

jakemac53:
note that we don't support strong mode tests or anything but this still helps validate that any form of null safe tests work

Fixes https://github.com/dart-lang/build/issues/2774


build_modules/lib/src/scratch_space.dart
```diff
@@ -92,8 +94,16 @@ String _scratchSpacePackageConfig(String rootConfig) {
   var packages =
       (parsedRootConfig['packages'] as List).cast<Map<String, dynamic>>();
   for (var package in packages) {
-    package['rootUri'] = '../packages/${package['name']}';
-    package['packageUri'] = '';
+    var rootUri = packageConfigUri.resolve(package['rootUri'] as String);
+    if (rootUri == _currentDirUri) {
```
natebosch, line 98: Would it make sense to track whether we saw any package with this root URI? If it's missing does that indicate a problem?

jakemac53: It does indicate a problem yes, we should see exactly one. Multiple would cause other explicit failures (you can't have identical root uris for mutliple packages).

I could add something in but probably wouldn't put it above a warning.

natebosch: Warning SGTM. My thinking is that if something goes wrong a warning here will be more clear than a strange failure in the compiler.

jakemac53:
Fixing a bunch of other testing related issues now, and adding dart2js support since that was never added (but is trivial)

jakemac53:
Actually, I think it probably makes sense to make this test be more isolated so we don't have to pollute a whole bunch of the integration tests with `--enable-experiment` flags.

build_modules/lib/src/scratch_space.dart
jakemac53: warning added

jakemac53:
PTAL @natebosch I added a new package for null safety testing to keep things more isolated, and added the warning

jakemac53:
Actually it looks like multi-root schemes are not supported by dart2js.... I filed https://github.com/dart-lang/build/pull/2777.

I can't think of a better option here other than to create two separate package config files?

_test_null_safety/test/opted_in_test.dart
```diff
@@ -0,0 +1,13 @@
...
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:test/test.dart';
+
+import 'common/message.dart';
+
+void main() {
+  test('can use null safety', () {
+    expect(message!.isNotEmpty, true);
```
natebosch, line 11: [nit] I know that the intention of the test is to use a `!` but I always get nervous to see non-idiomatic code spread. If this were a real test we'd write it as `expect(message, isNot(isEmpty)`... I wonder if there is a more idiomatic test case we could write that would still include a `!`?

jakemac53: 🤷‍♂️ I don't think it really matters. I could change this to an assert.

jakemac53:
This needs some additional support in dart2js to work with windows, see https://dart-review.googlesource.com/c/sdk/+/158900

jakemac53:
The currently failing tests do pass locally on my windows box :/. I am hoping that this is just not quite picking up the latest dev release for some reason... I can't tell what version it actually got.

jakemac53:
huh... I was able to find the dart version and it does confirm it is getting the expected version..... I am pretty stumped.

natebosch:
Is this something we should be pushing on?

jakemac53:
I do need to pick this back up and continue pushing on it yes

PR #2779, opened 2020-08-03
Try out using msgpack for asset graph

natebosch:
Towards #41

See if this makes a meaningful difference for performance.


natebosch:
We typically prefer to avoid taking on a dependency on third party packages from our core tools. We need to decide between accepting a dep, building our own msgpack implementation, or maintaining the use of json...

jakemac53:
We should also do more investigation into where the slowness is happening here as well - if it is simply the writing of the file to disk we could for instance gzip it.

PR #2782, opened 2020-08-07
add some null safety docs for builders

jakemac53:
I am starting a doc/guide for builder authors, outlining how decisions should be made in terms of what level of null safety support to add, how to write tests, etc. 

Closes https://github.com/dart-lang/build/issues/2658

### DIFF
--- a/docs/null_safety_migration.md
+++ b/docs/null_safety_migration.md
@@ -1,0 +1,166 @@
+# Migrating a Builder to support null safety
+
+The null safety feature in Dart presents several challenges for Builder
+authors, this is a living document which will address some of the core issues,
+as well as potential solutions.
+
+## Language versioning
+
+Language versioning is a new feature introduced to allow an incremental
+migration to null safety. It allows an individual package or even library
+to set its "language version". This determines what features the library is
+able to use.
+
+How this affects your Builder depends on what type of builder you have
+implemented, you may choose to only generate null safe code or you may want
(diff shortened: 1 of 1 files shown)


docs/null_safety_migration.md
```diff
@@ -0,0 +1,166 @@
+# Migrating a Builder to support null safety
+
+The null safety feature in Dart presents several challenges for Builder
+authors, this is a living document which will address some of the core issues,
+as well as potential solutions.
+
+## Language versioning
+
+Language versioning is a new feature introduced to allow an incremental
```
natebosch, line 9: Are there other language versioning docs we should link to?

jakemac53: cc @leafpetersen do we have a canonical doc?

natebosch: Perhaps not yet. https://github.com/dart-lang/site-www/issues/2554

leafpetersen: There is a feature specification for this in the language repository, but I don't know about anything broader than that.

jakemac53: Ya I don't really want to link to the full spec here, although we could. I am also concerned it will move around and don't want to have to keep the link up to date (it is currently under `accepted/future-releases`).

jakemac53: Going to merge as is for now and we can add a more canonical user facing link when we have it.

Issue #2723, opened 2020-06-10
Support building/running in strong or weak null safety mode.


jakemac53:
- [ ] Each library has to be compiled separately for each of these modes
  - Need separate builders, with different file extensions but we can likely re-use the module files
  - When creating libraries we need to pass the correct sdk dill outline file
- [ ] The entrypoint builders will determine the mode (strong/weak) based on the language version, and will require the correct corresponding JS files (which in turn require the correct dill files)
- [ ] We want to compile the sdk on our own to avoid shipping duplicate JS files with the SDK
  - `dartdevc` should expose a command that takes an sdk dill + module type and compiles it (for the sdk)
    - we need to create these on demand (using a builder) based on which mode apps are running in, and cache the file

franklinyow:
Is this something we need for track for tech preview 2?

jakemac53:
This should not affect flutter, so no I don't believe so.

nshahan:
AFAIK once https://github.com/dart-lang/sdk/commit/9460ac3c8daaac958147b1b7d98afb381cb74cd5 is included in a dev release all the parts of this issue should be unblocked.

jakemac53:
Sounds good, I would like to wait to support this until the feature lands in a dev release (no experiment flag is required).

Otherwise the detection of whether a library is "null safe" or not will have to be continually updated etc.

natebosch:
> Otherwise the detection of whether a library is "null safe" or not will have to be continually updated etc.

I don't understand this - do you mean that you hope to avoid adopting the extra complexity that comes from this idea of there being 2 meanings of the `2.9` language version?

natebosch:
@nshahan - I talked with Jake offline about this. I'm fine taking this on once we have a published 2.10 dev SDK to try with.

jakemac53:
I am going to work on this in a separate branch and send multiple PRs to that, merging all at once when it is done https://github.com/dart-lang/build/tree/sound-null-safety-ddc

PR #2791, opened 2020-08-18
Build the sdk JS files from dill - including both the sound and unsound versions

jakemac53:
Part of https://github.com/dart-lang/build/issues/2723

This will later be improved to actually load the correct one (and only build that one).

### DIFF
--- a/build_modules/lib/src/workers.dart
+++ b/build_modules/lib/src/workers.dart
@@ -62,7 +62,6 @@ BazelWorkerDriver get _dartdevkDriver {
           p.join(sdkDir, 'bin', 'dart'),
           [
             p.join(sdkDir, 'bin', 'snapshots', 'dartdevc.dart.snapshot'),
-            '--kernel',
             '--persistent_worker'
           ],
           mode: _processMode,
--- a/build_web_compilers/lib/builders.dart
+++ b/build_web_compilers/lib/builders.dart
@@ -56,10 +56,12 @@ Builder ddcKernelBuilder(BuilderOptions options) {
 Builder sdkJsCopyRequirejs(_) => SdkJsCopyBuilder();
 Builder sdkJsCompileSound(_) => SdkJsCompileBuilder(
     sdkKernelPath: 'lib/_internal/ddc_platform_sound.dill',
-    outputPath: 'lib/src/dev_compiler/dart_sdk.sound.js');
+    outputPath: 'lib/src/dev_compiler/dart_sdk.sound.js',
+    soundNullSafety: true);
 Builder sdkJsCompileUnsound(_) => SdkJsCompileBuilder(
     sdkKernelPath: 'lib/_internal/ddc_platform.dill',
-    outputPath: 'lib/src/dev_compiler/dart_sdk.js');
+    outputPath: 'lib/src/dev_compiler/dart_sdk.js',
+    soundNullSafety: false);
 
 // Dart2js related builders
 Builder dart2jsMetaModuleBuilder(BuilderOptions options) =>
--- a/build_web_compilers/lib/builders.dart
+++ b/build_web_compilers/lib/builders.dart
@@ -51,11 +51,8 @@ Builder ddcKernelBuilder(BuilderOptions options) {
       experiments: _readExperimentOption(options));
 }
 
+// Copies the `require.js` file from the sdk into this package.
 Builder sdkJsCopyBuilder(_) => SdkJsCopyBuilder();
-PostProcessBuilder sdkJsCleanupBuilder(BuilderOptions options) =>
-    FileDeletingBuilder(
-        ['lib/src/dev_compiler/dart_sdk.js', 'lib/src/dev_compiler/require.js'],
-        isEnabled: options.config['enabled'] as bool ?? false);
 
 // Dart2js related builders
 Builder dart2jsMetaModuleBuilder(BuilderOptions options) =>
--- a/build_web_compilers/lib/builders.dart
+++ b/build_web_compilers/lib/builders.dart
@@ -11,6 +11,7 @@ import 'package:path/path.dart' as p;
 import 'build_web_compilers.dart';
 import 'src/common.dart';
 import 'src/platforms.dart';
+import 'src/sdk_js_compile_builder.dart';
 import 'src/sdk_js_copy_builder.dart';
 
 // Shared entrypoint builder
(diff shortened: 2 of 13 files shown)


build_modules/lib/src/workers.dart
```diff
@@ -62,7 +62,6 @@ BazelWorkerDriver get _dartdevkDriver {
           p.join(sdkDir, 'bin', 'dart'),
           [
             p.join(sdkDir, 'bin', 'snapshots', 'dartdevc.dart.snapshot'),
-            '--kernel',
```
natebosch, line 65: Can/should this go out separately?

build_web_compilers/lib/builders.dart
```diff
@@ -51,11 +52,16 @@ Builder ddcKernelBuilder(BuilderOptions options) {
       experiments: _readExperimentOption(options));
 }
 
-Builder sdkJsCopyBuilder(_) => SdkJsCopyBuilder();
-PostProcessBuilder sdkJsCleanupBuilder(BuilderOptions options) =>
-    FileDeletingBuilder(
-        ['lib/src/dev_compiler/dart_sdk.js', 'lib/src/dev_compiler/require.js'],
-        isEnabled: options.config['enabled'] as bool ?? false);
+// Copies the `require.js` file from the sdk into this package.
```
natebosch, line 55: [nit] Is this comment necessary? Should it be a doc comment? Should it mention the other JS files?

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,130 @@
...
+  AssetId jsOutputId,
+  bool soundNullSafety,
+) async {
+  var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
+  var jsOutputFile = scratchSpace.fileFor(jsOutputId);
+
+  var request = WorkRequest()
+    ..arguments.addAll([
+      '--no-summarize',
+      '--module-name=dart-sdk',
```
nshahan, line 87: I don't use the module-name flag when building inside the sdk repo and just accept the default. Is there an important dependency on `dart-sdk` vs `dart_sdk` that I should be aware of?

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
```
natebosch, line 1: [nit] 2020

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,130 @@
...
+
+import 'common.dart';
+import 'dev_compiler_builder.dart' show fixSourceMapSources;
+import 'errors.dart';
+
+const _jsSourceMapExtension = '.js.map';
+
+/// A builder which can compile the SDK to JS from kernel.
+class SdkJsCompileBuilder implements Builder {
+  /// The sdk kernel file for the current platform.
```
natebosch, line 23: [nit] 
```suggestion
  /// The SDK kernel file for the current platform.
```

natebosch, line 23: What does "current platform" mean here? Is this sound vs unsound? Maybe we can mention that in the overall class doc comment?

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,130 @@
...
+  var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
+  var jsOutputFile = scratchSpace.fileFor(jsOutputId);
+
+  var request = WorkRequest()
+    ..arguments.addAll([
+      '--no-summarize',
+      '--module-name=dart-sdk',
+      '--multi-root-scheme=org-dartlang-sdk',
+      '--modules=amd',
+      '--${soundNullSafety ? '' : 'no-'}sound-null-safety',
```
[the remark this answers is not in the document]
jakemac53: Right it is not necessary for the SDK given the allow list

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,130 @@
...
+  bool soundNullSafety,
+) async {
+  var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
+  var jsOutputFile = scratchSpace.fileFor(jsOutputId);
+
+  var request = WorkRequest()
+    ..arguments.addAll([
+      '--no-summarize',
+      '--module-name=dart-sdk',
+      '--multi-root-scheme=org-dartlang-sdk',
```
natebosch, line 87: I'm a little surprised we need to specify this. Do you know why?

build_web_compilers/lib/src/sdk_js_compile_builder.dart
nshahan: Oh and actually the codepath that builds the sdk from dill accepts the flag but never uses the value. So if it is important I need to actually support it.

jakemac53: Oh actually this should have almost certainly been `dart_sdk` haha - the fact that this didn't break anything for me probably means it isn't important.

I don't know what exactly this does apparently 😆

jakemac53: I removed this flag

build_web_compilers/lib/src/sdk_js_compile_builder.dart
```diff
@@ -0,0 +1,129 @@
...
+  String librariesPath,
+  AssetId jsOutputId,
+  bool soundNullSafety,
+) async {
+  var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
+  var jsOutputFile = scratchSpace.fileFor(jsOutputId);
+
+  var request = WorkRequest()
+    ..arguments.addAll([
+      '--no-summarize',
```
nshahan, line 86: It's the same story with `--no-summarize` as well, it's accepted but never read. The summary is not being created by default since in the sdk build we create multiple output js module formats. We don't want copies of the same summary .dill for every output and never added a mechanism to specify the summary output path, so we just don't write the file.

Overall this seems totally confusing and it looks like we should cleanup the arguments that are accepted when compiling the sdk from dill.

build_web_compilers/lib/src/sdk_js_compile_builder.dart
nshahan: I'm not 100% sure but I believe this is used in source mapping where we want a uniform root for the sdk in sourcemaps and stack traces.

build_modules/lib/src/workers.dart
jakemac53: This is still supported generally, but the entrypoint which compiles the entire SDK does not support it. There isn't really any value in backporting this.

build_web_compilers/lib/src/sdk_js_compile_builder.dart
jakemac53: Ya I don't know exactly, the dill file has this scheme in it and if I don't provide the same scheme it complains

build_web_compilers/lib/builders.dart
jakemac53: I just removed it

build_web_compilers/lib/src/sdk_js_compile_builder.dart
jakemac53: removed

PR #2797, opened 2020-08-26
Throw when resolving files with syntax errors

### LINKED ISSUE #2624: Should we automatically detect syntax errors in `.dart` files?
A builder which has a `.dart` file as a primary input will almost certainly be unable to do anything useful if there are syntax errors (errors that would make `dartfmt` fail, don't need type resolution). Even if it could do something, later parts of the build will almost certainly fail. As builder authors look to harden their behavior around errors to avoid stack traces or confusing errors we should consider handling this case of errors for them since it is universal.

simolus3:
I added a new `SyntaxErrorInAssetException` that will be thrown when attempting to use `libraryFor` on an asset id that contains syntax errors. The `allowSyntaxErrors` flag can be used to opt-out of this new behavior. This fixes #2624.

One potential UX issue that came to mind is how this behaves when using numerous builders that resolve everything (not uncommon with `source_gen`). In those cases, we'd print the same error many times if some file contains syntax errors. Maybe the `source_gen` package could keep track of files with syntax errors to avoid duplicate logs then.

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.4.0
+
+The resolver will now throw an `SyntaxErrorInAssetException` when attempting to
+resolve an asset with syntax errors. You can opt-out of this new behavior by
+passing `allowSyntaxErrors: true` to `libraryFor`.
+
 ## 1.3.0
 
 Added a new experiments library which exposes a list of language experiments
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build
-version: 1.4.0-dev
+version: 1.4.0
 description: A build system for Dart.
 homepage: https://github.com/dart-lang/build/tree/master/build
 
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -23,3 +23,11 @@ dev_dependencies:
   build_vm_compilers: ">=0.1.0 <2.0.0"
   pedantic: ^1.0.0
   test: ^1.2.0
+
+dependency_overrides:
+  build_resolvers:
+    path: ../build_resolvers
+  build_runner:
+    path: ../build_runner
+  build_runner_core:
+    path: ../build_runner_core
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build
-version: 1.3.0
+version: 1.4.0-dev
 description: A build system for Dart.
 homepage: https://github.com/dart-lang/build/tree/master/build
 
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -15,6 +15,8 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/file_system/physical_file_system.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:analyzer/src/dart/analysis/driver.dart' show AnalysisDriver;
+import 'package:analyzer/src/dart/error/syntactic_errors.dart'
+    show ParserErrorCode;
 import 'package:analyzer/src/generated/engine.dart'
     show AnalysisOptions, AnalysisOptionsImpl;
 import 'package:build/build.dart';
@@ -86,10 +88,11 @@ class PerActionResolver implements ReleasableResolver {
   }
 
   @override
-  Future<LibraryElement> libraryFor(AssetId assetId) async {
+  Future<LibraryElement> libraryFor(AssetId assetId,
+      {bool allowSyntaxErrors = false}) async {
     if (!await _step.canRead(assetId)) throw AssetNotFoundException(assetId);
     await _resolveIfNecesssary(assetId);
-    return _delegate.libraryFor(assetId);
+    return _delegate.libraryFor(assetId, allowSyntaxErrors: allowSyntaxErrors);
   }
 
   Future<void> _resolveIfNecesssary(AssetId id) async {
@@ -129,7 +132,8 @@ class AnalyzerResolver implements ReleasableResolver {
   }
 
   @override
-  Future<LibraryElement> libraryFor(AssetId assetId) async {
+  Future<LibraryElement> libraryFor(AssetId assetId,
+      {bool allowSyntaxErrors = false}) async {
     var path = assetPath(assetId);
     var uri = assetId.uri;
     var source = _driver.sourceFactory.forUri2(uri);
@@ -138,7 +142,18 @@ class AnalyzerResolver implements ReleasableResolver {
     }
     var kind = await _driver.getSourceKind(path);
     if (kind != SourceKind.LIBRARY) throw NonLibraryAssetException(assetId);
-    return _driver.getLibraryByUri(assetId.uri.toString());
+
+    if (!allowSyntaxErrors) {
+      final result = await _driver.getErrors(path);
+
+      final parserErrors =
+          result.errors.where((e) => e.errorCode is ParserErrorCode).toList();
+      if (parserErrors.isNotEmpty) {
+        throw SyntaxErrorInAssetException(assetId, parserErrors);
+      }
+    }
+
+    return _driver.getLibraryByUri(uri.toString());
   }
 
   @override
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -8,15 +8,15 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:isolate';
 
+import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/sdk/build_sdk_summary.dart';
+import 'package:analyzer/error/error.dart';
 import 'package:analyzer/src/dart/analysis/experiments.dart';
 import 'package:analyzer/dart/analysis/features.dart';
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/file_system/physical_file_system.dart';
 import 'package:analyzer/src/generated/source.dart';
 import 'package:analyzer/src/dart/analysis/driver.dart' show AnalysisDriver;
-import 'package:analyzer/src/dart/error/syntactic_errors.dart'
-    show ParserErrorCode;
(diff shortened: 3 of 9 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -138,7 +142,18 @@ class AnalyzerResolver implements ReleasableResolver {
     }
     var kind = await _driver.getSourceKind(path);
     if (kind != SourceKind.LIBRARY) throw NonLibraryAssetException(assetId);
-    return _driver.getLibraryByUri(assetId.uri.toString());
+
+    if (!allowSyntaxErrors) {
+      final result = await _driver.getErrors(path);
```
jakemac53, line 147: Here is what we have internally, the main difference is it does support part files:

```dart
  @override
  Future<LibraryElement> libraryFor(AssetId assetId) async {
    ...
    var element = await _analysisDriver.getLibraryByUri(assetId.uri.toString());
    // Fail resolving if the library contains syntactic errors. This is to
    // prevent running code generators on syntactically invalid Dart inputs.
    var errorToResult = await _syntacticErrorsFor(element);
    if (errorToResult.isNotEmpty) {
      throw InvalidDartSyntax(errorToResult);
    }
    return element;
  }

  /// Returns a map of any syntactic errors to per-file results in [element].
  Future<Map<AnalysisError, ErrorsResult>> _syntacticErrorsFor(
      LibraryElement element) async {
    // TODO(b/152322246): We can't get errors for summarized sources.
    if (element.source is BasicSource) return {};
    var paths = await Future.wait([
      _pathForElement(element),
      for (var part in element.parts)
        // The source may be null if the part doesn't exist. This may be valid
        // as the part could be produced by a code generator.
        if (part.source != null)
          _pathForElement(part)
    ]);
    var errorToResult = <AnalysisError, ErrorsResult>{};
    for (var path in paths) {
      var result = await _analysisDriver.getErrors(path);
      if (result != null) {
        for (var error in result.errors) {
          if (error.errorCode.type == ErrorType.SYNTACTIC_ERROR) {
            errorToResult[error] = result;
          }
        }
      }
    }
    return errorToResult;
  }

  Future<String> _pathForElement(Element element) =>
      assetIdForElement(element).then(assetPath);
```

The exception class we have there is as follows:

```dart
import 'package:analyzer/error/error.dart';
import 'package:analyzer/dart/analysis/results.dart';

/// Thrown when resolving a Dart library with syntactic errors.
class InvalidDartSyntax implements Exception {
  InvalidDartSyntax(this.errorToResult)
      : assert(errorToResult.keys
            .every((e) => e.errorCode.type == ErrorType.SYNTACTIC_ERROR));

  /// A map from analysis error to per-file results.
  final Map<AnalysisError, ErrorsResult> errorToResult;

  @override
  String toString() => errorToResult.keys.join('\n');
}
```

It would be good to fully unify these in terms of the exception thrown at least, so we want the exception in `package:build` as you have done here, but I do think we want to keep the part file support like we have internally.  

cc @leonsenft assuming this lands is angular catching this exception today or would changing the location/name of it break anything you are aware of?

jakemac53: Also the toString in this implementation is a lot more descriptive than what we have internally right now - would that cause any problems? (are we relying on other build steps for the more descriptive output and we would now be duplicating it possibly?)

leonsenft: Thanks, Jake. One quick note, I think you copied this from an outdated client, because it looks like an older version with a bug I later fixed (using `assetIdForElement` could throw in some cases). The updated code looks like: 

```dart
  /// Returns a map of any syntactic errors to per-file results in [element].
  Future<Map<AnalysisError, ErrorsResult>> _syntacticErrorsFor(
      LibraryElement element) async {
    var paths = _assetPathsFor(element);
    var errorToResult = <AnalysisError, ErrorsResult>{};
    for (var path in paths) {
      var result = await _analysisDriver.getErrors(path);
      if (result != null) {
        for (var error in result.errors) {
          if (error.errorCode.type == ErrorType.SYNTACTIC_ERROR) {
            errorToResult[error] = result;
          }
        }
      }
    }
    return errorToResult;
  }

  /// Returns the cached asset paths for [element] and its parts.
  ///
  /// Returns an empty iterable if [element] is backed by a summary.
  Iterable<String> _assetPathsFor(LibraryElement element) {
    var elements = [
      element,
      // The source may be null if the part doesn't exist. This may be valid as
      // the part could be produced by a code generator.
      for (var part in element.parts)
        if (part.source != null) part,
    ];
    return elements
        .map((e) => _assetResolver.lookupAsset(e.source.uri))
        .whereType<AssetId>() // Filter out null.
        .map(assetPath);
  }
```

> cc @leonsenft assuming this lands is angular catching this exception today or would changing the location/name of it break anything you are aware of?

Angular is not relying on this exception at the moment; however, as I mentioned in https://github.com/dart-lang/build/issues/2624#issuecomment-680322041 we probably will start to in some cases so we don't report them as "unhandled" exceptions and instead are treated like proper builder failures. However, we should implement a base failure type to catch and rethrow instead of this specific exception: https://github.com/dart-lang/build/issues/981.

> Also the toString in this implementation is a lot more descriptive than what we have internally right now - would that cause any problems?

No. I don't believe we rely on the `toString()` implementation internally _at all_. The reason it's so sparse (and unused) internally is because we catch these errors in the build runner in order to report them with a specific format. This is necessary because the source URLs returned from `AnalysisDriver.getErrors()` need to be mapped back to the source location using the package map accessible only from the build runner.

simolus3: I agree that considering syntax errors from existing part files is necessary here, but I'm not sure why the error has to be a `Map<AnalysisError, ErrorsResult>` (as opposed to a `List<AnalysisError>` from multiple files). Do your internal builders use the associated path or the `AnalysisSession` from the error result?  Does it make sense to expose a `Map<AnalysisError, AssetId>` instead?

leonsenft: > I agree that considering syntax errors from existing part files is necessary here

Another subtle point that wasn't mentioned is the way this code checks for errors will also include `ScannerErrorCode`s which are another type of syntactic error.

>  Do your internal builders use the associated path or the AnalysisSession from the error result? 

Close, we can still get the associated source from the `AnalysisError` using [source](https://github.com/dart-lang/sdk/blob/52f1502191e8668e3ec216eb52cedc47e574c334/pkg/analyzer/lib/error/error.dart#L878). What we need from the `ErrorsResult` is the [lineInfo](https://github.com/dart-lang/sdk/blob/52f1502191e8668e3ec216eb52cedc47e574c334/pkg/analyzer/lib/dart/analysis/results.dart#L73) which we use to group and report errors in a custom format. 

I believe the issue with relying on the `AnaylsisError`'s own `toString()` is that it uses character offsets, instead of column and line numbers as seen [here](https://github.com/dart-lang/sdk/blob/52f1502191e8668e3ec216eb52cedc47e574c334/pkg/analyzer/lib/error/error.dart#L992-L997). Using the lineInfo, we're able to construct the proper column and line numbers.

Admittedly in hindsight, filtering the syntax errors into a map in this manner wasn't super clear. We could have just returned any `ErrorsResult`s that contain a syntax error, and then filter them again when reporting them?

I'll let @jakemac53 weigh in here, but it might not be worth worrying too much about the compatibility at first, since we have an entirely different implementation of the resolver internally. I think we can deal with merging the approaches afterwards?

leonsenft: Note that the analyzer CLI itself doesn't even rely on `AnalysisError.toString()`. They format the errors in a similar manner and rely on using the lineInfo as well: https://github.com/dart-lang/sdk/blob/f08c0e36c6dfed4d44029156818e0a649e45469a/pkg/analyzer_cli/lib/src/error_formatter.dart#L283.

jakemac53: I don't have strong opinions here, but we should try to make sure that the shared exception type that we expose from package:build has all the required information that we would need both internally and externally so that we can ultimately unify things.

simolus3: I changed it to a `List<ErrorResult>` and added a getter compatible to what you're using internally. I also changed the format to use the location with line and column number.

leonsenft: This looks sufficient for our needs internally, thanks!

jakemac53:
Sorry that this one has been left stagnant for so long - taking another look now.

build/pubspec.yaml
```diff
@@ -23,3 +23,11 @@ dev_dependencies:
...
   pedantic: ^1.0.0
   test: ^1.2.0
+
+dependency_overrides:
+  build_resolvers:
+    path: ../build_resolvers
+  build_runner:
+    path: ../build_runner
+  build_runner_core:
+    path: ../build_runner_core
```
jakemac53, line 33:
```suggestion
    path: ../build_runner_core

```

build/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build
-version: 1.3.0
+version: 1.4.0-dev
```
jakemac53, line 2: I think we can go ahead and remove the `-dev` here, we can publish once this is landed.

build_resolvers/pubspec.yaml
```diff
@@ -22,3 +22,7 @@ dev_dependencies:
   build_runner: ^1.0.0
   build_vm_compilers: ">=0.1.0 <2.0.0"
   build_modules: any
+
+dependency_overrides:
+  build:
+    path: ../build
```
jakemac53, line 28:
```suggestion
    path: ../build

```

jakemac53 (approved):
LGTM with just a few suggestions

build/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.4.0
+
+The resolver will now throw an `SyntaxErrorInAssetException` when attempting to
```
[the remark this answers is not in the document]
natebosch: I think it's very unlikely to break anyone in practice and the cost of churn probably isn't worth it.

natebosch (approved):

build/pubspec.yaml
simolus3: I also removed the `-dev` suffix in the `build_resolvers` package.

PR #2800, opened 2020-08-28
Make build_web_compilers output full kernel in debug mode

annagrin:
Make build_web_compilers output full kernel files and fixup metadata paths, in preparation for expression 
evaluation feature in webdev:

- make build_web_compilers output full kernel files in debug mode
  This is preparation work for expression evaluation in webdev.
  depends on the following SDK change:
   https://dart-review.googlesource.com/c/sdk/+/160944

- update paths inside metadata to match buildstep copying them to a final
  location

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_builder.dart
+++ b/build_web_compilers/lib/src/dev_compiler_builder.dart
@@ -181,9 +181,9 @@ Future<void> _createDevCompilerModule(
     ..arguments.addAll([
       '--dart-sdk-summary=$sdkSummary',
       '--modules=amd',
-      // output full dill
       '--no-summarize',
-      '--experimental-output-compiled-kernel',
+      if (generateFullDill)
+        '--experimental-output-compiled-kernel',
       '-o',
       jsOutputFile.path,
       debugMode ? '--source-map' : '--no-source-map',
--- a/build_web_compilers/lib/src/dev_compiler_builder.dart
+++ b/build_web_compilers/lib/src/dev_compiler_builder.dart
@@ -21,6 +21,7 @@ const jsModuleErrorsExtension = '.ddc.js.errors';
 const jsModuleExtension = '.ddc.js';
 const jsSourceMapExtension = '.ddc.js.map';
 const metadataExtension = '.ddc.js.metadata';
+const fullKernelExtension = '.ddc.full.dill';
 
 /// A builder which can output ddc modules!
 class DevCompilerBuilder implements Builder {
@@ -73,6 +74,7 @@ class DevCompilerBuilder implements Builder {
             jsModuleErrorsExtension,
             jsSourceMapExtension,
             metadataExtension,
+            fullKernelExtension,
           ],
         },
         environment = environment ?? {},
@@ -130,9 +132,6 @@ Future<void> _createDevCompilerModule(
     Map<String, String> environment,
     Iterable<String> experiments,
     {bool debugMode = true}) async {
-
-  var currentKernelId = module.primarySource.changeExtension(ddcKernelExtension);
-  await buildStep.canRead(currentKernelId);
   var transitiveDeps = await buildStep.trackStage('CollectTransitiveDeps',
       () => module.computeTransitiveDependencies(buildStep));
   var transitiveKernelDeps = [
@@ -145,7 +144,6 @@ Future<void> _createDevCompilerModule(
   await buildStep.trackStage(
       'EnsureAssets', () => scratchSpace.ensureAssets(allAssetIds, buildStep));
   var jsId = module.primarySource.changeExtension(jsModuleExtension);
-  var metadataId = module.primarySource.changeExtension(metadataExtension);
   var jsOutputFile = scratchSpace.fileFor(jsId);
   var sdkSummary =
       p.url.join(dartSdk, sdkKernelPath ?? 'lib/_internal/ddc_sdk.dill');
@@ -169,7 +167,9 @@ Future<void> _createDevCompilerModule(
     ..arguments.addAll([
       '--dart-sdk-summary=$sdkSummary',
       '--modules=amd',
+      // output full dill
       '--no-summarize',
+      '--experimental-output-compiled-kernel',
       '-o',
       jsOutputFile.path,
       debugMode ? '--source-map' : '--no-source-map',
@@ -229,10 +229,14 @@ Future<void> _createDevCompilerModule(
     }
 
     // Copy the output back using the buildStep.
-    // await scratchSpace.copyOutput(currentKernelId, buildStep);
     await scratchSpace.copyOutput(jsId, buildStep);
-    // await scratchSpace.copyOutput(metadataId, buildStep);
+
     if (debugMode) {
+      // copy and current full kernel
+      var currentFullKernelId =
+          module.primarySource.changeExtension(fullKernelExtension);
+      await scratchSpace.copyOutput(currentFullKernelId, buildStep);
+
       // We need to modify the sources in the sourcemap to remove the custom
       // `multiRootScheme` that we use.
       var sourceMapId =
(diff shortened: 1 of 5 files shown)


build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -227,8 +230,13 @@ Future<void> _createDevCompilerModule(
 
     // Copy the output back using the buildStep.
     await scratchSpace.copyOutput(jsId, buildStep);
-    await scratchSpace.copyOutput(metadataId, buildStep);
+
     if (debugMode) {
+      // copy current full kernel
```
jakemac53, line 235: nit: unnecessary comment

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -238,6 +246,15 @@ Future<void> _createDevCompilerModule(
       var json = jsonDecode(content);
       json['sources'] = fixSourceMapSources((json['sources'] as List).cast());
       await buildStep.writeAsString(sourceMapId, jsonEncode(json));
+
+      // remove temp directories from metadata
```
jakemac53, line 250: nit: I would be a bit more clear here:

```dart
// Copy the metadata output, modifying its contents to remove the temp directory from paths
```

Also, I would expect to see the `multi-root` scheme or something here, the metadata files might need to be updated - this will be harder to clean up in blaze I think (there isn't just one possible dir to strip out).

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -295,3 +312,12 @@ String ddcModuleName(AssetId jsId) {
       : jsId.path;
   return jsPath.substring(0, jsPath.length - jsModuleExtension.length);
 }
+
+void fixupMetadata(Map<String, dynamic> json, String scratchPath, String path) {
```
jakemac53, line 316: It doesn't look like the `path` arg is ever passed anything but `''` here, might as well remove it.

Also i would make this a private method.

natebosch:
> We should probably be hiding this behavior behind some special flag, at least for now.

The downside to a flag is that we'll have to remove it when we want this to be the default.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: Is the same code running in blaze?

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: will do, thanks!

build_web_compilers/lib/src/dev_compiler_builder.dart
natebosch: No `build_web_compilers` is external only. The blaze integration uses custom starlark build rules, it doesn't go through the `Builder` interface.

annagrin:
> You'll need to update the min SDK and update the expects in tests as they will now fail with the new files.

Still working on my SDK changes, will update as soon as the PR lands.

build_web_compilers/lib/src/dev_compiler_builder.dart
jakemac53: Right, I just mean that if you are having to write this here it will be harder to replicate there.

jakemac53:
> The downside to a flag is that we'll have to remove it when we want this to be the default.

I am not sure this should ever be the default though - we potentially only want to enable it conditionally? Or at least have the option to disable I think.

natebosch:
> I am not sure this should ever be the default though - we potentially only want to enable it conditionally? Or at least have the option to disable I think.

I'm not sure what that would look like or when we'd tell people to use it? Do you imagine that if folks complain about memory use or performance we'd tell them to pass an option but if they do they lose expression evaluation?

I'd think by default we would want expression evaluation to work.

grouma:
> I'm not sure what that would look like or when we'd tell people to use it? Do you imagine that if folks complain about memory use or performance we'd tell them to pass an option but if they do they lose expression evaluation?

Do we anticipate this having much of an impact on performance? My understanding is that DDC already computes this information already and it is simply dumping the information to an additional file. Shouldn't that have minimal impact?

jakemac53:
> Do we anticipate this having much of an impact on performance? My understanding is that DDC already computes this information already and it is simply dumping the information to an additional file. Shouldn't that have minimal impact?

We end up serializing it to a file once in DDC, and then copying it back inside the Builder, so we write the file at least twice. Then there is the complication that we also process the file to md5 hash it, and we cache it in memory.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: I need to change the ddc code to generate metadata with a multiroot scheme, currently it has full paths inside, which is not what we what for part of build output. 

Created an issue to modify metadata: https://github.com/dart-lang/sdk/issues/43309
Created an issue to update the code according to the change in build_web_compilers: https://github.com/dart-lang/build/issues/2802

annagrin:
@natebosch re:flag: 

We can follow the same strategy we did in flutter - hide the functionality under a flag, at first disabled by default. Experiment, dogfood, and fix until the feature is stable enough.  

After that, make the flag enabled by default  but do not remove the flag so the users have the option to disable the feature if causes problems. We can eventually remove the flag when the feature becomes super stable, if we prefer to have less flags in the system.

Does this sound reasonable?

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -21,11 +21,22 @@ const jsModuleErrorsExtension = '.ddc.js.errors';
...
   final bool useIncrementalCompiler;
 
+  /// Make ddc generate full dill for libraries it compiles.
+  ///
+  /// Full dill is only generated in debug mode and does not include any
+  /// dependent libraries or SDK dill.
+  ///
+  /// Full dill is used by the expression compilation service run by the
+  /// debugger to compile expressions in debugging worklows that use modular
+  ///  build, such as webdev and google3.
```
grouma, line 37: This package isn't used inside Google3 at all so we can probably leave out this extra bit of detail.

grouma (approved):
LGTM. Let's wait for approval from @natebosch or @jakemac53 first though.

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -21,11 +21,22 @@ const jsModuleErrorsExtension = '.ddc.js.errors';
 const jsModuleExtension = '.ddc.js';
 const jsSourceMapExtension = '.ddc.js.map';
 const metadataExtension = '.ddc.js.metadata';
+const fullKernelExtension = '.ddc.full.dill';
 
 /// A builder which can output ddc modules!
 class DevCompilerBuilder implements Builder {
   final bool useIncrementalCompiler;
 
+  /// Make ddc generate full dill for libraries it compiles.
```
natebosch, line 30: [nit] Should use a noun phrase.

```suggestion
  /// Whether to generate full dill file outputs for each module.
```

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -21,11 +21,22 @@ const jsModuleErrorsExtension = '.ddc.js.errors';
...
+  ///
+  /// Full dill is only generated in debug mode and does not include any
```
natebosch, line 32: This comment about "debug mode" feels misplaced. At best we can say "Full dill is only generated by default in debug mode," but I don't think even that is true, since we haven't updated the `build.yaml` file to set this option by default.

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -21,11 +21,22 @@ const jsModuleErrorsExtension = '.ddc.js.errors';
...
+  /// dependent libraries or SDK dill.
+  ///
+  /// Full dill is used by the expression compilation service run by the
```
natebosch, line 35: I think somewhere in this comment we should mention that it implies an extra file. As currently phrased it feels to me like different content in the same file.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: the comment is no longer true, I will remove it

annagrin: I got confused by debug mode - apparently debug mode settings is always on in my case since ddc is only used in debug mode. I will remove the comment.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: will do

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: thanks, fixed

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: thanks, fixed

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: fixed

jakemac53 (approved):

natebosch (approved):

annagrin:
Thanks everyone!

PR #2801, opened 2020-08-29
remove analyzer override for _test dir

jakemac53:
currently causing travis failures because dart_style requires a newer analyzer


jakemac53:
Go ahead and merge on approval

PR #2805, opened 2020-09-09
Mention applies_builder in docs for runs_before

natebosch:
We were careful to document that the configuration doesn't impact
builder application, but failed to highlight the proper configuration to
achieve that goal.


natebosch:
I'm a little unsure I like the phrasing of "steps are not skipped", but I couldn't think of a more clear way to describe this.

jakemac53 (approved):

grouma (approved):

PR #2806, opened 2020-09-09
prep packages for release

jakemac53:
We need to update all packages to allow the latest `build` release, and I also updated them to the latest `analyzer` so that we can get the sdk/experiment version fixes and lower the sdk constraint in build_runner.

Note that this does mean I will need to publish a new build version (1.4.1) which I created here, and effectively `1.4.0` will be a dead version.


natebosch (changes requested):
I do think we should hold this until we resolve the internal failures. I don't think it should be a blocker, but just in case we decide we want the opposite default or something it would be better to wait.

jakemac53:
> I do think we should hold this until we resolve the internal failures. I don't think it should be a blocker, but just in case we decide we want the opposite default or something it would be better to wait.

The internal failures have been resolved now.

natebosch (approved):

PR #2807, opened 2020-09-09
Sort imports in resolvers.dart

natebosch:
This will make it easier to have a consistent internal patch file.


grouma (approved):

natebosch:
I may end up having to merge on red travis. I'm hoping to unblock an internal sync which will then unblock us publishing: https://github.com/dart-lang/build/pull/2806

natebosch:
Ah, it's still green for this PR 😄

PR #2810, opened 2020-09-10
udpate build_resolvers version back to dev since it can't be published right now

jakemac53:
See https://github.com/dart-lang/build/pull/2806 where I was trying to work things out.

We are in a super bad state right now and I don't have the time to try and resolve it. Given that there is no rush we can circle back to this when 2.10 stable releases.


jakemac53:
cc @simolus3

natebosch (approved):

PR #2817, opened 2020-09-15
Add compilationUnitFor api to Resolver

jakemac53:
- This is a cheap api which does not resolve transitive sources, and only gives you a parsed AST for a single asset
- Updates the `SyntaxErrorInAsset` class to contain a list of `AnalysisResultWithErrors` instead of `ErrorResult`, which is the common base class for errors throws during parsing and resolving (it also has the exact same api surface).
- Also stops the eager resolving of `buildStep.inputId` when creating a `Resolver`, to ensure you can use this api cheaply.
- Also as a side-effect of this the `isLibrary` call is now faster - it does not resolve transitive sources either which it used to.

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,12 @@
+## 1.5.0-dev
+
+Added the `Future<CompilationUnit> compilationUnitFor(AssetId id)` api to the
+`Resolver` class, which returns only the parsed AST for the given asset.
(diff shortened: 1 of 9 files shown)


jakemac53:
Once this lands I will work on the internal impl since at that point I will be able to easily pull it in, but should be trivial.

natebosch (approved):

PR #2818, opened 2020-09-15
Support sound null safety for kernel (build_modules) and ddc (build_web_compilers)

jakemac53:
- Adds sound null safety support to KernelBuilder and DevCompilerBuilder
- Adds entrypoint detection to WebEntrypointBuilder (but not passed to dart2js yet)
  - Do we want to add a way to disable sound null safety? Maybe only on request?
- Enable null safety in the build_web_compilers tests and run them in both sound and unsound modes

Apologies for the messy commit history here, there were some merges from master and things that dirtied it up.

### DIFF
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -74,7 +74,10 @@ class KernelBuilder implements Builder {
   /// Experiments to pass to kernel (as --enable-experiment=<experiment> args).
   final Iterable<String> experiments;
 
-  /// Whether or not strong null safety should be enabled.
+  /// Whether or not sound null safety should be enabled.
+  ///
+  /// All dependencies of a given kernel output must be compiled in the
+  /// same mode as the kernel output itself.
   final bool soundNullSafety;
 
   KernelBuilder(
--- a/build_modules/lib/src/kernel_builder.dart
+++ b/build_modules/lib/src/kernel_builder.dart
@@ -74,11 +74,15 @@ class KernelBuilder implements Builder {
   /// Experiments to pass to kernel (as --enable-experiment=<experiment> args).
   final Iterable<String> experiments;
 
+  /// Whether or not strong null safety should be enabled.
+  final bool soundNullSafety;
+
   KernelBuilder(
       {@required this.platform,
       @required this.summaryOnly,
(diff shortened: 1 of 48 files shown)


jakemac53:
CLA's are fine, these commits already exist in master the history is just borked.

jakemac53:
cc @nshahan

build_web_compilers/lib/src/common.dart
```diff
@@ -18,6 +18,12 @@ final sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
 String defaultAnalysisOptionsArg(ScratchSpace scratchSpace) =>
     '--options=${scratchSpace.fileFor(defaultAnalysisOptionsId).path}';
 
+String sdkDdcKernelPath(bool soundNullSafety) => p.url.join('lib', '_internal',
+    soundNullSafety ? 'ddc_outline_sound.dill' : 'ddc_sdk.dill');
+
+String soundnessExt(bool soundNullSafety) =>
+    soundNullSafety ? '.sound' : '.unsound';
```
nshahan, line 25: This choice doesn't bother me, I just wanted to point out that everywhere else this is referred to as "weak" instead of "unsound".

build_modules/lib/src/kernel_builder.dart
```diff
@@ -74,11 +74,15 @@ class KernelBuilder implements Builder {
   /// Experiments to pass to kernel (as --enable-experiment=<experiment> args).
   final Iterable<String> experiments;
 
+  /// Whether or not strong null safety should be enabled.
```
natebosch, line 77: [nit]
```suggestion
  /// Whether or not sound null safety should be enabled.
```

Can we expand the docs while we are at it? Maybe we should say something about how we need to always use sound kernel files as inputs to build sound outputs, and vice versa?

natebosch (approved):

build_web_compilers/lib/src/common.dart
jakemac53: hmm, `unsound` seems to make a lot more sense lol. cc @natebosch wdyt?

natebosch: I prefer "sound" and "unsound", if we come across places still using "strong" and "weak" we should change them.

PR #2820, opened 2020-09-16
add dart2js sound mode support

jakemac53:
Passes the correct sound or unsound null safety flags based on whether the entrypoint is opted in.

Also adds support for experiments


natebosch (approved):
I expected dart2js would determine this on its own. Does it not?

cc @sigmundch

jakemac53:
Oh right, good question. I will try removing that flag and just passing the experiment.

jakemac53:
It works, updating this to just add experiment support but leaving the tests in to validate it.

natebosch (approved):

PR #2823, opened 2020-09-18
Update to latest analyzer and build packages

jakemac53:
Updates to allow analyzer 0.40.x and build 1.5.x across all packages.

### DIFF
--- a/build_test/test/test_builder_test.dart
+++ b/build_test/test/test_builder_test.dart
@@ -169,13 +169,15 @@ void main() {
           '.dart': ['.fake']
         },
         build: (step, _) async {
-          var lib = await step.inputLibrary;
-          var errors = await lib.session.getErrors(lib.source.fullName);
-          expect(
-              errors.errors.map((e) => e.message),
-              contains(contains(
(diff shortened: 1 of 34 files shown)


build_test/test/test_builder_test.dart
```diff
@@ -169,13 +169,15 @@ void main() {
...
-          var errors = await lib.session.getErrors(lib.source.fullName);
-          expect(
-              errors.errors.map((e) => e.message),
-              contains(contains(
-                  'This requires the \'extension-methods\' language feature to be '
-                  'enabled.')));
+          try {
+            await step.inputLibrary;
+          } on SyntaxErrorInAssetException catch (e) {
+            expect(
```
natebosch, line 175: Not sure if I missed it in a previous review or if this is new, this expectation won't work if it stops throwing.

```dart
expect(step.inputLibrary, throwsA(isA<SyntaxErrorInAssetException>().having(...
```

natebosch (approved):

build_test/test/test_builder_test.dart
jakemac53: Oh right, ill update that

PR #2830, opened 2020-09-22
remove dependency overrides now that versions are published

jakemac53:
Also reverts to running some tests from source, and removes the dev dependencies for running tests with build_runner.

### DIFF
--- a/build_resolvers/pubspec.yaml
+++ b/build_resolvers/pubspec.yaml
@@ -20,3 +20,4 @@ dependencies:
 
 dev_dependencies:
   test: ^1.0.0
+  build_test: ^1.0.0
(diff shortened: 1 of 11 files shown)


build_resolvers/pubspec.yaml
```diff
@@ -20,19 +20,3 @@ dependencies:
 
 dev_dependencies:
   test: ^1.0.0
-  build_test: ^1.0.0
```
natebosch, line 23: [nit] Deleting non-overrides should be mentioned in the commit message.

natebosch (approved):

build_resolvers/pubspec.yaml
jakemac53: will update when I merge

Issue #2787, opened 2020-08-13
Update/remove experiment warnings


jakemac53:
See the meta issue https://github.com/dart-lang/sdk/issues/43031

Given that experiments no longer require the latest language version, and that the analyzer is changing to detect the "current version" from the actual sdk it is provided, we can remove the warnings about trying to use an experiment when the analyzer has a newer language version than the current sdk [here](https://github.com/dart-lang/build/blob/master/build_resolvers/lib/src/resolver.dart#L363) .

leafpetersen:
Is this a blocker for completion of the feature or just a nice to have?  Where are we with this?

jakemac53:
This is just a nice to have, not a blocker.

PR #2840, opened 2020-09-25
update language version check when using experiments

jakemac53:
We no longer require an exact match, only that the sdk version is not ahead of the analyzer version.

Also:

* Prepares to release build_resolvers 1.4.1
* Run all packages that have tests which require experiment flags on 2.10 dev instead of 2.11 dev since analyzer doesn't yet have an analyzer version for 2.11.
* Closes https://github.com/dart-lang/build/issues/2787

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,17 +112,17 @@ jobs:
       env: PKGS="build_modules"
       script: ./tool/travis.sh command_4
     - stage: unit_test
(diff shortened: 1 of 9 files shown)


jakemac53:
cc @scheglov

natebosch (approved):

jakemac53:
Hmm, I am going to end up having to skip a almost all our integration testing here to make this work....

jakemac53:
I think I am going to go with a different strategy instead of skipping tests I will just run most tests on the previous 2.10 dev release and not 2.11 which is basically unusable.

jakemac53:
aaand of course the flutter tests just clone master lol so they get the latest, ill have to skip those

PR #2842, opened 2020-09-26
Remove more unnecessary dart:async imports

MichaelRFairhurst:
I missed these in the previous pass. Since Dart 2.1, Future and Stream
have been exported from dart:core.


jakemac53 (approved):

MichaelRFairhurst:
Thanks!

PR #2845, opened 2020-09-28
Fix loss of edges when replacing synthetic nodes with generated ones

jakemac53:
When we replace a synthetic node with a generated node, we first delete the synthetic node which deletes the node from the `inputs` sets of its `outputs`. This re-adds those when we are building up the `outputs` again for the new generated asset node.

- Also adds a regression test

Fixes https://github.com/dart-lang/build/issues/1804

### DIFF
--- a/build_runner_core/lib/src/asset_graph/graph.dart
+++ b/build_runner_core/lib/src/asset_graph/graph.dart
@@ -525,10 +525,7 @@ class AssetGraph {
       if (existing != null) {
         newNode.outputs.addAll(existing.outputs);
         // Ensure we set up the reverse link for NodeWithInput nodes.
-        existing.outputs
-            .map(get)
-            .whereType<NodeWithInputs>()
-            .forEach((node) => node.inputs.add(output));
+        _addInput(existing.outputs, output);
       }
       builderOptionsNode.outputs.add(output);
       _add(newNode);
@@ -542,6 +539,14 @@ class AssetGraph {
   // TODO remove once tests are updated
   void add(AssetNode node) => _add(node);
   Set<AssetId> remove(AssetId id) => _removeRecursive(id);
+
+  /// Adds [input] to all [outputs] if they represent [NodeWithInputs] nodes.
+  void _addInput(Iterable<AssetId> outputs, AssetId input) {
+    for (var output in outputs) {
+      var node = get(output);
+      if (node is NodeWithInputs) node.inputs.add(input);
+    }
+  }
 }
 
 /// Computes a [Digest] for [buildPhases] which can be used to compare one set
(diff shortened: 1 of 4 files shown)


build_runner_core/lib/src/asset_graph/graph.dart
```diff
@@ -524,6 +524,11 @@ class AssetGraph {
           isHidden: isHidden);
       if (existing != null) {
         newNode.outputs.addAll(existing.outputs);
+        // Ensure we set up the reverse link for NodeWithInput nodes.
+        existing.outputs
+            .map(get)
+            .whereType<NodeWithInputs>()
+            .forEach((node) => node.inputs.add(output));
```
natebosch, line 531: [nit] avoid `forEach` with a function literal.

```dart
var reverseDeps = existing.outputs.map(get).whereType<NodeWithInputs>();
for(var reverseDep in reverseDeps) reversDep.inputs.add(output);
```

natebosch (approved):
Needs a CHANEGLOG.

build_runner_core/lib/src/asset_graph/graph.dart
jakemac53: Does that need to apply to vm only code? I intentionally structured this to trick that lint into not complaining lmao because I think this is a fair bit cleaner 🤣 . I am happy to change it though if you would like.

natebosch: Yeah I think it still applies. The rule isn't specific to any platform.

https://dart.dev/guides/language/effective-dart/usage#avoid-using-iterableforeach-with-a-function-literal

I guess it is an "Avoid" and not a "Don't" though, so 🤷 

I am curious how you managed to trick the lint though.

jakemac53: The lint doesn't trigger if you have a chain of iterables like this, it is only if you have a standalone `<iterable>.forEach`.

jakemac53: Although it is pretty funny to me that to work around the lint I made it even slower 🤣

jakemac53: Ok, I refactored this to call a private method so I didn't have to deal with shadowing etc which was the primary reason I had for not wanting to use a for in loop in the first place.

build_runner_core/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_runner_core
-version: 6.0.2
+version: 6.0.2+1
```
natebosch, line 2:
```suggestion
version: 6.0.3
```

natebosch (approved):

PR #2846, opened 2020-09-29
Add docs for configuring environment variables

natebosch:


build_web_compilers/README.md
```diff
@@ -64,6 +64,34 @@ targets:
...
+  build_web_compilers:ddc:
+    options:
+      environment:
+        SOME_VAR: some value
+```
+
+For dart2js, use the `dart2js_args` option. This may be configured globally, or
+per-target.
+
+```
```
kevmoo, line 85: want to put `yaml` here, too?

kevmoo, line 85:
```suggestion
```yaml
```

build_web_compilers/README.md
```diff
@@ -64,6 +64,34 @@ targets:
           - -O2
 ```
 
+### Configuring -D environment variables
```
jakemac53, line 67: We probably should just support these similar to experiments.... 🤣.

Anyways I would list the command line usage as well because I think that would be the most common/useful?

jakemac53 (approved):

build_web_compilers/README.md
natebosch: > We probably should just support these similar to experiments....

Would we invalidate the entire build when they change? We could potentially use the zone to know which builders read them...

Want to file a feature issue?

> Anyways I would list the command line usage as well because I think that would be the most common/useful?

I'm not really sure which is the common usage. The `--define` approach is ugly, I added an example below.

PR #2847, opened 2020-09-29
Fix parsing of non-string environment for DDC

natebosch:


natebosch:
This is a followup to #2846. The diff will be smaller after that is merged.

jakemac53 (approved):

PR #2849, opened 2020-10-01
Fix up some mono_repo config

### LINKED ISSUE #2841: Update jobs to run on 2.11 dev once we can
See https://github.com/dart-lang/build/pull/2840 which is switching a bunch of tasks to run on 2.10 dev until we have an updated analyzer, and also removes the flutter testing.

natebosch:
- Indent lists to the same position as their containing key. This
  matches the style of other Google yaml files.
- Remove the extra `dartanalyzer` command from the `build_web_compilers`
  config.
- Remove an unnecessary `group` usage for a `test` job that does not
  have multiple tasks.
- Skip using a list when specifying a single Dart version for a job.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,16 +22,16 @@ jobs:
       env: PKGS="_test _test_common"
       script: ./tool/travis.sh dartanalyzer_0
     - stage: analyze_and_format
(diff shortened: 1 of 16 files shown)


natebosch:
Looks like we still have a problem with the opt-in for tests.

kevmoo (approved):

jakemac53 (approved):

PR #2851, opened 2020-10-02
Expand readable assets for non-root packages, make them configurable

### LINKED ISSUE #2819: Allow reading non-lib/ assets outside of the root package
Currently, attempting to read a non-`lib/` asset outside of the root package throws an `InvalidInputException`.
In fact, those assets aren't included in the build at all, since the `AssetTracker` doesn't pick them up:

https://github.com/dart-lang/build/blob/7a3d15df9b04ec18507455bacb1d23e81677569e/build_runner_core/lib/src/generate/build_definition.dart#L171-L172

I couldn't find any documentation describing _why_ this restriction exists. So I think that, either

- `InvalidInputException` should have a more descriptive error message explaining why this isn't possible
- or, if possible, the current restrictions should be loosened

As an example of a use-case that can't be written today, consider a builder writing all `LICENSE` files across dependencies into a Dart string literal. Another common scenarios are multi-package builds: Let's say I had a web app and server in a monorepo. Building the server should also build the app, and copy the generated js into a directory for static files. If the app applies a builder like [`build_version`](https://pub.dev/packages/build_version), the build fails with an `InvalidInputException`. I'd expect the build to behave exactly like if I only built the app.

simolus3:
Instead of throwing an `InvalidInputException` for non-lib, non-root packages, we now allow the following files by default:
- `lib/**`
- `bin/**`
- `LICENSE`
- `README*`
- `pubspec.yaml`

Each package can expose additional globs with the `additional_public_assets` build option. I opted to not support exluding assets because
- excluding `lib/` might lead to unexpected behavior
- that package's default target includes `additional_public_assets` by default. If there were two `excludes`, it's not clear in which priority they should interfere with each other

In the root package, all files are still allowed of course. I moved the check to throw an `InvalidInputException` down to the `SingleStepReader`. Computing the globs of available sources per package now happens in the `TargetGraph` since that's where the build config is read. That feels a bit unclean since the _target_ graph now decides which assets are visible for a _package_, please let me know if I should move that logic to the `PackageGraph` or elsewhere.

Closes #2819. I also wrote [an example](https://github.com/simolus3/build_license_collector/blob/main/lib/src/crawler.dart) to crawl licenses across dependencies.

### DIFF
--- a/_test_common/lib/test_phases.dart
+++ b/_test_common/lib/test_phases.dart
@@ -96,7 +96,7 @@ Future<BuildResult> testBuilders(
   String logPerformanceDir,
   String expectedGeneratedDir,
 }) async {
-  packageGraph ??= buildPackageGraph({rootPackage('a'): []});
+  packageGraph ??= _packageGraphFromInputs(inputs.keys);
   writer ??= InMemoryRunnerAssetWriter();
   reader ??= InMemoryRunnerAssetReader.shareAssetCache(writer.assets,
       rootPackage: packageGraph?.root?.name);
@@ -162,6 +162,27 @@ Future<BuildResult> testBuilders(
         expectedGeneratedDir: expectedGeneratedDir);
   }
   return result;
+}
+
+/// Builds the default package graph from input names.
+///
+/// The package graph will contain a root package named `a` depending on every
+/// other package present in [inputIds].
+PackageGraph _packageGraphFromInputs(Iterable<String> inputIds) {
+  final root = rootPackage('a');
+  final additionalPackageNames = <String>{};
+
+  for (final id in inputIds) {
+    final assetId = makeAssetId(id);
+    if (assetId.package != root.name) {
+      additionalPackageNames.add(assetId.package);
+    }
+  }
+
+  return buildPackageGraph({
+    root: additionalPackageNames.toList(),
+    for (final name in additionalPackageNames) package(name): [],
+  });
 }
 
 /// Translates expected outptus which start with `$$` to the build cache and
--- a/_test_common/lib/test_phases.dart
+++ b/_test_common/lib/test_phases.dart
@@ -96,7 +96,7 @@ Future<BuildResult> testBuilders(
   String logPerformanceDir,
   String expectedGeneratedDir,
 }) async {
-  packageGraph ??= _packageGraphFromInputs(inputs.keys);
+  packageGraph ??= buildPackageGraph({rootPackage('a'): []});
   writer ??= InMemoryRunnerAssetWriter();
   reader ??= InMemoryRunnerAssetReader.shareAssetCache(writer.assets,
       rootPackage: packageGraph?.root?.name);
@@ -162,27 +162,6 @@ Future<BuildResult> testBuilders(
         expectedGeneratedDir: expectedGeneratedDir);
   }
   return result;
-}
-
-/// Builds the default package graph from input names.
-///
-/// The package graph will contain a root package named `a` depending on every
-/// other package present in [inputIds].
-PackageGraph _packageGraphFromInputs(Iterable<String> inputIds) {
-  final root = rootPackage('a');
-  final additionalPackageNames = <String>{};
-
-  for (final id in inputIds) {
-    final assetId = makeAssetId(id);
-    if (assetId.package != root.name) {
-      additionalPackageNames.add(assetId.package);
-    }
-  }
-
-  return buildPackageGraph({
-    root: additionalPackageNames.toList(),
-    for (final name in additionalPackageNames) package(name): [],
-  });
 }
 
 /// Translates expected outptus which start with `$$` to the build cache and
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.6.0-dev
+## 1.5.1-dev
 
 - Expose a set of valid inputs in `InvalidInputException`.
 - Build implementations are now expected to throw `InvalidInputException`s, the
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 1.6.0-dev
+
+- Expose a set of valid inputs in `InvalidInputException`.
+- Build implementations are now expected to throw `InvalidInputException`s, the
+  check has been removed from this package.
+
 ## 1.5.0
 - Allow the latest analyzer version `0.40.x`.
 - Added the `Future<CompilationUnit> compilationUnitFor(AssetId id)` api to the
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## 1.5.1-dev
 
 - Expose a set of valid inputs in `InvalidInputException`.
-- Build implementations are now expected to throw `InvalidInputException`s, the
-  check has been removed from this package.
+- `AssetReader` implementations are now expected to throw 
+  `InvalidInputException`s when reading invalid inputs.
+  The check has been removed from the build step implementation.
 
 ## 1.5.0
 - Allow the latest analyzer version `0.40.x`.
--- a/build/lib/src/asset/exceptions.dart
+++ b/build/lib/src/asset/exceptions.dart
@@ -41,7 +41,7 @@ class InvalidInputException implements Exception {
   /// choose to allow additional assets via a `build.yaml` file.
   final List<String> allowedGlobs;
 
-  InvalidInputException(this.assetId, [this.allowedGlobs = const ['lib/**']]);
+  InvalidInputException(this.assetId, {this.allowedGlobs = const ['lib/**']});
 
   @override
   String toString() {
--- a/build/lib/src/asset/exceptions.dart
+++ b/build/lib/src/asset/exceptions.dart
@@ -40,8 +40,8 @@ class InvalidInputException implements Exception {
   String toString() => 'InvalidInputException: $assetId\n'
       'For package dependencies, only public files like `lib/**` may be used '
       'as inputs.\n'
-      'A package can mark a file as public by adding it to its `public_assets` '
-      'in a build.yaml file.';
+      'A package can mark a file as public by including it in its'
+      '`additional_public_assets` in a build.yaml file.';
 }
 
 class BuildStepCompletedException implements Exception {
--- a/build/lib/src/asset/exceptions.dart
+++ b/build/lib/src/asset/exceptions.dart
@@ -48,10 +48,12 @@ class InvalidInputException implements Exception {
     final allowedBuffer = StringBuffer();
 
     for (var i = 0; i < allowedGlobs.length; i++) {
-      if (i == allowedGlobs.length - 1) {
-        allowedBuffer.write(' and ');
-      } else if (i != 0) {
-        allowedBuffer.write(', ');
+      if (i > 0) {
+        if (i == allowedGlobs.length - 1) {
+          allowedBuffer.write(' or ');
+        } else {
+          allowedBuffer.write(', ');
+        }
       }
 
       allowedBuffer.write(allowedGlobs[i]);
--- a/build/lib/src/asset/exceptions.dart
+++ b/build/lib/src/asset/exceptions.dart
@@ -32,16 +32,37 @@ class InvalidOutputException implements Exception {
 }
 
 class InvalidInputException implements Exception {
+  /// The invalid asset that couldn't be read.
   final AssetId assetId;
 
-  InvalidInputException(this.assetId);
+  /// The list of readable globs in the package dependency.
+  ///
+  /// This includes public files like `lib/` by default, but a package can
+  /// choose to allow additional assets via a `build.yaml` file.
+  final List<String> allowedGlobs;
+
+  InvalidInputException(this.assetId, [this.allowedGlobs = const ['lib/**']]);
 
   @override
-  String toString() => 'InvalidInputException: $assetId\n'
-      'For package dependencies, only public files like `lib/**` may be used '
-      'as inputs.\n'
-      'A package can mark a file as public by including it in its '
-      '`additional_public_assets` in a build.yaml file.';
+  String toString() {
+    final allowedBuffer = StringBuffer();
+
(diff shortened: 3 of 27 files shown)


jakemac53 (commented):
Will take a longer look on Monday, nothing major stands out though.

I think the TargetGraph is the correct place to manage this - that manages what the actual visible individual files are for the package already effectively.

_test_common/lib/test_phases.dart
```diff
@@ -96,7 +96,7 @@ Future<BuildResult> testBuilders(
   String logPerformanceDir,
   String expectedGeneratedDir,
 }) async {
-  packageGraph ??= buildPackageGraph({rootPackage('a'): []});
+  packageGraph ??= _packageGraphFromInputs(inputs.keys);
```
simolus3, line 99: Some tests call `testBuilders` without a custom package graph and `inputs` outside of the default `a` package. This caused test failures since we can't know which inputs are valid without looking at the package graph. I changed the default package graph to contain packages for all inputs. If this is not desirable I could also fix the offending tests instead.

docs/build_yaml_format.md
```diff
@@ -8,12 +8,13 @@ to see examples or goal oriented docs you may want to look at the
 
 ## BuildConfig
 
-key                   | value                                                                      | default
---------------------- | -------------------------------------------------------------------------- | -------
-targets               | Map<String, [BuildTarget](#buildtarget)>                                   | a single target with the same name as the package
-builders              | Map<String, [BuilderDefinition](#builderdefinition)>                       | empty
-post_process_builders | Map<String, [PostProcessBuilderDefinition](#postprocessbuilderdefinition)> | empty
-global_options        | Map<String, [GlobalBuilderOptions](#globalBuilderOptions)>                 | empty
+key                      | value                                                                      | default
```
simolus3, line 11: I should mention that this markdown document has other tables whose columns are all aligned. Since you have an internal formatting tool, maybe you could quickly run that over this file so that I don't have to manually fix them?

simolus3 (commented):
There's a problem with the new approach and placeholder nodes like `test/$test$` that are added to every package.
When we call `runBuilder` [here](https://github.com/dart-lang/build/blob/6900853dcb5164118d2e86b2b08391cbf30ed8f8/build_runner_core/lib/src/generate/build_impl.dart#L485), we don't set the `rootPackage` argument. This causes the `BuildStepImpl` to believe that the primary input is in fact the root package. Is this a bug? I'm getting test failures when a `TestBuilder` (applied to multiple packages) calls `canRead('nonRootPackage|test/$test$')`, which is now considered an invalid input.

So what do we do to fix this? I've considered
1. disabling the check on placeholder nodes since they can't be read anyways
2. Not running a builder on primary inputs that are now invalid

To me, option 2 sounds more reasonable, but it might be considered breaking since builders applied to all packages would no longer run for some placeholder assets. What do you think?

jakemac53 (commented):
> There's a problem with the new approach and placeholder nodes like `test/$test$` that are added to every package.
> When we call `runBuilder` [here](https://github.com/dart-lang/build/blob/6900853dcb5164118d2e86b2b08391cbf30ed8f8/build_runner_core/lib/src/generate/build_impl.dart#L485), we don't set the `rootPackage` argument. This causes the `BuildStepImpl` to believe that the primary input is in fact the root package. Is this a bug? I'm getting test failures when a `TestBuilder` (applied to multiple packages) calls `canRead('nonRootPackage|test/$test$')`, which is now considered an invalid input.
> 
> So what do we do to fix this? I've considered
> 
> 1. disabling the check on placeholder nodes since they can't be read anyways
> 2. Not running a builder on primary inputs that are now invalid
> 3. Making `test/$test$` public by default
> 
> To me, option 2 sounds the most reasonable, but it might be considered breaking since builders applied to all packages would no longer run for some placeholder assets. What do you think?

I am not quite fully understanding the issue here - it does seem like a bug that we are not passing the `rootPackage` argument through though. Does doing that fix the problem?

natebosch:
> cc @natebosch I think might have something he uses here? I don't have a tool I use :/

I do, unfortunately it is internal only. I can fix it up before we merge.

build_runner_core/lib/src/package_graph/target_graph.dart
```diff
@@ -84,12 +106,55 @@ class TargetGraph {
...
+      ];
+    }
+  }
+
+  /// Whether the [id] is visible in a build.
+  ///
+  /// All assets in the root package are visible. For non-root packages, an
+  /// asset is visible if it's in a conceptually public folders of a Dart
+  /// package (like `lib/` or `bin/`), or if the enclosing package made that
+  /// asset public by including it in [BuildConfig.additionalPublicAssets].
```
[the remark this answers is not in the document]
simolus3: Right, I added a note for that.

build/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.6.0-dev
```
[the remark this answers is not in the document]
simolus3: That's unfortunate but understandable. I changed the version to `1.5.1`.

simolus3:
> I am not quite fully understanding the issue here

Consider [this test](https://github.com/dart-lang/build/blob/99540a0826416026a8b63a7e7d005eaae8956f33/build_runner_core/test/generate/build_test.dart#L58-L80) for example. We'd invoke the `TestBuilder` on `a|test/$test`, `b|test/$test` and `$sdk$|test/$test` (among others). However, only `a|test/$test$` is visible in the build and `TestBuilder` calls `buildStep.canRead(buildStep.inputId)`, so we now get an `InvalidInputException` on the other two invocations.

If we only passed the `rootPackage` argument we'd get the same error. Arguably, that test should have failed before too since the inputs have always been invalid. Since those builders invocations are useless, I added an additional check to `_matchingPrimaryInputs` so that we only run the builder if their primary input is visible in the build. That alone should fix the problem, but I also pass the `rootPackage` argument to avoid potential other inconsistencies.

I brought this up because I'm not sure if we want to consider this a bugfix or a breaking change. Passing the `rootPackage` argument also means that `findAssets` will now run on the root package instead of the package enclosing the primary input.

build/CHANGELOG.md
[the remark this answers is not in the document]
jakemac53: One thing I wanted to confirm here - if people get this latest version of `build` but not the latest `build_runner_core` (which would now be possible) - they will still get an exception when trying to read files outside of lib right? It would just be an `AssetNotFoundException` instead of an `InvalidInputException`?

jakemac53:
> However, only `a|test/$test$` is visible in the build and `TestBuilder` calls `buildStep.canRead(buildStep.inputId)`, so we now get an `InvalidInputException` on the other two invocations.

Ah - I think the issue here then is that `canRead` should not throw in this case. It should return `false`. This api should never throw, as it is the way you ask if it is possible to read some file safely.

jakemac53:
> Passing the `rootPackage` argument also means that `findAssets` will now run on the root package instead of the package enclosing the primary input.

Ya we definitely wouldn't want this behavior.

build/CHANGELOG.md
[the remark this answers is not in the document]
simolus3: Yes exactly. In the old `build_runner_core` version, the non-lib files aren't considered by the `AssetTracker` at all. So one would definitely get an `AssetNotFoundException`.

simolus3:
> I think the issue here then is that `canRead` should not throw in this case

Note that `canRead` called `_checkInput` before, so it did throw for invalid inputs in the past. It just missed some (_most?_) cases due to the `rootPackage` argument not being set. I can remove that check for `canRead` now if you want, but I want to make sure that's the behavior we want here.

>> Passing the `rootPackage` argument also means that `findAssets` will now run on the root package instead of the package enclosing the primary input.

> Ya we definitely wouldn't want this behavior.

My initial statement was inaccurate, we actually pass `inputId?.package ?? rootPackage` if we have a `MultiPackageAssetReader`. However, `inputId` is guaranteed to be non-null if the `BuildStepImpl` was created through `runBuilder` and that would be the last usage of `rootPackage`, so should I just remove that field and deprecate the parameter?

jakemac53:
> Note that `canRead` called `_checkInput` before, so it did throw for invalid inputs in the past. It just missed some (_most?_) cases due to the `rootPackage` argument not being set. I can remove that check for `canRead` now if you want, but I want to make sure that's the behavior we want here.

Oh interesting, I don't think this behavior was intentional - cc @natebosch can you think of why that would have been intentional?

> My initial statement was inaccurate, we actually pass `inputId?.package ?? rootPackage` if we have a `MultiPackageAssetReader`. 

I am a bit confused about this part, where do you see this code?

> However, `inputId` is guaranteed to be non-null if the `BuildStepImpl` was created through `runBuilder` and that would be the last usage of `rootPackage`, so should I just remove that field and deprecate the parameter?

Yes it sounds to me like the `rootPackage` param on `runBuilder` should be deprecated - I don't see us passing that anywhere.

simolus3:
>I am a bit confused about this part, where do you see this code?

Sorry for being unclear, I was talking about usages of `rootPackage` in `BuildStepImpl`:

https://github.com/dart-lang/build/blob/e7b3d4e544be76b77a08bf81fa471eb0440012a4/build/lib/src/builder/build_step_impl.dart#L109-L114

However, `inputId` is guaranteed to be non-null: We never set the `rootPackage` argument and the default is `inputId.package` from [here](https://github.com/dart-lang/build/blob/e7b3d4e544be76b77a08bf81fa471eb0440012a4/build/lib/src/generate/run_builder.dart#L49). Since it was never used, I'm going to remove the field and parameter entirely.

jakemac53:
> However, `inputId` is guaranteed to be non-null: We never set the `rootPackage` argument and the default is `inputId.package` from [here](https://github.com/dart-lang/build/blob/e7b3d4e544be76b77a08bf81fa471eb0440012a4/build/lib/src/generate/run_builder.dart#L49). Since it was never used, I'm going to remove the field and parameter entirely.

Ok, I understand now :). I agree with your assessment as well, and I don't see any direct instantiations of `BuildStepImpl` outside of this package either so it should be safe to remove.

We should similarly deprecate the argument on `runBuilder` since all it does is pass it through to this constructor.

jakemac53 (commented):
This is starting to look pretty close, I want to do another pass before we merge, which I will try to do today. I am on vacation next week though - so either I would need @natebosch to finalize the landing/publishing of this or I can pick it back up on/after 10/19 when I am back.

Speaking of, @natebosch did you want to do a review here as well?

build/lib/src/asset/exceptions.dart
```diff
@@ -32,13 +32,39 @@ class InvalidOutputException implements Exception {
...
   final AssetId assetId;
 
-  InvalidInputException(this.assetId);
+  /// The list of readable globs in the package dependency.
+  ///
+  /// This includes public files like `lib/` by default, but a package can
+  /// choose to allow additional assets via a `build.yaml` file.
+  final List<String> allowedGlobs;
+
+  InvalidInputException(this.assetId, [this.allowedGlobs = const ['lib/**']]);
```
jakemac53, line 44: nit: Change to a named optional argument (more extensible in the future)

```suggestion
  InvalidInputException(this.assetId, [this.allowedGlobs = const ['lib/**']]);
```

build_runner_core/lib/src/generate/options.dart
```diff
@@ -19,8 +19,22 @@ import '../package_graph/target_graph.dart';
 import '../util/hash.dart';
 import 'exceptions.dart';
 
+/// The default list of files visible for non-root packages.
+///
+/// This is also the default list of files for targets in non-root packages when
+/// an explicit include is not provided.
+const List<String> defaultNonRootVisibleAssets = [
```
jakemac53, line 26: @natebosch particularly it might be good to get another set of eyes here - this list seems pretty reasonable to me though.

Possibly we should add `'CHANGELOG*'`?

jakemac53 (approved):

jakemac53:
Hey, sorry I just saw this one is still pending. Are you waiting on anything from us? Just a final review or did you have any more work pending here?

simolus3:
I just fixed the conflicts, so now I don't have any pending work here :)

docs/build_yaml_format.md
natebosch: The toll I was using is internal. I went ahead and pushed a commit which runs it on this file, but looks like most of the formatting had already been handled. Sorry for the difficulty.

natebosch:
> Oh interesting, I don't think this behavior was intentional - cc @natebosch can you think of why that would have been intentional?

I can't think of any reason we'd want `canRead` to throw...

build_runner_core/lib/src/generate/options.dart
natebosch: This list looks fine to me. I'd be in favor of adding `CHANGELOG*`

build/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 1.5.1-dev
+
+- Expose a set of valid inputs in `InvalidInputException`.
+- Build implementations are now expected to throw `InvalidInputException`s, the
+  check has been removed from this package.
```
natebosch, line 5: I don't understand this. Can we rephrase or expand it?

natebosch (commented):
Overall this looks good to me. I'm also running this through some internal tests to see if there is anything breaking that we didn't intend.

natebosch (approved):
Internal tests look good.

build/CHANGELOG.md
jakemac53: The `BuildStepImpl` used to have a check that banned reads for all files outside of `lib` from the non-root package. It no longer has enough information to do this check itself (and no longer knows the root package either).

The underlying asset reader that it is given must implement this now.

jakemac53:
It looks like one failing test here and then we can merge/publish (possibly update the changelog as Nate suggests as well)

build/CHANGELOG.md
simolus3: I adapted the changelog to point out that `AssetReader` implementations are supposed to perform the check now, and that it has been removed from the `BuildStepImpl`.

PR #2852, opened 2020-10-05
add sound_null_safety and null_assertions options

jakemac53:
- Only supported for the to the build_web_compilers|entrypoint builder
- Works for ddc and dart2js
- Adds some basic e2e tests

### DIFF
--- a/_test_null_safety/test/common/message.dart
+++ b/_test_null_safety/test/common/message.dart
@@ -5,5 +5,5 @@
 // Uses `?` to ensure null safety is enabled
 final String? message = 'hello';
 
-// Prints a non-nullable message, used in null assertion tests.
-void printString(String message) => print(message);
+// Used in null assertion tests.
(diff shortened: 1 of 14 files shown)


_test_null_safety/test/common/message.dart
```diff
@@ -4,3 +4,6 @@
 
 // Uses `?` to ensure null safety is enabled
 final String? message = 'hello';
+
+// Prints a non-nullable message, used in null assertion tests.
+void printString(String message) => print(message);
```
natebosch, line 9: [nit] how about `printNonNullable` so that the name is a closer match to the intent?

natebosch (approved):

_test_null_safety/test/common/message.dart
jakemac53: Sure

nshahan (approved):
Thanks for the additions!

PR #2864, opened 2020-10-19
Add support for native non-null asserts to build_web_compilers:entrypoint builder

jakemac53:
Defaults to disable these for now, but the plan will be to enable them by default later on once things have stabilized a bit and we have good error messages that highlight how to bypass these checks.

Part of https://github.com/dart-lang/sdk/issues/43780

cc @srujzs


jakemac53:
I left a comment [here](https://github.com/dart-lang/sdk/issues/42536#issuecomment-712470594) about the flag not being negatable, it probably isn't a big deal.

natebosch (approved):

PR #2866, opened 2020-10-20
Fix handling of changeFile for the analysis driver

natebosch:
Fixes an issue where we may fail to notify the `AnalysisDriver` of a
newly available asset that it has previously read as missing before we
provided it. This could happen in some versions of the analyzer package
when calling `parseFile`, and it can happen in all versions when we use
the resolver with an import to an asset that is not generated yet, so
not visible to the build step.

See https://github.com/dart-lang/sdk/issues/43744

From the standpoint of the analyzer, we need to perform two steps in
order for the AnalysisDriver to see the correct content for a file after
it has previously tried to read it and it was missing:
1 - MemoryResourceManager.newFile
2 - AnalysisDriver.changeFile

In the case where the analyzer had never cached any state for that file,
we can get by with just `newFile`. We do currently guarantee that
`newFile` is called in time for analysis, but we don't guarantee that
`changeFile` has been called.

Remove the concept of deciding concretely whether to call `changeFile`
during the crawl. Instead track the asset paths that currently need to
be updated in the `AnalysisDriver` in a separate set. The first build
step to see the asset, or changed asset, during a transitive crawl will
record it as needing an update, and the first build step to finish a
crawl that visited that asset will call `changeFile`. Often these will
be the same build step.

Resolver changes to fix:
- Add a `_needsChangeFile` set. The extra memory used should scale with
  ongoing work, not with the total size of the project.
- Simplify the `_AssetState` class.
- After any crawl, find the intersection between the transitive
  dependencies crawled, and the assets that need an update, and call
  `changeFile` for each.

Also, as a separate optimization: Add an extra check against the cached
digests after reading the file content. This prevents duplicate work of
`_parseDirectives` in the case where multiple build steps saw an
uncached digest and start to await the `readAsString` call.

Add a test with builders forcing a specific series of interactions that
demonstrates this problem:

- g1 builder: reads a file which imports the out it has not yet written.
  Puts the analyzer into the cached bad state.
- root builder: (next non-optional builder) uses the resolver and starts
  a transitive crawl.
  - This crawl sees `.g1.dart` and calls `newFile`. It also caches the
    digest.
  - This crawl asks for `.slow.dart` and`.g2.dart`. This triggers a
    lazy build in both.
- The builder for `.slow.dart` won't complete, therefore the
  `crawlAsync` being done for the resolver intended for the "root"
  builder won't complete, therefore no `changeFile` are called.
- The builder for `.g2.dart` tries to use the resolver. During it's
  transitive dependency crawl it found that the digest for `.g1.dart`
  was cached, so it skips this asset for `changeFile`. It sees "null"
  for the annotation value, instead of what it would get if the
  `.g1.dart` import were correctly visible.

Remove the `invariant_booleans` lint rule for false positives.

### DIFF
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -32,6 +32,10 @@ class BuildAssetUriResolver extends UriResolver {
   /// changed.
   final _cachedAssetDigests = <AssetId, Digest>{};
 
+  /// Asset paths which have been updated in [resourceProvider] but not yet
+  /// updated in the analysis driver.
+  final _needsChangeFile = <String>{};
+
   final resourceProvider = MemoryResourceProvider(context: p.posix);
 
   /// The assets which are known to be readable at some point during the current
@@ -64,16 +68,17 @@ class BuildAssetUriResolver extends UriResolver {
             uncrawledIds,
             (id) => _updateCachedAssetState(id, buildStep,
                 transitivelyResolved: transitivelyResolved),
-            (id, state) => state.directives.where(notCrawled)).toList()
+            (id, state) => state.dependencies.where(notCrawled)).toList()
         : [
             for (var id in uncrawledIds)
               await _updateCachedAssetState(id, buildStep),
           ];
 
-    assetStates
-        .where((state) => state.isAssetUpdate)
-        .map((state) => state.path)
-        .forEach(driver.changeFile);
+    for (final state in assetStates) {
+      if (_needsChangeFile.remove(state.path)) {
+        driver.changeFile(state.path);
+      }
+    }
   }
 
   /// Updates the internal state for [id], if it has changed.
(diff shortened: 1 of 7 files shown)


jakemac53 (approved):

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -102,26 +107,30 @@ class BuildAssetUriResolver extends UriResolver {
...
-      return _AssetState.unchanged(path, _cachedAssetDependencies[id]);
+      return _AssetState(path, _cachedAssetDependencies[id]);
     } else {
       final isChange = _cachedAssetDigests.containsKey(id);
       final content = await buildStep.readAsString(id);
-      _cachedAssetDigests[id] = digest;
-      final dependencies =
-          _cachedAssetDependencies[id] = _parseDirectives(content, id);
+      // ignore: invariant_booleans
+      if (_cachedAssetDigests[id] == digest) {
```
jakemac53, line 121: I think the lint was actually correctly triggering here? This is effectively unreachable code (will always evaluate to false?)

jakemac53: NVM I see now that this is to check if the asset digest was updated after the `await buildStep.readAsString`.

I would add a comment about that to clarify

natebosch: I don't think it was correct. I need to file a bug against the lint. The `await` means that someone else could have updated `_cachedAssetDigests` since we checked it last.

jakemac53 (approved):

build_resolvers/lib/src/build_asset_uri_resolver.dart
natebosch: This is somewhat similar to https://github.com/dart-lang/linter/issues/1858

jakemac53 (approved):

PR #2867, opened 2020-10-20
[DISCUSSION ONLY] make the build apis return FutureOr

jakemac53:
This is an investigation into making `canRead`, `readAsString`, `readAsBytes`, and `digest` return a `FutureOr`. The primary motivation is to investigate the performance impact if we are able to take advantage of this in a few key places (namely, in build_resolvers which is done here).

On the angular_forms repo I saw around a 10% performance improvement here - from ~22 seconds to ~20 seconds.

If we actually were to roll this out we may want to create entirely new apis, to avoid breaking existing `Builder` code, and we could also then possibly avoid a breaking change in the `build` package - although that would be a bit questionable.

### DIFF
--- a/build_modules/lib/src/module_cache.dart
+++ b/build_modules/lib/src/module_cache.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:async/async.dart';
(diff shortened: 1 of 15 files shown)


build_modules/lib/src/module_cache.dart
```diff
@@ -63,17 +64,18 @@ class DecodingCache<T> {
     if (!_cached.containsKey(id)) {
       entry = _cached[id] = _Entry()
         ..needsCheck = false
-        ..value = Result.capture(reader.readAsBytes(id).then(_fromBytes))
-        ..digest = Result.capture(reader.digest(id));
+        ..value =
+            Result.capture(toFuture(reader.readAsBytes(id)).then(_fromBytes))
```
natebosch, line 68: I wonder if it would be sensible for `Result.capture` to take a `FutureOr` even outside of this specific use.

natebosch (commented):
I wonder if another route we could explore is bundling up multiple pieces into a single API.

For instance one of the sources of slowness is we currently need to `await` the digest, and _also_ `await` the content (which will always be available synchronously if the digest is available I think). If we had some API that returned both the content and the digest at the same time, would that help?

I think my preference would lean towards some additional API like `SyncAssetReader` which would not be available to builders, but which we could use for the resolver. Not sure if that would add any more or less complexity than this.

In either case, if we do decide to move forward I think it would be worth spending some time with the profiler to make sure we understand how and why this is improving things, and see if there are other places that should also be updated to take advantage of it.

jakemac53:
Another interesting anecdote here - I decided to try this on a larger package (angular components gallery example) - and the difference was much smaller. Only saves about 2s out of 75s.

Probably a coincidence that it is the same total savings of 2s? pretty weird coincidence haha.

build_modules/lib/src/module_cache.dart
jakemac53: that might be nice yes

jakemac53:
I am going to go ahead and just close this given that the impact doesn't seem to scale for larger apps for whatever reason.

kevmoo:
Can delete branch @jakemac53 ?

PR #2873, opened 2020-10-23
Strip temp directory from full kernel path in metadata

annagrin:
Strip temp directory from full kernel path in metadata, 
in preparation to addition of full kernel path to metadata.

Related: https://github.com/dart-lang/sdk/issues/43684


jakemac53 (commented):
Please update the pubspec and changeling as well

jakemac53 (approved):

Issue #2790, opened 2020-08-17
AssetNotFoundException: my-app|.dart_tool/package_config.json


JanMate:
I got this stacktrace, so it can be useful for future debugging.

I launched:
flutter pub run build_runner watch --delete-conflicting-outputs

and after a few time it ends with. Maybe it terminated during flutter pub get or during anything else.

AssetNotFoundException: my-app|.dart_tool/package_config.json
package:build_runner_core/src/asset/file_based.dart 134:18  _fileForOrThrow.<fn>
dart:async                                                  _completeOnAsyncReturn
package:pool/pool.dart                                      Pool.withResource
package:build_runner_core/src/asset/file_based.dart 133:26  _fileForOrThrow
package:build_runner_core/src/asset/file_based.dart 33:48   FileBasedAssetReader.readAsBytes
package:build_runner/src/generate/watch_impl.dart 272:46    WatchImpl._run.<fn>

pub finished with exit code 1

jakemac53:
This looks like it definitely could have been a result of running a `flutter pub get` while using the `watch` command - it tried to re-run before the package config was re-written.

Do you see this happen often or just once? I am a bit surprised I haven't seen a report of this before...

JanMate:
Sometimes I catch this problem. It is nothing severe. Re-run "flutter pub run build_runner watch --delete-conflicting-outputs" always helps and works like a charm. :)

jakemac53:
I went ahead and sent out a pr with a simple fix here to keep trying to read the file for at least 1 second. That should generally work although it isn't perfect obviously :).

PR #2875, opened 2020-10-27
Mitigate a race condition when package config files are overwritten

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2790

### DIFF
--- a/build_runner/test/generate/watch_test.dart
+++ b/build_runner/test/generate/watch_test.dart
@@ -27,6 +27,7 @@ void main() {
   final copyABuildApplication = applyToRoot(
       TestBuilder(buildExtensions: appendExtension('.copy', from: '.txt')));
   final defaultBuilderOptions = const BuilderOptions({});
+  final packageConfigId = makeAssetId('a|.dart_tool/package_config.json');
   InMemoryRunnerAssetWriter writer;
 
   setUp(() async {
@@ -35,8 +36,7 @@ void main() {
 # Fake packages file
 a:file://fake/pkg/path
 ''');
-    await writer.writeAsString(makeAssetId('a|.dart_tool/package_config.json'),
-        jsonEncode(_packageConfig));
+    await writer.writeAsString(packageConfigId, jsonEncode(_packageConfig));
   });
 
   group('watch', () {
@@ -261,7 +261,7 @@ a:file://fake/pkg/path
         var expectedGraph = await AssetGraph.build(
             [],
             <AssetId>{},
-            {AssetId('a', '.dart_tool/package_config.json')},
+            {packageConfigId},
             buildPackageGraph({rootPackage('a'): []}),
             InMemoryAssetReader(sourceAssets: writer.assets));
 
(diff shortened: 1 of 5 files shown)


build_runner/test/generate/watch_test.dart
```diff
@@ -404,9 +404,41 @@ a:file://different/fake/pkg/path
...
+        checkBuild(result, outputs: {'a|web/a.txt.copy': 'a'}, writer: writer);
+
+        await writer.delete(packageConfigId);
+
+        // Wait for it to try reading the file twice to ensure it will retry.
+        await _readerForState[buildState]
+            .onCanRead
+            .where((id) => id == packageConfigId)
+            .take(2)
+            .toList();
```
natebosch, line 437: [nit] 
```suggestion
            .drain();
```

build_runner/test/generate/watch_test.dart
```diff
@@ -816,6 +848,9 @@ Future<BuildState> startWatch(List<BuilderApplication> builders,
       logLevel: logLevel,
       onLog: onLog,
       skipBuildScriptCheck: true);
+  // Some tests need access to `reader` so we expose it through an expando.
```
natebosch, line 851: This feels a little strange to me.

natebosch (commented):
Do we expect in the typical case the build will be invalidated? Would it be sensible to catch the error instead and show a nicer error message instead of a stack trace? Will this still have a stack trace if it is missing for a full second?

build_runner/test/generate/watch_test.dart
jakemac53: Ya it is but its just a test and its a low boilerplate/simple way to get ahold of this reader from the test.

I could also create a whole new class which has the reader and BuildState object and returning that (and that is what I would do if this was public code that needed this), but that just seemed overkill for this use case.

jakemac53:
> Do we expect in the typical case the build will be invalidated?

This is specifically to handle the case where somebody runs `pub get/upgrade` while running the server or watcher.

>  Would it be sensible to catch the error instead and show a nicer error message instead of a stack trace?

We could throw a nicer error here, although I do not expect the error case to be hit any more in practice so it didn't seem worth it.

> Will this still have a stack trace if it is missing for a full second?

Not as useful of one, but still useful enough (if somebody were to paste that stack trace with just this one function we could easily still trace what happened). This is never the direct result of user code, its an internal state error essentially.

natebosch:
> This is specifically to handle the case where somebody runs `pub get/upgrade` while running the server or watcher.

Right, I'm wondering whether in the typical case for this situation once we "recover" and read the file after 1 second, we'll throw away the whole build anyway. I can't remember if we restart the whole watcher at that point, or if it exits entirely.

If we expect that most of the time we would hit the case where something changed and we need to bail out, we could do that eagerly instead of waiting a second to see the new file content and decide to do that.

If we think things can continue smoothly in the typical case when this file has changed we can go with this approach.

natebosch (approved):

jakemac53:
It looks like even in the case of a `pub get` this will terminate, because pub includes the date the file was generated in the package config 🤦‍♂️ . Given that, ya it might just be good to remove all the stuff trying to determine if the package config has actually changed and just exit eagerly.

jakemac53:
@natebosch ok I went ahead and simplified this a bunch, removed the new test and made it bail immediately on any change to these files.

natebosch (approved):

jakemac53:
Ok, it turns out any call to `pub run` modifies the `.packages` file, causing this to exit. This breaks tests and some real world usage so I reverted the simplification commits and went back to the file content checks and retries.

natebosch (approved):

Issue #2878, opened 2020-10-28
build_web_compilers doesn't party well with dart2js sound null safe arg


kevmoo:
v2.13.0

```yaml
# See https://github.com/dart-lang/build/tree/master/build_web_compilers#configuration
targets:
  $default:
    builders:
      build_web_compilers|entrypoint:
        options:
          dart2js_args:
            - --dump-info
            - --no-frequency-based-minification
            - --no-source-maps
            - --show-package-warnings
            - -O4
            - --enable-experiment=non-nullable
            - --sound-null-safety
```

```
[INFO] Generating build script completed, took 347ms
[INFO] Reading cached asset graph completed, took 287ms
[INFO] Checking for updates since last build completed, took 685ms
[SEVERE] build_web_compilers:entrypoint on web/main.dart:
ExitCode:1
StdOut:
Usage: dart2js [options] dartfile

Compiles Dart to JavaScript.

Common options:
  -o <file> Generate the output into <file>.
  -m        Generate minified output.
  -h        Display this message (add -v for information about all options).

Error: Error: cannot specify both --sound-null-safety and --no-sound-null-safety.

StdErr:

[INFO] Running build completed, took 912ms
[INFO] Caching finalized dependency graph completed, took 196ms
[SEVERE] Failed after 1.1s
```

jakemac53:
@kevmoo You want to configure this with the top level `sound_null_safety` option for this builder, so:

```yaml
# See https://github.com/dart-lang/build/tree/master/build_web_compilers#configuration
targets:
  $default:
    builders:
      build_web_compilers|entrypoint:
        options:
          sound_null_safety: true
          dart2js_args:
            - --dump-info
            - --no-frequency-based-minification
            - --no-source-maps
            - --show-package-warnings
            - -O4
            - --enable-experiment=non-nullable
```

Note that enabling the experiment should also be done on the command line, with `pub run build_runner --enable-experiment=non-nullable <command>`.

We could try to examine the provided args here and notify people of this problem though.

kevmoo:
> We could try to examine the provided args here and notify people of this problem though.

Yes, please!

I figured it out, but it was a mess.

Also: I don't see the args to dart2js printed any more. On purpose?

jakemac53:
No, that wasn't intentional. We had to move away from batch mode as it wasn't compatible with changing modes and we were told not to use it generally haha. The logging existed in that batch mode code and wasn't copied over.

 I can go ahead and add that back, and add some detection for these args in there.

PR #2879, opened 2020-10-29
Dart2js cleanup

jakemac53:
Restores support for the BUILD_DART2JS_VM_ARGS environment variable and the logging of dart2js arguments.

Adds warnings when configuring things through `dart2js_args` that should be configured through top level options.

Closes https://github.com/dart-lang/build/issues/2878

### DIFF
--- a/build_web_compilers/lib/src/dart2js_bootstrap.dart
+++ b/build_web_compilers/lib/src/dart2js_bootstrap.dart
@@ -67,7 +67,7 @@ Future<void> _bootstrapDart2Js(
           .join('\n');
       log.warning('''
 Skipping compiling ${buildStep.inputId} with dart2js because some of its
-transitive libraries have sdk dependencies that not supported on this platform:
+transitive libraries have sdk dependencies that are not supported on this platform:
 
 $librariesString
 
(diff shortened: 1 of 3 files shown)


kevmoo (approved):

kevmoo:
Thanks!

build_web_compilers/lib/src/dart2js_bootstrap.dart
```diff
@@ -162,3 +165,37 @@ Future<void> _copyIfExists(
 }
 
 final _resourcePool = Pool(maxWorkersPerTask);
+
+const _dart2jsVmArgsEnvVar = 'BUILD_DART2JS_VM_ARGS';
```
natebosch, line 169: Can you call this out in the commit message? Why do we need this?

natebosch (approved):

build_web_compilers/lib/src/dart2js_bootstrap.dart
jakemac53: Oh yes this was old behavior I am also restoring, workiva I believe needed it to increase ram available to dart2js.

build_web_compilers/lib/src/dart2js_bootstrap.dart
```diff
@@ -88,8 +88,11 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
             ? 'packages/${dartUri.path}'
             : dartUri.path.substring(1)) +
         jsEntrypointExtension;
+    var librariesSpec = p.joinAll([sdkDir, 'lib', 'libraries.json']);
+    _validateUserArgs(dart2JsArgs);
     args = dart2JsArgs.toList()
```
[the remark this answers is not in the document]
jakemac53: normally I would but ya with travis as slow as it is right now ill probably pass

PR #2884, opened 2020-11-02
load the sound ddc sdk in sound mode

jakemac53:


natebosch (approved):

jakemac53:
landed on red due to other failures

PR #2886, opened 2020-11-03
Fix travis

jakemac53:
- fixes dev compiler tests


natebosch (approved):

jakemac53:
Merged and cancelled the travis build based on the fact that it did run these tests and they went green

PR #2887, opened 2020-11-03
add native null assertions configuration doc

jakemac53:
Part of https://github.com/dart-lang/sdk/issues/43961


natebosch (approved):
Is there an existing doc this would fit in with? I hope to avoid a pattern where we collect many small docs in this directory.

jakemac53:
This is a specific doc for a specific SDK error message to link to so I think it makes sense to be its own (and the link already exists)

Issue #2862, opened 2020-10-15
Document BUILD_MAX_WORKERS_PER_TASK variable


alanknight-wk:
There's no documentation for this environment variable that I, or Google, can find. The only references to it are in issues in this package.

PR #2889, opened 2020-11-06
document BUILD_MAX_WORKERS_PER_TASK

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2862.

We could move this to the general FAQ as well... not sure where makes the most sense.


natebosch:
I do think the FAQ might be a better place, especially if we combine it with a description of `--low-resources-mode`.

jakemac53:
Moved to the FAQ and added a section about --low-resources-mode as well which links to it

natebosch (approved):

jakemac53:
merged on red, fixing the existing failure now

PR #2896, opened 2020-11-11
Fix 500 response for cached failure nodes in serve mode

jakemac53:
Respond with UnreadableReason.notOutput for failure nodes that don't match any build filters, which makes the server respond with a 404 instead of 500.


natebosch (approved):

jakemac53:
landing on red due to existing failure, which is already being fixed in https://github.com/dart-lang/build/pull/2895

Issue #2902, opened 2020-11-17
Sdk lines aren't collapsed for opted in programs on web


jakemac53:
Can be repro'd by adding an error to the `_test_null_safety/test/opted_in_test.dart` test and running `pub run build_runner test -- test/opted_in_test.dart`.

The sdk file is named `dart_sdk.sound.js` which I think is probably the reason (I think we look specifically for `dart_sdk.js`).

Issue #2901, opened 2020-11-17
build_runner test sourcemaps are ignored


daniel-v:
**Short summary:**

Running `pub run build_runner test -r -- -p chrome` command, I'd expect Dart stack traces to be reported in console. It is not the case.

**Things I tried**

In build yaml, I have the following to ensure that after building the assets, I retain the sourcemaps.

```yaml
global_options:
  build_web_compilers|dart_source_cleanup:
    release_options:
      enabled: false
  build_web_compilers|dart2js_archive_extractor:
    release_options:
      filter_outputs: false
```

I verified that they are in fact loaded, by starting tests with `--pause-after-load` and opening a dart file in chrome's dev toolbar. I'm not sure where the issue is but:

- I have sourcemaps
- they get loaded in browser
- I see non-mapped JS stack trace in console for errors

Anything I can do to get Dart traces?

Extra info:

* Dart SDK version: 2.9.3 (stable)
```
  build_runner:
    version: "1.9.0"
  build_runner_core:
    version: "5.1.0"
  build_test:
    version: "1.0.0"
  build_web_compilers:
    version: "2.10.0"
```
* Using it for an Angular 5.3.1 app

This issue might be related to https://github.com/dart-lang/build/issues/2377 but I'm not sure.

jakemac53:
I can repro this at HEAD, it looks like we aren't cleaning up the uris in the source maps like we need to, so they have `org-dartlang-app` scheme uris.

jakemac53:
I can fix this issue when running the test manually, but it looks like we don't support stack trace mapping for dart2js the way the test package wants it set up (we need to register a stack trace mapper on the page I think, which will be a fair bit of work given the current setup). I am not sure dart2js actually supports it either like DDC does.

We may be able to register a standard stack trace mapper on the package:test side of things, when in precompiled mode, which we don't do today.

daniel-v:
As far as I can remember, the `*_test.browser.dart` files (might not remember the exact extension) wraps each of the tests in some code (channel back to the test runner eg.). Most importantly, I saw some code related to a mapper. First glance, it seemed that there is a mapper registered on client side which attempts to map stack traces around serialization. 

Not sure if this helps but figured I'd drop it here if you were so quick to jump on this issue.

[...]

PR #2903, opened 2020-11-17
Stack trace fixes for sound mode ddc and dart2js

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2902

- Relaxes the check to identify the dart sdk and collapse lines to also match the sound mode dart sdk js file.

Helps with https://github.com/dart-lang/build/issues/2901 by fixing up the source maps, but the test package still doesn't report properly. Manually debugged tests work now though.

### DIFF
--- a/build_web_compilers/lib/src/common.dart
+++ b/build_web_compilers/lib/src/common.dart
@@ -58,13 +58,13 @@ void validateOptions(Map<String, dynamic> config, List<String> supportedOptions,
   }
 }
 
-/// Copied to `web/stack_trace_mapper.dart`, these need to be kept in sync.
-///
-/// Given a list of [uris] as [String]s from a sourcemap, fixes them up so that
-/// they make sense in a browser context.
+/// Fixes up the [uris] from a source map so they make sense in a browser
+/// context.
 ///
 /// - Strips the scheme from the uri
 /// - Strips the top level directory if its not `packages`
+///
+/// Copied to `web/stack_trace_mapper.dart`, these need to be kept in sync.
 List<String> fixSourceMapSources(List<String> uris) {
   return uris.map((source) {
     var uri = Uri.parse(source);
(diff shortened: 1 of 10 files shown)


build_web_compilers/lib/src/common.dart
```diff
@@ -57,3 +57,22 @@ void validateOptions(Map<String, dynamic> config, List<String> supportedOptions,
...
+///
+/// Given a list of [uris] as [String]s from a sourcemap, fixes them up so that
+/// they make sense in a browser context.
+///
+/// - Strips the scheme from the uri
+/// - Strips the top level directory if its not `packages`
+List<String> fixSourceMapSources(List<String> uris) {
+  return uris.map((source) {
+    var uri = Uri.parse(source);
+    // We only want to rewrite multi-root scheme uris.
```
natebosch, line 71: should we check for specific schemes?

build_web_compilers/lib/src/common.dart
```diff
@@ -57,3 +57,22 @@ void validateOptions(Map<String, dynamic> config, List<String> supportedOptions,
         'longer have any effect and should be removed.');
   }
 }
+
+/// Copied to `web/stack_trace_mapper.dart`, these need to be kept in sync.
```
natebosch, line 61: [nit] add a header sentence before the detail about how it is copied. Maybe the next sentence could serve as the header, it could be a bit shorter too.

```dart
/// Fixes up the [uris] from a source map so they make sense in a browser context.
```

natebosch (approved):

build_web_compilers/lib/src/common.dart
jakemac53: Idk, it might just be more brittle tbh. The only schemes we _should_ ever see are the multi root ones. And definitely no scheme other than `http(s)` would be valid in a browser context, which is what this is trying to fix.

PR #2908, opened 2020-11-17
build_config: support the latest release of pkg:yaml

kevmoo:
Pass Uri instead of String when parsing BuildConfig

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -109,7 +109,7 @@ class BuildConfig {
         (map) =>
             BuildConfig.fromMap(packageName, dependencies, map ?? const {}),
         allowNull: true,
-        sourceUrl: configYamlPath,
+        sourceUrl: configYamlPath == null ? null : Uri.file(configYamlPath),
       );
     } on ParsedYamlException catch (e) {
       throw ArgumentError(e.formattedMessage);
--- a/build_config/pubspec.yaml
+++ b/build_config/pubspec.yaml
@@ -21,8 +21,6 @@ dev_dependencies:
   test: ^1.6.0
 
 dependency_overrides:
-  # just for testing...
-  yaml: 3.0.0-nullsafety.0
   # Required due to the tight dependency on this package from these packages
   build_runner:
     path: ../build_runner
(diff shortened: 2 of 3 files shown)


build_config/pubspec.yaml
```diff
@@ -21,6 +21,8 @@ dev_dependencies:
   test: ^1.6.0
 
 dependency_overrides:
+  # just for testing...
```
kevmoo, line 24: Forcing 3.0.0 here – just to sanity check CI

build_config/lib/src/build_config.dart
```diff
@@ -109,7 +109,7 @@ class BuildConfig {
         (map) =>
             BuildConfig.fromMap(packageName, dependencies, map ?? const {}),
         allowNull: true,
-        sourceUrl: configYamlPath,
+        sourceUrl: Uri.file(configYamlPath),
```
jakemac53, line 112: This can be null it looks like, at least in tests

kevmoo: yup...

kevmoo: fixed

kevmoo:
I'm hitting analyzer version issues. I think we've proved that the fix is okay – just hitting bumps w/ the versions of packages

kevmoo:
@jakemac53 – this should work now!

build_config/pubspec.yaml
```diff
@@ -21,6 +21,8 @@ dev_dependencies:
...
+  yaml: 3.0.0-nullsafety.0
```
jakemac53, line 25: I would go ahead and remove this now

kevmoo: Yup! just making sure we were green first

natebosch (approved):

jakemac53 (approved):

build_config/pubspec.yaml
kevmoo: done

build_config/pubspec.yaml
```diff
@@ -21,6 +21,8 @@ dev_dependencies:
...
   # Required due to the tight dependency on this package from these packages
   build_runner:
```
kevmoo: @jakemac53 – should I remove these, too?

PR #2910, opened 2020-11-18
Allow latest build_config in build_runner, remove various dep overrides

jakemac53:
General post publish cleanup CL, removes overrides and prep to publish a build_runner version that allows the latest build_config so people can actually get it :).

Also opts in the `_test_null_safety` e2e package to null safety without the experiment.

Will land as a rebase/merge with separate commits.

### DIFF
--- a/build_config/pubspec.yaml
+++ b/build_config/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_config
-version: 0.4.3-dev
+version: 0.4.3
 description: Support for parsing `build.yaml` configuration.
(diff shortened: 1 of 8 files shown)


build_config/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_config
-version: 0.4.3
+version: 0.4.3-dev
```
natebosch:
```suggestion
version: 0.4.4-dev
```

natebosch (approved):

PR #2915, opened 2020-11-18
Update mono_repo generated output

kevmoo:


kevmoo:
Feel free to squash/merge and delete the branch if you're happy

.github/workflows/dart.yml
```diff
@@ -1,9 +1,10 @@
-# Created with package:mono_repo v3.1.0-beta.2
+# Created with package:mono_repo v3.1.0-beta.3
 name: Dart CI
 on:
   push:
     branches:
-      - $default-branch
+      - main
```
natebosch, line 6: What is this doing?

kevmoo: `$default-branch` is only valid for **templates** – learned that the hard way

Defaulting these to master & main to cover the common default cases.

Unlikely someone will have both. Eliminates the need for a lot of custom logic/config

natebosch: I don't understand, isn't that what this file is? The existing config is working...

natebosch: Synced off-thread. LGTM.

natebosch (approved):

Issue #2933, opened 2020-12-07
Create an FAQ for configuring generate_for to speed up builds


jakemac53:
Often when large projects are experiencing long build times, this is the best solution. Some examples include https://github.com/dart-lang/build/issues/2928, https://github.com/dart-lang/build/issues/2815, https://github.com/dart-lang/build/issues/2861.

We should have an FAQ entry which would allow users to discover this on their own, and we could also link it when people file issues.

PR #2935, opened 2020-12-07
Add section on `generate_for` to the FAQ

jakemac53:
Closes https://github.com/dart-lang/build/issues/2933.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,3 +1,24 @@
+## How can I speed up my build?
+
+Slow builds are often the result of builders that run on all Dart files in your
+package, and analyze them. In this case you can greatly speed up your builds by
+telling those builders exactly which files they should run on.
+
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -1,3 +1,24 @@
+## How can I speed up my build?
```
natebosch, line 1: I worry this title is too general - how about something like "How do I avoid running builders for unnecessary inputs?"

docs/faq.md
```diff
@@ -1,3 +1,24 @@
...
+
+Slow builds are often the result of builders that run on all Dart files in your
+package, and analyze them. In this case you can greatly speed up your builds by
```
natebosch, line 4: [nit] drop "greatly"

natebosch (approved):

PR #2937, opened 2020-12-08
use latest mono_repo, add chat pings, and drop travis

jakemac53:
This is going to fail on the self validate stage until we publish mono_rep, at which point I will re-run.

cc @kevmoo


kevmoo (approved):

natebosch (approved):

kevmoo:
When this lands I'll delete the travis integration hooks, etc

jakemac53:
cc @kevmoo landed!

PR #2940, opened 2020-12-09
build_runner: fix version and SDK version constraint

kevmoo:
Update mono_repo config to test supported SDKs

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -37,16 +37,16 @@ jobs:
       - name: mono_repo self validate
         run: pub global run mono_repo generate --validate
   job_002:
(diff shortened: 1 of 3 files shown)


jakemac53 (commented):
changelog?

kevmoo:
> changelog?

already happened. This just updated the pubspec to match!

natebosch (approved):

jakemac53 (approved):

kevmoo:
Looks like the build_runner e2e test is now failing on 2.10 – this is due to me adding back testing on 2.10

kevmoo:
Thoughts on what we should do there?

jakemac53:
It looks like a test that can never run on 2.10 because it sets 2.12 as the language version, but I don't see any reason it needs to do that. We can just remove the language marker probably.

jakemac53:
https://github.com/dart-lang/build/pull/2941

PR #2941, opened 2020-12-09
Change language marker in test to be an opt out

jakemac53:
This should be consistent with all the sdks we run our tests on.


kevmoo (approved):

natebosch (approved):

jakemac53:
Actually it looks like the easiest thing here will to be enforce it being unsound, which should work on both sdks. Trying that now.

Issue #2946, opened 2020-12-10
Expose a safe api for getting the AST node for an Element


jakemac53:
This would provide a path for most builders currently using the AnalysisSession to migrate off of it entirely, which is the root cause of the InconsistentAnalysisSession errors.

See https://github.com/dart-lang/build/issues/2689, https://github.com/dart-lang/build/issues/2634, https://github.com/rrousselGit/freezed/issues/312.

At a minimum this would allow us to canonicalize the workaround so builder authors don't have to think about it, but I think we can also avoid the issue entirely, I have three possible ideas for avoiding it:

- Use the newish `compilationUnitFor` api to parse the library for the element, and search for the same node in the ast.
  - This would not give you a _resolved_ ast, just a parsed one, I am not sure if that is enough or not.
- Use the session approach but lock all other uses of the resolver  during the call so that the session does not become invalidated.
- See if we can execute the function synchronously, since we can have direct access to the AnalysisDriver instance, we might be able to avoid the problem by not allowing any asynchronous gaps between getting the session and using it to get the ast node.

cc @rrousselGit @davidmorgan @scheglov @natebosch

PR #2947, opened 2020-12-10
Expose a new api to get the AstNode for an Element

jakemac53:
The api is `Future<AstNode> astNodeFor(Element element, {bool resolve: false})`, and is exposed on `Resolver`.

This api can be used avoid using the AnalysisSession directly, which is unsafe to do in Builder code and results in `InconsistentAnalysisException`s. Internally we can avoid this issue due to having synchronous access to the current analysis session.

Fixes https://github.com/dart-lang/build/issues/2946.

cc @rrousselGit @davidmorgan @eernstg I know  you had builders hitting this issue, can try verifying this fix? Does it fully solve all use cases you had for using the `session`? Does it resolve the issue with that?

You can test it out with these dependency overrides:

```
dependency_overrides:
  build:
    git:
      path: build
      ref: safe-ast
  build_resolvers:
    git:
      path: build_resolvers
      ref: safe-ast
```

You may also need to override build_runner_core, build_runner, and build_test (if you depend on that) in a similar manner but I don't _think_ so.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -92,6 +92,10 @@ class PerActionResolver implements ReleasableResolver {
   }
 
   @override
+  Future<AstNode> astNodeFor(Element element, {bool resolve = false}) =>
+      _delegate.astNodeFor(element, resolve: resolve);
+
+  @override
   Future<CompilationUnit> compilationUnitFor(AssetId assetId,
       {bool allowSyntaxErrors = false}) async {
     if (!await _step.canRead(assetId)) throw AssetNotFoundException(assetId);
@@ -149,6 +153,23 @@ class AnalyzerResolver implements ReleasableResolver {
     return source != null &&
         source.exists() &&
         (await _driver.getSourceKind(assetPath(assetId))) == SourceKind.LIBRARY;
+  }
+
+  @override
+  Future<AstNode> astNodeFor(Element element, {bool resolve = false}) async {
+    var session = _driver.currentSession;
+    var library = element.library;
+
+    if (resolve) {
+      return (await session.getResolvedLibraryByElement(library))
+          .getElementDeclaration(element)
+          .node;
+    } else {
+      return session
+          .getParsedLibraryByElement(library)
+          .getElementDeclaration(element)
+          .node;
+    }
   }
 
   @override
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -161,12 +161,12 @@ class AnalyzerResolver implements ReleasableResolver {
     var library = element.library;
 
     if (resolve) {
-      return (await session.getResolvedLibraryByElement(library))
+      return (await session.getResolvedLibrary(library.source.fullName))
           .getElementDeclaration(element)
           .node;
     } else {
       return session
-          .getParsedLibraryByElement(library)
+          .getParsedLibrary(library.source.fullName)
           .getElementDeclaration(element)
           .node;
     }
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -158,17 +158,14 @@ class AnalyzerResolver implements ReleasableResolver {
   @override
   Future<AstNode> astNodeFor(Element element, {bool resolve = false}) async {
     var session = _driver.currentSession;
-    var library = element.library;
+    var path = element.library.source.fullName;
 
     if (resolve) {
-      return (await session.getResolvedLibrary(library.source.fullName))
+      return (await session.getResolvedLibrary(path))
           .getElementDeclaration(element)
           .node;
     } else {
-      return session
-          .getParsedLibrary(library.source.fullName)
-          .getElementDeclaration(element)
-          .node;
+      return session.getParsedLibrary(path).getElementDeclaration(element).node;
     }
   }
 
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -92,7 +92,8 @@ class PerActionResolver implements ReleasableResolver {
   }
 
   @override
-  Future<AstNode> astNodeFor(Element element) => _delegate.astNodeFor(element);
+  Future<AstNode> astNodeFor(Element element, {bool resolve = false}) =>
+      _delegate.astNodeFor(element, resolve: resolve);
 
   @override
   Future<CompilationUnit> compilationUnitFor(AssetId assetId,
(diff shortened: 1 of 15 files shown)


jakemac53:
I will also follow up with a test here

build_resolvers/lib/src/resolver.dart
```diff
@@ -151,6 +155,23 @@ class AnalyzerResolver implements ReleasableResolver {
...
   }
 
+  @override
+  Future<AstNode> astNodeFor(Element element, {bool resolve = false}) async {
+    var session = _driver.currentSession;
+    var library = element.library;
+
+    if (resolve) {
+      return (await session.getResolvedLibraryByElement(library))
+          .getElementDeclaration(element)
```
jakemac53, line 165: @scheglov is it possible for this call to fail with an `InconsistentAnalysisException`? I could drop support for resolved ast nodes if it might.

davidmorgan:
Thanks! Will aim to take a look tomorrow.

jakemac53:
Going to need to push some fixes here it looks like this doesn't quite work, ill ping once I have the tests passing

jakemac53:
Ok this should be good to go now

build_resolvers/lib/src/resolver.dart
natebosch: I would be in favor of starting without support for resolved nodes unless we know someone is asking for it today.

build_resolvers/test/resolver_test.dart
```diff
@@ -680,4 +681,52 @@ int? get x => 1;
...
+        var lib = await resolver.libraryFor(entryPoint);
+        var x = lib.topLevelElements.firstWhere((x) => !x.isSynthetic);
+        expect(x.name, 'x');
+        expect(await resolver.isLibrary(AssetId('a', 'web/other.dart')), true);
+
+        // Validate that direct session usage would throw
+        expect(() => lib.session.getParsedLibraryByElement(x.library),
+            throwsA(isA<InconsistentAnalysisException>()));
+
+        var astNode = await resolver.astNodeFor(x);
```
natebosch, line 726: We should have a version of this same test that passes `resolve: true` since that is the case we are unsure about.

natebosch (approved):

build_resolvers/test/resolver_test.dart
jakemac53: Added, although it doesn't actually check the behavior we care about.

I added an additional test that copies the code and does check that behavior also.

build_resolvers/lib/src/resolver.dart
jakemac53: I added a test and it seems fine to use these results after the session is invalidated

jakemac53:
Alternatively to this, I was just thinking it might make more sense to just expose the `ParsedLibraryResult` and/or `ResolvedLibraryResult` directly. These objects seem safe to use after the session is invalidated, and that would add more flexibility.

jakemac53:
@scheglov is it much more expensive to get the resolved ast nodes here as opposed to just the parsed one, assuming we already have a `LibraryElement` and so it has already been resolved? Is it just plugging in the `definingElement` fields or is it recreating those elements?

build_resolvers/lib/src/resolver.dart
scheglov: As it is now, `getElementDeclaration` never fails with `InconsistentAnalysisException`. Parsed/ResolvedLibraryResult has all its parsed/resolved units, and its own copy of the element model.

But ideally my dream is to keep a single copy of the element model, and do fine grained patching as some declarations in files change. This will be a breaking change of course.

scheglov:
> @scheglov is it much more expensive to get the resolved ast nodes here as opposed to just the parsed one, assuming we already have a `LibraryElement` and so it has already been resolved? Is it just plugging in the `declaredElement` fields on the nodes or is it recreating those elements?

Resolution is much more expensive. Yes, you apply existing (or resynthesized from the same source) `LibraryElemennt` to declarations. But there are more than declarations in files - all method bodies also resolved.

jakemac53:
> Resolution is much more expensive. Yes, you apply existing (or resynthesized from the same source) `LibraryElemennt` to declarations. But there are more than declarations in files - all method bodies also resolved.

Ah ok thanks, that makes sense. I think we will just support the unresolved AST for now then, and we can add an option for a resolved AST later if somebody requires it.

rrousselGit:
I've tested this and can confirm that it removes the exception – at least from what I've seen

eernstg:
@jakemac53 wrote:
> I think we will just support the unresolved AST for now then, and we
> can add an option for a resolved AST later if somebody requires it.

Reflectable encounters the `InconsistentAnalysisException` when it obtains the `getResolvedLibraryByElement`, and the results obtained from that library are used for further lookups. For example, we need the syntactic information in order to be able to generate code for an expression that yields the same value as a given constant in client code, and we need the resolution data in order to know which libraries to import into the generated library, such that this expression is well defined. I believe that we cannot do these things based on a feature that only delivers a library that hasn't been resolved.

jakemac53:
@eernstg you still have the resolved `Element` model, just not the fully resolved AST. I thought that `reflectable` only looks at annotations, which would be available on the `Element` model?

eernstg:
There is a difference between (1) looking at metadata in order to determine whether a given declaration (say, a class) should be supported by reflection related data, and (2) supporting reflective access to metadata (as requested by `metadataCapability`).

For (2), I believe we have to create an expression in the generated library that evaluates to the same object as the value of the given metadata. Creating "the same expression" in a different library is the core of the task that requires access to both syntax and resolution data.

jakemac53:
@eernstg Ah just to confirm here -  you are saying that you need this in order to re-construct annotations inside of code generated by `reflectable` such that their values can be retrieved and returned (if `metadataCapability` is enabled at least)?

Presumably you need to know what the required imports would be for all nodes and can't just copy in the ast? Or maybe even copy the actual values of certain items because they might be private to the library?

eernstg:
> you need this in order to re-construct annotations inside of code
> generated by reflectable such that their values can be retrieved and returned

Exactly, and that does potentially require the addition of import prefixes to several subexpressions.

> copy the actual values of certain items because they might be private

Yes, that may happen as well. In a number of cases reflectable simply reports that private access isn't supported, but we do handle a few special cases.

jakemac53:
I will go ahead and add back the resolved AST option and land this

jakemac53:
This is now published in package:build 1.6.0, feel free to start migrating any usage and removing your custom workarounds!

build_resolvers/lib/src/resolver.dart
eernstg: Using this from reflectable, I get null errors at `.node`. The code looks like what I have in reflectable now, except for `?.node` at this point rather than the plain `.node`.

eernstg: Looks like line 168 could have the same issue.

jakemac53: I thought I had checked this case and this returned essentially an "empty" element declaration, and not `null`. But I certainly might have missed something here, I will add the `?.`.

PR #2954, opened 2020-12-16
Use ?. on the element declaration in astNodeFor

jakemac53:
cc @eernstg who was seeing an issue in reflectable here.

In my testing I had thought that you got an empty element declaration back and not null but I must have missed some case.

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.5.1
+
+- Fix a potential nullability issue in the `astNodeFor` api.
+
 ## 1.5.0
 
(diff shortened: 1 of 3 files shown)


natebosch (approved):

natebosch:
If we could find an example and add a test that would be good, not a huge deal if that's hard to write.

jakemac53:
> If we could find an example and add a test that would be good, not a huge deal if that's hard to write.

Unfortunately I couldn't figure one out, but perhaps @eernstg can suggest one, since reflectable does apparently trigger this :).

Going to land as I want to get this out before the winter break, but happy to follow up and add a test here.

PR #2944, opened 2020-12-10
Migrate off of $test$ and $web$

natebosch:
Towards #2575

PR #2956, opened 2020-12-16
Add $package$ to default generate_for

natebosch:
This was missed in #2944

After changing the input extension from `$test$` to `$package$` it is
critical to also include it in the default `generate_for`, otherwise the
builder will not run.

### DIFF
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -23,7 +23,8 @@ final _builders = <_i1.BuilderApplication>[
       [_i2.debugIndexBuilder, _i2.debugTestBuilder, _i2.testBootstrapBuilder],
       _i1.toRoot(),
       hideOutput: true,
-      defaultGenerateFor: const _i3.InputSet(include: [r'test/**'])),
+      defaultGenerateFor:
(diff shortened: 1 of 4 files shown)


jakemac53 (commented):
How do we want to manage this and the bug in build_runner? This will break old users....

natebosch:
> How do we want to manage this and the bug in build_runner? This will break old users....

We could bump the constraint on `build` to `^1.6.0` even though it doesn't use the new API. This would limit the range of `build_runner` to just `1.10.10` and `1.10.11`, since there is no difference between them otherwise a `pub upgrade` should always work to get compatible.

This still leaves a broken `build_test` where the `index.html` file won't work for some subset of users, but that's a smaller issue than a compile error with the generated build script... WDYT?

jakemac53:
That is probably a reasonable solution

kevmoo:
Might be nice to have test validation so this doesn't regress again...right? 🤷

natebosch:
> Might be nice to have test validation so this doesn't regress again...right? 🤷

Yeah we can file an issue. It's not super pressing and I don't want to block a fix on it.

jakemac53 (approved):

Issue #2959, opened 2020-12-19
Issue In abstract class Builder Documentation


ZainUrRehmanKhan:
Issue in Documentation:
     - [Build](https://github.com/dart-lang/build/blob/master/build/README.md)
     

**Actual:** 
In Section `Implementing your own Builders`:

![image](https://user-images.githubusercontent.com/67098758/102682665-26445880-41ed-11eb-94a9-f4da0a987eb6.png)
![image](https://user-images.githubusercontent.com/67098758/102682899-c189fd80-41ee-11eb-9b6b-65d54ba15549.png)

**Expected:**
This should be FurtureOr like in the Source Code: 

`FutureOr` in place of `Future`

Thank you for taking the time to file an issue!
 - Not a Problem.

PR #2960, opened 2020-12-19
fix: Issue In abstract class Builder Documentation 

ZainUrRehmanKhan:
Related Issue:
https://github.com/dart-lang/build/issues/2959

**Actual:** 
In Section `Implementing your own Builders`:

image
image

**Expected:**
This should be FurtureOr like in the abstract class: 

`FutureOr` in place of `Future`

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -50,7 +50,7 @@ class CopyBuilder implements Builder {
   };
 
   @override
-  Future<void> build(BuildStep buildStep) async {
+  FutureOr<void> build(BuildStep buildStep) async {
     // Each `buildStep` has a single input.
     var inputId = buildStep.inputId;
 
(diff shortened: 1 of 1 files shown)


build/README.md
```diff
@@ -50,7 +50,7 @@ class CopyBuilder implements Builder {
   };
 
   @override
-  Future<void> build(BuildStep buildStep) async {
+  FutureOr<void> build(BuildStep buildStep) async {
```
kevmoo, line 53: It's totally valid to change the return type to `Future<void>` since this is an async function – it awaits something.

Making a return type more specific in an override is fine.

ZainUrRehmanKhan: Yeah, you are absolutely right! It's fine to have a return type more specific in an override. But for a new user, it's confusing to have the implementation with `Future`

natebosch: > But for a new user, it's confusing to have the override with `Future`

It's always a bit of a struggle to choose between writing idiomatic code for the purposes of serving as a good example and something which can be copied, vs writing non-idiomatic code that elaborates things more than necessary.

In this case `Future<void>` was chosen over `FutureOr<void>` because it's more idiomatic to write it that way - as a definition of an signature that must be implemented `FutureOr` can be friendly, however as a definition of a concrete method signature it's preferable to tighten that to either `Future<void>` or `void` as appropriate. It is uncommon to have a synchronous builder so our examples focus on async builders, but they are possible so we do allow them.

I think for this example we should keep it as is since it will look more like the code that users will write.

PR #2965, opened 2021-01-04
Return empty for findAssets on missing package

### LINKED ISSUE #2923: Have an option in `PackageAssetReader` to return an empty Stream instead of throwing
https://github.com/dart-lang/build/blob/8fbdb405f17ff20a8391cb454d40e8cc32fe2883/build_test/lib/src/package_reader.dart#L112-L115

This makes it possible to use this in `testBuilder` where some packages are in-memory.

It's super nice to be able to just pass in trivial `sourceAssets` to `testBuilder`. But in my case, I need those files to import packages, which fails without also providing a `MultiPackageAssetReader` to the `reader` param.

`testBuilder` already combines the provided reader with other readers here.

https://github.com/dart-lang/build/blob/8fbdb405f17ff20a8391cb454d40e8cc32fe2883/build_test/lib/src/test_builder.dart#L180-L184

I just need it to not throw so the `inMemoryReader` result can be used!

...the alternative – using `test_descirptor` greatly increases the code complexity.

natebosch:
We have checks in the real build system to keep packages on the rails.
Those checks don't necessarily need to be included in test code and for
fake environments within the test.

Closes #2923

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.3.3
+
+- Return an empty stream instead of throwing from
+  `PackageAssetReader.findAssets` when passed a non-existing package name.
+
 ## 1.3.2
 
 - Fix the `test/index.html` not generating by default.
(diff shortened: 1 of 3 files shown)


kevmoo:
Ship it!

kevmoo (approved):

jakemac53 (approved):

PR #2967, opened 2021-01-05
allow the null safe built_* package versions

jakemac53:
This overrides the deps to point at the git repo since these are not yet published for now.

A few open questions before we land/publish though:

- Is the generated code from the null safe built_value_generator compatible with the older (7.x.x) built_value and built_collection versions? If not we won't be able to do a range like this.
- Do we want to publish built_value/built_value_generator or this package first? One or the other will be in a bit of a weird state at least initially (publishing a version that nobody can get immediately), due to the dependency cycle.
- Who should own migrating `code_builder`? That also uses built_value/built_collection and we also depend on it.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 2.1.5
+
+- Allow the null safe built_collection, built_value, and built_value_generator
+  packages, and regenerate serialization code with them. 
+
 # 2.1.4
 
 - Remove dependency on `package:package_resolver`.
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,7 +1,8 @@
-# 2.1.5
+# 2.1.5-dev
 
-- Allow the null safe built_collection, built_value, and built_value_generator
-  packages, and regenerate serialization code with them. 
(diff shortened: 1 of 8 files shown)


jakemac53:
Tests answered the question for me, lol https://github.com/dart-lang/build/pull/2967/checks?check_run_id=1651376388. So the newly generated code is _not_ compatible.

I will try the old generated code with the new libs next.

jakemac53:
It looks like the new libs with the old generated code does work - so we can go ahead with that if @davidmorgan promises not to break that compatibility :).

natebosch:
I'm OK with the approach as long as @davidmorgan is.

davidmorgan (approved):

davidmorgan:
Yes, old generated code will continue to work with new `built_value`; I hadn't thought about it before, but I guess that's going to be needed by other packages with generated code checked in, otherwise they won't be able to depend on ranges and everything will be very stuck.

Thanks.

jakemac53:
I realized that there is a published version of built_value so I am trying CI with that and then I can remove the overrides and land this assuming it is all good.

natebosch (approved):

davidmorgan:
Yes, should have mentioned built_value was published. LG, thanks!

PR #2972, opened 2021-01-13
Support custom build scripts

### LINKED ISSUE #2963: Move parts of build script generation to build_runner_core
I'm trying to setup a semi-custom build environment to embed Dart builds in another build system. For this, I basically want the regular `build_runner` behavior, expect that I need a custom `BuildEnvironment` and build configuration overrides. At the moment, I have to write my own build scripts and manage the builder applications by hand.

It would be great if we could have an api in `build_runner_core` that takes a package graph and overridden build configurations and returns everything necessary to generate custom build scripts. Some mechanism to detect when a build script needs to be invalidated would be great too, but I'm not sure how feasible that is when allowing custom build configurations.

I'm happy to work on this if you agree that this is something worth having.

simolus3:
This adds the `BuildScriptGenerationOptions` class, which can be extended to customize how build scripts are generated. `generateAndRun` and `generateBuildScript` optionally take an instance of that class.

I also export `generateAndRun` now, since I think that behavior is generally useful for build runners. I added another argument to support custom branding on unexpected crashes.

I'm not sure what the best way to test this would look like. I added a very simple test that checks imports and custom builder definitions. If we want some kind of integration test for that I can add one too, but I didn't know where to put one.

Closes #2963

### DIFF
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -13,10 +13,9 @@ import 'package:build_modules/builders.dart' as _i5;
 import 'package:build_vm_compilers/builders.dart' as _i6;
 import 'package:build_web_compilers/builders.dart' as _i7;
 import 'package:build/build.dart' as _i8;
-import 'dart:async' as _i9;
-import 'dart:isolate' as _i10;
-import 'package:build_runner/build_runner.dart' as _i11;
-import 'dart:io' as _i12;
+import 'dart:isolate' as _i9;
+import 'package:build_runner/build_runner.dart' as _i10;
+import 'dart:io' as _i11;
 
 final _builders = <_i1.BuilderApplication>[
   _i1.apply(
@@ -139,8 +138,8 @@ final _builders = <_i1.BuilderApplication>[
   _i1.applyPostProcess(
       r'provides_builder:some_post_process_builder', _i4.somePostProcessBuilder)
 ];
-_i9.Future<void> main(List<String> args, [_i10.SendPort sendPort]) async {
-  var result = await _i11.run(args, _builders);
+void main(List<String> args, [_i9.SendPort sendPort]) async {
+  var result = await _i10.run(args, _builders);
   sendPort?.send(result);
-  _i12.exitCode = result;
+  _i11.exitCode = result;
 }
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -13,9 +13,10 @@ import 'package:build_modules/builders.dart' as _i5;
 import 'package:build_vm_compilers/builders.dart' as _i6;
 import 'package:build_web_compilers/builders.dart' as _i7;
 import 'package:build/build.dart' as _i8;
-import 'dart:isolate' as _i9;
-import 'package:build_runner/build_runner.dart' as _i10;
-import 'dart:io' as _i11;
+import 'dart:async' as _i9;
+import 'dart:isolate' as _i10;
+import 'package:build_runner/build_runner.dart' as _i11;
+import 'dart:io' as _i12;
 
 final _builders = <_i1.BuilderApplication>[
   _i1.apply(
@@ -138,8 +139,8 @@ final _builders = <_i1.BuilderApplication>[
   _i1.applyPostProcess(
       r'provides_builder:some_post_process_builder', _i4.somePostProcessBuilder)
 ];
-Future<void> main(List<String> args, [_i9.SendPort sendPort]) async {
-  var result = await _i10.run(args, _builders);
+_i9.Future<void> main(List<String> args, [_i10.SendPort sendPort]) async {
+  var result = await _i11.run(args, _builders);
   sendPort?.send(result);
-  _i11.exitCode = result;
+  _i12.exitCode = result;
 }
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -138,7 +138,7 @@ final _builders = <_i1.BuilderApplication>[
   _i1.applyPostProcess(
       r'provides_builder:some_post_process_builder', _i4.somePostProcessBuilder)
 ];
-void main(List<String> args, [_i9.SendPort sendPort]) async {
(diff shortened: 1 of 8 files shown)


build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -18,13 +18,26 @@ final _logger = Logger('Bootstrap');
...
 /// Returns the exit code from running the build script.
 ///
 /// If an exit code of 75 is returned, this function should be re-ran.
-Future<int> generateAndRun(List<String> args, {Logger logger}) async {
+Future<int> generateAndRun(
+  List<String> args, {
+  Logger logger,
+  BuildScriptGenerationOptions generationOptions =
+      const BuildScriptGenerationOptions(),
+  void Function(Stdout output) printUncaughtError,
```
jakemac53, line 37: This is a bit of a weird api - really it just looks like it is allowing you to customize the message?

Should we change it to `void Function(Object error, StackTrace stackTrace) handleUncaughtError` which would be the more typical way of allowing users to handle uncaught errors?

I would probably implement that like:

```dart
handleUncaughtError ??= (e, _) {
  printUncaughtError(stderr);
  stderr..writeln(error)..writeln(Trace.parse(trace).terse);
};
...
errorListener = errorPort.listen((e) {
  final error = e[0];
  final trace = Trace.parse(e[1] as String).terse;
  handleUncaughtError(error, trace);
  if (scriptExitCode == 0) scriptExitCode = 1;
});
```

Possibly also allowing a function with a single error argument but im not sure its worth the complexity and loss of static validation on the function type.

simolus3: > Should we change it to `void Function(Object error, StackTrace stackTrace) handleUncaughtError` which would be the more typical way of allowing users to handle uncaught errors?

Yes, definitely. I just wanted to be able to remove the "Report this at dart-lang/build" message and missed the obvious way.

jakemac53 (approved):
cc @natebosch any concerns before we land this?

@simolus3 can you also update the pubspec/changelog :)

jakemac53:
also @simolus3 for the failing golden test you can run `cd _test && pub run build_runner generate-build-script && cp .dart_tool/build/entrypoint/build.dart test/goldens/generated_build_script.dart` to update the golden

_test/test/goldens/generated_build_script.dart
```diff
@@ -13,9 +13,10 @@ import 'package:build_modules/builders.dart' as _i5;
 import 'package:build_vm_compilers/builders.dart' as _i6;
 import 'package:build_web_compilers/builders.dart' as _i7;
 import 'package:build/build.dart' as _i8;
-import 'dart:isolate' as _i9;
-import 'package:build_runner/build_runner.dart' as _i10;
-import 'dart:io' as _i11;
+import 'dart:async' as _i9;
```
natebosch, line 16: Why do we need this import?

_test/test/goldens/generated_build_script.dart
```diff
@@ -138,8 +139,8 @@ final _builders = <_i1.BuilderApplication>[
   _i1.applyPostProcess(
       r'provides_builder:some_post_process_builder', _i4.somePostProcessBuilder)
 ];
-void main(List<String> args, [_i9.SendPort sendPort]) async {
-  var result = await _i10.run(args, _builders);
+_i9.Future<void> main(List<String> args, [_i10.SendPort sendPort]) async {
```
natebosch, line 142: Why did this return type need to change?

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -21,23 +21,79 @@ const scriptSnapshotLocation = '$scriptLocation.snapshot';
...
-    return DartFormatter().format('''
+/// Options to generate custom build scripts.
+///
+/// The build script is a generated Dart file in [scriptLocation]. It contains
+/// all builder definitions part of the build and is used by the top-level
+/// `build_runner` commands to run a build with the correct builders.
+///
+/// By providing custom generation options, it's possible to tweak parts of a
+/// build script.
+class BuildScriptGenerationOptions {
```
natebosch, line 32: This class feels a little clumsy to me. I'm not a fan of enumerating each bit of customization like this. I would prefer to build smaller blocks that can be put together, rather than a large block that allows fine-grained tweaks to small details.

My preference would be to do something like

```dart
Future<int> generateAndRun(
  List<String> args, {
  Logger logger,
  Future<String> Function() generateBuildScript = generateBuildScript,
  void Function(Object error, StackTrace stackTrace) handleUncaughtError,
}) async {
```

Then we can make `Future<Iterable<Expression>> _findBuilderApplications()` public and the consuming code could use that along with the APIs from `code_builder` to generate any type of build script they want.

We can also make the `PackageGraph`, and `Map<String, BuildConfig>` arguments to `findBuilderApplications` and let them be passed in rather than plumb through a `Future<PackageGraph> Function()` and `Future<Map<String, BuildConfig>> Function(PackageGraph, RunnerAssetReader)`.

This would impose a cost of some extra complexity to consumers - they will have to write the `main` themselves - but it gives maximal flexibility.

@jakemac53 WDYT?

_test/test/goldens/generated_build_script.dart
simolus3: I thought it might be more appropriate given the `async` modifier, but now that we have full customization over the build script, that doesn't matter much to me anymore. I've reverted the change.

build_runner/lib/src/build_script_generate/build_script_generate.dart
jakemac53: I don't have a strong opinion here - I did find the class a little bit surprising (more in the way that you implement it as opposed to passing it the options), but I was also OK with it.

simolus3:
Is there an update on this? I think I've addressed all review comments now. I replaced the options class with a `generateBuildScript` parameter as Nate suggested.

jakemac53:
Sorry just forgot about it, thanks for the ping :)

jakemac53:
@natebosch you happy with this as is now?

natebosch (approved):

PR #2974, opened 2021-01-13
Allow null safe stream_transform

### LINKED ISSUE #2973: build_test: update dependency on stream_transform to allow the latest null safe version
Specifically 2.0.0-nullsafety.0

This is blocking our external angular roll

natebosch:
Closes #2973


kevmoo (approved):
Thanks! Let's just be super careful about breaking changes

natebosch:
> Let's just be super careful about breaking changes

Nothing breaking here. Just the migration.

PR #2985, opened 2021-02-01
Fix AssetId.resolve for empty uris

### LINKED ISSUE #2984: Web compilers fail on libraries importing themselves
The build system can't compile Dart files importing themselves via an empty import.

### How to reproduce

Run `stagehand web-simple`. Then, replace `web/main.dart` with 

```dart
import 'dart:html';

import '' as self;

void main() {
  querySelector('#output').text = 'Your Dart app is running.';
  self.foo();
}

void foo() {}
```

Both `dart2js` and `dartdevc` happily compile that file, and there are no analysis errors, but a build fails with

```
[SEVERE] build_web_compilers:entrypoint on web/main.dart:
Unable to find modules for some sources, this is usually the result of either a
bad import, a missing dependency in a package (or possibly a dev_dependency
needs to move to a real dependency), or a build failure (if importing a
generated file).

Please check the following imports:

`import '' as self;` from build_repro|web/main.dart at 3:1
```

This example is obviously nonsense, but self-imports are useful to call a top-level function in a class which has a method with the same name. It's valid Dart too, so the builders shouldn't fail.

### Package versions

I'm using the following packages that might be relevant:

- `build_web_compilers`: 2.16.1
- `build_modules`: 3.0.3

simolus3:
An empty uri resolves to whatever it's resolved against, but `AssetId.resolve` would resolve it to its parent.

Closes #2984.


build_modules/test/meta_module_test.dart
```diff
@@ -447,4 +447,21 @@ void main() {
...
+    var assets = makeAssets({
+      'myapp|lib/a.dart': '''
+            import 'a.dart';
+          ''',
+    });
+
+    var a = AssetId('myapp', 'lib/a.dart');
+
+    var expectedModules = [
+      matchesModule(Module(a, [a], [], defaultPlatform, true)),
```
simolus3, line 461: @jakemac53 I made sure that this module importing itself does not have any dependencies. Does that clear your concerns? I didn't see any additional errors in manual testing either.

jakemac53: Testing that it doesn't list itself in deps is sufficient yes :)

jakemac53 (approved):

PR #2990, opened 2021-02-04
Allow nullsafety version(s) of bazel_worker

aaronlademann-wf:
@jakemac53 my apologies if I'm jumping the gun here, but we're working through some `2.12.x` dev readiness items at Workiva, and we've run into an issue where we can't upgrade past `analyzer 0.39.x` on the `2.12.x` dev build of the SDK unless `build_modules`, and subsequently `build_web_compilers` are updated to support the nullsafety release(s) of `bazel_worker`.

When using these version constraints locally - `pub get` resolves successfully.

If there are other work items that are needed besides version range changes - please feel free to close this PR without explanation :)

### DIFF
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_modules
-version: 3.0.3
+version: 3.0.4
 description: Builders for Dart modules
 repository: https://github.com/dart-lang/build/tree/master/build_modules
 
@@ -9,7 +9,7 @@ environment:
 dependencies:
   analyzer: ">0.36.4 <0.42.0"
   async: ^2.0.0
-  bazel_worker: ^0.1.20
+  bazel_worker: ">=0.1.18 <2.0.0"
   build: ">=1.3.0 <2.0.0"
   build_config: ">=0.3.0 <0.5.0"
   collection: ^1.0.0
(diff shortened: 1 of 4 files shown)


build_modules/pubspec.yaml
```diff
@@ -9,7 +9,7 @@ environment:
 dependencies:
   analyzer: ">0.36.4 <0.42.0"
   async: ^2.0.0
-  bazel_worker: ^0.1.20
+  bazel_worker: ">=0.1.18 <2.0.0"
```
jakemac53, line 12:
```suggestion
  bazel_worker: ">=0.1.20 <2.0.0"
```

Unless you need 0.1.18 to be selectable? Or we can try to confirm that 0.1.18 does work, I don't know if this was bumped unnecessarily or not.

jakemac53:
This is fine - we aren't in a rush to update these ourselves (its a bit of whack-a-mole and not a lot of concrete benefit for us) but if you put in the work to bump them and send prs then I am happy to review/land/publish them. Same goes for any of our other deps that you need bumped.

If you file issues we will also eventually bump them (probably in <=1 week) - but its definitely faster if you send prs like this.

build_modules/pubspec.yaml
aaronlademann-wf: oh woops - thanks.

aaronlademann-wf:
Awesome @jakemac53! I restored the lower bound of `bazel_worker` that I accidentally changed...

jakemac53 (approved):
I can land/publish then when I'm at my computer next

Issue #1899, opened 2018-11-08
Make `AssetId.parse` harder to misuse


natebosch:
The `AssetId.parse` constructor only handles the format created by `AssetId.toString()` and was mostly meant as a cheaper serialization format when writing to json. It was never really meant to be used outside of build system implementations that deal with serializing large numbers of IDs.

Unfortunately it's leaked a bit into some build users. For instance Angular has a dependency on the fact that the `Resolver` returns elements where the `element.source.fullName` is a serialized `AssetId`. We happen to implement it that way in our resolvers:

https://github.com/dart-lang/build/blob/0e79b63c6387adbb7e7f4c4f88d572b1242d24df/bazel_codegen/lib/src/summaries/build_asset_uri_resolver.dart#L52-L53

https://github.com/dart-lang/build/blob/0e79b63c6387adbb7e7f4c4f88d572b1242d24df/build_resolvers/lib/src/resolver.dart#L266-L267

And Angular depends on it here:

https://github.com/dart-lang/angular/blob/cbd6150a145457cb8580e61105506ce03ee6895c/angular_compiler/lib/src/outliner/collect_type_parameters.dart#L61

I propose we either:

- Deprecate `AssetId.parse` and transition users to `AssetId.resolve`, make `AssetId.resolve` handle `package|path` serialization format for backwards compatibility, and transition serialization to use `package:` or `asset:` URIs using `AssetId.uri` and `AssetId.resolve` to serialize/deserialize. This would come with some efficiency losses but I don't think they'll have a huge impact. Our asset graph serialization already uses something different and even more space efficient where it doesn't repeat package names. In the next breaking change release for `build` we'd change `toString()` to use `.uri.toString()` since that format is more useful to humans.
- Make `AssetId.parse` handle `asset:` and `package:` URI by forwarding to `AssetId.resolve` when it doesn't have a `|` character. We wouldn't need to deprecate it and could live with both APIs being nearly interchangeable except where there is a `from` argument for relative imports. We could leave existing serialization usage as is. We had originally proposed expanding the allowed format for `AssetId.parse` in https://github.com/dart-lang/build/issues/287 and decided instead to implement as a separate `resolve` API, but don't have a recorded discussion on why we made that choice.

cc @jakemac53 @matanlurey

natebosch:
For completeness I should note that I'm pretty sure it was me who recommended this approach in the angular builder. 😳

jakemac53:
[...]

PR #2996, opened 2021-02-08
Migrate user-facing apis to null-safety

### LINKED ISSUE #2920: Add support for nullsafety
I am upgrading my package (Kiwi) vanlooverenkoen/kiwi#51

But it seems that build & build_runner is not yet migrated to support nullsafety. Is there a timeline on when this could be added?

simolus3:
Migrate packages typically imported by build authors to null-safety (part 1 /3). Closes #2920

In this PR, I migrate

- the `build` package
- the `build_test` package (or rather, the parts of that package that don't import `build_resolvers`)
- the `scratch_space` package

After the `graphs` package is migrated, I'll open another PR to migrate the `build_resolvers` package. When that one is done too, we can start to opt the tests of `build` and `build_test` into null-safety and publish, thus enabling build authors to write and test their builders with strong null-safety. 
Migrating the rest of the build system is not a goal for the near future, so we won't run builds with strong null-safety.

### DIFF
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -5,7 +5,7 @@ environment:
   sdk: ">=2.3.0 <3.0.0"
 
 dev_dependencies:
-  analyzer: ^0.42.0-nullsafety.0
+  analyzer: ^1.0.0
   path: ^1.4.2
   test: ^1.6.2
   test_process: ^1.0.4
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -5,7 +5,7 @@ environment:
   sdk: ">=2.3.0 <3.0.0"
 
 dev_dependencies:
-  analyzer: ^0.41.0
+  analyzer: ^0.42.0-nullsafety.0
   path: ^1.4.2
   test: ^1.6.2
   test_process: ^1.0.4
@@ -13,6 +13,7 @@ dev_dependencies:
     path: pkgs/provides_builder/
 
 dependency_overrides:
+  analyzer: ^0.42.0-nullsafety.0 # pkg:test doesn't support the latest analyzer yet
   build:
     path: ../build
   build_config:
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -17,7 +17,6 @@ dev_dependencies:
   build_test: any
 
 dependency_overrides:
-  analyzer: ^0.42.0-nullsafety.0 # pkg:test doesn't support the latest analyzer yet
   build:
     path: ../build
   build_config:
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -11,6 +11,10 @@ dev_dependencies:
   test_process: ^1.0.4
   provides_builder:
     path: pkgs/provides_builder/
+  build_vm_compilers: any
+  build_web_compilers: any
+  build_runner: any
+  build_test: any
 
 dependency_overrides:
   analyzer: ^0.42.0-nullsafety.0 # pkg:test doesn't support the latest analyzer yet
--- a/build/lib/src/generate/run_builder.dart
+++ b/build/lib/src/generate/run_builder.dart
@@ -30,21 +30,20 @@ import 'expected_outputs.dart';
 /// function with the associated primary input.
 Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
     AssetReader reader, AssetWriter writer, Resolvers resolvers,
-    {Logger logger,
-    ResourceManager resourceManager,
-    @Deprecated('The rootPackage argument is unused') String rootPackage,
+    {Logger? logger,
+    ResourceManager? resourceManager,
     StageTracker stageTracker = NoOpStageTracker.instance,
-    void Function(AssetId input, Iterable<AssetId> assets)
+    void Function(AssetId input, Iterable<AssetId> assets)?
         reportUnusedAssetsForInput}) async {
   var shouldDisposeResourceManager = resourceManager == null;
-  resourceManager ??= ResourceManager();
+  final resources = resourceManager ?? ResourceManager();
   logger ??= Logger('runBuilder');
   //TODO(nbosch) check overlapping outputs?
   Future<void> buildForInput(AssetId input) async {
     var outputs = expectedOutputs(builder, input);
     if (outputs.isEmpty) return;
     var buildStep = BuildStepImpl(
-        input, outputs, reader, writer, resolvers, resourceManager,
+        input, outputs, reader, writer, resolvers, resources,
         stageTracker: stageTracker,
         reportUnusedAssets: reportUnusedAssetsForInput == null
             ? null
@@ -59,7 +58,7 @@ Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
   await scopeLogAsync(() => Future.wait(inputs.map(buildForInput)), logger);
 
   if (shouldDisposeResourceManager) {
-    await resourceManager.disposeAll();
-    await resourceManager.beforeExit();
+    await resources.disposeAll();
+    await resources.beforeExit();
   }
 }
--- a/build/lib/src/generate/run_builder.dart
+++ b/build/lib/src/generate/run_builder.dart
@@ -29,7 +29,7 @@ import 'expected_outputs.dart';
 /// `BuildStep.reportUnusedAssets` in [builder] will be forwarded to this
 /// function with the associated primary input.
 Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
-    AssetReader reader, AssetWriter writer, Resolvers resolvers,
+    AssetReader reader, AssetWriter writer, Resolvers? resolvers,
     {Logger? logger,
     ResourceManager? resourceManager,
     StageTracker stageTracker = NoOpStageTracker.instance,
--- a/build/test/generate/run_builder_test.dart
+++ b/build/test/generate/run_builder_test.dart
@@ -1,6 +1,7 @@
 // Copyright (c) 2017, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+// @dart=2.9
 @TestOn('vm')
 import 'dart:async';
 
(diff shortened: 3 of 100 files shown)


build_test/lib/src/resolve_source.dart
```diff
@@ -118,13 +118,13 @@ Future<T> resolveSource<T>(
 /// - by default, [PackageAssetReader.currentIsolate]. A custom [packageConfig]
 /// may be provided to map files not visible to the current package's runtime.
 Future<T> resolveSources<T>(
-  Map<String, String> inputs,
+  Map<String, String>? inputs,
```
jakemac53, line 121: Lets go ahead and make this non-nullable - it was already doing a null check right away :).

You can use `ArgumentError.checkNotNull` to retain the check for legacy codebase interactions, if desired (but I wouldn't bother)

build_test/lib/src/stub_reader.dart
```diff
@@ -22,7 +22,8 @@ class StubAssetReader extends AssetReader implements MultiPackageAssetReader {
       Future.value(null);
 
   @override
-  Stream<AssetId> findAssets(Glob glob, {String package}) => null;
+  Stream<AssetId> findAssets(Glob glob, {String? package}) =>
```
jakemac53, line 25: I am confused how the other methods don't need to be updated?

build_test/lib/src/test_bootstrap_builder.dart
```diff
@@ -7,7 +7,7 @@ import 'dart:async';
 import 'package:build/build.dart';
 import 'package:crypto/crypto.dart';
 import 'package:path/path.dart' as p;
-// ignore: deprecated_member_use
+// ignore: deprecated_member_use, import_of_legacy_library_into_null_safe
```
jakemac53, line 10: We don't want to be ignoring any of these at least not under `lib` - we can push on migrating `test_core/backend.dart` sooner than later, but for now I would opt out this file and any file that imports it.

Otherwise when that library does migrate it could break this library.

build_test/pubspec.yaml
```diff
@@ -4,26 +4,31 @@ version: 1.3.6
 homepage: https://github.com/dart-lang/build/tree/master/build_test
 
 environment:
-  sdk: ">=2.9.0 <3.0.0"
+  sdk: ">=2.12.0-0 <3.0.0"
 
 dependencies:
   async: ">=1.2.0 <3.0.0"
```
jakemac53, line 10: We should up all deps to require the null safe versions (we are likely relying on them in practice)

```suggestion
  async: ^2.5.0
```

jakemac53:
Thanks for starting on this! As you have probably discovered this is going to be a large and somewhat complicated effort. My thoughts were that we should be able to migrate mostly in a bottom up fashion, one package at a time. I do believe it is feasible to do that, although I am not sure all of the packages are unblocked yet, and we will want to wait on that probably before publishing _any_ of them (I think it just gets confusing otherwise, and it doesn't unblock anybody). 

Tests would initially all be opted out (or run in --no-sound-null-safety mode). Maybe the latter is better.

The order would be roughly [`build`, `build_config`, `build_daemon`], [`build_resolvers`, `scratch_space`], `build_test`, `build_runner_core`, `build_runner`, `build_modules`, [`build_web_compilers`, `build_vm_compilers`].

Then we would go back through and enable sound null safety on all the tests, fixing things up as needed.

I think probably we would want to do all of this before we publish any of them, get them all passing tests and happy, etc. That means a somewhat long feature freeze most likely. Whenever one is migrated you would have to go add overrides for it in all the ones that depend on it.

Another option we may want to consider, is splitting this into two chunks for publishing. If we get `build`, `build_config`, `build_resolvers`, and `build_test` migrated, that should be pretty much everything any builders depend on, and most builders would at that point be unblocked from migrating. I think we could run most tests with sound null safety at that point also. We could then publish those.

The rest of the packages are more related to the build script itself, which can be tackled as a separate thing, and is honestly just a non-goal for the immediate future.

jakemac53:
cc @natebosch

simolus3:
Thanks for the quick feedback!

> Another option we may want to consider, is splitting this into two chunks for publishing. If we get `build`, `build_config`, `build_resolvers`, and `build_test` migrated, that should be pretty much everything any builders depend on, and most builders would at that point be unblocked from migrating

From a builder-author perspective, this is what I was aiming for. I want to write and test my builders with sound null safety, but I don't care about whether the actual build runner is using null safety or not (similar to the story with `package:test`).
Maybe we can encourage users to run builds with weak null assertions if that helps.
And on a another thought we might not even need `build_config` for the first step: It's a non-dev dependency, but we typically don't import that into `lib/`.

`build`, `build_test` and `build_resolvers` have circular (dev-)dependencies, so we kind of have to migrate them at once. When that's done, we can run the majority of tests with sound null safety too.

build_test/lib/src/stub_reader.dart
simolus3: They all return `Future.value(null)` which apparently isn't a static error, but would fail at runtime.

`StubAssetReader`/`StubAssetWriter` seem to be exclusively used in other tests of the build system, for instance to ensure that a `BuildStepImpl` doesn't allow certain reads or writes. I wasn't sure how migrate the reader: Should the methods throw an `UnsupportedError('stub')`? Some bogus value?

build/test/generate/run_builder_test.dart
```diff
@@ -1,3 +1,4 @@
+//@dart=2.9
```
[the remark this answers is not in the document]
simolus3: They were importing `package:build_test` before I migrated it. I think most of these can be removed as soon as the "core" packages are all migrated.

build/lib/src/generate/run_builder.dart
```diff
@@ -30,21 +30,20 @@ import 'expected_outputs.dart';
...
-    @Deprecated('The rootPackage argument is unused') String rootPackage,
+    {Logger? logger,
+    ResourceManager? resourceManager,
     StageTracker stageTracker = NoOpStageTracker.instance,
-    void Function(AssetId input, Iterable<AssetId> assets)
+    void Function(AssetId input, Iterable<AssetId> assets)?
         reportUnusedAssetsForInput}) async {
   var shouldDisposeResourceManager = resourceManager == null;
-  resourceManager ??= ResourceManager();
+  final resources = resourceManager ?? ResourceManager();
```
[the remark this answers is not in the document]
simolus3: The problem was that it's later used in the local `buildForInput` method, in which it wasn't promoted.

jakemac53:
> `build`, `build_test` and `build_resolvers` have circular (dev-)dependencies, so we kind of have to migrate them at once. When that's done, we can run the majority of tests with sound null safety too.

They can be migrated separately - just the tests have to be opted out (or ran without sound null safety and add an ignore to the imports).

jakemac53:
I am OK though with a larger PR that migrates them all at once, but its more difficult to manage and review etc. It is easier to have a few smaller PRs and then go back and enable null safety in a followup imo.

simolus3:
Fair enough, I'll open another PR for `build_resolvers` and to migrate the tests later. I already migrated most of `build`, `build_test` and `scratch_space` though, if it's not too much work for you I'd prefer to not split those up. 

And unfortunately I had to add lots of dependency overrides now: Since `_test` wants all build packages from path, they all need to support the latest analyzer and glob, and due to breaking changes I couldn't really expand the version range. So the overrides are kind of starting to cascade now...

jakemac53:
Ya I am fine with keeping this PR as is at this point and not splitting it up.

jakemac53:
> And unfortunately I had to add lots of dependency overrides now: Since `_test` wants all build packages from path, they all need to support the latest analyzer and glob, and due to breaking changes I couldn't really expand the version range. So the overrides are kind of starting to cascade now...

Ya, this mono_repo is definitely more than a bit crazy like that

build_test/lib/src/stub_reader.dart
jakemac53: Ya we would have to look more at how the tests are used.... this type of pattern is definitely more challenging with null safety.

simolus3:
I think we might want to put this on hold until the test packages support the latest analyzer. The dependency conflicts are getting out of hand.

jakemac53:
> I think we might want to put this on hold until the test packages support the latest analyzer. The dependency conflicts are getting out of hand.

sgtm

simolus3:
I'll wait for #3003 to land before fixing conflicts (otherwise I'd effectively have to do it twice). Apart from that, I think this should be ready now.

jakemac53:
> I'll wait for #3003 to land before fixing conflicts (otherwise I'd effectively have to do it twice). Apart from that, I think this should be ready now.

Sounds good, we probably also want to actually do a round of releases as well before merging and releasing this.

simolus3:
> Sounds good, we probably also want to actually do a round of releases as well before merging and releasing this.

Definitely. And if we end up doing a `build: 2.0.0` release for this, we might want to integrate [other breaking changes](https://github.com/dart-lang/build/labels/next-breaking-release) too.

jakemac53:
@simolus3 we should be ready to proceed here if you want to merge with master and I can take another pass

jakemac53:
Note that our tests red on master right now, I am working on fixing them.

jakemac53:
Re-running the tests now, should go green

jakemac53:
@simolus3 I think maybe you need to merge with master again to get the test fixes?

_test/pubspec.yaml
```diff
@@ -17,6 +17,7 @@ dev_dependencies:
   build_test: any
 
 dependency_overrides:
+  analyzer: ^0.42.0-nullsafety.0 # pkg:test doesn't support the latest analyzer yet
```
jakemac53, line 20: This shouldn't be necessary any more (we should be on `1.x.x` everywhere now)

build_test/mono_pkg.yaml
```diff
@@ -7,7 +7,7 @@ stages:
     - dartfmt: sdk
     - dartanalyzer: --fatal-infos --fatal-warnings .
 - unit_test:
-  - test: --test-randomize-ordering-seed=random
+  - command: dart --no-sound-null-safety test --test-randomize-ordering-seed=random
```
jakemac53, line 10: We shouldn't need this if the tests themselves are opted out (this should be the default for opted out tests).

build_test/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.4.0-dev
+
+- Migrate to null safety
```
simolus3, line 3: Just for completeness, I want to point out that this package is not fully migrated yet (and shouldn't be released as-is). We're still blocked by `build_resolvers`, which is blocked by `graphs`.

jakemac53: Understood, did you plan to look at those or should one of us (either is fine, just want to not duplicate work :) )

jakemac53 (approved):

build_test/CHANGELOG.md
simolus3: One of you would have to take a look at https://github.com/dart-lang/graphs/pull/52. As soon as that package is migrated I'm happy to work on `build_resolvers` and the rest of `build_test`.

NarHakobyan:
are you planning to release?

jakemac53:
Probably sometime next week, there is another small breaking change we are rolling as a part of this release (https://github.com/dart-lang/build/issues/1899) but its going to take a little bit to roll that out (and I am OOO for the rest of this week).

Issue #2998, opened 2021-02-10
build_runner hangs on "Creating build script snapshot......"


JostSchenck:
Dart SDK version: 2.12.0-259.8.beta (beta) (Tue Feb 2 15:59:29 2021 +0100) on "windows_x64"
using build_runner: ^1.11.1

Since some time, in a project of mine build_runner hangs on "Creating build script snapshot......" forever (hours). The only builder I'm using is from objectbox_generator package; it worked before (version 0.10.0), but even if I revert from version 0.11.0 back to it, build_runner hangs. I'm not experienced with build_runner (not that experienced with flutter either, I admit). I tried getting any output with -verbose but I do not get any more information on what it is hanging with.
Following advices I found for people having a hanging build_runner I did many a `flutter clean`, `flutter upgrade`, `flutter pub get`. I used the flutter install that used to work till somewhen last week, I used a fresh flutter from current dev branch (I'm developing for desktop, so stable is not an option for me).

I created a new flutter example application with flutter create, added the objectbox libraries and build_runner and ran it. Build_runner hangs at the same step.

I'm clueless and would be grateful for any ideas.

```flutter pub run build_runner build
[INFO] Generating build script...
[INFO] Generating build script completed, took 253ms

[INFO] Creating build script snapshot......
```

jakemac53:
Can you try manually snapshotting the build script? You can do this with the following command:

`dart --snapshot=test.dart.snapshot $(pub run build_runner generate-build-script)`

JostSchenck:
Thank you! I tried (`dart --snapshot=test.dart.snapshot $(flutter pub run build_runner generate-build-script)`). I have no idea why, because I didn't change anything knowingly, but my app as well as the test app I created now complain because of missing null safety. I tried adding a `--no-sound-null-safety`, but this didn't help.

```
Failed to precompile build_runner:build_runner:
/C:/Users/Jost%20Schenck/AppData/Local/Pub/Cache/hosted/pub.dartlang.org/build_runner-1.11.1/bin/build_runner.dart:89:9: Warning: Operand of null-aware operation '?.' has type 'StreamSubscription<dynamic>' which excludes null.
 - 'StreamSubscription' is from 'dart:async'.
  await logListener?.cancel();
        ^
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:build_runner_core
 - package:io
 - package:build_runner
 - package:build
 - package:glob
 - package:pool
 - package:crypto
[...]

PR #3005, opened 2021-02-16
fix snapshot generation hanging on windows

jakemac53:
Fixes https://github.com/dart-lang/build/issues/2998

Logs at `fine` level everything on stdout, and `warning` everything on stderr. Adds an info level message if anything is visible on stdout telling you to run in `--verbose` mode.

This fixes the issue by actually draining the stdout stream. I am unsure why this is only presenting itself as an issue on windows, but seems like a good change anyways.

### DIFF
--- a/build_runner/lib/src/build_script_generate/bootstrap.dart
+++ b/build_runner/lib/src/build_script_generate/bootstrap.dart
@@ -156,27 +156,40 @@ Future<int> _createSnapshotIfNeeded(Logger logger) async {
     }
   }
 
-  String stderr;
   if (!await snapshotFile.exists()) {
     var mode = stdin.hasTerminal
         ? ProcessStartMode.normal
         : ProcessStartMode.detachedWithStdio;
+    var hadStdOut = false;
     await logTimedAsync(logger, 'Creating build script snapshot...', () async {
       var snapshot = await Process.start(Platform.executable,
           ['--snapshot=$scriptSnapshotLocation', scriptLocation],
           mode: mode);
-      stderr = (await snapshot.stderr
-              .transform(utf8.decoder)
-              .transform(LineSplitter())
-              .toList())
-          .join('');
+      await Future.wait([
+        snapshot.stderr
+            .transform(utf8.decoder)
+            .transform(LineSplitter())
+            .listen((l) {
+          logger.warning('stderr: $l');
+        }).asFuture(),
+        snapshot.stdout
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -156,27 +156,40 @@ Future<int> _createSnapshotIfNeeded(Logger logger) async {
...
+            .transform(LineSplitter())
+            .listen((l) {
+          hadStdOut = true;
+          logger.fine('stdout: $l');
+        }).asFuture(),
+      ]);
     });
+    if (hadStdOut) {
+      logger.info('There was output on stdout while compiling the build script '
+          'snapshot, run with `--verbose` to see the it\n.');
```
natebosch, line 186: Will they need to do a clean first or something? If it was successful but with output I don't think the next run with rebuild the snapshot will it?

jakemac53: Hmmm right, they would have to clean. Should we suggest a `clean` do you think? Just not emit a message?

natebosch: I guess maybe suggest a `clean` for now, and file an issue to add a `--force-snapshot` flag.

jakemac53: I don't think we actually want to add a `--force-snapshot` flag - just a normal clean in this scenario should be fine IMO.

build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -156,27 +156,41 @@ Future<int> _createSnapshotIfNeeded(Logger logger) async {
...
+            .transform(LineSplitter())
+            .listen((l) {
+          hadStdOut = true;
+          logger.fine('stdout: $l');
+        }).asFuture(),
+      ]);
     });
+    if (hadStdOut) {
+      logger.info('There was output on stdout while compiling the build script '
+          'snapshot, run with `--verbose` to see the it (you will need to run '
```
natebosch, line 186:
```suggestion
          'snapshot, run with `--verbose` to see it (you will need to run '
```

natebosch (approved):

jakemac53:
The test failure here I am consistently seeing on master now as well. There is something broken with `dev`, but I am going to go ahead and merge this and try to address that separately.

PR #3007, opened 2021-02-17
Remove overrides, expand build_config constraint

jakemac53:
All the packages are published now so this cleans things up and removes the overrides.

I also expanded the build_config constraint in build_runner and build_runner_core which uses a tight constraint and didn't allow the latest version.

Also updates the build_modules dep in build_vm_compilers.

### DIFF
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -21,13 +21,3 @@ dev_dependencies:
   build_test: ^1.0.0
   pedantic: ^1.0.0
   test: ^1.2.0
-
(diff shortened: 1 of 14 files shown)


jcollins-g (approved):

jakemac53:
All the packages are published now so this cleans things up and removes the overrides.

I also expanded the build_config constraint in build_runner and build_runner_core which uses a tight constraint and didn't allow the latest version.

Also updates the build_modules dep in build_vm_compilers.

PR #3013, opened 2021-02-18
fix integration tests by adding real deps

jakemac53:
A recent build_runner_core update made dependency overrides not be treated as immediate deps, this pr udpates the integration testing framework we use to add all deps as both immediate deps and overrides so that builders are still applied.


natebosch (approved):

jakemac53:
Going to merge as this does fix some tests, I was hoping it would fix them all but the other ones are still present. Will address those next.

PR #3015, opened 2021-02-18
Fix clean and generate commands

jakemac53:
This effectively does add some integration test coverage for `clean`, but we should follow up with some explicit integration tests for both of these commands.

Filed https://github.com/dart-lang/build/issues/3016

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -28,9 +28,9 @@ Future<void> main(List<String> args) async {
       BuildCommandRunner([], await PackageGraph.forThisPackage());
   var localCommands = [CleanCommand(), GenerateBuildScript()];
   var localCommandNames = localCommands.map((c) => c.name).toSet();
-  localCommands.forEach(commandRunner.addCommand);
   for (var command in localCommands) {
-    // This flag is added to each command indivudally
+    commandRunner.addCommand(command);
+    // This flag is added to each command individually and not the top level.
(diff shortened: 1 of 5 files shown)


build_runner/bin/build_runner.dart
```diff
@@ -28,6 +29,14 @@ Future<void> main(List<String> args) async {
   var localCommands = [CleanCommand(), GenerateBuildScript()];
   var localCommandNames = localCommands.map((c) => c.name).toSet();
   localCommands.forEach(commandRunner.addCommand);
+  for (var command in localCommands) {
```
natebosch, line 32: [nit] Move the line above into the loop

```dart
for (var command in localCommands) {
  commandRunner.addCommand(command);
  command.argParser.addFlag(verboseOption,
  //...
```

build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -184,7 +184,7 @@ Future<int> _createSnapshotIfNeeded(Logger logger) async {
     if (hadStdOut) {
       logger.info('There was output on stdout while compiling the build script '
           'snapshot, run with `--verbose` to see it (you will need to run '
-          'a `clean` first to re-snapshot)\n.');
+          'a `clean` first to re-snapshot).\n');
```
natebosch, line 187: 🙃

natebosch (approved):

PR #3017, opened 2021-02-19
Migrate build_resolvers

simolus3:
I missed the return type of `astNodeFor` in `package:build`, it should be nullable as indicated in the documentation.


build_resolvers/lib/src/resolver.dart
```diff
@@ -320,12 +334,12 @@ class AnalyzerResolvers implements Resolvers {
...
-          await loadPackageConfigUri(await Isolate.packageConfig);
-      var driver = await analysisDriver(_uriResolver, _analysisOptions,
-          await _sdkSummaryGenerator(), _packageConfig);
-      _resolver = AnalyzerResolver(driver, _uriResolver);
+      final uriResolver = _uriResolver = BuildAssetUriResolver();
+      final loadedConfig = _packageConfig ??=
+          await loadPackageConfigUri((await Isolate.packageConfig)!);
+      var driver = await analysisDriver(uriResolver, _analysisOptions,
+          await _sdkSummaryGenerator(), loadedConfig);
+      return _resolver = AnalyzerResolver(driver, uriResolver);
```
jakemac53, line 342: nit: We shouldn't need the `return` here.

build_resolvers/lib/src/resolver.dart
```diff
@@ -451,7 +465,7 @@ final _dartUiPath =
 
 /// The current feature set based on the current sdk version and enabled
 /// experiments.
-FeatureSet _featureSet({List<String> enableExperiments}) {
+FeatureSet _featureSet({List<String>? enableExperiments}) {
   enableExperiments ??= [];
```
jakemac53, line 469:
```suggestion
FeatureSet _featureSet({List<String> enableExperiments = const []}) {
```

We used to use the old pattern as a defensive thing to guard against passing null, but now we can actually be sure we don't get it (well, assuming sound null safety but good enough).

build_resolvers/pubspec.yaml
```diff
@@ -1,28 +1,31 @@
 name: build_resolvers
-version: 1.5.4
+version: 1.6.0
```
jakemac53, line 2: We will want to bump to `2.0.0`

build_resolvers/test/resolver_test.dart
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+//
+// @dart=2.9
```
jakemac53, line 5: This is blocked on build_test right? Plan is to just come back and opt these in after landing both prs?

simolus3: I can migrate this test in the second PR, looks like I just missed it.

simolus3: Actually I already did: https://github.com/dart-lang/build/pull/3018/commits/273e80bdf687551880574fcccc5335cfdfbbbe5f

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -198,17 +197,20 @@ String assetPath(AssetId assetId) =>
...
-        })
-        .map((d) => AssetId.resolve(d.uri.stringValue, from: from))
-        .where((id) => id != null));
+Set<AssetId> _parseDirectives(String content, AssetId from) => HashSet.of(
+      parseString(content: content, throwIfDiagnostics: false)
+          .unit
+          .directives
+          .whereType<UriBasedDirective>()
+          .map((directive) => directive.uri.stringValue)
+          .where((uriContent) {
```
jakemac53, line 206: We could do another `.whereType<String>()` to avoid the null check in the body here and the `!` in the following map function, but either way sgtm.

build_resolvers/lib/src/resolver.dart
```diff
@@ -282,13 +296,13 @@ class AnalyzerResolvers implements Resolvers {
   final Future<String> Function() _sdkSummaryGenerator;
 
   // Lazy, all access must be preceded by a call to `_ensureInitialized`.
-  AnalyzerResolver _resolver;
-  BuildAssetUriResolver _uriResolver;
+  late final AnalyzerResolver _resolver;
+  BuildAssetUriResolver? _uriResolver;
```
jakemac53, line 300: It looks to me like this and `_packageConfig` could both be `late` and non-nullable as well?

simolus3: In `reset()`, we use `_uriResolver?.reset()` so we need to know whether it was initialized or not.  `_packageConfig` could be set in the constructor or in `_ensureInitialized`, so we need to check that too (and we can't move the initializer into a `late` field declaration because its async). I don't think we can actually avoid the null checks here?

jakemac53: Ya if we need to do null checks then we need to keep this as you have it, sgtm

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -56,21 +55,23 @@ class BuildAssetUriResolver extends UriResolver {
...
-        ? await crawlAsync<AssetId, _AssetState>(
+        ? await crawlAsync<AssetId, _AssetState?>(
             uncrawledIds,
             (id) => _updateCachedAssetState(id, buildStep,
-                transitivelyResolved: transitivelyResolved),
-            (id, state) => state.dependencies.where(notCrawled)).toList()
+                transitivelyResolved: transitivelyResolved), (id, state) {
+            if (state == null) return const [];
+            return state.dependencies.where(notCrawled);
+          }).where((state) => state != null).cast<_AssetState>().toList()
```
jakemac53, line 71: Can we replace the null check + cast with a `whereType<_AssetState>()`?

build_runner/test/server/build_updates_client/reloading_manager_test.dart
```diff
@@ -30,7 +30,7 @@ void main() {
           methods.reloadModule,
           methods.getModuleLibraries,
           methods.reloadPage,
-          (id) => moduleParentsGraph[id],
+          (id) => moduleParentsGraph[id] ?? const [],
```
simolus3, line 33: In `graphs` 1.x, the `edges`  function in `stronglyConnectedComponents` is not allowed to return `null` anymore. That sounds reasonable to me, but it's tricky to catch in non-migrated packages.

I don't know if `build_runner/web/hot_reload_client.dart` needs changes too, is there an integration test to find out?

jakemac53: I don't think we have a hot reload test any longer in build_runner (I don't think we do any webdriver testing which this would require). Webdev is generally where tests like that live but it has its own hot module reloading script which is a bit different if I recall correctly.

I believe it wouldn't return null though, all nodes in the graph should have parents, or else they wouldn't be in the app at all. That assumes we do skip doing anything if you edit a module not depended on by the app.

jakemac53: Honestly we could just finally remove entirely the ability to hot reload from this package, we really expect people to go through webdev for that now and the implementation in this package is likely to bitrot.

simolus3:
Too early for a ping, but from my end this looks ready now :)

jakemac53 (approved):

jakemac53:
Thanks @simolus3 🥳

PR #3018, opened 2021-02-19
Migrate build_test, tests of migrated packages, example

simolus3:
This migrates:

- `package:build_test`
- the tests of `package:build`, `package:build_resolvers` and `package:scratch_space`
- the example in this repo

With this, builder authors will be able to write and test their builders with sound null safety :tada:

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.4.0-dev
 
-- Migrate to null safety
+- Migrate `package:build_test/build_test.dart` to null safety.
 
 ## 1.3.7
 
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 1.4.0-dev
+## 2.0.0
 
 - Migrate `package:build_test/build_test.dart` to null safety.
 
--- a/build_test/lib/src/test_builder.dart
+++ b/build_test/lib/src/test_builder.dart
@@ -38,7 +38,7 @@ AssetId _passThrough(AssetId id) => id;
 /// association to a package pass [mapAssetIds] to translate from the logical
 /// location to the actual written location.
 void checkOutputs(
-    Map<String, /*List<int>|String|Matcher<List<int>>*/ dynamic>? outputs,
+    Map<String, /*List<int>|String|Matcher<List<int>>*/ Object>? outputs,
     Iterable<AssetId> actualAssets,
     RecordingAssetWriter writer,
     {AssetId Function(AssetId id) mapAssetIds = _passThrough}) {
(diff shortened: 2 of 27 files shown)


build_test/CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## 1.4.0-dev
```
jakemac53, line 1: I think I missed this before but we will want this to be `2.0.0` as well

build_test/lib/src/test_builder.dart
```diff
@@ -39,7 +38,7 @@ AssetId _passThrough(AssetId id) => id;
 /// association to a package pass [mapAssetIds] to translate from the logical
 /// location to the actual written location.
 void checkOutputs(
-    Map<String, /*List<int>|String|Matcher<List<int>>*/ dynamic> outputs,
+    Map<String, /*List<int>|String|Matcher<List<int>>*/ dynamic>? outputs,
```
jakemac53, line 41: Generally we are switching over to `Object` or `Object?` from `dynamic` where we notice it. Not a big thing but does help prevent accidental dynamic calls and forces explicit casts instead.

```suggestion
    Map<String, /*List<int>|String|Matcher<List<int>>*/ Object>? outputs,
```

build_test/lib/src/test_builder.dart
```diff
@@ -127,15 +126,15 @@ void checkOutputs(
 /// `withEnabledExperiments` method from package:build.
 Future testBuilder(
     Builder builder, Map<String, /*String|List<int>*/ dynamic> sourceAssets,
```
jakemac53, line 128:
```suggestion
    Builder builder, Map<String, /*String|List<int>*/ Object> sourceAssets,
```

build_test/lib/src/test_builder.dart
```diff
@@ -127,15 +126,15 @@ void checkOutputs(
...
-    Map<String, /*String|List<int>|Matcher<List<int>>*/ dynamic> outputs,
-    void Function(LogRecord log) onLog,
-    void Function(AssetId, Iterable<AssetId>) reportUnusedAssetsForInput,
-    PackageConfig packageConfig}) async {
+    {Set<String>? generateFor,
+    bool Function(String assetId)? isInput,
+    String? rootPackage,
+    MultiPackageAssetReader? reader,
+    RecordingAssetWriter? writer,
+    Map<String, /*String|List<int>|Matcher<List<int>>*/ dynamic>? outputs,
```
jakemac53, line 134:
```suggestion
    Map<String, /*String|List<int>|Matcher<List<int>>*/ Object>? outputs,
```

jakemac53 (approved):

jakemac53:
@simolus3 I think we do want to look at adding in this breaking change also while we can https://github.com/dart-lang/build/issues/1899.

I need to do a bit of work to investigate how breaking that might be internally first though.

PR #3020, opened 2021-03-01
Remove support for module reloads

simolus3:
Following the suggestion from @jakemac53:

>  Honestly we could just finally remove entirely the ability to hot reload from this package, we really expect people to go through webdev for that now and the implementation in this package is likely to bitrot.

This avoids some uncertainty related to a `pkg:graphs` change.


simolus3:
How should we fix the markdown link failure? Remove the reference from older changelog entries?

jakemac53:
> How should we fix the markdown link failure? Remove the reference from older changelog entries?

That seems reasonable ya

jakemac53:
@grouma just to confirm - webdev isn't relying on this flag right?

grouma:
> @grouma just to confirm - webdev isn't relying on this flag right?

Shouldn't be. The hot restart logic is now located in `package:dwds` and used by webdev / Flutter Tools.

jakemac53 (approved):

PR #3021, opened 2021-03-02
Support the latest pubspec_parse and checked_yaml

kevmoo:


build_config/pubspec.yaml
```diff
@@ -1,17 +1,17 @@
 name: build_config
-version: 0.4.6
+version: 0.4.7-dev
 description: Support for parsing `build.yaml` configuration.
 repository: https://github.com/dart-lang/build/tree/master/build_config
 
 environment:
-  sdk: '>=2.9.0 <3.0.0'
+  sdk: '>=2.11.99 <3.0.0'
```
kevmoo, line 7: @jakemac53 @natebosch – since this depends on these other packages which require latest dart, it seems this package should, too!

jakemac53 (approved):

build_config/pubspec.yaml
natebosch: >  since this depends on these other packages which require latest dart, it seems this package should, too!

Which? It looks like most of the lower bounds still allow non-null safe. Are there any where the lower bound for a dependency results in this SDK constraint?

kevmoo: I can get pub to solve on `2.9.0` using `_PUB_TEST_SDK_VERSION=2.9.0 pub upgrade` (when removing overrides)

I figured if we can't resolve – let along test – we shouldn't claim to support a version

natebosch (approved):

PR #3028, opened 2021-03-07
Support latest web socket packages

### LINKED ISSUE #3024: Bump web_socket_channel in build_runner
* Dart SDK Version (`dart --version`): `Dart SDK version: 2.12.0 (stable) (Thu Feb 25 19:50:53 2021 +0100) on "linux_x64"`
* What package(s) from this repo you are using: `build_runner 1.11.5`

I am maintaining a package that uses `web_socket_channel` and `build_runner`. Sadly, since build runner depends on `^1.0.0` I cannot upgrade `web_socket_channel` to v2. The reason I want to upgrade is to be able to go sound null safe.

Thus this issue is a request to bump the `web_socket_channel` package on `build_runner` to v2 in order to enable people like me upgrading to sound null safety

### LINKED ISSUE #3025: Unable to upgrade dependencies because of build_runner dependencies
Hello, 
I'm unable to update the dependencies of my flutter app because build_runner still depends on web_socket_channel 1.0.9 while I want to use version 2.0.0. I'm using build_runner 1.11.5.

`dart --version`:
```
Dart SDK version: 2.12.0 (stable) (Thu Feb 25 19:50:53 2021 +0100) on "linux_x64"
```

`flutter packages get`:
```
Because build_runner 1.11.5 depends on web_socket_channel ^1.0.9 and no versions of build_runner match >1.11.5 <2.0.0, build_runner ^1.11.5 requires web_socket_channel ^1.0.9.
So, because xyz depends on both web_socket_channel ^2.0.0 and build_runner ^1.11.5, version solving failed.                             
pub get failed (1; So, because xyz depends on both web_socket_channel ^2.0.0 and build_runner ^1.11.5, version solving failed.)
```

If any info is missing, I'm happy to add it but I think this should be everything you need. Thanks in advance!

simolus3:
Support the latest version of `web_socket_channel` and `shelf_web_socket` in `build_runner` and `build_daemon`.

Closes #3024. Closes #3025.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.1.9-dev
+
+- Support version `1.x` of `shelf_web_socket` and `2.x` of `web_socket_channel`
+
 ## 2.1.8
 
 - Begin conversion to use analyzer 1.0.0.
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -3,6 +3,7 @@
 - __Breaking__: Remove support for hot-reloads. Use `package:webdev` instead.
 - Support version 2.0.x of the `build` package
 - Update to graphs `1.x`
+- Support version `1.x` of `shelf_web_socket` and `2.x` of `web_socket_channel`
 
 ## 1.11.5
 
--- a/build_daemon/pubspec.yaml
+++ b/build_daemon/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_daemon
-version: 2.1.8
+version: 2.1.9-dev
 description: A daemon for running Dart builds.
 homepage: https://github.com/dart-lang/build/tree/master/build_daemon
 
@@ -15,10 +15,10 @@ dependencies:
   path: ^1.6.2
   pool: ^1.3.6
   shelf: '>=0.7.4 <2.0.0'
-  shelf_web_socket: ^0.2.2+4
+  shelf_web_socket: '>=0.2.2+4 <2.0.0'
   stream_transform: ">=0.0.20 <3.0.0"
   watcher: '>=0.9.7 <2.0.0'
-  web_socket_channel: ^1.0.9
+  web_socket_channel: '>=1.0.9 <3.0.0'
 
 dev_dependencies:
   build_runner: ^1.0.0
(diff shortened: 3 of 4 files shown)


jlnrrg:
@simolus3 It might be beneficial to set the same package versions across the multiple build packages.
For example (state rn):
`build_runner` depends on `pubspec_parse: ^1.0.0`
`build_config` depends on `pubspec_parse: '>=0.1.5 <2.0.0'`

build_daemon/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.9-dev
```
jakemac53, line 1: We should be able to publish this right away I think so we can drop the `-dev` (in the pubspec also).

build_runner/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - __Breaking__: Remove support for hot-reloads. Use `package:webdev` instead.
 - Support version 2.0.x of the `build` package
 - Update to graphs `1.x`
+- Support version `1.x` of `shelf_web_socket` and `2.x` of `web_socket_channel`
```
jakemac53, line 6: Should we do this in a non-breaking release? Maybe we should actually do this current release as non-breaking? It is only the hot reload support which was dropped which isn't really "breaking".

I don't want to do multiple breaking changes in quick succession here, and we will have another for null safety.

cc @natebosch wdyt?

jakemac53 (commented):
Releasing these is somewhat gated on build 2.0 and friends right?

simolus3:
I think `build_daemon` is good to go as you mentioned. It has a worryingly high number of overrides, but from a quick look they all seem to be for dev dependencies. `build_runner` depends on `build` 2.x unless we rebase this on a pre-nnbd state which IMO isn't worth it.

jakemac53:
> `build_runner` depends on `build` 2.x unless we rebase this on a pre-nnbd state which IMO isn't worth it.

Ya I would like to avoid that, there isn't a lot of value. We should focus on getting out the build 2.0 and friends releases.

simolus3:
I've changed the version of `build_runner` back to 1.12-dev for now. It looks like the build failure is due to Chrome not starting in time, can you restart the checks?

PR #3040, opened 2021-03-12
prep to release all packages with build 2.0 support

jakemac53:
cc @simolus3

With this all user facing packages should be migrated to null safety (build, build_resolvers, and package:build_test/build_test.dart).

And then it also bumps the deps in other packages to allow the new versions where there were breaking releases.

I also bumped `scratch_space` to 1.0.0

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.0.0-dev
+## 2.0.0
 
 - Migrate to null-safety
 - __Breaking__: Remove the deprecated `rootPackage` argument to `runBuilder`
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build
-version: 2.0.0-dev
+version: 2.0.0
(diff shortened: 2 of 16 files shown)


natebosch (approved):

erf:
dependency fails with `json_serializable:` now

```
dart pub install
Resolving dependencies... (2.3s)
Because no versions of json_serializable match >4.0.2 <5.0.0 and json_serializable 4.0.2 depends on build
  ^1.6.1, json_serializable ^4.0.2 requires build ^1.6.1.
And because build_runner >=1.12.0 depends on build >=2.0.0 <2.1.0, json_serializable ^4.0.2 is incompatible with
  build_runner >=1.12.0.
So, because xxx depends on both build_runner ^1.12.1 and json_serializable ^4.0.2, version solving failed.
```

jakemac53:
@erf you should relax your `build_runner` dependency to not require the latest. It will take a while for all builders to support `build` version 2.0.

Feel free to file issues on any builders (json_serializable in this case) that need to be updated though.

Issue #3046, opened 2021-03-13
NullSafety "flutter pub run build_runner build" problem


yfer:
I'm using new null-safety feature with dart and had a problem with build_runner.

```
 Error: Operator '[]' cannot be called on 'ArgResults?' because it is potentially null.
 - 'ArgResults' is from 'package:args/src/arg_results.dart' ('/C:/flutter/.pub-cache/hosted/pub.dartlang.org/args-2.0.0/lib/src/arg_results.dart').
    var verbose = parsedArgs.command['verbose'] as bool ?? false;
```
and
```
Failed to precompile build_runner:build_runner:
/c:/flutter/.pub-cache/hosted/pub.dartlang.org/build_runner-1.12.1/bin/build_runner.dart:101:9: Warning: Operand of null-aware operation '?.' has type 'StreamSubscription<dynamic>' which excludes null.
 - 'StreamSubscription' is from 'dart:async'.
  await logListener?.cancel();
```

**Fixed both of them, hence PR.** But of course I still got 

```
Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:build_runner_core
 - package:io
 - package:build_runner
 - package:build_config
 - package:build_daemon
 - package:timing
 - package:code_builder
 - package:dart_style
 - package:http_multi_server
 - package:shelf_web_socket
 - package:web_socket_channel
```
Tried to run `flutter pub run build_runner build --no-sound-null-safety`, it had no effect.

**Solution!!!**
The only thing that helped is manually insert `// @dart=2.9` at the beginning of bin/build_runner.dart, after that builder is working.

**Should we just include `// @dart=2.9` at the beginning of bin/build_runner.dart, so you at least don't see this problem untill null-safety is fully supported?**

```
- Flutter version 2.0.2 at C:\flutter
- Framework revision 8962f6dc68 (2 days ago), 2021-03-11 13:22:20 -0800
- Engine revision 5d8bf811b3
- Dart version 2.12.1
- What package(s) from this repo you are using: **build_runner 1.11.5**, **build_runner 1.12.1** and master
- What builder(s) you are using: **freezed 0.14.0+1**, but without any builders it fails the same
- Whether you are using? Windows
```

simolus3:
@yfer, how do you run `build_runner`? The package itself doesn't opt-in to null safety, so running `dart run build_runner build_runner build` should work (replace `dart` with `flutter pub` if you're using Flutter).

[...]

PR #3047, opened 2021-03-13
/bin/build_runner.dart is null-safe

yfer:
Described the [issue](https://github.com/dart-lang/build/issues/3046)

this commit basically made /bin/build_runner.dart itself null-safe.

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -88,7 +88,7 @@ Future<void> main(List<String> args) async {
       }
     });
   } else {
-    var verbose = parsedArgs.command['verbose'] as bool ?? false;
+    var verbose = parsedArgs.command?['verbose'] as bool;
     if (verbose) Logger.root.level = Level.ALL;
     logListener =
         Logger.root.onRecord.listen(stdIOLogListener(verbose: verbose));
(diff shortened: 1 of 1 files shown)


yfer:
Hmm, i see. So the build process on build_runner package explicitly setup not to use not-nullable feature.

But as a user nothing should stop me to use this package in my project as a tool. 
Wouldn't it be better then to use `// @dart=2.9` on /bin/build_runner.dart to ignore these inconsistencies for the ones who want to use not-nullable in their code?

dancamdev:
The CI Failed with the following message.

`error • This requires the 'non-nullable' language feature to be enabled. • bin/build_runner.dart:91:37 • experiment_not_enabled`

Could it be related to the dependencies or sdk version not being upgraded?

higorlapacw:
I think so, @dancamdev 

```
environment:
  sdk: ">=2.11.99 <3.0.0"
```

jakemac53:
This package is _not_ migrated to null safety yet, so we can't use these features (nor opt in because we would then be importing opted out libraries).

jakemac53:
(something else is going on in the original issue - there should be no problem with build_runner being opted out for now)

PR #3052, opened 2021-03-16
Fix analysis warnings for non-null error result

natebosch:


jakemac53 (approved):

jakemac53 (approved):

kevmoo (approved):

kevmoo:
This due to a change in pkg:analyzer, right?

Do we need to bump the min version of analyzer supported? Something is non-nullable now that was nullable...

jakemac53:
Oh geeze ya you are right @kevmoo.

jakemac53 (dismissed):
see kevins comment :)

natebosch:
> Do we need to bump the min version of analyzer supported?

Yes we do. I guess it's more helpful if I push my commits rather than just site with them locally...

PR #3063, opened 2021-03-29
migrate build_config to null safety

jakemac53:
Note that error handling around null values changes a bit here. We could add more specialized handling to explicitly handle null values in explicit `include` or `exclude` globs (a cast error surfaces today).  However I think it might make the most sense to try and get something added upstream here to improve the message generally.

### DIFF
--- a/build_config/lib/src/build_config.dart
+++ b/build_config/lib/src/build_config.dart
@@ -74,7 +74,7 @@ class BuildConfig {
   @JsonKey(name: 'global_options')
   final Map<String, GlobalBuilderConfig> globalOptions;
 
-  @JsonKey(name: 'additional_public_assets')
+  @JsonKey(name: 'additional_public_assets', defaultValue: [])
   final List<String> additionalPublicAssets;
 
   /// The default config if you have no `build.yaml` file.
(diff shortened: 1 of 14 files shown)


build_config/lib/src/build_config.dart
```diff
@@ -124,19 +127,21 @@ class BuildConfig {
   }
 
   BuildConfig({
-    String packageName,
-    @required Map<String, BuildTarget> buildTargets,
-    Map<String, GlobalBuilderConfig> globalOptions,
-    Map<String, BuilderDefinition> builderDefinitions,
-    Map<String, PostProcessBuilderDefinition> postProcessBuilderDefinitions =
+    String? packageName,
```
natebosch, line 130: I'm surprised we allows a missing package name - I guess since this wasn't `@required` before we have some usages with no name?

natebosch (approved):

build_config/lib/src/build_config.dart
jakemac53: I can't actually find usages of this, I will try just making it required.

jakemac53: Ah, its because the generated code uses it.

PR #3093, opened 2021-04-21
fix remaining analyzer deprecation warnings

jakemac53:
Some places I did change the code to utilize the new methods better and add some type checks, others I just added an explicit cast.

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -152,8 +152,8 @@ class AnalyzerResolver implements ReleasableResolver {
   Future<bool> isLibrary(AssetId assetId) async {
     var source = _driver.sourceFactory.forUri2(assetId.uri);
     if (source == null || !source.exists()) return false;
-    var result = _driver.getFileSync2(assetPath(assetId));
-    return result is FileResult && !result.isPart;
+    var result = _driver.getFileSync2(assetPath(assetId)) as FileResult;
+    return !result.isPart;
   }
 
   @override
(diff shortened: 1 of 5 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -152,9 +151,9 @@ class AnalyzerResolver implements ReleasableResolver {
   @override
   Future<bool> isLibrary(AssetId assetId) async {
     var source = _driver.sourceFactory.forUri2(assetId.uri);
-    return source != null &&
-        source.exists() &&
-        (await _driver.getSourceKind(assetPath(assetId))) == SourceKind.LIBRARY;
+    if (source == null || !source.exists()) return false;
+    var result = _driver.getFileSync2(assetPath(assetId));
+    return result is FileResult && !result.isPart;
```
natebosch, line 156: one thing I was a little worried about is changing behavior subtly in cases where we get an invalid result.

If we introduce a bug down the line, will we start treating some libraries and non-libraries when the real situation is using the analyzer incorrectly?

natebosch (approved):

jakemac53:
It looks like this did also change the behavior in some cases. I will take a look at those tests.

jakemac53:
I went ahead and changed the logic more back to how it was, it looks like it was important for us to explicitly re-parse the file or we get stale results.

natebosch (approved):

jakemac53:
merging this and following up with a bug fix for the other issue showing up.

build_resolvers/lib/src/resolver.dart
natebosch: Do we already know that non-Dart files won't show up here? Could a non-part, non-library file show up as if it was a library?

cc @scheglov - this is similar to the internal code so the same concerns likely apply.

PR #3097, opened 2021-04-29
Use frontend_server for snapshots

simolus3:
Generate build script snapshots by spawning a frontend server instead of invoking a non-incremental compiler. This makes creating snapshots much faster.

### DIFF
--- a/build_runner/lib/src/build_script_generate/bootstrap.dart
+++ b/build_runner/lib/src/build_script_generate/bootstrap.dart
@@ -169,7 +169,7 @@ Future<int> _createKernelIfNeeded(Logger logger) async {
         // Note: We're logging all output with a single log call to keep
         // annotated source spans intact.
         final logOutput = result?.compilerOutputLines.join('\n');
-        if (logOutput != null) {
+        if (logOutput != null && logOutput.isNotEmpty) {
           hadOutput = true;
           if (hadErrors) {
             // Always show compiler output if there were errors
--- a/build_runner/lib/src/build_script_generate/bootstrap.dart
+++ b/build_runner/lib/src/build_script_generate/bootstrap.dart
@@ -16,7 +16,7 @@ import 'package:stack_trace/stack_trace.dart';
 
 final _logger = Logger('Bootstrap');
 
-/// Generates the build script, snapshots it if needed, and runs it.
+/// Generates the build script, precompiles it if needed, and runs it.
 ///
 /// The [handleUncaughtError] function will be invoked when the build script
 /// terminates with an uncaught error.
@@ -66,7 +66,7 @@ Future<int> generateAndRun(
       return ExitCode.config.code;
     }
 
-    scriptExitCode = await _createSnapshotIfNeeded(logger);
+    scriptExitCode = await _createKernelIfNeeded(logger);
     if (scriptExitCode != 0) return scriptExitCode!;
 
     exitPort = ReceivePort();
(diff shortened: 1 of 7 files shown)


build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -150,31 +150,24 @@ Future<int> _createSnapshotIfNeeded(Logger logger) async {
...
-            .transform(LineSplitter())
-            .listen((l) {
-          hadStdOut = true;
-          logger.fine('stdout: $l');
-        }).asFuture(),
-      ]);
+      try {
+        final result = await client.compile();
+        result?.compilerOutputLines.forEach(logger.fine);
+        hadOutput = result == null || result.errorCount > 0;
```
jakemac53, line 165:
```suggestion
        hadOutput = result == null | result.compilerOutputLines.isNotEmpty;
```

jakemac53 (commented):
It would be nice to clean up some of the wording here as well where it is no longer quite accurate. Basically I think replace `snapshot` where used as a verb with `precompile` and where used as a noun with `kernel`.

simolus3:
I've applied your code and wording suggestions.

One thing that stands out is a VM crash when we compile a Kernel file and the sources are missing imports. This causes [a test failure](https://github.com/dart-lang/build/pull/3097/checks?check_run_id=2469317323) now. That might be me using the frontend server wrong and not checking for errors correctly, but I've opened https://github.com/dart-lang/sdk/issues/45865 because I think the VM shouldn't crash here.

jakemac53:
Haha, yes a seg fault in the VM like that is pretty much always a bug. Interesting catch, hopefully it gets fixed soon.

jakemac53:
It does seem though like we are somehow not catching the error - it should have compiled properly in the first place?

simolus3:
> it should not have compiled properly in the first place

It looks like the frontend inserts an invalid expression instead of refusing to compile:

```
static final field core::List<app::BuilderApplication*>* _builders = core::_GrowableList::_literal1<app::BuilderApplication*>(app::apply("a:fake", core::_GrowableList::_literal1<(bui::BuilderOptions*) →* bui::Builder*>(invalid-expression ".dart_tool/build/entrypoint/build.dart:14:29: Error: Getter not found: 'myFactory'.\n  _i1.apply(r'a:fake', [_i2.myFactory], _i1.toNoneByDefault(), hideOutput: true)\n                            ^^^^^^^^^"), app::toNoneByDefault(), hideOutput: true));
```

We also get those errors in `compilerOutputLines`, but they're easy to overlook since they're not being logged by default.

I've changed the implementation to throw the Kernel file away if it exists and there were compilation errors (`result.errorCount > 0`).  In that case, we also log all output lines with `logger.warning` because it's likely that something's wrong and the user should probably take a look at that. When there are no errors but we did get output lines, we continue to log them with `logger.fine` and inform the user of `--verbose`. I think this should overall match the previous behavior which would reject invalid sources too, except that we can't distinguish between stdout and stderr anymore. It also avoids the VM crash.

build_runner/lib/src/build_script_generate/bootstrap.dart
```diff
@@ -160,20 +160,40 @@ Future<int> _createKernelIfNeeded(Logger logger) async {
...
       try {
         final result = await client.compile();
-        result?.compilerOutputLines.forEach(logger.fine);
-        hadOutput = result == null || result.compilerOutputLines.isNotEmpty;
+        hadErrors = result == null || result.errorCount > 0;
+
+        // Note: We're logging all output with a single log call to keep
+        // annotated source spans intact.
+        final logOutput = result?.compilerOutputLines.join('\n');
+        if (logOutput != null) {
```
jakemac53, line 172:
```suggestion
        if (logOutput != nul && logOutput.isNotEmpty) {
```

jakemac53:
@simolus3  I sent https://github.com/simolus3/build/pull/5 as an update to this which keeps around old dill files and uses them on future compiles, so we get the incremental builds here.

simolus3:
@googlebot I consent.

jakemac53:
It looks like a few tests need to be updated for the wording that I changed sorry!

jakemac53 (approved):

jakemac53:
thanks @simolus3 !

PR #3103, opened 2021-05-12
fix serve mode crash when the build script is out of date

jakemac53:
Fixes a bug discovered in this test failure https://github.com/dart-lang/build/pull/3102/checks?check_run_id=2563152879#step:6:247

Hooray for randomized test ordering! 🥳

### DIFF
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -188,7 +188,7 @@ class WatchImpl implements BuildState {
   /// Pending expected delete events from the build.
   final Set<AssetId> _expectedDeletes = <AssetId>{};
 
-  late FinalizedReader _reader;
+  FinalizedReader? _reader;
 
   /// Only non-null after [ready] completes without error.
   FinalizedReader? get reader => _reader;
--- a/build_runner/lib/src/generate/watch_impl.dart
+++ b/build_runner/lib/src/generate/watch_impl.dart
@@ -366,7 +366,7 @@ class WatchImpl implements BuildState {
         _readyCompleter.completeError(e, s);
       }
       if (build != null) {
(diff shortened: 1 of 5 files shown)


build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -340,26 +345,31 @@ class WatchImpl implements BuildState {
...
+        _readyCompleter.completeError(e, s);
+      } on BuildScriptChangedException catch (e, s) {
         _terminateCompleter.complete();
 
         firstBuild = BuildResult(BuildStatus.failure, [],
             failureType: FailureType.buildScriptChanged);
+        _readyCompleter.completeError(e, s);
+      }
+      if (build != null) {
+        assert(_readyCompleter.isCompleted);
```
natebosch: If this assert passes I would expect line 371 to fail.

Should this be `assert(!_readyCompleter.isCompleted)`?

build_runner/lib/src/generate/watch_impl.dart
```diff
@@ -186,7 +189,9 @@ class WatchImpl implements BuildState {
   final Set<AssetId> _expectedDeletes = <AssetId>{};
 
   FinalizedReader? _reader;
-  FinalizedReader get reader => _reader!;
+
+  /// Only non-null after [ready] completes without error.
+  FinalizedReader? get reader => _reader;
```
natebosch, line 194: What do you think of combining this with `ready`?

`Future<FinalizedReader> get ready`

(or keep the name `reader` but make it a `Future` and drop the `ready` field)

jakemac53: Hmm ya it might make more sense to just make it a Future, idk why it wasn't one already.

natebosch (approved):

PR #3107, opened 2021-05-18
add more context to the outdated analyzer version warning

jakemac53:
Related to https://github.com/dart-lang/sdk/issues/46041.

We now reach out to pub to find the latest analyzer version, and compare that against the current version. We will only suggest filing an issue if you are on the latest. Otherwise we suggest upgrading.

This does require making an http get request to pub to know the latest version, as well as reading the pubspec from the current analyzer, which requires a dep on `yaml`. As a result the error will be asynchronously logged some time after the resolver is created.

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,4 +1,7 @@
-## 2.0.2-dev
+## 2.0.2
+
+- Add more context to the outdated analyzer version message. It now provides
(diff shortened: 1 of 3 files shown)


natebosch (approved):

scheglov (approved):

srawlins:
Very cool! Thanks for exploring this route!

PR #3111, opened 2021-05-24
Fix determining null-safety for relative imports

simolus3:
There's a bug in the code that determines whether a build can run with sound null safety when a package uses builders with relative imports from its pubspec (e.g. without the `../` prefix).
Currently, it tried to read the asset by resolving it relative to the generated build script which is not correct in that case. For instance, consider a `tool/my_builder.dart` import in a `build.yaml`. We'd then try to read `.dart_tool/build/entrypoint/tool/my_builder.dart` to determine that file's language version. As that fails, we bail out and run the build without null safety. We now use the existing logic to determine the proper import.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.0.4
+
+- Allow builders defined as relative imports from the root of a package to run
+  with sound null safety.
+
 ## 2.0.3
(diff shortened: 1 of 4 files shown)


jakemac53 (approved):

jakemac53:
Thanks! This is published now.

PR #3112, opened 2021-05-24
Label assets without .dart as not Dart libraries

natebosch:
The implementation for `isLibrary` considers any `.dart` file that does
not contain a `part of` directive to be a Dart library file (potentially
with errors if it has non-Dart source code content). Clarify this
behavior with a test.


natebosch:
Hmm, maybe this isn't doing what I thought and I wasn't running the test correctly.

It looks like the current behavior is to treat any readable asset that doesn't contain `part of` to be a Dart library.

@jakemac53 - do you think it would be worthwhile to add an extra check around the file extension?

jakemac53:
It seems reasonable to add that check... I do wonder if we will accidentally break some workflows though.

We could add it and then just be prepared to revert if we break somebody I suppose?

jakemac53:
@natebosch should we just close this, or did you want to pick it back up?

natebosch:
@jakemac53 - I haven't found anything broken by this and I think it's probably going to be more helpful than harmful. I think it makes sense to go forward with it. PTAL and let me know what you think.

natebosch:
This also already matches the behavior of our bazel resolver implementation.

jakemac53 (approved):

PR #3120, opened 2021-06-03
Add an option for generating debug symbols from DDC

annagrin:
Add an option for generating debug symbols from DDC

- add `--output-debug-symbols` option
- pass `--output-debug-symbols` to DDC
- copy the `.symbols` files generated by DDC to the final output destination.
- fix a bug where `.full.dill` files did not have the paths adjusted to the final location.

Related:
https://github.com/dart-lang/sdk/issues/40273

Towards:
https://github.com/dart-lang/webdev/issues/982


build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+## 3.1.0-dev
+
+- Add the `generate-debug-symbols` option for the `build_web_compilers:ddc`
+  builder. The sdebug ymbols output is used by the debugger in webdev for
```
jakemac53, line 4:
```suggestion
  builder. The debug symbols output is used by the debugger in webdev for
```

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -193,6 +207,7 @@ Future<void> _createDevCompilerModule(
       '--modules=amd',
       '--no-summarize',
       if (generateFullDill) '--experimental-output-compiled-kernel',
+      if (generateDebugSymbols) '--experimental-output-debug-symbols',
```
jakemac53, line 210: Should we be retaining the `--experimental` name in the flag we expose from build_web_compilers as well? Usually the names would line up exactly, although I see we didn't do that for the generate-full-dill option...

build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -282,6 +297,17 @@ Future<void> _createDevCompilerModule(
...
+
+      // Copy the symbols output, modifying its contents to remove the temp
+      // directory from paths
+      if (generateDebugSymbols) {
+        var symbolsId = module.primarySource
+            .changeExtension(symbolsExtension(soundNullSafety));
+        file = scratchSpace.fileFor(symbolsId);
+        content = await file.readAsString();
+        json = jsonDecode(content);
+        await buildStep.writeAsString(symbolsId, jsonEncode(json));
```
jakemac53, line 309: Since we don't need to modify the content we can just use `copyOutput`:

```suggestion
        await scratchSpace.copyOutput(symbolsId, buildStep);
```

build_web_compilers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 3.0.0
+version: 3.1.0-dev
```
jakemac53, line 2: If you are ready to release this we can go ahead and do that and you can drop the `-dev` (from the changelog also).

jakemac53 (commented):
Do we also need to up the min sdk constraint? It is currently at 2.12.0.

build_web_compilers/pubspec.yaml
annagrin: Not ready yet as we need the SDK to actually produce symbols first. Will update when ready.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: I should probably just make the flags non-experimental in the sdk as well, as we don't have a special way of creating experimental flags, they could be just hidden flags, disabled by default.  My current plan is to just remove them as corresponding feature become mature and are enabled by default.

How does 'output-full-dill' and 'output-debug-symbols' sound for both SDK and build_web_compilers?

I would need to change `output-compiled-kernel`  in a different set of PRs though as changing flags between repos require a handshake between repos:)

annagrin:
> Do we also need to up the min sdk constraint? It is currently at 2.12.0.

Yes, I will update the constraint as soon as the SDK change producing the symbols is in dev channel.

build_web_compilers/lib/src/dev_compiler_builder.dart
annagrin: Changed to `output-debug-symbols` for both SDK and build_web_compilers, will clean up the existing option in a separate PR.

jakemac53: Note that changing the flag in the SDK is breaking for old versions of this package so if we are going to change (or remove) those we really should also ensure the old ones still work, or at least don't throw...

annagrin: Sure, I would need a series of changes over a long period of time:)

build_web_compilers/pubspec.yaml
jakemac53: Ok, are you fine with waiting to land this until then? I don't want to risk an accidental release of this in a state where it doesn't have the proper constraints.

annagrin: Definitely!

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+## 3.1.0-dev
+
+- Add the `emit-debug-symbols` option for the `build_web_compilers:ddc`
+  builder. The debug symbols output is used by the debugger in webdev for
+  variable inspection. This setting is disabled by default but can be
```
natebosch, line 5: Should we consider enabling this by default, or even making it non-configurable?

How many users do we expect to want this?

annagrin: Eventually everyone will use it, but the feature will not be ready for some time. I am hiding it now since it is still in development stage, and plan to enable it after dogfooding.

annagrin:
@jakemac53 @natebosch The required SDK changes are in the release dev channel, so I updated the min SDK constraint. Is it ok to publish this once it lands, as a dev version?

jakemac53:
> Is it ok to publish this once it lands, as a dev version?

We don't generally publish dev versions of these packages (generally, nobody will get them so there isn't much point).

We could however publish a stable version? We can just ignore the pub warning about the dev sdk constraint.

jakemac53 (approved):

PR #3122, opened 2021-06-07
Revert "update mono_repo (#3121)", broke CI

jakemac53:
This reverts commit 36110150863e3641b27437442d0991c8188bffa6.


kevmoo:
How does it break? Can I have an issue in mono_repo to fix?

kevmoo (approved):

jakemac53:
I think this is just a behavior difference between `dart analyze` and `dartanalyzer`, with respect to validation of analysis_options.yaml files. It failed on the original PR as well...

PR #3130, opened 2021-06-17
Migrate uses of followedBy to collection literals

natebosch:
In the past I've looked for other common patterns that look nicer as
literals and on a hunch I wanted to check which uses of `followedBy` are
more readable as subsequent elements in a collection. In these cases I
think it is an improvement.

Most migrated cases either were immediately gathered into a collection
to `.toList` or `.toSet`. In a few places where I trust that the
collection will be fairly small and always iterated I also switched from
using a lazy iterable into an eager collection.

Some uses of `followedBy` remain. Those cases are either where an eager
collection could be non-performant, or cases where the iterable chains
subsequent `.map` or other calls that would make the collection literal
more noisy.

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -149,12 +149,12 @@ List<Module> _mergeModules(Iterable<Module> modules, Set<AssetId> entrypoints) {
     }
   }
 
-  return mergedModules.values
-      .map(Module.merge)
-      .map(_withConsistentPrimarySource)
-      .followedBy(entrypointModuleGroups.values.map(Module.merge))
-      .followedBy(standaloneModules)
-      .toList();
+  return [
+    for (var module in mergedModules.values)
+      _withConsistentPrimarySource(Module.merge(module)),
+    for (var module in entrypointModuleGroups.values) Module.merge(module),
+    ...standaloneModules
+  ];
 }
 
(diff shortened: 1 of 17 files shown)


build_modules/lib/src/meta_module.dart
```diff
@@ -252,14 +252,15 @@ MetaModule _coarseModulesForLibraries(
...
 MetaModule _fineModulesForLibraries(
     AssetReader reader, List<ModuleLibrary> libraries, DartPlatform platform) {
-  var modules = libraries
-      .map((library) => Module(
+  var modules = [
+    for (var library in libraries)
+      Module(
           library.id,
-          library.parts.followedBy([library.id]),
+          [...library.parts, library.id],
```
natebosch, line 259: This is one place where I'm not fully confident that the eager collection won't be less performant than a lazy Iterable.

The `Module` constructor will do `UnmodifiableSetView(HashSet.of(source))` so it's going to be iterated and copied into a new collection immediately. Since these are small and we're allocating a list for `[library.id]` anyway I don't expect this to be impactful, but I might be overlooking something.

jakemac53: Ya it is something you would have to measure to really know. I agree it is very unlikely to matter though, given these are small lists, and we aren't creating any extra list allocations. The majority of files have no parts anyways so the list will be identical. LGTM

jakemac53 (approved):

Issue #3129, opened 2021-06-16
[build_runner] Do not sent the file contents with the response to a HEAD request.


annagrin:
See https://github.com/dart-lang/webdev/issues/1345.

Some change in SDK uncovered a problem where build_runner sends the whole file contents to the client, but the 
client does not read it, causing broken pipe errors in some configurations but not the others.

The HEAD request spec mentions:
> The HEAD method is identical to GET except that the server MUST NOT return a message-body in the response.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

jakemac53:
Is this specifically for build_daemon?

annagrin:
> Is this specifically for build_daemon?
yes

annagrin:
Thanks Jake, this helps us improve expression evaluation startup time!

PR #3134, opened 2021-06-21
dont set the content length or body when responding to HEAD requests

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3129

Note that in the test I am getting out `0` and `''` for the content length/body (instead of null), but I am not setting those. I assume that shelf itself is plugging in some empty values here for convenience? cc @kevmoo not sure if you know.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 2.0.5-dev
+## 2.0.5
+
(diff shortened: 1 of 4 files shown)


kevmoo:
Sadly, I do not!

annagrin (commented):
LGTM!

annagrin (approved):

natebosch (approved):

PR #3144, opened 2021-07-14
Log requests and request failures in build daemon asset server

annagrin:
- Add `--log-requests` flag to build runner daemon command.
- Log request failures in asset server.
- Configure http client in tests to limit max connections per host,
  to prevent CI from failing with reset connection errors.

Extra logging will help diagnose CI flakes in build:

https://github.com/dart-lang/build/runs/3062926301

and webdev:

Related: https://github.com/dart-lang/webdev/issues/1345

### DIFF
--- a/_test/test/serve_integration_test.dart
+++ b/_test/test/serve_integration_test.dart
@@ -25,7 +25,10 @@ void main() {
   group('basic serve', () {
     setUpAll(() async {
       // These tests depend on running `test` while a `serve` is ongoing.
-      await startServer(ensureCleanBuild: true);
+      await startServer(
+        ensureCleanBuild: true,
+        buildArgs: ['--verbose', '--log-requests'],
+      );
     });
 
     tearDownAll(() async {
@@ -135,6 +138,8 @@ void main() {
       'web',
       '--build-filter',
       'web/sub/main.dart.js',
+      '--verbose',
+      '--log-requests',
       '--define',
       'build_web_compilers|ddc=environment={"message": "goodbye"}',
     ]);
--- a/_test/test/serve_integration_test.dart
+++ b/_test/test/serve_integration_test.dart
@@ -15,7 +15,13 @@ void main() {
   late HttpClient httpClient;
 
   setUpAll(() {
-    httpClient = HttpClient();
+    // Configure the client so it is possible to download a large number
+    // of files simultaneously. This prevents getting SocketException on
+    // too many connections.
+    httpClient = HttpClient()
+      ..maxConnectionsPerHost = 200
+      ..idleTimeout = Duration(seconds: 30)
+      ..connectionTimeout = Duration(seconds: 30);
   });
 
   tearDownAll(() {
@@ -27,7 +33,9 @@ void main() {
       // These tests depend on running `test` while a `serve` is ongoing.
       await startServer(
         ensureCleanBuild: true,
-        buildArgs: ['--verbose', '--log-requests'],
+        buildArgs: [
+          '--verbose',
+        ],
       );
     });
 
@@ -139,7 +147,6 @@ void main() {
       '--build-filter',
       'web/sub/main.dart.js',
       '--verbose',
-      '--log-requests',
       '--define',
       'build_web_compilers|ddc=environment={"message": "goodbye"}',
     ]);
@@ -161,5 +168,49 @@ void main() {
         await (await httpClient.get('localhost', 8080, 'main.sound.ddc.js'))
             .close();
     expect(await utf8.decodeStream(ddcFileResponse), contains('"goodbye"'));
+  });
+
+  test('should serve files in parallel', () async {
+    await startServer(buildArgs: [
+      'web',
+      '--build-filter',
+      'web/sub/main.dart.js',
+      '--verbose',
+      '--define',
+      'build_web_compilers|ddc=generate-full-dill=true',
+    ]);
+
+    addTearDown(() async {
+      await stopServer();
+    });
+
+    Future<void> read(String path) async {
+      try {
+        var request = httpClient.get('localhost', 8080, path);
+        var response = await (await request).close();
+        expect(
+          response.statusCode,
+          HttpStatus.ok,
+          reason: '$path ${response.reasonPhrase}',
+        );
+        await response.drain();
+      } catch (e, s) {
+        print('Test error on: $e:$s');
+        rethrow;
+      }
+    }
+
+    const n = 1000;
+    var futures = [
+      for (var i = 0; i < n; i++) read('main.sound.ddc.js'),
+      for (var i = 0; i < n; i++) read('main.sound.ddc.js.map'),
+      for (var i = 0; i < n; i++) read('main.sound.ddc.dill'),
+      for (var i = 0; i < n; i++) read('main.sound.ddc.full.dill'),
+      for (var i = 0; i < n; i++) read('sub/message.sound.ddc.js'),
+      for (var i = 0; i < n; i++) read('sub/message.sound.ddc.js.map'),
+      for (var i = 0; i < n; i++) read('sub/message.sound.ddc.dill'),
+      for (var i = 0; i < n; i++) read('sub/message.sound.ddc.full.dill'),
+    ];
+    await Future.wait(futures);
   });
 }
--- a/_test/test/serve_integration_test.dart
+++ b/_test/test/serve_integration_test.dart
@@ -4,7 +4,13 @@
 
 @TestOn('vm')
 import 'dart:convert';
-import 'dart:io' show HttpClient, HttpHeaders, HttpStatus;
+import 'dart:io'
+    show
+        HttpClient,
+        HttpClientRequest,
+        HttpClientResponse,
+        HttpHeaders,
+        HttpStatus;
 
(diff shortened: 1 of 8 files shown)


_test/test/serve_integration_test.dart
```diff
@@ -135,6 +138,8 @@ void main() {
       'web',
       '--build-filter',
       'web/sub/main.dart.js',
+      '--verbose',
```
annagrin, line 141: Trying to make the server emit verbose messages

build_runner/lib/src/server/server.dart
```diff
@@ -651,7 +658,7 @@ shelf.Handler _logRequests(shelf.Handler innerHandler) {
           request.requestedUri, request.method, watch.elapsed);
       logFn(msg);
       return response;
-    }, onError: (Object error, StackTrace stackTrace) {
+    }).catchError((Object error, StackTrace stackTrace) {
```
annagrin, line 661: This change keeps correct the original stacktrace from the inner handler, as opposed to  creating a new one originating in _logResuests

grouma (approved):

_test/test/serve_integration_test.dart
```diff
@@ -15,7 +15,13 @@ void main() {
   late HttpClient httpClient;
 
   setUpAll(() {
-    httpClient = HttpClient();
+    // Configure the client so it is possible to download a large number
+    // of files simultaneously. This prevents getting SocketException on
+    // too many connections.
+    httpClient = HttpClient()
```
grouma, line 21: My concern is that this is just hiding an underlying SDK issue. Have you shared this repro with the SDK team?

_test/test/serve_integration_test.dart
natebosch: How about we go with https://github.com/dart-lang/build/pull/3145 and enable this more centrally.

_test/test/serve_integration_test.dart
```diff
@@ -157,4 +169,48 @@ void main() {
...
+        var request = httpClient.get('localhost', 8080, path);
+        var response = await (await request).close();
+        expect(
+          response.statusCode,
+          HttpStatus.ok,
+          reason: '$path ${response.reasonPhrase}',
+        );
+        await response.drain();
+      } catch (e, s) {
+        print('Test error on: $e:$s');
```
natebosch, line 198: Why do we need this?

build_runner/lib/src/server/server.dart
natebosch: Can you elaborate on this? I'm not sure how this impacts the `StackTrace`.

_test/test/serve_integration_test.dart
annagrin: I created this repro: https://github.com/dart-lang/sdk/issues/46647
Will try to find someone to redirect it too as well.

_test/test/serve_integration_test.dart
annagrin: Oh forgot to remove, thanks!

build_runner/lib/src/server/server.dart
annagrin: Just from my experiments, adding a `throw` statement inside the build runner server handler: 
- onError reports a stack trace originating in the `_logRequests` handler.
- catchError reports the original stack trace, starting in the server's handler.

I am not sure why the behavior is different though, I would expect it to be the same.

build_runner/lib/src/watcher/graph_watcher.dart
```diff
@@ -55,11 +55,13 @@ class PackageGraphWatcher {
                 .watch()
                 .where(_nestedPathFilter(w.node))
                 .handleError((Object e, StackTrace s) {
-              _logger.severe(
-                  'Error from directory watcher for package:${w.node.name}\n\n'
-                  'If you see this consistently then it is recommended that '
-                  'you enable the polling file watcher with '
-                  '--use-polling-watcher.');
+              _logger
+                ..severe('Error in watcher for package: ${w.node.name}: $e:$s')
```
jakemac53, line 59: We probably don't want separate logs here, and the stack trace and exception should be passed using the explicit parameters for those

build_runner/lib/src/server/server.dart
natebosch: I'm not able to to reproduce this. I tried adding a throw in `AssetHandler._handle` and the stack trace points there with either `onError:` or `.catchError`. Was this possibly related to other changes you had made? Can you still see a behavior difference?

annagrin:
@natebosch @grouma I am thinking about cancelling this PR, as I actually only need one change in it (logging the exception that happens in the asset server handler). I realized that I can log requests from webdev as well, without adding flags to the build daemon, and the rest of the changes  are just to enable more logging to debug the failures I encountered in this PR. 

Anyway, it helped me narrow down the actual issue and find the solution for it so there was a benefit:) I'll submit a new one just for logging exceptions.

build_runner/lib/src/watcher/graph_watcher.dart
annagrin: This is the only change I actually want to keep:) the exception propagates to the shelf handler, and the stack trace gets shortened intentionally sp not much information is left. It looks like:

``
SocketException: Write failed (OS Error: Connection reset by peer, errno = 104), address = localhost, port = 35146
package:shelf/src/middleware/logger.dart  logRequests.<fn>.<fn>.<fn>
``

Or it would have a one-liner stack pointing to the shelf handler if I disable logging requests.

annagrin: Update - sorry, my mistake, this is just my addition for debugging, I don't need to keep it.

build_runner/lib/src/server/server.dart
```diff
@@ -317,53 +317,60 @@ class AssetHandler {
...
+      }
+      List<int>? body;
+      if (request.method != 'HEAD') {
+        body = await _reader.readAsBytes(assetId);
+        headers[HttpHeaders.contentLengthHeader] = '${body.length}';
+      }
+      return shelf.Response.ok(body, headers: headers);
+    } catch (e, s) {
+      _logger.finest('Error on request '
+          '${request.method} ${request.requestedUri}: $e:$s');
```
annagrin, line 371: basically I only need this and the tests. will update.

build_runner/lib/src/server/server.dart
natebosch: Let's go with https://github.com/dart-lang/build/pull/3149

If you can still repro a confusing stack trace after that one and #3148 are in we can revisit. I'm not sure I saw the _same_ confusing stack traces as you, but I'm pretty confident that the async/await pattern stands the best chance at giving us useful stack traces in the most situations on the VM since it has better handling for `await` than raw futures.

annagrin: sound good, thanks!

_test/test/serve_integration_test.dart
annagrin: removed.

build_runner/lib/src/server/server.dart
annagrin: done

_test/test/serve_integration_test.dart
annagrin: replaced by 'fail' with extra information

annagrin:
@natebosch @jakemac53 @grouma I addressed (i think) all comments.

Changes I kept - let me know if you would like me to remove any of them:

- add a --log-requests flag to build daemon
- test parallel download of many files in _test/test/serve_integration_tests.dart
- test parallel download of many files in build_runner/test/serve_integration_tests.dart
- log failures in asset server handler
- limit simultaneous connections on http client in tests to prevent flakes in CI.

build_runner/lib/src/server/server.dart
```diff
@@ -317,53 +317,60 @@ class AssetHandler {
...
+      rethrow;
```
jakemac53, line 372:
```suggestion
      _logger.finest(
          'Error on request ${request.method} ${request.requestedUri}',
          e, s);
      rethrow;
```

annagrin: I'll submit it in another commit so i can format it, thanks!

jakemac53 (approved):

grouma (approved):

_test/test/serve_integration_test.dart
```diff
@@ -15,7 +21,13 @@ void main() {
   late HttpClient httpClient;
 
   setUpAll(() {
-    httpClient = HttpClient();
+    // Configure the client so it is possible to download a large number
```
natebosch, line 24: When you merge please update the commit message to include this change.

build_runner/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## 2.0.7-dev
```
natebosch, line 1:
```suggestion
## 2.1.0-dev
```

Since we are adding a new feature we should bump the second number.

natebosch (approved):

_test/test/serve_integration_test.dart
annagrin: Will do!

PR #3148, opened 2021-07-19
Avoid folding frames for build system bugs

natebosch:
The intent of the `.foldFrames` call is to reduce the noise for stack
traces when an exception is thrown inside a builder implementation. The
folding makes internal bugs harder to diagnose though, because it can
mask the source of exceptions that are thrown within `build_runner`
code.

Check if the lowest frame of the trace is within one of the build
system implementation packages, and if it is do no folding.

Add `build_runner_core` to the list of packages treated as
implementation of the build system.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,5 @@
+## 2.0.7-dev
+
 ## 2.0.6
 
 - Allow analyzer version 2.x.x.
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -36,8 +36,8 @@ StringBuffer colorLog(LogRecord record, {required bool verbose}) {
   if (record.stackTrace != null && verbose) {
     var trace = Trace.from(record.stackTrace!);
     const buildSystem = {'build_runner', 'build_runner_core', 'build'};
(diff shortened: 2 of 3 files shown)


natebosch:
I noticed this bad UX while trying to reproduce the behavior in https://github.com/dart-lang/build/pull/3144/files/95be61b41858c603be5983914c3ea285b81d7bd4#r669994466

The first thing I had to do to see this, though, was change the non-sensible `log.severe` call to pass the `StackTrace` to the logger instead of trying to format the string manually.

https://github.com/dart-lang/build/blob/97533977ba4a2260b55117eb91156f702178a554/build_runner/lib/src/server/server.dart#L658

With that fixed and the trace folding coming in to play I was able to see a difference between `onError:` and `.catchError` in that the first frame on the folded trace changed from `package:build_runner                _logRequests.<fn>` to `package:build_runner                ServeHandler.handlerFor.<fn>`. Neither of these is all that use, it's still masking the actual origin of the exception.

Keeping that fix, and also applying the change in this PR the first frame in the trace is now the correct origin for the exception `package:build_runner/src/server/server.dart 319:55  AssetHandler._handle`. This is with the case with either `onError:` or `.catchError`.

I'm not sure if this is actually the same problem hit in that PR, but it's likely to be useful in general for any debugging of bugs/exceptions in the build system itself.

jakemac53 (approved):

PR #3152, opened 2021-07-23
Support capture groups in build extensions

### LINKED ISSUE #1689: Enhance the expressiveness of `buildExtensions` with allowing directory moves
Build extensions covers a large set of use cases and has a simple implementation, but doesn't cover more complex situations.

We should consider selectively expanding what we can express with specific use cases that are likely to happen. One that has come up a couple times is the idea of parallel directory structure. For instance writing outputs to `/web/docs` when the inputs are in `/docs` - but keeping the remaining paths identical.

When we [originally proposed](https://github.com/dart-lang/build/issues/222) the build extension scheme we did suggest the configuration language could be expanded - our criteria was just that it was static configuration over a snippet of code.

Pros:
- Not overly complicated to support for a big jump in possibilities.
- Still high on convention so does not drastically increase how hard it is to understand the build (with the exception of `generate_for`/`sources` interaction noted below). May make some builds easier to understand if they are today working around this with hacks.

Cons:
- Would be a breaking change to introduce a new field on `Builder` to hold this information.
- Will be an extra configuration option that most builders probably won't use - expanding the configuration surface area is always costly.
- We deal with build extensions in a few different places and we'd need to be careful that they are all updated in sync.
- The interaction between `generate_for` and a target's `sources` is already hard to understand - this would make it even harder.

I think if we can find a few use cases that would be solved by something simple here we should consider it.

simolus3:
Support capture groups for builders.

Closes #1689.


build/lib/src/generate/expected_outputs.dart
```diff
@@ -7,22 +7,115 @@ import '../builder/builder.dart';
...
+      // The input does not contain a capture group, so we should simply match
+      // all files whose paths ends with the desired input.
+      return _ParsedBuildOutput(
+          RegExp('${RegExp.escape(input)}\$'), outputs, false);
+    }
+
+    final match = groups.current;
+
+    if (groups.moveNext()) {
+      throw ArgumentError(
```
jakemac53, line 54: This error (and others in this function) should mention the builder as well in this error message.

build/lib/src/generate/expected_outputs.dart
```diff
@@ -7,22 +7,115 @@ import '../builder/builder.dart';
...
+      // input path.
+      regexBuffer
+        ..write('^')
+        ..write(RegExp.escape(input.substring(1, match.start)));
+    } else {
+      regexBuffer.write(RegExp.escape(input.substring(0, match.start)));
+    }
+
+    regexBuffer
+      ..write('(.+)') // capture group
```
jakemac53, line 85: A bit of an open question if this should be `.+` or `.*` I suppose, but I think probably requiring some matching text here is the correct choice.

build/lib/src/generate/expected_outputs.dart
```diff
@@ -7,22 +7,115 @@ import '../builder/builder.dart';
...
+      throw ArgumentError(
+        'The builder declares an input "$input" which contains multiple '
+        'groups ("{{}}"). This is not supported.',
+      );
+    }
+
+    // When using a capture group in the build input, it must also be used in
+    // every output to ensure outputs have unique names.
+    for (final output in outputs) {
+      if (!_captureGroup.hasMatch(output)) {
```
jakemac53, line 63: We want to assert exactly one match here also.

build/lib/src/generate/expected_outputs.dart
```diff
@@ -7,22 +7,115 @@ import '../builder/builder.dart';
...
+  _ParsedBuildOutput(this._pathMatcher, this._outputs, this._usesCaptureGroup);
+
+  factory _ParsedBuildOutput.parse(String input, List<String> outputs) {
+    final groups = _captureGroup.allMatches(input).iterator;
+
+    if (!groups.moveNext()) {
+      // The input does not contain a capture group, so we should simply match
+      // all files whose paths ends with the desired input.
+      return _ParsedBuildOutput(
+          RegExp('${RegExp.escape(input)}\$'), outputs, false);
```
jakemac53, line 48: It would be good to understand the performance difference between this RegExp and a simple `endsWith` check.

We could probably avoid using a regex here by making the path matcher be a Function instead of a RegExp?

build/test/generate/expected_outputs_test.dart
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) 2021, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'package:build/build.dart';
+import 'package:build_test/build_test.dart';
+import 'package:test/test.dart';
+
+void main() {
```
jakemac53, line 9: Lets also add some tests validating the behavior when there is a capture group at the start or end of an input string, as well as when the match is potentially ambiguous, to demonstrate the eagerness (or not) of the matching.

build_runner_core/lib/src/asset_graph/graph.dart
```diff
@@ -374,17 +374,20 @@ class AssetGraph {
...
     if (!action.generateFor.matches(input)) return false;
-    Iterable<String> inputExtensions;
+
     if (action is InBuildPhase) {
-      inputExtensions = action.builder.buildExtensions.keys;
+      if (expectedOutputs(action.builder, input).isEmpty) return false;
     } else if (action is PostBuildAction) {
-      inputExtensions = action.builder.inputExtensions;
+      // todo: Do we want to allow capture groups for post process builders?
+      // I guess not?
```
jakemac53, line 382: They can already output whatever they want, but it would allow some more interesting matching on input extensions....

If it isn't supported (fine for now imo), we should throw if we see them so people know it isn't supported.

build_runner_core/lib/src/asset_graph/graph.dart
```diff
@@ -374,17 +374,20 @@ class AssetGraph {
   bool _actionMatches(BuildAction action, AssetId input) {
     if (input.package != action.package) return false;
     if (!action.generateFor.matches(input)) return false;
-    Iterable<String> inputExtensions;
+
     if (action is InBuildPhase) {
-      inputExtensions = action.builder.buildExtensions.keys;
+      if (expectedOutputs(action.builder, input).isEmpty) return false;
```
jakemac53, line 379: We should consider separating out a different function that just checks for matching input extensions, and doesn't also build up the outputs.

example/build.yaml
```diff
@@ -25,3 +25,10 @@ builders:
     builder_factories: ["cssBuilder"]
     build_extensions: {"$package$": ["web/generated.css"]}
     auto_apply: root_package
+
+  movingBuilder:
```
jakemac53, line 29: nit: the name of this builder is a bit misleading, it isn't moving files around. Maybe `fileDiagnosticsBuilder`? (same for the class etc).

example/lib/example.dart
```diff
@@ -76,3 +76,25 @@ Visible libraries: $visibleLibraries
     '.dart': ['.dart.info']
   };
 }
+
+class MovingBuilder implements Builder {
```
jakemac53, line 80: It might be good to add some comments on this builder explaining what it does and why that is interesting as an example :).

jakemac53 (commented):
This is generally looking good! Nice to see that only one place really had to be updated.

We should consider I think still whether multiple capture groups should be supported - but that could also just be added later if people need it? We could consider `{{}}` to be an "unnamed" capture group, and then allow named groups like `{{x}}`, which wouldn't conflict.

build/lib/src/generate/expected_outputs.dart
simolus3: Is this really necessary though? If I have `build_extensions: {'lib/{{}}.dart': ['lib/generated/{{}}/{{}}.g.dart']}` then I'd generate files in weird locations, but the generated files are still unique so it might not be problem? It's totally fair if we just want to avoid a footgun, I'm just wondering if there's something else I might be missing here.

jakemac53: I think we just want to avoid the footgun here - we could always add support for that if there was a useful case that comes up, but we can't as easily remove the support.

build/lib/src/generate/expected_outputs.dart
simolus3: I've opted to split the functionality for inputs using capture groups / no capture groups into different classes. A function for the path matcher would have been a bit of a hassle since we need a `Match` in the regex case but a `bool` for the simple suffixes.

simolus3:
Thanks for your feedback. Regarding the docs and the example, I've mainly used that for testing purposes for now. I'll try to come up with something more practical and add a new section in the documentation on builders.

> We should consider I think still whether multiple capture groups should be supported - but that could also just be added later if people need it?

My understanding was that this maps poorly to Bazel? But on the other hand that's not my problem to worry about :D I can add support for multiple or named capture groups if we want to support this now.

jakemac53:
Note that I tried using this for protobuf (to make a basic protobuf builder), and it does seem to work 👍 https://github.com/jakemac53/protobuf/tree/proto-builder.

> Regarding the docs and the example, I've mainly used that for testing purposes for now. I'll try to come up with something more practical and add a new section in the documentation on builders.

👍 

> My understanding was that this maps poorly to Bazel? But on the other hand that's not my problem to worry about :D I can add support for multiple or named capture groups if we want to support this now.

It would be harder to implement in bazel, but should be feasible. However, I would probably just stick with the one capture group for now.

build/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.1.0-dev
+
+- Expand the expressiveness of `buildExtensions` to allow directory moves.
```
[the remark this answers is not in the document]
simolus3: I've expanded the section on outputs in `writing_a_builder.md` and linked the relevant parts from the changelog.

docs/writing_a_builder.md
```diff
@@ -31,6 +31,48 @@ will be passed as a primary input, and each build step may produce two assets.
...
 `some_file.txt.foo` and `some_file.txt.bar`.
 
+Keys in `buildExtensions` match a suffix in the path of potential inputs. That
+is, a builder will run when an input ends with its input extension.
+Valid outputs are formed by replacing the matched suffix with values in that 
+map. For instance, `{'.dart': ['.g.dart']}` matches all files ending with
+`.dart` and allows the builder to write a file  with the same name but with a
+`.g.dart` extension instead.
+
+Builders can declare more complex inputs and outputs by using a capture group
```
jakemac53, line 41: I think it would be good to add a header for this section:

```suggestion
### Capture Groups

Builders can declare more complex inputs and outputs by using a capture group
```

jakemac53 (commented):
This is looking great! Let me know once you think it is ready for a final review. I do think we want at least a few basic integration tests as well (something using `testBuilders` should be sufficient).

example/lib/example.dart
simolus3: I've replaced it with a new builder that reads strings from somewhere in `assets/` and generates matching Dart constants in `lib/`. I think this may be closer to a practical use of this feature as well.

simolus3:
I think it should be ready now :) I've added a simple integration test for capture groups and one to ensure that the right source is recognized when using optional builders with a capture group. Let me know if you have more ideas for integration tests.

I have some ideas to build on this feature, but I'll open separate issues from them.

jakemac53 (commented):
Lets bump the version for build_runner_core to `7.1.0` as well, as this is effectively a new feature we are supporting.

There is a bit of a weird versioning issue here also... somebody could get a new version of build, and a builder that relies on this feature, but in an older version of build_runner. I really don't think we want to do a breaking change in build_runner though, so I am not sure the best way to resolve the issue.

simolus3:
> somebody could get a new version of build, and a builder that relies on this feature, but in an older version of build_runner_core

Good catch, I don't have a good solution for this either. We could be optimistic and assume that people get new versions of their builders with `pub upgrade`, so then they hopefully get the latest `build_runner_core` as well.

jakemac53:
> Good catch, I don't have a good solution for this either. We could be optimistic and assume that people get new versions of their builders with `pub upgrade`, so then they hopefully get the latest `build_runner_core` as well.

I spoke with @natebosch about this (in person, what a concept!). We decided the best approach would be to leave the versions as is, but bump the analyzer constraint in `build` to match the highest analyzer constraint from any dependency that build_runner_core has. Typically analyzer would be the main reason for somebody to be held back on an earlier build_runner_core, so that should mitigate most issues (by not allowing those users to get the latest build, either).  Since we only care about real deps and not dev deps there, it looks like that means updating the analyzer constraint to `>=1.5.0 <3.0.0' in `build`.

jakemac53 (approved):

build/lib/src/generate/expected_outputs.dart
```diff
@@ -7,22 +7,148 @@ import '../builder/builder.dart';
...
+    // Start writing the regex up to the capture group
+    if (input.startsWith('^')) {
+      // Build inputs starting with `^` start matching at the beginning of the
+      // input path.
+      regexBuffer
+        ..write('^')
+        ..write(RegExp.escape(input.substring(1, match.start)));
+    } else {
+      regexBuffer.write(RegExp.escape(input.substring(0, match.start)));
+    }
```
natebosch, line 90: What cases is this conditional impacting the end result?  We know the `.substring(1` call is always removing exactly `^`, and that is what we replace it with.

Can this be

```dart
final regexBuffer = StringBuffer(input.substring(0, match.start));
```

docs/writing_a_builder.md
```diff
@@ -31,6 +36,43 @@ will be passed as a primary input, and each build step may produce two assets.
 For the primary input `some_file.txt` the allowed outputs are
 `some_file.txt.foo` and `some_file.txt.bar`.
 
+### Capture groups
+
+Builders can declare more complex inputs and outputs by using a capture group
+in their build input. Capture groups allow builders to move files across
```
natebosch, line 42: [nit] I worry that "files" could be misconstrued to include the input.

```suggestion
in their build input. Capture groups allow builders to move outputs across
```

docs/writing_a_builder.md
```diff
@@ -31,6 +36,43 @@ will be passed as a primary input, and each build step may produce two assets.
...
+- Capture groups match at least one character in the input path, but may match
+  arbitrarily many characters.
+- When the input uses a capture group, every output must reference that capture
+  group as well.
+- It is not currently allowed to declare multiple capture groups in one input
+  extension.
+- Capture groups match as many characters as possible. In the proto example,
+  the asset `proto/nested/proto/test.proto` would match with `{{}}` capturing
+  `nested/proto/test`. The shorter suffix match with `{{}}` capturing just
+  `test` is not considered.
```
natebosch, line 62: Can we add a sentence with something like. "If the `{{}}` group is at the start of the input it will match the full path from the root of the package. To match only file names start the input expression with `/{{}}`.

natebosch (approved):

docs/writing_a_builder.md
jakemac53: Or maybe
```suggestion
in their build input. Capture groups allow builders to copy files across
```

docs/writing_a_builder.md
jakemac53: Because of the eager matching, this would actually match the entire path after the first `/` today.

natebosch: Oh right.

Hmm, does that mean this implementation does not cover this use case?
https://github.com/dart-lang/build/issues/1689#issuecomment-728194831

build/lib/src/generate/expected_outputs.dart
simolus3: What if there are other characters from `1` to `match.start` that might need escaping? If `input` was `^.hidden/{{}}`, I think we'd want to generate a `^\.hidden/(.*)` regex, right? We can't escape all of it either since we need the `^` at the beginning.

natebosch: Oh right, I completely overlooked the escaping. 😳

docs/writing_a_builder.md
jakemac53: Good point, I do think we need to think about how to enable that case. The current design works well for moving files from one top level directory to another, but not for moving files more locally as in this example.

jakemac53: I think if we had a way for you to configure whether the match should match directory separators, that could enable this?

Should we have another type of capture group possibly? Or can we think of a better way to configure the existing type?

jakemac53: One option could be just allowing multiple capture groups:

```yaml
build_extensions:
  '{{0}}/{{1}}.graphql':
    - '{{0}}/graphql/{{1}}.data.gql.dart'
```

docs/writing_a_builder.md
simolus3: I went with "Capture groups can be used to write outputs in a different directory than the primary input." which stays closer to build's common terminology.

docs/writing_a_builder.md
simolus3: I think I prefer multiple capture groups if we can support them on all build implementations. They're more powerful and efficient than lazily-matched groups. Should I do that in this PR or open another one?

-----

Even if we allowed a lazy capture group, `\/(.+?)\.graphql$` would still start matching after the first `/`:

```dart
void main() {
  final regex = RegExp(r'\/(.+?)\.graphql$');
  
  // only finds a single match, with the group capturing `foo/bar/baz `
  regex.allMatches('lib/foo/bar/baz.graphql').forEach((match) {
    print(match.group(1));
  });
}
```

I don't know if there's a way to make a group capture the least amount of chars on the right, I'm not that good at regex :D We could implicitly start the regex with `.*` if there's no `^` in the input, but then we'd have problems with capture groups at the beginning.

jakemac53: I think I prefer the multiple capture groups approach as well.

For bazel we can't use regex, and we also can't use recursion, which makes things a bit interesting. But I think it should be possible to implement still.

We almost certainly could support some hard coded number of capture groups in any case, which would be sufficient.

So I think just supporting multiple (named) capture groups is the way to go. I think leaving unnamed capture groups is still also fine.

simolus3: I've added support for multiple capture groups in 2349c1f7e516da07291374d2f2dba9eb4ae86458. I also added another subsection showing how to use them for the "generate in relative subdirectory" usecase.

build/lib/src/generate/expected_outputs.dart
```diff
@@ -42,59 +42,74 @@ extension on AssetId {
 }
 
 abstract class _ParsedBuildOutputs {
-  static final RegExp _captureGroup = RegExp('{{}}');
+  static final RegExp _captureGroup = RegExp('{{([\\w\\d]*)}}');
```
jakemac53, line 45:
```suggestion
  static final RegExp _captureGroup = RegExp('{{(\\w*)}}');
```

word characters already include 0-9 :)

build/lib/src/generate/expected_outputs.dart
```diff
@@ -42,59 +42,74 @@ extension on AssetId {
...
 
   _ParsedBuildOutputs();
 
   factory _ParsedBuildOutputs.parse(
       Builder builder, String input, List<String> outputs) {
-    final groups = _captureGroup.allMatches(input).iterator;
+    final groups = _captureGroup.allMatches(input);
```
jakemac53, line 51:
```suggestion
    final groups = _captureGroup.allMatches(input).toList();
```

Otherwise we end up actually performing the match multiple times. The work actually happens late in the [iterator](https://github.com/dart-lang/sdk/blob/master/sdk/lib/_internal/vm/lib/regexp_patch.dart#L373), so calling methods that grab a new iterator causes it to redo all the logic each time.

I confirmed this with [a benchmark](https://gist.github.com/jakemac53/e62fb721f6cf122099c0d818af2742ff) as well and creating the list is still more efficient, although there may be a way to restructure the method so that we don't have to create the list either, but probably not a big deal.

jakemac53: I think we should consider renaming this variable `matches`

build/lib/src/generate/expected_outputs.dart
```diff
@@ -147,7 +168,22 @@ class _CapturingBuildOutputs extends _ParsedBuildOutputs {
...
     return _outputs.map((output) {
-      final resolvedOutput = output.replaceFirst('{{}}', match.group(1)!);
+      final resolvedOutput = output.replaceAllMapped(
+        _ParsedBuildOutputs._captureGroup,
+        (outputMatch) {
+          final name = outputMatch.group(1)!;
+          final index = _groupNames.indexOf(name);
+
+          if (index < 0) {
+            // The output refers to a capture group that does not appear in the
```
jakemac53, line 178: I think we should make it an error to have a capture group in outputs that doesn't match one from inputs. I don't think we should support `{{}}` as raw text in output extensions, its much more likely to just be a typo. 

If somebody really needs that we could allow escaping it later or something haha, but I highly doubt this would come up.

build/lib/src/generate/expected_outputs.dart
```diff
@@ -42,59 +42,74 @@ extension on AssetId {
...
+      positionInInput = 1;
     }
 
-    // When using a capture group in the build input, it must also be used in
-    // every output to ensure outputs have unique names.
-    for (final output in outputs) {
-      if (_captureGroup.allMatches(output).length != 1) {
+    // Builders can have multiple capture groups, which are disambiguated by
+    // their name. Names can be empty as well: `{{}}` is a valid capture group.
+    final names = <String>{};
```
jakemac53, line 68: I would actually just use a list here, and do a `contains` check below. There will never be a large number of things in it, so a list is actually more efficient. And then we don't need to convert it to a list later.

We also rely on the insertion order iteration, which not all sets do. If we do leave it as a set I would explicitly use the `LinkedHashSet` constructor (even though this is the default type), with a comment about insertion order iteration. Just so it is very explicit.

docs/writing_a_builder.md
```diff
@@ -74,6 +72,39 @@ the following noteworthy properties:
   If the builder had used `^proto/{{}}.proto` as an input, it would not have
   run on strict suffix matches.
 
+#### Using multiple groups
```
jakemac53, line 75:
```suggestion
#### Using multiple capture groups
```

docs/writing_a_builder.md
```diff
@@ -74,6 +72,39 @@ the following noteworthy properties:
...
+
+A builder may use multiple capture groups in an input. Groups can be given a
```
jakemac53, line 77:
```suggestion
A builder may use multiple capture groups in an input. Groups must be given a
```

jakemac53:
This feature is 🔥 - excited to get it out to users!

I do think another feature later on might be useful to allow users to somehow configure the output extensions of builders they depend on, as discussed in the other issue you filed. But it is I think a totally separate issue, and would take some thought as to how best to implement it. The builders need to be compatible with it so they output the right files also.

build/lib/src/generate/expected_outputs.dart
simolus3: I agree that being more explicit about insertion order is important, I've changed this to use a list.

docs/writing_a_builder.md
simolus3: I wasn't sure how to describe this since you can technically have one unnamed group (`{{}}/{{file}}.proto` is valid), but I agree that "must" fits better here :)

simolus3:
> But it is I think a totally separate issue, and would take some thought as to how best to implement it. The builders need to be compatible with it so they output the right files also.

I guess we'll make most build users happy if `sourceGen|combiningBuilder` had this option. For most builds that's the only visible source of generated files, so most other builders wouldn't have to change. We could add configured output extensions to `BuilderOptions` and encourage build authors to use that value for `buildExtensions`. Then it's just a matter of making outputs configurable in `TargetBuilderConfig`. But I agree that we should probably fletch this out in another issue.

build/test/generate/expected_outputs_test.dart
```diff
@@ -0,0 +1,162 @@
...
+        ),
+        throwsArgumentError,
+      );
+    });
+
+    test('output may not use capture groups not used in the input', () {
+      expect(
+        () => expectedOutputs(
+          TestBuilder(buildExtensions: {
+            '{{input}}.txt': ['{{output}}.foo']
```
jakemac53, line 95:
```suggestion
            '.txt': ['{{output}}.foo']
```

(this test wasn't currently distinguishing between two potential errors)

simolus3: Good catch, this check was also missing in the code because we just use raw suffixes if there's no capture group in the input.

docs/writing_a_builder.md
jakemac53: thanks, this LGTM 👍

PR #3170, opened 2021-08-16
log a better and more consistent message when we encounter dart: uris in conditional imports

jakemac53:
Fixes https://github.com/dart-lang/sdk/issues/46918

Example output:

```
[SEVERE] build_modules:module_library on test/configurable_uri_test.dart:

Invalid argument(s): Unsupported conditional import of `dart:io` found in _test|test/configurable_uri_test.dart:

import 'dart:io';

This environment does not support direct conditional imports of `dart:` libraries. Instead you must create a separate library which imports (or exports) the `dart:` library that you want to use, and conditionally import (or export) that library.

### DIFF
--- a/build_modules/lib/src/module_library.dart
+++ b/build_modules/lib/src/module_library.dart
@@ -78,12 +78,24 @@ class ModuleLibrary {
       if (directive is! UriBasedDirective) continue;
       var path = directive.uri.stringValue;
       if (path == null) continue;
+
+      List<Configuration>? conditionalDirectiveConfigurations;
+      if (directive is ImportDirective && directive.configurations.isNotEmpty) {
+        conditionalDirectiveConfigurations = directive.configurations;
+      } else if (directive is ExportDirective &&
+          directive.configurations.isNotEmpty) {
+        conditionalDirectiveConfigurations = directive.configurations;
+      }
+
       var uri = Uri.parse(path);
       if (uri.isScheme('dart-ext')) {
         // TODO: What should we do for native extensions?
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jakemac53** [comment]
Note that I am using the `AST.toSource()` method from the analyzer to log the directive text, which doesn't appear to log the conditional imports portion of a directive. cc @scheglov

**Comment by scheglov** [comment]
I will enhance `ToSourceVisitor` to support conditional URIs.

**Comment by scheglov** [comment]
https://dart-review.googlesource.com/c/sdk/+/210281

**Review by natebosch — APPROVED** [review]

File: build_modules/lib/src/module_library.dart [hunk 1]
```diff
@@ -127,6 +129,23 @@ class ModuleLibrary {
...
+    if (parsedUri == null) {
+      throw ArgumentError(
+          'Unsupported conditional import with non-constant uri found in $id:'
+          '\n\n${node.toSource()}');
+    } else if (parsedUri.scheme == 'dart') {
+      throw ArgumentError(
+          'Unsupported conditional import of `$parsedUri` found in $id:\n\n'
+          '${node.toSource()}\n\nThis environment does not support direct '
+          'conditional imports of `dart:` libraries. Instead you must create '
+          'a separate library which imports (or exports) the `dart:` library '
```
mraleph, line 143: consider `which unconditionally imports (or exports) the \`dart:\` library` to make it crystal clear.

mraleph (approved):
Thanks!

PR #3171, opened 2021-08-16
Add integration test for changing build extensions

### LINKED ISSUE #683: BuilderOptions can't change buildExtensions
Example use case: angular_components want SASS to output `.scss.css` while other packages want it to output `.css`. We have not solution for this with config driven generation.

In bazel this is handled by using custom skylark code, but as we look to revisit dazel we don't have a solution for package authors providing us custom skylark.

We have a few options:
1. Disallow varying the build extensions by config. angular_components can either switch to `.css`, or the SASS package can vend two builders - one which uses `.css` and one which uses `.scss.css` and angular_components can choose to apply the latter and disable the former.
1. Write a configuration language that allows us to understand this statically and be able to use it in both bazel and build_runner. IMO the complexity of this makes it a non-option.
1. Allow this in build_runner and make it clear that builders which use the flexibility won't work in bazel. This let's us punt on the problem and we can eventually allow this in bazel by solving the custom skylark problem. The downside is that any package which takes advantage of this will be blocked from using bazel until we solve that problem - and anything that depends on them will be blocked too.

simolus3:
Adds a test for a builder changing its extensions after re-running a build script. I also clarified the dart docs on `buildExtensions` to point out that extensions may be chosen based on builder options and linked to the detailed documentation for outputs.

I guess this closes #683?

### DIFF
--- a/build/lib/src/builder/builder.dart
+++ b/build/lib/src/builder/builder.dart
@@ -17,10 +17,13 @@ abstract class Builder {
   /// from the values in this map are expected as outputs.
   ///
   /// - If an empty key exists, all inputs are considered matching.
-  /// - A builder must always return the same configuration. Typically this will
-  /// be `const` but may vary based on build arguments.
+  /// - An instance of a builder must always return the same configuration.
+  ///   Typically, a builder will return a `const` map. Builders may also choose
+  ///   extensions based on [BuilderOptions].
   /// - Most builders will use a single input extension and one or more output
-  /// extensions.
+  ///   extensions.
+  /// - For more information on build extensions, see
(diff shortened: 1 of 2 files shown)


jakemac53:
There are still some general questions around this functionality, but I suppose that an integration test ensuring we don't break what works today is probably good either way (especially since we know some builders are exploiting it). @natebosch wdyt?

natebosch:
> There are still some general questions around this functionality, but I suppose that an integration test ensuring we don't break what works today is probably good either way (especially since we know some builders are exploiting it). @natebosch wdyt?

Yeah I think at least making sure we're aware and intentional if we do end up having to break this is a good idea.

jakemac53 (approved):

PR #3174, opened 2021-08-19
Update build_web_compiler:entrypoint outputs to support multiple DDC apps on the same page

evanweible-wf:
## Changes
Currently, serving a page that loads more than one `.dart.js` entrypoint (when they are compiled with DDC) results in only the last one actually executing. This is due to the way the entrypoint JS loads requirejs and queues up the loading of its bootstrap file. Previously, it used the [`data-main` attribute](https://requirejs.org/docs/api.html#data-main) on the require.js `<script>` element, but that only works if there's only one main to call. In the case of multiple entrypoints, each one effectively overwrites the previous `data-main`.

This commit changes the way the bootstrap module is loaded and executed to a way that is compatible with multiple DDC entrypoints on the same page. This is accomplished by generating an additional `.bootstrap.require.js` output from the `build_web_compilers|entrypoint` builder that will be responsible for calling `require([...])` with the bootstrap module name, and then updating the entrypoint JS output to load this new asset after loading `require.js`.

It also changes the logic responsible for populating `window.$dartLoader` so that it appends entrypoint-specific data rather than setting it only once. This allows multiple entrypoints to run and augment that global instead of only the first.

---

Functionally, this accomplishes our goal of being able to at least load and run multiple DDC entrypoints on the same page without any errors. But there may be a better way to do this that maybe doesn't require an extra builder output. I originally tried adding that call to `require(["<module_name>"]);` via an inline script, but because the `require.js` script is deferred loaded, the inline script was running too early. I also tried setting up an onload listener for `require.js`, but that got complicated because each entrypoint adds its own <script> to load it, and if I understood correctly, the onload event wouldn't fire if the script had already been loaded. Ultimately, it seemed that the simplest option for the runtime logic was to load a script that would call `require()` in the same way we were loading the other scripts to ensure that they'd be run in order.

Open to suggestions on any and all of this!

## Testing

For testing purposes, I've been using a very simple app:

```html
<!-- web/index.html -->
<!DOCTYPE html>
<html>
  <head>
    <title>App with multiple DDC entrypoints</title>
  </head>
  <body>
    <h1>App</h1>
    <p id="one">Loading one.dart...</p>
    <p id="two">Loading two.dart...</p>
    <script defer src="one.dart.js"></script>
    <script defer src="two.dart.js"></script>
  </body>
</html>
```

```dart
// web/one.dart
import 'dart:async';
import 'dart:html';

void main() {
  querySelector('#one').innerText = 'one.dart loaded';
  print('one');
}
```

```dart
// web/two.dart
import 'dart:async';
import 'dart:html';

void main() {
  querySelector('#two').innerText = 'two.dart loaded';
  print('two');
}
```

Without these changes, you'd see only the second entrypoint run. With the changes, both run.
_Note: if serving with webdev, you'll also need `--no-injected-client` as the webdev server and injected client code make some similar assumptions about only having one entrypoint on the page. I'm going to put up a similar PR to webdev._

### DIFF
--- a/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
+++ b/build_web_compilers/lib/src/dev_compiler_bootstrap.dart
@@ -106,6 +106,8 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   }
 
   var bootstrapId = dartEntrypointId.changeExtension(ddcBootstrapExtension);
+  var bootstrapRequireId =
+      dartEntrypointId.changeExtension(ddcBootstrapRequireExtension);
   var bootstrapModuleName = _context.withoutExtension(_context.relative(
       bootstrapId.path,
       from: _context.dirname(dartEntrypointId.path)));
@@ -137,6 +139,9 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
             oldModuleScope: oldAppModuleScope));
 
   await buildStep.writeAsString(bootstrapId, bootstrapContent.toString());
+
+  var bootstrapRequireContent = 'require(["$bootstrapModuleName"]);';
+  await buildStep.writeAsString(bootstrapRequireId, bootstrapRequireContent);
 
   var entrypointJsContent = _entryPointJs(bootstrapModuleName);
   await buildStep.writeAsString(
@@ -258,17 +263,34 @@ String _entryPointJs(String bootstrapModuleName) => '''
   var mainUri = _currentDirectory + "$bootstrapModuleName";
 
   if (typeof document != 'undefined') {
+    // Load stack trace mapper
     var el = document.createElement("script");
     el.defer = true;
     el.async = false;
     el.src = mapperUri;
     document.head.appendChild(el);
 
+    // Load require.js
     el = document.createElement("script");
     el.defer = true;
     el.async = false;
     el.src = requireUri;
-    el.setAttribute("data-main", mainUri);
+    document.head.appendChild(el);
+
+    // Load the bootstrap script for this entrypoint, which defines the module
+    // and its dependencies with requirejs.
+    el = document.createElement("script");
+    el.defer = true;
+    el.async = false;
+    el.src = mainUri + '.js';
+    document.head.appendChild(el);
+
+    // Load the require script for this entrypoint, which triggers the execution
+    // of this entrypoint's main() via requirejs.
+    el = document.createElement("script");
+    el.defer = true;
+    el.async = false;
+    el.src = mainUri + '.require.js';
     document.head.appendChild(el);
   } else {
     importScripts(mapperUri, requireUri);
@@ -319,10 +341,12 @@ String _dartLoaderSetup(Map<String, String> modulePaths, String appDigests,
     '''
 $_currentDirectoryScript
 $_baseUrlScript
+let rootDirectory = window.location.origin + baseUrl;
 let modulePaths = ${const JsonEncoder.withIndent(" ").convert(modulePaths)};
+let appDigests = _currentDirectory + '$appDigests';
 if(!window.\$dartLoader) {
    window.\$dartLoader = {
-     appDigests: _currentDirectory + '$appDigests',
+     appDigests: new Array(),
      moduleIdToUrl: new Map(),
      urlToModuleId: new Map(),
      rootDirectories: new Array(),
@@ -355,8 +379,11 @@ if(!window.\$dartLoader) {
      getModuleLibraries: null, // set up by _initializeTools
    };
 }
+if (window.\$dartLoader.rootDirectories.indexOf(rootDirectory) === -1) {
+  window.\$dartLoader.rootDirectories.push(rootDirectory);
+}
+window.\$dartLoader.appDigests.push(appDigests);
 let customModulePaths = {};
-window.\$dartLoader.rootDirectories.push(window.location.origin + baseUrl);
 for (let moduleName of Object.getOwnPropertyNames(modulePaths)) {
   let modulePath = modulePaths[moduleName];
   if (modulePath != moduleName) {
--- a/build_web_compilers/build.yaml
+++ b/build_web_compilers/build.yaml
@@ -102,6 +102,7 @@ builders:
     build_extensions:
       .dart:
         - .dart.bootstrap.js
+        - .dart.bootstrap.require.js
         - .dart.js
         - .dart.js.map
         - .dart.js.tar.gz
--- a/build_web_compilers/lib/src/web_entrypoint_builder.dart
+++ b/build_web_compilers/lib/src/web_entrypoint_builder.dart
@@ -15,6 +15,7 @@ import 'dart2js_bootstrap.dart';
 import 'dev_compiler_bootstrap.dart';
 
 const ddcBootstrapExtension = '.dart.bootstrap.js';
+const ddcBootstrapRequireExtension = '.dart.bootstrap.require.js';
 const jsEntrypointExtension = '.dart.js';
 const jsEntrypointSourceMapExtension = '.dart.js.map';
 const jsEntrypointArchiveExtension = '.dart.js.tar.gz';
@@ -118,6 +119,7 @@ class WebEntrypointBuilder implements Builder {
   final buildExtensions = const {
     '.dart': [
       ddcBootstrapExtension,
+      ddcBootstrapRequireExtension,
       jsEntrypointExtension,
       jsEntrypointSourceMapExtension,
       jsEntrypointArchiveExtension,


jakemac53:
I will work on another PR to fix up the recent analyzer deprecations in parallel here to get the build green first.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -319,10 +341,12 @@ String _dartLoaderSetup(Map<String, String> modulePaths, String appDigests,
     '''
 $_currentDirectoryScript
 $_baseUrlScript
+let rootDirectory = window.location.origin + baseUrl;
 let modulePaths = ${const JsonEncoder.withIndent(" ").convert(modulePaths)};
+let appDigests = _currentDirectory + '$appDigests';
 if(!window.\$dartLoader) {
    window.\$dartLoader = {
-     appDigests: _currentDirectory + '$appDigests',
+     appDigests: new Array(),
```
evanweible-wf, line 349: This does change the type of `window.$dartLoader.appDigests` from a string to an array of strings, but I couldn't find anything in here or in webdev that is actually using this value, so I thought it was safe.

jakemac53: @annagrin do you know if anything is using this?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -258,17 +263,34 @@ String _entryPointJs(String bootstrapModuleName) => '''
   var mainUri = _currentDirectory + "$bootstrapModuleName";
 
   if (typeof document != 'undefined') {
+    // Load stack trace mapper
```
jakemac53, line 266: Both this script and the `require.js` script would ideally only get injected once (assuming that does work with `require.js`). Maybe assign an ID to these and query to see if they already exist before adding them?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: I think maybe this used to be used by hot reload but we dropped the support for that in webdev? Maybe we should stop adding this altogether.

jakemac53 (commented):
My main concern here is around the global setup that we are doing, and how these apps might step on each other. Some of that I think you were working on addressing in webdev though?

It looks to me like the `window.$dartLoader` object is probably ok, as we kind of build that up incrementally and don't bash over it if it already exists.

The more concerning bit is the call to `require.config`. I thought that was a global thing so I think we might be bashing over the previous app config with the new one. I have a feeling this probably presents some race conditions in app loading.

I think it is probably worth trying out using the `config` parameter https://requirejs.org/docs/api.html#multiversion. I have no experience with it but I think it should allow us to do what we want here.

Note also that today all apps on the page share state with each other, see the example I added. I think using the multiversion stuff with require.js might resolve that?

evanweible-wf:
> The more concerning bit is the call to require.config. I thought that was a global thing so I think we might be bashing over the previous app config with the new one. I have a feeling this probably presents some race conditions in app loading.

I had this concern, too, but in stepping through the require.js code it seemed to be safe enough. Each call configures three things:

```
require.config({
    baseUrl: baseUrl,
    waitSeconds: 0,
    paths: modulePaths
});
```

Afaict, the baseUrl should be the same for each entrypoint because it is specific to the document and not the scripts. The waitSeconds is hardcoded to 0. And the `paths` get merged together if `config()` is called more than once.

> I think it is probably worth trying out using the config parameter https://requirejs.org/docs/api.html#multiversion. I have no experience with it but I think it should allow us to do what we want here.

I saw that as well, but didn't spend much time playing with it. I'll see if I can get something working with that approach.

> Note also that today all apps on the page share state with each other, see the example I added. I think using the multiversion stuff with require.js might resolve that?

Interesting, didn't realize that! Probably not what we want :)

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
evanweible-wf: Good idea, I can do that 👍

jakemac53:
Actually I just pushed an update here to use a custom context argument - it does give each app its own global state now and appears to work as expected 👍 .

jakemac53:
@googlebot I consent

jakemac53:
> Interesting, didn't realize that! Probably not what we want :)

Ya, especially since in dart2js they wouldn't share state. I heard people like their development compiler semantics to match up with their release compiler semantics generally :D

jakemac53:
Looks like we need to merge master here to pick up the analyzer fixes, I will let you do that though to avoid too many merge conflicts between us haha.

evanweible-wf:
Looks like I have a couple tests to update (and maybe some to add, too). I've got meetings for the next couple of hours but will take a stab at it afterwards.

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
annagrin: I found this use in legacy strategy (for google3 users): https://github.com/dart-lang/webdev/blob/0274a13ffd50a77a4808c294d843ec1f7dcd06cd/dwds/lib/src/loaders/legacy.dart#L97

and that's where it is called: https://github.com/dart-lang/webdev/blob/a0b8e2b062fa8cfafc3c410259e934322cc4dde6/dwds/lib/src/handlers/injector.dart#L201

jakemac53:
One other note here, if this breaks `webdev serve` we should release it as a breaking change. And then `webdev` will need to update the build_web_compilers constraints it checks for (since its globally activated it does its own checks for the current version of build _web_compilers and complains if it isn't compatible).

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
evanweible-wf: Sorry, maybe I'm missing something - those are usages of `$dartLoader`, but not of this `appDigests` field, right? I think `$dartLoader.forceLoadModule()` would still behave the same

evanweible-wf:
If the changes to `$dartLoader` are breaking, I think we could also probably work around that by introducing new fields instead of changing the existing ones and then webdev could switch to using the new ones and update the minimum constraint on build_web_compilers.

jakemac53:
Ya, it would be ideal if we can release it as non-breaking. But breaking changes in this package aren't too painful (other packages don't really depend on it).

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -258,26 +265,54 @@ String _entryPointJs(String bootstrapModuleName) => '''
...
+      el = document.createElement("script");
+      el.id = mapperId;
+      el.defer = true;
+      el.async = false;
+      el.src = mapperUri;
+      document.head.appendChild(el);
+    }
+
+    // Load require.js if not already loaded
+    var requirejsId = 'dart-requirejs';
```
evanweible-wf, line 282: Not sure if these IDs are specific enough to avoid any potential collisions, thoughts?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -454,11 +494,15 @@ String _requireJsConfig(bool soundNullSafety) => '''
...
-});
+if (!window.\$dartRequire) {
+  window.\$dartRequire = new Map();
+}
+window.\$dartRequire.set("$contextId", require.config({
+  baseUrl: baseUrl,
+  context: "$contextId",
+  waitSeconds: 0,
+  paths: customModulePaths
+}));
```
evanweible-wf, line 505: Is this what you were thinking for namespacing these custom requirejs contexts?

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: I think this is fine 👍

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -265,19 +265,29 @@ String _entryPointJs(String bootstrapModuleName) => '''
...
-    var el = document.createElement("script");
-    el.defer = true;
-    el.async = false;
-    el.src = mapperUri;
-    document.head.appendChild(el);
+    var el;
+
+    // Load stack trace mapper if not already loaded
+    var mapperId = 'dart-stack-trace-mapper';
+    if (!document.querySelector('#' + mapperId)) {
```
jakemac53, line 272: nit: use `getElementById` for these

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -265,19 +265,29 @@ String _entryPointJs(String bootstrapModuleName) => '''
...
+      el = document.createElement("script");
```
jakemac53, line 273: might as well make `el` be a local variable to each of these blocks now instead of a shared one we overwrite

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: Yes 👍

evanweible-wf:
I'm not sure what's up with that test failure. Running the same command locally works for me, so maybe it's windows-specific?

jakemac53:
I can try locally on windows

jakemac53:
Unfortunately this also passes for me locally on windows. Not sure what could be causing it to fail on github?

jakemac53:
Actually looking at the error it is failing to load this file: `http://localhost:50021/Vma2b75L2B5pZ90MHWi%2F3N6QvT8Wk16f/test/subdir_test.dart.browser_test.dart.bootstrap.js`.

That looks to be missing the `sub-dir/` directory so it is trying to load the wrong path.

jakemac53:
Ok I figured this out - the bootstrap module isn't actually defined yet by the time this require.js script is running, in github - on windows only. Basically its a race condition in the loading of the scripts. And I guess github windows has a crappy internet connection or loads scripts reliably differently haha.

You can reproduce the error by throttling your network to "Slow 3g" in chrome devtools. That causes the require bootstrap to load first.

Maybe if the bootstrap.js script loaded the require bootstrap script (after it has defined that module), instead of the .dart.js script injecting them both, then it would work out?

If the regular bootstrap script isn't yet known by the time the require bootstrap runs, then it tries to download it by name and fails since the name doesn't match the path.

jakemac53:
This does violate my understanding of how `defer` scripts work... I thought they still ran in source order but that doesn't seem to be the case. I am suprised we don't see similar issues with requirejs itself or the stack trace mapper not being loaded yet, as those are larger scripts.

jakemac53:
We could also possibly just remove the `defer` attribute from all but the require bootstrap script?

jakemac53:
Also now I can't get it to reproduce lol... but it did reproduce once for me. 🤕

jakemac53:
Ok I pushed another change here that removes the extra bootstrap file, and instead inlines it in the original bootstrap file, after defining the bootstrap module. So it basically just immediately runs it after defining it.

jakemac53:
@evanweible-wf any concerns around the new approach? It seems to make windows happy.

evanweible-wf:
Nope, this approach looks great to me! I also started seeing this issue this morning until I pulled your latest commits, which fixed it. Although like you, I'm a bit confused as to why this was an issue.. the MDN docs say:

> Scripts with the defer attribute will execute in the order in which they appear in the document.

¯\_(ツ)_/¯  regardless, this approach is simpler and seems to work 👍

jakemac53:
Ok, do you mind pushing a cleanup commit to remove the changes to the build extensions then? And then I can take another pass here :).

jakemac53:
Ok, this generally LGTM at this point to me. I think we need to finalize the release strategy still (is this actually breaking for external webdev users or not?).

It would also be nice to get a multi-app test going, but it might be difficult to do. The testing package isn't really set up for that, but we could probably still manage something. I will see if I can come up with something that works here.

jakemac53:
@evanweible-wf should we consider making this feature opt-in behind some builder configuration? That would make it non-breaking, and would also allow us to land it without proper webdev support in place yet.

evanweible-wf:
Yeah that's a good idea! On the webdev side, would we read that same builder config to check for opt-in or just inspect the window/build outputs to determine whether it's being used?

Also, sorry I stepped away from this for a bit - we've been heads down getting to 2.13, but that's almost done now and I'll be able to dedicate more time to the webdev side of things.

jakemac53:
> Yeah that's a good idea! On the webdev side, would we read that same builder config to check for opt-in or just inspect the window/build outputs to determine whether it's being used?

I don't think webdev could reasonably check the actual config - it could either infer it by looking at the outputs or we could expose something specific on the window that tells it which mode to be in?

kevmoo:
Thanks for the update @evanweible-wf – looks like @annagrin may run w/ this a bit

Monassar96:
**image**image

PR #3194, opened 2021-09-29
Mark mono_repo outputs as generated

natebosch:
Add a mono_repo outputs to the .gitattributes as generated which allows
GitHub diff views to leave them collapsed by default.


kevmoo (approved):
Sure!

kevmoo:
> Should we consider making mono_repo add these automatically if they don't exist?

PRs accepted! 😄

PR #3197, opened 2021-09-30
Drop package:pedantic for package:lints 

jakemac53:
Also enables the recommended lints and fixes them.

Had to add a lot of ignores for private imports especially - although I did just fix up some that were just...bad.


build/mono_pkg.yaml
```diff
@@ -1,5 +1,5 @@
 dart:
-- 2.12.0
+- 2.14.0
```
kevmoo, line 2: If we bump this, we should (ideally) bump the min SDK in the package.

build_config/lib/src/builder_definition.dart
```diff
@@ -55,7 +55,7 @@ class BuilderDefinition {
   /// The name of the dart_library target that contains `import`.
   ///
   /// May be null or unreliable and should not be used.
-  @deprecated
+  @Deprecated('Do not use')
```
kevmoo, line 58: Uh...could we be more clear why? 😄

kevmoo (approved):

build_config/lib/src/builder_definition.dart
natebosch: It's not very user facing and the doc comment says it's unreliable.  I don't think we need to be super specific with these.

natebosch (approved):

PR #3199, opened 2021-10-01
Cleanup tasks

jakemac53:
- Move to `modifyFile` from `updateFile` analyzer api.
- Remove an unnecessary dependency_override in build_runner.

### DIFF
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -125,7 +125,7 @@ class BuildAssetUriResolver extends UriResolver {
         return _AssetState(path, _cachedAssetDependencies[id]!);
       }
       if (isChange) {
-        resourceProvider.updateFile(path, content);
+        resourceProvider.modifyFile(path, content);
       } else {
         resourceProvider.newFile(path, content);
       }
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -125,7 +125,7 @@ class BuildAssetUriResolver extends UriResolver {
         return _AssetState(path, _cachedAssetDependencies[id]!);
       }
       if (isChange) {
-        resourceProvider.modifyFile(path, content);
+        resourceProvider.updateFile(path, content);
       } else {
         resourceProvider.newFile(path, content);
       }
--- a/build_resolvers/test/resolver_test.dart
+++ b/build_resolvers/test/resolver_test.dart
@@ -10,6 +10,7 @@ import 'package:analyzer/dart/analysis/results.dart';
 import 'package:analyzer/dart/analysis/session.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/file_system/memory_file_system.dart';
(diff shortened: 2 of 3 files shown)


jakemac53:
Note that I reverted the analyzer change after discussion with the analyzer folks - it looks like there was a mistake and `updateFile` is the one we should continue to use, and `modifyFile` will be deprecated.

natebosch (approved):

natebosch (approved):

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -125,7 +125,7 @@ class BuildAssetUriResolver extends UriResolver {
         return _AssetState(path, _cachedAssetDependencies[id]!);
       }
       if (isChange) {
-        resourceProvider.updateFile(path, content);
+        resourceProvider.modifyFile(path, content);
```
kevmoo, line 128: sure the pkg:analyzer version we have has this?

jakemac53: yes

jakemac53:
cc @scheglov it looks like the new API doesn't throw the `InconsistentAnalysisException` as expected, is that intentional? See https://github.com/dart-lang/build/pull/3199/checks?check_run_id=3792799647#step:6:98.

scheglov:
Yes, it is intentional. We now try to keep libraries that are not affected by the change to a file. So, we never discard the whole session. But we mark individual libraries are invalid.

See https://dart-review.googlesource.com/c/sdk/+/214502

However it seems that I did not change places where we used to check for session to checking this new flag. I will need to make these updates.

jakemac53:
Ok so I think this test would need some additional fixes then - it will need to modify the file and re-invalidate it.

I tried adding this line to replace where we used to do a `isLibrary` check on another library (that used to invalidate the session):

```dart
        (x.library!.session.resourceProvider as MemoryResourceProvider)
            .modifyFile('/a/web/main.dart', 'int y;');
```

That doesn't appear to work today but it sounds like it might be related to the other issue you mentioned?

jakemac53:
Just skipped those tests for now

scheglov:
Yes, methods that accept `LibraryElement` should be updated to check that the element is still valid, and return an `InvalidResult`, in my work-in-progress CL it would be:
```dart
/// The type of [InvalidResult] returned when the given element was invalidated
/// since it was created and returned to the client, because of a change to
/// a file that affected it.
///
/// Clients may not extend, implement or mix-in this class.
class InvalidatedElementResult
    implements
        InvalidResult,
        SomeParsedLibraryResult,
        SomeResolvedLibraryResult {}
```

scheglov:
https://dart-review.googlesource.com/c/sdk/+/215482

natebosch:
@scheglov - is the analyzer team still considering a change in this design?

scheglov:
No, we don't.
https://dart-review.googlesource.com/c/sdk/+/230040 was implemented instead.

scheglov:
Both should throw, as you can see from https://dart-review.googlesource.com/c/sdk/+/230040/3/pkg/analyzer/test/src/dart/analysis/session_test.dart#300

But be aware that `modifyFile` only changes the file system, and `AnalysisDriver` does not watch the file system, it is up to the higher level clients to do and call `AnalysisDriver.changeFile()`.

Issue #3196, opened 2021-09-29
Handle asynchronous errors in `serveRequests` calls


annagrin:
`serveRequests` does not run its code in a `runZonedGuarded` if there is another zone in the call chain, which changes the logic of the calling code. Also, would help diagnose problems if the call was wrapped in `Chain.capture` to provide all async traces on exception:

See discussion in https://github.com/dart-lang/shelf/issues/202
See similar issue fixes in https://github.com/dart-lang/webdev/pull/1417

PR #3203, opened 2021-10-06
Make build daemon serve requests in an error zone

annagrin:
We have a large number of DDR server crashes due to `serveRequests` not running in a proper error zone.

See https://github.com/dart-lang/shelf/issues/202 for discussion.

**Related issues**

Similar issue https://github.com/dart-lang/webdev/issues/1418
Similar issue flutter/devtools#3415

Closes:  dart-lang/build#3196


annagrin:
@jakemac53 @natebosch This is needed to unblock DDR before the proper fix is developed in https://github.com/dart-lang/shelf/issues/202

build_daemon/lib/src/server.dart
```diff
@@ -90,7 +92,11 @@ class Server {
     });
 
     var server = _server = await HttpMultiServer.loopback(0);
-    serveRequests(server, handler);
+    // Serve requests in an error zone to prevent failures
+    // when running from another error zone.
+    runZonedGuarded(() => serveRequests(server, handler), (e, _) {
+      _logger.warning('Error serving requests: $e');
```
kevmoo, line 98: doesn't the logger have an exception and stack argument?

jakemac53: Yes, I would pass those assuming we properly log them for warnings but I think we should (or we should update the logging code if not).

build_daemon/lib/src/server.dart
```diff
@@ -29,6 +30,7 @@ import 'managers/build_target_manager.dart';
 /// Handles notifying clients of logs and results for registered build targets.
 /// Note the server will only notify clients of pertinent events.
 class Server {
+  final _logger = Logger('BuildDaemon Server');
```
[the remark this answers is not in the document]
annagrin: will change, thanks!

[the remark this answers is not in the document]
annagrin: done.

build_daemon/lib/src/server.dart
annagrin: done.

build_daemon/lib/src/server.dart
```diff
@@ -29,6 +30,7 @@ import 'managers/build_target_manager.dart';
 /// Handles notifying clients of logs and results for registered build targets.
 /// Note the server will only notify clients of pertinent events.
 class Server {
+  static final _logger = Logger('BuildDaemon Server');
```
grouma, line 33: I don't think we want to use a `Logger` here but instead the `logHandler` that is part of this package. The reason being that the daemon process doesn't log to a user visible output while the `logHandler` pipes logs to the clients for display.

PR #3207, opened 2021-10-14
Allow resource value reuse when disposed

natebosch:
The intent with resources was to be able to do things like cache
expensive computations across builds, with the `dispose` hook helping to
keep things as hermetic as possible. Throwing away the entire instance
defeats the purpose.

Only discard the instance if there was no dispose callback
available.

Change the handling of disposing resources that haven't been used to an
`assert`. The prior fallback behavior was never hit, and we don't want to
 silently create resources that aren't needed solely to dispose them if we
 introduce a bug that would cause it.

Expand the class level docs to make the usage and intent more clear.

### DIFF
--- a/build/lib/src/resource/resource.dart
+++ b/build/lib/src/resource/resource.dart
@@ -112,8 +112,8 @@ class Resource<T> {
   /// If a [_userDispose] was provided, invoke it and assume the state can be
   /// retained for the next build.
   Future<void> _dispose(ResourceManager manager) {
-    if (!_instanceByManager.containsKey(manager)) return Future.value(null);
-    var oldInstance = _fetch(manager);
+    assert(_instanceByManager.containsKey(manager));
+    var oldInstance = _instanceByManager[manager]!;
     if (_userDispose != null) {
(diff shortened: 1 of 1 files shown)


natebosch:
> We should probably create some examples too?

Yeah I was trying to think of how complete or realistic I wanted to be. What do you think of putting a skeleton like this:


```dart
final someResource = Resource<SomeResource>(() => SomeResource._(),
    dispose: (something) => something._dispose(),
    beforeExit: (something) => something._beforeExit());

class SomeResource {
  SomeResource._();

  Future<String> somethingUsefulForBuilders(AssetReader assetReader) async {
    // Any information returned to the caller should be derived from the contents
    // read through `assetReader`. Calling `assetReader.digest` is sufficient for
    // the build system to track dependencies.
  }

  void _dispose() {
    // Clear or invalidate any cached state that was read during the build.
  }

  void _beforeExit() {
    // Shutdown or clean up any externally held resources.
  }
}
```

@kevmoo - would this have helped?

kevmoo:
> @kevmoo - would this have helped?

Absolutely!

PR #3214, opened 2021-11-07
Use raw string literals in more places when generating the build script

simolus3:
When generating the build script, some string literals like the builder's name are generated as raw strings already. However, strings that might be part of a builder's options are not escaped.

For example, a build script generated for the following builder options would fail to run:

```yaml
options:
  args: "-shared -o $output $input"
```

With this PR, we use raw string literals in more places. I have updated the golden file as a test, let me know if we need more tests for this.

### DIFF
--- a/_test/test/goldens/generated_build_script.dart
+++ b/_test/test/goldens/generated_build_script.dart
@@ -109,14 +109,14 @@ final _builders = <_i1.BuilderApplication>[
         r'test/**.node_test.dart',
         r'test/**.vm_test.dart'
       ]),
-      defaultOptions: _i8.BuilderOptions(const {
+      defaultOptions: const _i8.BuilderOptions({
         r'dart2js_args': [r'--minify']
       }),
-      defaultDevOptions: _i8.BuilderOptions(const {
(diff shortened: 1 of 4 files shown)


_test/test/goldens/generated_build_script.dart
```diff
@@ -109,13 +109,14 @@ final _builders = <_i1.BuilderApplication>[
         r'test/**.node_test.dart',
         r'test/**.vm_test.dart'
       ]),
-      defaultOptions: _i8.BuilderOptions({
-        'dart2js_args': ['--minify']
+      defaultOptions: _i8.BuilderOptions(const {
```
jakemac53, line 112: Note that `BuilderOptions` has a const constructor - I would probably try making the builder options themselves be const if we want to move towards more const things here.

jakemac53 (approved):
I am mildly concerned we will run into some issues with poorly behaved builders that were trying to modify the options map here, but I am willing to give this a go. They really shouldn't be doing that anyways so it would be good to catch cases where they are.

PR #3216, opened 2021-11-17
Implement UriResolver.pathToUri()

scheglov:
Prepare for https://dart-review.googlesource.com/c/sdk/+/220480

### DIFF
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -196,11 +196,15 @@ class BuildAssetUriResolver extends UriResolver {
     var pathSegments = resourceProvider.pathContext.split(path);
     var packageName = pathSegments[1];
     if (pathSegments[2] == 'lib') {
-      var packagePath = p.url.joinAll(pathSegments.skip(3));
-      return Uri.parse('package:$packageName/$packagePath');
+      return Uri(
+        scheme: 'package',
+        pathSegments: [packageName].followedBy(pathSegments.skip(3)),
+      );
     } else {
-      var packagePath = p.url.joinAll(pathSegments.skip(2));
-      return Uri.parse('asset:$packageName/$packagePath');
+      return Uri(
+        scheme: 'asset',
+        pathSegments: [packageName].followedBy(pathSegments.skip(2)),
+      );
     }
   }
 
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -191,8 +191,23 @@ class BuildAssetUriResolver extends UriResolver {
   }
 
   @override
-  Uri? restoreAbsolute(Source source) =>
-      lookupCachedAsset(source.uri)?.uri ?? source.uri;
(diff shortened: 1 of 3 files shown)


build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -191,8 +191,27 @@ class BuildAssetUriResolver extends UriResolver {
   }
 
   @override
-  Uri? restoreAbsolute(Source source) =>
-      lookupCachedAsset(source.uri)?.uri ?? source.uri;
+  // ignore: override_on_non_overriding_member
```
jakemac53, line 194: nit: add a comment here about why its necessary?

jakemac53 (commented):
I assume this is a new method that we will need to implement, but note that this could be breaking for some users even if we have an implementation in build_resolvers pre-released, because they could potentially get a newer analyzer but older build_resolvers for some reason.

Usually the reason for being stuck on an older build_resolvers would be being stuck on an old analyzer for some reason though, so it shouldn't be super common. But some people also likely have build_resolvers pinned to an old version, and would be broken (unless they also pin analyzer).

scheglov:
I will update analyzer to make it backward compatible.
https://dart-review.googlesource.com/c/sdk/+/220480/5/pkg/analyzer/lib/src/generated/source.dart#426

jakemac53:
> I will update analyzer to make it backward compatible.

Ah I thought we just implemented this api but I see we extend it so LGTM

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -191,8 +191,27 @@ class BuildAssetUriResolver extends UriResolver {
...
+  Uri pathToUri(String path) {
+    var pathSegments = resourceProvider.pathContext.split(path);
```
jakemac53, line 196: It looks like we always use the `posix` context to create these paths. We should either use the resource providers context there as well or just use posix directly here probably?

scheglov: There is some contradiction here. The `ResourceProvider` and its `pathContext` should correspond to the style of path(s). If we always build Posix path(s), then `pathContext` must be Posix. Otherwise I don't understand why it would work.

jakemac53: It seems reasonable to me to update to use the `pathContext` when constructing the paths, do you mind updating that here as well?

scheglov: Hm... We also use `posix` in `assetPath`, and we don't have access to `ResourceProvider`. In this case, it is better to use `posix` directly here as well.

jakemac53: sgtm

jakemac53 (approved):

jakemac53:
anything else on your end? I pushed up a change to prep for release as well.

scheglov:
Nothing that I can think of.

jakemac53:
ok ill merge this once its happy and then merge internally, and publish when that lands

natebosch (approved):

PR #3220, opened 2021-12-02
Describe build schema setup for common tools

### LINKED ISSUE #3217: `build.yaml` schema
It would be nice to provide a publicly available schema for `build.yaml` as currently VS Code extension assumes it's a completely different schema (when using schemastore.org) which leads to analyzer reporting errors incorrectly.

> Dart SDK Version (`dart --version`)
```
Dart SDK version: 2.15.0-116.0.dev (dev) (Thu Sep 16 09:47:01 2021 -0700) on "macos_x64"
```
> What package(s) from this repo you are using, and the version (i.e. `build_runner 0.7.12`)

`build_runner: ^2.1.4`, however, it's more concerned to `build_config` I think.

> What builder(s) you are using (or writing yourself). Try to give a short summary of what they do.

Not applicable.

> Any other packages or constraints we should know about

No.

simolus3:
Closes #3217.

Maybe we want to encourage users to enable that for `build.yaml` files overriding packages too? I think all patterns would be `build.yaml`, `*.build.yaml`, `build.*.yaml` and `*.build.*.yaml` then, right? I'm not sure if the last one is valid (overriding a package only under a specific `--config` flag?).


jakemac53:
> I'm not sure if the last one is valid (overriding a package only under a specific `--config` flag?).

Haha, I don't know either 😆

jakemac53:
The linked image doesn't appear to work, but otherwise it looks good to me. Suggesting the other patterns also seems like a good idea (but maybe not the last one, haha).

jakemac53 (approved):

simolus3:
> The linked image doesn't appear to work, but otherwise it looks good to me.

I think that may be GitHub resolving links in previews weirdly? It appears to work when viewed [on the branch](https://github.com/simolus3/build/blob/build-yaml-verification-guide/docs/faq.md#intellij-based-editors-eg-idea-android-studio-webstorm).

But I'll add the two other file patterns too.

jakemac53:
Thanks! I just used these instructions to set it up for myself and it was a breeze and its quite nice 👍

kevmoo:
THIS IS AWESOME! Thank you, @simolus3 !!

PR #3222, opened 2021-12-16
update to the latest analyzer across all packages

jakemac53:
If we are green I will just remove the overrides and publish with the expanded range, won't need to wait on dependencies to update first.

### DIFF
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -23,7 +23,6 @@ dev_dependencies:
   test_process: ^2.0.0
 
 dependency_overrides:
-  analyzer: ^3.0.0
   build:
     path: ../build
   build_config:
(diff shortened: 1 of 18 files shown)


kevmoo (approved):

natebosch (approved):

AlexV525:
Why should a major version constraint be changed in a minor version publishment?

kevmoo:
@AlexV525 – if there were no changes to the APIs that were used by the package, it's fine.

The exception is if APIs are exported from one package to the other....

AlexV525:
> The exception is if APIs are exported from one package to the other....

@kevmoo Then it should be easy to figure out that packages are very breakable from one to the other.

jakemac53:
Expanding to allow a new major version of a package should not be a breaking change if your package is not affected directly, and your package does not export the package whose range was expanded. That criteria is met here.

@AlexV525 were you broken by this? Can you elaborate on the breakage if so?

jakemac53:
It looks like the breakage (I see it now in https://github.com/google/json_serializable.dart/issues/1083) was not actually related to this PR.

It was a result of a breakage in analyzer `3.1.0`.

PR #3241, opened 2022-01-31
Use ??= over contains and `!` for map defaulting

natebosch:
Avoid a non-null assertion by chaining access from a null conditional
assignment instead of separately inserting to the map, then reading it
back out for further operations.

Also refactor an iterable `.toList()` to a List literal.

### DIFF
--- a/build_modules/lib/src/meta_module.dart
+++ b/build_modules/lib/src/meta_module.dart
@@ -238,14 +238,13 @@ MetaModule _coarseModulesForLibraries(
   var librariesByDirectory = <String, Map<AssetId, ModuleLibrary>>{};
   for (var library in libraries) {
     final dir = _topLevelDir(library.id.path);
-    if (!librariesByDirectory.containsKey(dir)) {
-      librariesByDirectory[dir] = <AssetId, ModuleLibrary>{};
-    }
-    librariesByDirectory[dir]![library.id] = library;
+    (librariesByDirectory[dir] ??= {})[library.id] = library;
   }
-  final modules = librariesByDirectory.values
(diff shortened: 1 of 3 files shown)


natebosch:
I noticed some stuff I could code-golf when I was trying to understand behavior, and after looking at it I think this is actually nicer despite being a bit dense.

build_modules/lib/src/meta_module.dart
```diff
@@ -238,14 +238,13 @@ MetaModule _coarseModulesForLibraries(
   var librariesByDirectory = <String, Map<AssetId, ModuleLibrary>>{};
   for (var library in libraries) {
     final dir = _topLevelDir(library.id.path);
-    if (!librariesByDirectory.containsKey(dir)) {
-      librariesByDirectory[dir] = <AssetId, ModuleLibrary>{};
-    }
-    librariesByDirectory[dir]![library.id] = library;
+    (librariesByDirectory[dir] ??= {})[library.id] = library;
```
jakemac53, line 241: we could use `putIfAbsent` also, but 🤷‍♂️

jakemac53 (approved):

build_modules/lib/src/meta_module.dart
natebosch: Yeah, I had thought about that pattern too but for some reason `??=` looked nicer to me.

Maybe it was something like dart2js emitting better code for `??=` than `putIfAbsent`? (even though this will not be compiled with dart2js)

jakemac53: I would usually use `putIfAbsent`, it's true that this doesn't require a closure though. I am fine with either way.

PR #3249, opened 2022-02-28
Update tests to use the 2.12 null safety version

jakemac53:
Looks like the experimental release version is removed now, so you need to be on 2.12


kevmoo (approved):
:shipit:

PR #3253, opened 2022-03-05
Await and skip a missed broken expectation

natebosch:
These expectations can fail as an unhandled async error and the
exception lost because of the error handling zone running the `Builder`.
The test should have been failing, add a `skip` for now instead of
silently missing the failure.


jakemac53 (approved):

jakemac53:
Do we need to file an issue here?

PR #3254, opened 2022-03-08
Require analyzer 2.8.0, remove UriResolver.restoreAbsolute() implementation.

scheglov:
See https://dart-review.googlesource.com/c/sdk/+/235765


jakemac53:
Merged on red - it looks like we have some existing failures that cropped up.

PR #3255, opened 2022-03-08
fix a failing build_daemon test

jakemac53:
Not sure why this was expecting unsound output files previously, but it should be outputting sound files.


jakemac53:
> how was this passing before? Both files exist?

Honestly no idea how it was passing before.

jakemac53:
Merged on red, can't get a passing build right now due to sdk and chrome flakes.

PR #3267, opened 2022-03-18
update to package:lints 2.0

jakemac53:
Also removed redundant lints from the analysis_options.yaml.

I had to disable two lints at least for now, because built_value generate code violates them. I did fix up the other violations though.


jakemac53:
Hmm, so i guess lints 2.0 requires the dev sdk. Do we want to hold off on this? Or I could make it be a range?

jakemac53:
I will try just making it be a range and see if everything passes. It might put us in a bit of a weird spot but as long as 2.0 only adds new lints, it should be OK.

natebosch (approved):

PR #3273, opened 2022-03-25
Report changed assets through build daemon

### LINKED ISSUE #3268: List of changed assets from daemon
I'm trying to combine a build daemon with the `frontend_server_client` package to incrementally compile Dart apps and to reload them after a build.

I want the front end to see assets in `.dart_tool/build/generated` too, so I'm just translating package uris to the asset server from the daemon and using those as HTTP entrypoints. This works well, but now I need to explicitly tell the frontend about changed URIs after a rebuild. It looks like there's no easy way to get informed about which assets have changed during a build, could this information be exposed in `BuildResult`?

simolus3:
Report a list of changed assets (outputs + changed inputs from watcher) through the build daemon. Because this list can be rather large, the changes are only reported to clients actually interested in them.

Closes #3268.

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -111,10 +111,18 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
               '${target.target}/**', _buildOptions.packageGraph.root.name));
       }
     }
+    Iterable<AssetId>? outputs;
+
     try {
       var mergedChanges = collectChanges([changes]);
       var result = await _builder.run(mergedChanges,
           buildDirs: buildDirs, buildFilters: buildFilters);
+
+      outputs = {
+        for (var change in changes) change.id,
+        ...result.outputs,
+      };
+
       for (var target in targets) {
         if (result.status == core.BuildStatus.success) {
           // TODO(grouma) - Can we notify if a target was cached?
@@ -140,7 +148,7 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
       }
       _logMessage(Level.SEVERE, 'Build Failed:\n${e.toString()}');
     }
-    _signalEnd(results);
+    _signalEnd(results, outputs?.map((e) => e.uri));
   }
 
   @override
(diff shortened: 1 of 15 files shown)


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -111,10 +111,18 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
+
     try {
       var mergedChanges = collectChanges([changes]);
       var result = await _builder.run(mergedChanges,
           buildDirs: buildDirs, buildFilters: buildFilters);
+
+      outputs = {
+        for (var change in changes) change.id,
+        ...result.outputs,
+      };
```
jakemac53, line 124: Ideally we wouldn't create this set either if we don't need it. Probably not a huge deal, but if it isn't too ugly to omit it in that case I would.

jakemac53 (commented):
Just the one comment, otherwise this looks good to me.

We will also need to look internally and see if this affects things at all, although we can probably just not support the feature there.

jakemac53:
cc @grouma can you comment on the internal stuff before I try to do a roll?

grouma:
I don't foresee any issues. In fact I vaguely recall that we wanted to add this kind of support a while back but it was never prioritized.

build_runner/lib/src/daemon/daemon_builder.dart
simolus3: My idea was to do this in `build_daemon` and to strip the value out in the server so that different implementations don't have to all do the same check. I thought that combining two collections we already have (both `changes` and `result.outputs` are available either way) into a single set was cheap enough to not worry about it, but I can change it if we want to avoid doing it unconditionally. As far as I'm concerned there's only a single build implementation either way :smile:

jakemac53: It just depends how easy it is to avoid - it definitely takes some time to allocate a new set and then its going to end up growing it a bunch of times etc as it adds items. So probably a measurable amount of time for larger builds, although dwarfed by overall build time.

If it would make the code really ugly though I would be ok with it.

simolus3: Good point, and it turned out to be relatively easy to avoid (55ff4a3b98b07e2086b40c57fe8d6f2bd7acad22).

jakemac53:
I will get an internal merge going here and then we can publish once that is successful

Issue #3281, opened 2022-04-11
Large data model for built_value causes large memory consumption and codegen runtime


michael-golfi:
Hello, @Markzipan is a friend of mine and encouraged me to post this issue I'm seeing with build_runner and built_value.

I'm running the following tooling versions:

* Dart SDK version: 2.16.2 (stable) (Tue Mar 22 13:15:13 2022 +0100) on "macos_x64"
* build_runner: 2.1.8
* built_value: 8.1.4

### Problem

Our data model is generated from a GraphQL schema and operations which describe the data being pulled from the API. The data model has quite a few references between built values within the output serializable types. These are the following packages we're using for covering that portion of the builder process.

* Other packages with versions for builders:
* gql_build: 0.4.1
* gql_code_builder: 0.5.1
* ferry_generator: 0.5.0-dev.10

Though this doesn't seem relevant to the issue, I am building this for Flutter on iOS and Android.

### My investigation

I started investigating our issues with codegen once I noticed that codegen started going up to roughly 22 mins on average and previously also went higher (eg: 1 hour). At the same time I'm seeing memory consumption of 22 Gb of RAM on a Macbook Pro (it only has 16 Gb of physical RAM). It seems like i/o also goes high and some of the files take several minutes to generate output from `built_value`. I started seeing messages such as the following:

```text
[WARNING] No actions completed for 15.0s, waiting on:
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/login.queries.data.gql.dart
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/login.queries.var.gql.dart
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/login.queries.req.gql.dart
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/delete_event.mutations.ast.gql.dart
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/delete_event.mutations.data.gql.dart
  .. and 11 more

...

[WARNING] No actions completed for 7m 54s, waiting on:
  - built_value_generator:built_value on lib/src/graphql/resolvers/__generated__/event_tracking.mutations.data.gql.dart
  .. and 11 more
```

I ran `--track-performance` to see the timing graph and noticed high latency for a few generation tasks (the same ones which are mentioned in the `[WARNING]` message above. More and more it started like there's some form of blocking / resource contention occurring due to waiting on certain operations to finish for others to proceed.

[...]

PR #3282, opened 2022-04-13
add performance tracking to all methods on PerActionResolver

jakemac53:
This helps when trying to figure out where the time is going inside a builder (spawned from investigation into https://github.com/dart-lang/build/issues/3281)

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,6 +1,7 @@
-## 2.0.7-dev
+## 2.0.7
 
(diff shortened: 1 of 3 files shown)


natebosch (approved):
Should we be worried about this making performance logs much longer?

jakemac53:
> Should we be worried about this making performance logs much longer?

I don't think so, in general this is all only enabled if/when people pass `--track-performance` anyways. The web viewer did a good job of handling this for the case I was looking at, usually these calls are sequential so they end up in one row on the timeline all together.

PR #3283, opened 2022-04-14
update analyzer deps, prepare to publish packages with public deps

jakemac53:


jakemac53:
Will remove the dependency overrides on analyzer once package:test and any other deps that need to be upgraded are published.

natebosch (approved):

jakemac53:
Note that the failures all looked related to some changes that were required in `package:source_gen`, that have already been fixed and are just not published.

Issue #3286, opened 2022-04-15
`^` syntax doesn't work without...something?


kevmoo:
Some weird behavior here.

Thinking I should be able to use `^pubspec.yaml` as a key and only hit a root `pubspec.yaml` file – but that doesn't work!

See https://github.com/kevmoo/build_version/pull/26

CC @jakemac53

PR #3287, opened 2022-04-15
support ^ in regular build extensions for full path matching

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3286

### DIFF
--- a/build/lib/src/generate/expected_outputs.dart
+++ b/build/lib/src/generate/expected_outputs.dart
@@ -21,6 +21,9 @@ Iterable<AssetId> expectedOutputs(Builder builder, AssetId input) sync* {
   }
 }
 
+// Regexp for capture groups.
+final RegExp _captureGroupRegexp = RegExp('{{(\\w*)}}');
+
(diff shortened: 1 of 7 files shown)


build/lib/src/generate/expected_outputs.dart
```diff
@@ -52,31 +55,97 @@ extension on AssetId {
...
+    if (!hasAnyOutputFor(input)) return const Iterable.empty();
+
+    // If we expect an output, the asset's path ends with the input extension.
+    // Expected outputs just replace the matched suffix in the path.
+    return outputExtensions
+        .map((extension) => AssetId(input.package, extension));
+  }
+}
+
+/// A simple build input/output set which matches file suffixes and doesn't use
```
jakemac53, line 112: This was not actually edited... the diff just got weird because of how I moved some of the other code around 🤷‍♂️

natebosch (approved):
I'm a little worried about furthering the appearance that we support general regex, but neither _actually_ supporting general regex, nor trying to find a non-regex-appearing way to express this are very appealing to me... This approach LGTM

PR #3291, opened 2022-04-28
Remove redundant import

natebosch:
Also imported by relative path below.


kevmoo (approved):
I thought we ditched all of these!

natebosch:
> I thought we ditched all of these!

I found this one accidentally... I wonder if we have a bug in in the analyzer rule which caused it to miss this.

jakemac53 (approved):

PR #3308, opened 2022-05-23
Check that the result of getErrors() is ErrorsResult.

scheglov:
It [will be](https://dart-review.googlesource.com/c/sdk/+/245680) `PartWithoutLibraryResult` if the file at which `part` directive points does not have the corresponding `part of`.


natebosch (approved):

scheglov:
Thank you :-)

PR #3311, opened 2022-05-25
Add include_defaults for sources in build.yaml to include all default sources

rotorgames:
Every time when we would like to include custom sources to `build.yaml` file, we also have to add all items from [defaultNonRootVisibleAssets](https://github.com/dart-lang/build/blob/995fc94d34c8a6c144002cf9e04a485c4348b878/build_runner_core/lib/src/generate/options.dart#L24) or [defaultRootPackageSources](https://github.com/dart-lang/build/blob/995fc94d34c8a6c144002cf9e04a485c4348b878/build_runner_core/lib/src/generate/options.dart#L37) manually. 

It's ok if we develop our own project but if we develop a package for public usage, we need to ask everyone who uses our package to create `build.yaml` file and define both their sources and all these items in the `sources`.

For me, it's not good approach because the list of these items can be extended in over time.

So, I suggest to add the `include_defaults` parameter to `sources` which includes all sources from [defaultNonRootVisibleAssets](https://github.com/dart-lang/build/blob/995fc94d34c8a6c144002cf9e04a485c4348b878/build_runner_core/lib/src/generate/options.dart#L24) or [defaultRootPackageSources](https://github.com/dart-lang/build/blob/995fc94d34c8a6c144002cf9e04a485c4348b878/build_runner_core/lib/src/generate/options.dart#L37) by default.

As result, we can keep our `build.yaml` clean and event if default sources will be changed in future we will be able easily update them in `build_runner_core` and provide them for all developer by new publishing.

Example:
```yaml
targets:
  $default:
    sources:
      include_defaults: true
      include:
        - my_custom_sources/**
```

### DIFF
--- a/build_config/lib/src/input_set.dart
+++ b/build_config/lib/src/input_set.dart
@@ -14,6 +14,8 @@ part 'input_set.g.dart';
 class InputSet {
   static const anything = InputSet();
 
+  final bool includeDefaults;
+
   /// The globs to include in the set.
   ///
   /// May be null or empty which means every possible path (like `'**'`).
@@ -24,7 +26,7 @@ class InputSet {
   /// May be null or empty which means every path in [include].
   final List<String>? exclude;
 
-  const InputSet({this.include, this.exclude});
+  const InputSet({this.include, this.exclude, this.includeDefaults = false});
 
   factory InputSet.fromJson(dynamic json) {
     if (json is List) {
@@ -38,7 +40,8 @@ class InputSet {
           'Expected a Map or a List but got a ${json.runtimeType}');
     }
     final parsed = _$InputSetFromJson(json as Map);
-    if (parsed.include != null && parsed.include!.any((s) => s.isEmpty)) {
+    if (parsed.include != null && parsed.include!.any((s) => s.isEmpty) ||
+        parsed.includeDefaults && parsed.include == null) {
       throw ArgumentError.value(
           parsed.include, 'include', 'Include globs must not be empty');
     }
@@ -56,6 +59,9 @@ class InputSet {
       result.write('any path');
     } else {
       result.write('paths matching $include');
+    }
+    if (includeDefaults) {
+      result.write(' with defaults');
     }
     if (exclude != null && exclude!.isNotEmpty) {
       result.write(' except $exclude');
--- a/build_runner_core/lib/src/package_graph/target_graph.dart
+++ b/build_runner_core/lib/src/package_graph/target_graph.dart
@@ -61,7 +61,7 @@ class TargetGraph {
     final modulesByKey = <String, TargetNode>{};
     final publicAssetsByPackage = <String, InputMatcher>{};
     final modulesByPackage = <String, List<TargetNode>>{};
-    final includedSourceDefaults = <InputSet>{};
+    final sourcesWithIncludedDefaults = <InputSet>{};
     late BuildConfig rootPackageConfig;
     for (final package in packageGraph.allPackages.values) {
       final config = overrideBuildConfig[package.name] ??
@@ -85,8 +85,8 @@ class TargetGraph {
       }
       final nodes = config.buildTargets.values.map((target) {
         if (target.sources.includeDefaults &&
-            !includedSourceDefaults.contains(target.sources)) {
-          includedSourceDefaults.add(target.sources);
+            !sourcesWithIncludedDefaults.contains(target.sources)) {
+          sourcesWithIncludedDefaults.add(target.sources);
           target.sources.include!.addAll(defaultInclude);
         }
         return TargetNode(target, package, defaultInclude: defaultInclude);
--- a/build_runner_core/lib/src/package_graph/target_graph.dart
+++ b/build_runner_core/lib/src/package_graph/target_graph.dart
@@ -82,8 +82,14 @@ class TargetGraph {
         publicAssetsByPackage[package.name] =
             InputMatcher(const InputSet(), defaultInclude: defaultInclude);
       }
-      final nodes = config.buildTargets.values.map((target) =>
-          TargetNode(target, package, defaultInclude: defaultInclude));
+      final nodes = config.buildTargets.values.map((target) {
+        if (target.sources.include?.contains(r'$defaults$') == true) {
+          target.sources.include!.removeWhere((s) => s == r'$defaults$');
+          target.sources.include!.addAll(defaultInclude);
+        }
+
+        return TargetNode(target, package, defaultInclude: defaultInclude);
+      });
       if (package.name != r'$sdk') {
         var requiredPackagePaths =
             package.isRoot ? requiredRootSourcePaths : requiredSourcePaths;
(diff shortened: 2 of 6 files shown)


rotorgames:
@jakemac53 I guess we can implement it without huge changes in the code.

My approach with `$defaults$` looks not too bad as well to be honest cuz we already have the `$package$` and the `$lib$`, so the `$defaults$` just extends the list of existed synthetic sources.

Maybe, we can vote for the best approach here?

jakemac53:
> so the `$defaults$` just extends the list of existed synthetic sources.

But `$default$` is _not_ a synthetic source, that has a specific meaning and purpose (it can be used as an actual input, for builders that are intended to run on entire packages and have no input to pin off of).

Some other solution needs to be chosen that doesn't conflate the two things.

rotorgames:
@jakemac53 I changed my approach to `include_defaults`. Please, take a look.

build_config/lib/src/input_set.dart
```diff
@@ -38,7 +40,8 @@ class InputSet {
           'Expected a Map or a List but got a ${json.runtimeType}');
     }
     final parsed = _$InputSetFromJson(json as Map);
-    if (parsed.include != null && parsed.include!.any((s) => s.isEmpty)) {
+    if (parsed.include != null && parsed.include!.any((s) => s.isEmpty) ||
```
jakemac53, line 43: You should be able to use `include_defaults` with `excludes` but no `includes`, which I think this won't allow for.

I would also provide a different message for the case where `include_defaults` is set but no other configuration exists, this message is a bit confusing.

Lastly, I would put explicit parenthesis instead of relying on operator precedence here, it will make the intent more clear for the next reader here.

build_runner_core/lib/src/package_graph/target_graph.dart
```diff
@@ -82,8 +83,14 @@ class TargetGraph {
         publicAssetsByPackage[package.name] =
             InputMatcher(const InputSet(), defaultInclude: defaultInclude);
       }
-      final nodes = config.buildTargets.values.map((target) =>
-          TargetNode(target, package, defaultInclude: defaultInclude));
+      final nodes = config.buildTargets.values.map((target) {
+        if (target.sources.includeDefaults &&
+            !sourcesWithIncludedDefaults.contains(target.sources)) {
```
jakemac53, line 88: I am not quite understanding the point of this `sourcesWithIncludedDefaults` set, can you explain?

If the issue is a shared `sources` list across targets, then I think we have a different issue here.

rotorgames: @jakemac53 When I tested my code the first time, I found that a map function called several times. I didn't do a deep digging to find why, but now, when you have pointed to it I have done and I found, that a `nodes` variable has `Iterable` type, so, when we try to get values from the `nodes` variable in lines https://github.com/dart-lang/build/blob/1dc5aa685909a126ae2ef6e527bf392b603b8ac4/build_runner_core/lib/src/package_graph/target_graph.dart#L212 and https://github.com/dart-lang/build/blob/1dc5aa685909a126ae2ef6e527bf392b603b8ac4/build_runner_core/lib/src/package_graph/target_graph.dart#L108
We call the map function again. 

So, we can easily fix this issue just if we call `toList` method after `map` but, maybe you know another reason why we need to create a new `TargetNode` instance in the map function again and again?

build_config/lib/src/input_set.dart
rotorgames: @jakemac53 good note about `exclude`. 

I can fix it by two ways. 

**The first**: To set default non null value for `include`, but then `May be null or empty which means every possible path (like '**').` definition will not be true as  `include` will never be null. I found some `include` null checks in the code, so, this change can break something. Also, I don't want to change the established behavior. 

**The second**
I can do something like that:
```
if (parsed.includeDefaults &&
    (parsed.include?.isEmpty ?? true) &&
    (parsed.exclude?.isEmpty ?? true)) {
  throw ArgumentError.value(parsed.includeDefaults, 'include_defaults',
      'Include globs or Exclude globs must be defined if include_defaults is true');
}
if (parsed.includeDefaults && parsed.include == null) {
  // include field could not be null if we would like to include defaults
  parsed =
      InputSet(include: [], exclude: parsed.exclude, includeDefaults: true);
}
```
So, `include` field will never be null if `include_defaults` is true. It allows us to fill `include` with defaults values in the next steps.

I prefer the second way.

jakemac53:
One other thing here, is the `InputSet` class is occasionally used for things other than `sources`, in particular it is also used for [generate_for](https://github.com/dart-lang/build/blob/7a384f3849eeea46c1a7b41b6892155b5d29a526/build_config/lib/src/build_target.dart#L84).

We probably want to do some validation that people aren't passing it for those types of targets. I could see either some validation on `TargetBuilderConfig` to check that people don't try to use it for `generateFor`, or possibly we can do something more general, in the `InputMatcher` class, where it doesn't allow an input set with `includeDefaults == true` but no actual defaults provided.

jakemac53:
> We probably want to do some validation that people aren't passing it for those types of targets. I could see either some validation on `TargetBuilderConfig` to check that people don't try to use it for `generateFor`, or possibly we can do something more general, in the `InputMatcher` class, where it doesn't allow an input set with `includeDefaults == true` but no actual defaults provided.

Or actually it occurs to me that we could support `include_defaults` for `generate_for`, but the "defaults" in that case would just be the `sources` globs? So that way you could easily exclude just a few sources for instance.

build_runner_core/lib/src/package_graph/target_graph.dart
rotorgames: @jakemac53 I don't want to add `sources` in `InputSet` directly as well. But if we don't do it, we will have a problem with graphs as the graph doesn't use `InputMatcher`. For example here: 
https://github.com/dart-lang/build/blob/d2b84fd34ca6825ced31f5a0ecf3b38bbb5747d2/build_runner_core/lib/src/package_graph/apply_builders.dart#L346

We can fix it by changing `node.target.sources` to something like that: 
```dart
return createPhase(
  node.package,
  options,
  InputSet(
      include: node.sourceIncludes.map((e) => e.toString()).toList()),
  builderConfig?.generateFor,
  isReleaseMode);
```

But I'm not sure that my changes don't break something else. @jakemac53 how do you think are these changes worth it? If it is, so, we need to spend more time for testing graphs and etc as well.

jakemac53: Ultimately, we always convert the `InputSet` to an `InputMatcher` before using it, so I don't see a real issue here with passing `node.target.sources` directly there, it is immediately converted to an `InputMatcher` though so we could also more eagerly convert it.

It is interesting though how the `InBuildPhase` does not pass a `defaultInclude` for the `InputMatcher` that it creates. I think we could probably change `InputMatcher` to require the `defaultInclude` parameter though, to flag all places where it isn't passed (looks like only InBuildPhase), and then pass in the desired defaults there, which would require some additional plumbing. Possibly the `InBuildPhase` would need a `PackageNode` instead of just a `String` with the package name (but it looks like this is available, although it would break a bunch of tests).

jakemac53: But note that it has been a very long time since I looked at any of this code so there is definitely some investigation required here.

rotorgames:
Have been done in https://github.com/dart-lang/build/pull/3390

Issue #3312, opened 2022-05-25
[analyzer] InconsistentAnalysisException: Requested result might be inconsistent with previously returned results in analyzer 4.1.0


noinskit:
After upgrading analyzer from 3.4.1 to 4.1.0 I started getting `InconsistentAnalysisException` exceptions on `flutter pub run build_runner build` such as:
```
2022/05/25 14:03:05 [SEVERE] freezed:freezed on lib/src/parfs/atrial_fibrillation.dart:
2022/05/25 14:03:05 
2022/05/25 14:03:05 InconsistentAnalysisException: Requested result might be inconsistent with previously returned results
2022/05/25 14:03:05 
2022/05/25 14:03:05 package:analyzer/src/dart/analysis/session.dart 133:7  AnalysisSessionImpl._checkConsistency
2022/05/25 14:03:05 package:analyzer/src/dart/analysis/session.dart 94:5   AnalysisSessionImpl.getParsedUnit
2022/05/25 14:03:05 package:build_resolvers/src/resolver.dart 204:32       AnalyzerResolver.compilationUnitFor
2022/05/25 14:03:05 package:build_resolvers/src/resolver.dart 111:26       PerActionResolver.compilationUnitFor.<fn>
2022/05/25 14:03:05 package:source_gen/src/builder.dart 373:18             _hasAnyTopLevelAnnotations
2022/05/25 14:03:05 package:source_gen/src/builder.dart 387:9              _hasAnyTopLevelAnnotations
2022/05/25 14:03:05 package:source_gen/src/builder.dart 89:11              _Builder.build
```

I also opened an issue against freezed first (https://github.com/rrousselGit/freezed/issues/671), but since freezed is not in the stack trace above at all, it might be a problem with the analyzer instead.

It does not happen every time and not always with the same set of files, also tends to differ e.g. between my laptop and my CI/CD env.

I also noticed that it seems related to a change mentioned in analyzer 4.0.0 changelog:
```
Changed synchronous AnalysisSession.getFile and getParsedUnit to fail if there are pending file changes, instead of reading. As for any other AnalysisSession method, await AnalysisContext.applyPendingFileChanges().
```
(`AnalyzerSession.getParsedUnit` is what's throwing the exception here)

I'm using flutter/dart with the following versions:
```
Flutter 3.0.1 • channel stable • https://github.com/flutter/flutter.git
Framework • revision fb57da5f94 (6 days ago) • 2022-05-19 15:50:29 -0700
[...]

PR #3313, opened 2022-05-26
Add global analysis driver pool to protect usages

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3312

If the driver is used in between calls to `changeFile` and before `applyPendingChanges` completes, then you can get `InconsistentAnalysisError`s. This change ensures we won't do that through a new single resource pool shared by all users of the driver.

- Guard all accesses to the driver and session with the new resource.
  - During the resolve, this only grabs the resource once it is fully ready for all of the `changeFile` calls and subsequent call to `applyPendingFileChanges`. This prevents deadlocks compared to a global resolution pool, which may call out to `package:build` apis which can deadlock.
  - This shouldn't cause significant slowdown, most of the code using the pool has a fixed number of async actions it waits on, and none of those actions might need to wait on a build step to complete (they are all async calls to analyzer apis, not build apis).
- Keeps in place the per action resolve pool, still needed for managing updates to `_entryPoints`.
- Temporarily overrides all packages to run with this version of build_resolvers, will revert that commit before landing

### DIFF
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -30,12 +30,6 @@ import 'analysis_driver.dart';
 import 'build_asset_uri_resolver.dart';
 import 'human_readable_duration.dart';
 
-// Used to protect all usages of the analysis driver from
-// `InconsistentAnalysisException` errors, which might occur if we are in the
-// middle of some calls to `changeFile` but have not yet completed
-// `applyPendingFileChanges`.
-final _driverPool = Pool(1);
-
 final _logger = Logger('build_resolvers');
 
 Future<String> _packagePath(String package) async {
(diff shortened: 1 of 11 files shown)


build_resolvers/lib/src/resolver.dart
```diff
@@ -29,6 +30,12 @@ import 'analysis_driver.dart';
 import 'build_asset_uri_resolver.dart';
 import 'human_readable_duration.dart';
 
+// Used to protect all usages of the analysis driver from
+// `InconsistentAnalysisException` errors, which might occur if we are in the
+// middle of some calls to `changeFile` but have not yet completed
+// `applyPendingFileChanges`.
+final _driverPool = Pool(1);
```
natebosch, line 37: Could we have this as a field on `AnalyzerResolver`?

We don't have any use cases for it, but the rest of the code does (or should?) allow for multiple instances of the resolver to be independent of one another.

natebosch (approved):
Can we expand the commit message with 2 things:

-    More details about why we need to guard access to reads from the driver.
-    A rough sketch of the way the code is changing - in particular the nested function argument used to lock only part of the work.

build_resolvers/lib/src/resolver.dart
jakemac53: Oh true, I think this could be an field on `AnalyzerResolver`, I can change that.

scheglov (approved):

build_resolvers/lib/src/resolver.dart
jakemac53: done

jakemac53:
> Can we expand the commit message with 2 things:

How does the updated text look?

natebosch:
> How does the updated text look?

Looks great to me. Thanks!

PR #3314, opened 2022-06-03
symlink example/README.md to README.md for a better pub score

jakemac53:
silly but it does seem to work, there is no API here so the readme is the best place to go.

### DIFF
--- a/build_runner/example/README.md
+++ b/build_runner/example/README.md
@@ -1,0 +1,1 @@
+../README.md


natebosch:
I think the intent is to describe the example. Is there any very brief summary we could put in?

jakemac53:
There is no example, because this is just a tool with no API?

If you are looking for how to use the package, I think the root README.md is the best place to go? That is generally how we expect to introduce people to the package.

evanweible-wf:
fwiw, I filed an issue for a similar scenario: https://github.com/dart-lang/pub-dev/issues/2931

I think it's silly to require a _separate_ example for all packages.

jakemac53:
> I think it's silly to require a _separate_ example for all packages.

Yeah, it is a bit silly, but symlinking the readme seems like a pretty easy workaround lol.

natebosch:
Ah, I didn't realize this is _creating_ the example. I thought there was an example directory and we were adding a README.

> pretty easy workaround

I think if we're going to ignore the spirit of the score it would be better to push more on changing the score criteria.
cc @mit-mit

jakemac53:
I don't have a strong opinion here, just noticed we are up to 40 points dinged now, although only 10 of those I could do anything about (these 10). The other 30 are all bogus.

jakemac53:
I did also consider just linking https://www.youtube.com/watch?v=rmABbHSOTqQ if we want to go full on cheeky mode :P

jakemac53:
@natebosch did you want me to just close this?

natebosch:
I think `build_runner_core` could get a copy.

PR #3342, opened 2022-07-25
migrate off of library.parts

jakemac53:


jakemac53:
Looks like links to direct source files on github are not working, causing the markdown checker to fail.

build_resolvers/lib/src/resolver.dart
```diff
@@ -256,7 +256,7 @@ class AnalyzerResolver implements ReleasableResolver {
   Future<List<ErrorsResult>> _syntacticErrorsFor(LibraryElement element) async {
     final existingElements = [
       element,
-      for (final part in element.parts)
+      for (final part in element.parts2)
```
jakemac53, line 259: It looks like this no longer gives the actual part source, but just the library source.

How can I get the part source? It seems hidden now.

jakemac53: Should I cast these to a `DirectiveUriWithSource`? Or just do a check for that type?

jakemac53: Ok yeah that looks like the fix, I updated to check for the `DirectiveUriWithSource` type and then use the source from that. LMK if I should do something else.

build_resolvers/lib/src/resolver.dart
```diff
@@ -254,17 +254,19 @@ class AnalyzerResolver implements ReleasableResolver {
...
-      for (final part in element.parts)
-        // The source may be null if the part doesn't exist. That's not
-        // important for us since we only care about syntax
-        if (part.source.exists()) part,
-    ];
+    final existingSources = [element.source];
+
+    for (final part in element.parts2) {
+      var uri = part.uri;
+      // The source may be null if the part doesn't exist. That's not
```
scheglov, line 261: No, if it is `DirectiveUriWithSource`, then the `.source` is not `null`.

This cycle could be `parts2.map((e) => e.uri).whereType<DirectiveUriWithSource>().map((e) => e.source)`, and used with `...` in the list literal.

scheglov (approved):

build_resolvers/test/resolver_test.dart
```diff
@@ -238,7 +238,7 @@ void main() {
               } ''',
       }, (resolver) async {
         var lib = await resolver.libraryFor(entryPoint);
-        expect(lib.imports.length, 2);
+        expect(lib.libraryImports.length, 2);
```
scheglov, line 241: BTW, see the comment below.

build_resolvers/lib/src/resolver.dart
jakemac53: I updated the comment a bit from the old language, I think its getting a bit much to try and jam into a list literal though.

build_resolvers/test/resolver_test.dart
jakemac53: Updated

PR #3351, opened 2022-08-12
Avoid deprecated analyzer APIs

natebosch:
- `getType` -> `getClass`
- `element` -> `element2`
- `name.name` -> `name2.lexeme`
- `declaredElement` -> `declaredElement2`


scheglov (approved):
Thank you!

Issue #3354, opened 2022-08-17
Replace `findPackageConfigUri` with `loadPackageConfigUri` in package_reader.dart


jensjoha:
https://github.com/dart-lang/build/blob/fcff629feab5c5e3210593176aae5a3a50000de9/build_test/lib/src/package_reader.dart#L75-L83

should not use `findPackageConfigUri` but instead use `loadPackageConfigUri`.

Currently it gets a specific uri for the package config, then asks `package:package_config` to discover which one to use, which looks in the directory of the uri by conversion filenames (i.e. tries `.dart_tools/package_config.json` and will read that as a json file, or `.packages` and will read that as an ini style format).

This for instance means that if the uri given points to a json file called `.packages` it will be found an attempted to be read as an old-style format, and that if it points to something called, say, `foo.bar` it will never be "found".

Note that this makes for a crash internally.

PR #3355, opened 2022-08-17
Make package configuration loading use the correct function.

lrhn:
When you have the URI of a package configuration file, you should load it directly.
It's not safe to use discovery, which assumes you start with a Dart script file.

Fixes https://github.com/dart-lang/build/issues/3354

### DIFF
--- a/build_test/CHANGELOG.md
+++ b/build_test/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## 2.1.6-dev
 
+- Fix bug in package config file loading.
+
 ## 2.1.5
 
 - Allow the latest analyzer.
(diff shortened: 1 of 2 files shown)


mit-mit:
Please update changelog and pubspec too.

jensjoha (approved):
LGTM.
Though "you don't need to use package configuration discovery." is misleading.
It's wrong to do it, and we've seen how it can crash depending on filenames.

lrhn:
Pubspec is at 2.1.6-dev. A bug-fix should be a safe addition to that release number, no need for a minor version increment.

If we want to release, we have to remove the -dev, but probably also other things, like references to this package from other packages in the same mono-repo. I'd prefer to leave that to people who know the repo.

mit-mit (approved):

jakemac53 (approved):

PR #3365, opened 2022-09-08
Fix stuck in LruCache after updating items

linroid:
Inside the [LruCache](build_runner_core/lib/src/asset/lru_cache.dart#L29), it always create a new link node when doing the `[]=` operator, what if we're doing the "updating" operation? 

The `_entries` will replace the new value to the same key entry, but the link list will have two nodes with a same key, once the capacity is exhausted, the application will stuck in removing item as the second removing with the same key changes nothing.

You can reproduce this case by executing the test cases provided by this PR.

### DIFF
--- a/build_runner_core/lib/src/asset/lru_cache.dart
+++ b/build_runner_core/lib/src/asset/lru_cache.dart
@@ -32,7 +32,7 @@ class LruCache<K, V> {
     if (entry.weight > _individualWeightMax) {
       return;
     }
-    
+
     if (_entries.containsKey(key)) {
       remove(key);
     }
--- a/build_runner_core/lib/src/asset/lru_cache.dart
+++ b/build_runner_core/lib/src/asset/lru_cache.dart
@@ -27,14 +27,11 @@ class LruCache<K, V> {
   }
 
   void operator []=(K key, V value) {
+    remove(key);
     var entry = _Link(key, value, _computeWeight(value));
     // Don't cache at all if above the individual weight max.
     if (entry.weight > _individualWeightMax) {
       return;
-    }
-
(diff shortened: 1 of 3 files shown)


build_runner_core/test/asset/lru_cache_test.dart
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
+@Timeout(Duration(seconds: 10))
```
linroid, line 4: The annotation actually won't stop the endless execution, anyone knows how to fix it?

jakemac53: The way timeouts work in this package is a bit weird - what it actually means is that the test must "make progress" in the timeout window. As long as it executes any code at all, the timeout resets. It only catches things that are stuck completely and not executing any async tasks.

build_runner_core/lib/src/asset/lru_cache.dart
```diff
@@ -32,6 +32,10 @@ class LruCache<K, V> {
     if (entry.weight > _individualWeightMax) {
       return;
     }
+    
+    if (_entries.containsKey(key)) {
```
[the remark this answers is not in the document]
linroid: It makes more sense, thanks:)

build_runner_core/test/asset/lru_cache_test.dart
linroid: I see, thanks for the explanation..

jakemac53 (approved):

PR #3368, opened 2022-09-13
Support latest package:analyzer

kevmoo:
Prepare to publish several packages

### DIFF
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -4,7 +4,7 @@ description: A package for authoring build_runner compatible code generators.
 repository: https://github.com/dart-lang/build/tree/master/build
 
 environment:
-  sdk: ">=2.14.0 <3.0.0"
+  sdk: ">=2.17.0 <3.0.0"
 
 dependencies:
   analyzer: ">=1.5.0 <6.0.0"
--- a/build/pubspec.yaml
+++ b/build/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build
-version: 2.3.0
+version: 2.3.1
 description: A package for authoring build_runner compatible code generators.
 repository: https://github.com/dart-lang/build/tree/master/build
 
@@ -7,7 +7,7 @@ environment:
   sdk: ">=2.14.0 <3.0.0"
 
 dependencies:
-  analyzer: ">=1.5.0 <5.0.0"
+  analyzer: ">=1.5.0 <6.0.0"
   async: ^2.5.0
   convert: ^3.0.0
   crypto: ^3.0.0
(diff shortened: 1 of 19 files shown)


build/pubspec.yaml
```diff
@@ -21,3 +21,9 @@ dev_dependencies:
   build_test: ^2.0.0
   lints: '>=1.0.0 <3.0.0'
   test: ^1.16.0
+
+dependency_overrides:
```
jakemac53, line 25: Do you need these?

jakemac53 (approved):
LGTM but I think you should be able to remove the overrides?

build/pubspec.yaml
kevmoo: I need them for latest analyzer to resolve for tests. I was going to drop when I publish.

kevmoo:
> LGTM but I think you should be able to remove the overrides?

I was going to before I publish. Should I land the PR w/out the overrides?

jakemac53:
> I was going to before I publish. Should I land the PR w/out the overrides?

If they aren't needed I would yes - I realize now you probably did this to allow you to get the latest analyzer? But we can see its green so I would now revert and then land.

jakemac53:
Oh actually - it looks like you still did not get the latest analyzer (at least for some of the runs).

You might want to add an override to the latest analyzer to force it.

jakemac53:
Scratch my last comment - looks like the dev builds did pick it up

build_resolvers/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
+version: 2.0.10
 description: Resolve Dart code in a Builder
 repository: https://github.com/dart-lang/build/tree/master/build_resolvers
 
 environment:
   sdk: ">=2.17.0 <3.0.0"
 
 dependencies:
-  analyzer: ^4.6.0
+  analyzer: '>=4.4.0 <6.0.0'
```
natebosch, line 10: Is this safe to downgrade the min version?

```suggestion
  analyzer: '>=4.6.0 <6.0.0'
```

build_runner_core/pubspec.yaml
```diff
@@ -27,7 +27,7 @@ dependencies:
   yaml: ^3.0.0
 
 dev_dependencies:
-  analyzer: ^4.6.0
+  analyzer: '>=4.4.0 <6.0.0'
```
natebosch, line 30:
```suggestion
  analyzer: '>=4.6.0 <6.0.0'
```

build_test/pubspec.yaml
```diff
@@ -24,6 +24,6 @@ dependencies:
   watcher: ^1.0.0
 
 dev_dependencies:
-  analyzer: ^4.6.0
+  analyzer: '>=4.4.0 <6.0.0'
```
natebosch, line 27:
```suggestion
  analyzer: '>=4.6.0 <6.0.0'
```

natebosch (approved):

kevmoo:
Thanks @natebosch – copy/paste error

PR #3386, opened 2022-10-18
specify platform support in the pubspec

jakemac53:


natebosch:
We'll need a version bump in both packages. (I'm not sure why we didn't get an package uploaded emails yet)

natebosch (approved):

Issue #3387, opened 2022-10-18
Resolving `dart:` SDK libraries


simolus3:
This is mostly a question, but if it turns out to be impossible, then please view this as a feature request. I'm happy to implement support for it in that case.

Basically, I'm looking for a way to reliably resolve a `dart:` library through the build resolver (so that I can get the `LibraryElement` for it). I know that ASTs for `dart:` will probably not work in all implementations, but getting the element model should be fine?

I know that I can lookup libraries by name (so doing that with `dart.core` would work), but I'm also looking for ways to resolve libraries not imported by any of the entrypoints used by the resolver. Is there a decent way to get hold of a `LibraryElement` for an SDK library at the moment?

jakemac53:
It seems to me like this works today? I tried adding the following test to build_resolvers which passes:

```dart
test('Can resolve sdk libraries that are not imported', () async {
    await resolveSources({
      'a|lib/a.dart': '',
    }, (resolver) async {
      var convert = await resolver.findLibraryByName('dart.convert');
      expect(convert, isNotNull);
      expect(convert!.getClass('Codec'), isNotNull);
    }, resolvers: AnalyzerResolvers());
  });
```

simolus3:
Huh, I didn't expect this to work. The docs mention that imported libraries are available, so I'd expect that only `dart:core` would be reachable. Is this because `dart:core` imports the other SDK libraries?

jakemac53:
I don't see any specific logic in there adding the SDK libraries, so I think that is probably the case. That would also explain why we don't see `dart:mirrors`, `dart:cli`, `dart:io`, etc, which I now realize are all missing :).

So yeah, I think there is a feature request here, maybe an api like `Resolver#libraryForUri`? I don't believe we have a way of creating an asset id for an sdk library, so you can't use `Resolver#libraryFor`. The naming becomes a bit unfortunate, but 🤷 .

jakemac53:
Actually though this would end up requiring a breaking change for `package:build`, since that interface is meant to be implemented :(.

jakemac53:
Given that, I think we should either make some way of talking about an SDK library with an `AssetId`, or allow `findLibraryByName` to search all available SDK libraries. If we do the latter we need to decide if they should all be added to `libraries` as well. 

Note that the API docs on `libraries` today actually says it does include all SDK libraries. It just lies :). But maybe just implementing that properly is the easiest fix.

simolus3:
[...]

PR #3389, opened 2022-10-21
Include all SDK libraries in resolver

simolus3:
This PR makes the following changes to the resolvers interface and the default implementation in this repository:

- I've updated the documentation of `libraries` to explicitly state which libraries are included.
- I've adapted the default implementation to always include everything from the Dart SDK.

https://github.com/dart-lang/build/issues/3387

### DIFF
--- a/build/lib/src/analyzer/resolver.dart
+++ b/build/lib/src/analyzer/resolver.dart
@@ -18,10 +18,15 @@ abstract class Resolver {
   /// or is a `part of` file (not a standalone Dart library).
   Future<bool> isLibrary(AssetId assetId);
 
-  /// All libraries recursively accessible from the entry point or subsequent
-  /// calls to [libraryFor] and [isLibrary].
+  /// All libraries resolved by this resolver.
   ///
-  /// **NOTE**: This includes all Dart SDK libraries as well.
+  /// This includes the following libraries:
+  ///  - The primary input of this resolver (in other words, the
+  ///   [BuildStep.inputId] of a build step).
+  ///  - Libraries resolved with a direct [libraryFor] call.
+  ///  - Every public `dart:` library part of the SDK.
+  ///  - All libraries recursively accessible from the mentioned sources, for
+  ///    instance because due to imports or exports.
   Stream<LibraryElement> get libraries;
 
   /// Returns the parsed [AstNode] for [Element].
@@ -58,8 +63,9 @@ abstract class Resolver {
   /// Returns the first resolved library identified by [libraryName].
   ///
   /// A library is resolved if it's recursively accessible from the entry point
-  /// or subsequent calls to [libraryFor] and [isLibrary]. If no library can be
-  /// found, returns `null`.
+  /// or subsequent calls to [libraryFor]. In other words, this searches for
+  /// libraries in [libraries].
+  /// If no library can be found, returns `null`.
   ///
   /// **NOTE**: In general, its recommended to use [libraryFor] with an absolute
   /// asset id instead of a named identifier that has the possibility of not
--- a/build_resolvers/lib/src/resolver.dart
+++ b/build_resolvers/lib/src/resolver.dart
@@ -261,7 +261,7 @@ class AnalyzerResolver implements ReleasableResolver {
   Future<List<ErrorsResult>> _syntacticErrorsFor(LibraryElement element) async {
     final existingSources = [element.source];
 
-    for (final part in element.parts2) {
+    for (final part in element.parts) {
       var uri = part.uri;
       // There may be no source if the part doesn't exist. That's not important
       // for us since we only care about existing file syntax.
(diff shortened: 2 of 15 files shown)


build/lib/src/analyzer/resolver.dart
```diff
@@ -18,10 +18,15 @@ abstract class Resolver {
   /// or is a `part of` file (not a standalone Dart library).
   Future<bool> isLibrary(AssetId assetId);
 
-  /// All libraries recursively accessible from the entry point or subsequent
-  /// calls to [libraryFor] and [isLibrary].
+  /// All libraries resolved by this resolver.
```
simolus3, line 21: I have updated the documentation of this getter and `findLibraryByName` to be consistent with the current implementation in `build_resolvers`.

In that implementation, neither `isLibrary` nor `compilationUnitFor` add an entrypoint. That behavior feels correct to me (and I'm not aware of any complaints about this), so I'm in favor of updating the documentation. I've also added a test for this behavior (calling `isLibrary` should not alter `libraries`).

build_resolvers/lib/src/resolver.dart
```diff
@@ -17,6 +17,8 @@ import 'package:analyzer/error/error.dart';
 import 'package:analyzer/file_system/physical_file_system.dart';
 // ignore: implementation_imports
 import 'package:analyzer/src/clients/build_resolvers/build_resolvers.dart';
+// ignore: implementation_imports
+import 'package:analyzer/src/dart/sdk/sdk.dart';
```
simolus3, line 21: I have a feeling the other library in the analyzer package has only been added so that we can avoid implementing other internal libraries here? Should a method to get a list of SDK library URIs be added to the `clients/build_resolvers` library then?

jakemac53: cc @scheglov

build/lib/src/analyzer/resolver.dart
jakemac53: SGTM

build_resolvers/lib/src/resolver.dart
```diff
@@ -297,6 +307,16 @@ class AnalyzerResolver implements ReleasableResolver {
     throw UnimplementedError();
   }
 
+  Stream<LibraryElement> get sdkLibraries {
+    return Stream.fromFutures(_sdkUris.map((uri) {
```
jakemac53, line 311: Can we cache these or is it not safe (will we get errors about stale analysis?)

jakemac53 (approved):

natebosch:
Design SGTM. Sounds like we might need to wait for a change in analyzer to surface this information from the SDK summary

build_resolvers/lib/src/resolver.dart
simolus3: I've migrated to the new API, but I'll mark this PR as WIP until the analyzer version is released.

jakemac53: SGTM, we will also need to increase the min analyzer version in the pubspec (not sure if we know the number yet, so you may have to wait)

scheglov:
Note also, https://dart-review.googlesource.com/c/sdk/+/265561
There are some issues with publishing `analyzer` right now, but once published, you will get even more APIs deprecated (all just renames, IIRC).

scheglov:
https://pub.dev/packages/analyzer/changelog#520

jakemac53:
cc @simolus3 do you mind also fixing up the deprecations in this PR? We could also do it separately, but its going to block running tests here until we do that otherwise, so might be easier to just do it all at once.

simolus3:
I've fixed the deprecation warnings and used 5.2.0 as a minimum supported analyzer version on those packages. I also had to raise the min SDK constraint to 2.18.0 because `mono_repo` likes to test with 2.17.0 otherwise and there is no version of `test_core` compatible with the latest analyzer _and_ Dart 2.17.

jakemac53 (approved):

build_runner_core/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: Core tools to organize the structure of a build and run Builders.
 repository: https://github.com/dart-lang/build/tree/master/build_runner_core
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
+  sdk: ">=2.18.0 <3.0.0"
```
jakemac53, line 7: We should also update the version to `7.2.8-dev` here and mention the min sdk bump

build_test/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ version: 2.1.6-dev
 repository: https://github.com/dart-lang/build/tree/master/build_test
 
 environment:
-  sdk: '>=2.17.0 <3.0.0'
+  sdk: '>=2.18.0 <3.0.0'
```
jakemac53, line 7: lets mention this bump in the changelog here also

scratch_space/pubspec.yaml
```diff
@@ -4,7 +4,7 @@ description: A tool to manage running external executables within package:build
 repository: https://github.com/dart-lang/build/tree/master/scratch_space
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
+  sdk: ">=2.18.0 <3.0.0"
```
jakemac53, line 7: and this one :)

PR #3390, opened 2022-10-22
Build hidden outputs if they're public

simolus3:
Some directories, like `lib/`, are special-cased by the build system in the sense that

1. Their assets are readable for non-root packages
2. Hidden assets are generated for them even if they're not part of any build filter or directory.

Some time ago, I've added the `additional_public_assets` configuration to expand this behavior to non-standard directories. As a motivation, consider building a C library with the build system and some builder generating a hidden file in `include/`. Similar to hidden outputs in `lib/`, we definitely want this builder to run for this asset since `include/` has a meaning somewhat comparable to `lib/` in Dart. The `additional_public_assets` build configuration hence makes it easier to integrate non-Dart components in the Dart package ecosystem and build them with Dart's standard build tools.

I am also running into this issue with a static site generator that I'm writing. It generates a (hidden) asset for markdown files in `pages/` containing HTML and a list of target URLs. A post-process builder later generates `.html` assets in `web/` based on these intermediate assets. This builder breaks when a build directory is present, since hidden assets in `pages/` are no longer built (see https://github.com/dart-lang/webdev/issues/1200, https://github.com/dart-lang/webdev/issues/1733).

Having public assets handled in a consistent way would really help with my static site generator. I also believe that it is generally the right thing to do in the implementation as it could avoid confusion for advanced users interested in building not-only-Dart modules or packages.
So, this PR makes the output check consistent with the notion of a public asset used in other parts of the build implementation.

### DIFF
--- a/build_runner_core/lib/src/util/build_dirs.dart
+++ b/build_runner_core/lib/src/util/build_dirs.dart
@@ -6,33 +6,39 @@ import 'package:build/build.dart';
 
 import '../generate/options.dart';
 import '../generate/phase.dart';
+import '../package_graph/target_graph.dart';
 
 /// Returns whether or not [id] should be built based upon [buildDirs],
-/// [phase], and optional [buildFilters].
+/// [phase], [targetGraph], and optional [buildFilters].
 ///
 /// The logic for this is as follows:
 ///
 /// - If any [buildFilters] are supplied, then this only returns `true` if [id]
 ///   explicitly matches one of the filters.
 /// - If no [buildFilters] are supplied, then the old behavior applies - all
-///   build to source builders and all files under `lib` of all packages are
-///   always built.
+///   build to source builders and all public assets (according to
+///   [TargetGraph.isPublicAsset]) are always built.
 /// - Regardless of the [buildFilters] setting, if [buildDirs] is supplied then
 ///   `id.path` must start with one of the specified directory names.
-bool shouldBuildForDirs(AssetId id, Set<String> buildDirs,
-    Set<BuildFilter>? buildFilters, BuildPhase phase) {
-  buildFilters ??= {};
+bool shouldBuildForDirs(
+  AssetId id, {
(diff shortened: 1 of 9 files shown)


build_runner_core/test/asset/finalized_reader_test.dart
```diff
@@ -18,8 +20,13 @@ void main() {
   group('FinalizedReader', () {
     FinalizedReader reader;
     late AssetGraph graph;
+    late TargetGraph targetGraph;
 
     setUp(() async {
+      final packageGraph = buildPackageGraph({rootPackage('foo'): []});
```
jakemac53, line 26: nit: we usually just use `a` as the default package name (same below)

jakemac53 (commented):
I do think this makes sense, just some nits.

build_runner_core/lib/src/util/build_dirs.dart
```diff
@@ -6,33 +6,39 @@ import 'package:build/build.dart';
...
+  required BuildPhase phase,
+  required TargetGraph targetGraph,
+  Set<BuildFilter> buildFilters = const {},
+}) {
   if (buildFilters.isEmpty) {
+    // Build non-hidden and public assets
     if (!phase.hideOutput) return true;
 
-    if (id.path.startsWith('lib/')) return true;
+    if (targetGraph.isPublicAsset(id)) return true;
```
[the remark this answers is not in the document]
simolus3: I've inlined the return statement in the two branches. I've also moved the `isPublicAsset` to the end of the or, so in most cases it won't be called at all.

build_runner_core/test/asset/finalized_reader_test.dart
jakemac53: Ah no I probably was just skimming too fast and didn't realize :)

jakemac53 (approved):

PR #3395, opened 2022-10-27
Fix daemon_test.dart

a-siva:
Remove @dart specification from daemon_test.dart in preparation for https://dart-review.googlesource.com/c/sdk/+/261140.
This is based on the observation from @emmanuel-p

### DIFF
--- a/build_daemon/test/daemon_test.dart
+++ b/build_daemon/test/daemon_test.dart
@@ -167,7 +167,6 @@ Future<String> _statusOf(Process daemon) async {
 
 Future<Process> _runDaemon(var workspace, {int timeout = 30}) async {
   await d.file('test.dart', '''
-    // @dart=2.9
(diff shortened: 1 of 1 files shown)


jakemac53:
That change is likely to break a whole lot more than this test.....

jakemac53:
I think that change likely requires some additional work in any package that spawns a dart isolate or program - at least if they do it from source?

Do the compile commands still do the auto-detection? We rely on this auto detection all over the place.

jakemac53 (approved):
I will go ahead and approve/merge this anyways as we should be able to run with sound null safety now

a-siva:
> 
> Do the compile commands still do the auto-detection? We rely on this auto detection all over the place.

They will continue to do auto detection in Dart 2.19 and the CL that disables auto detection will land after the branch cut for the final beta of Dart 2.19.

PR #3405, opened 2022-11-17
Persistent byte store for analysis driver

simolus3:
In a single build (e.g. not using `watch`), the resolver can take up the majority of the build time. When users are complaining about slow builds, the slowdown is virtually always caused by the analyzer having to analyze hundreds or thousands of files, with lots of transitive imports and a huge package graph being common for Flutter apps.

The analyzer can persist computed resolve results on disk to speed up subsequent resolves. At the moment, we're not benefiting from this as we're always using a memory byte store. So for each `build_runner build`, the analyzer needs to re-analyze all sources being resolved. If we can look up computation results from a previous run, my intuition is that build performance would increase.

I've done a (completely non-vigorous) benchmark on some of the projects I maintain and I've seen build times drop by around 20%. If you know of other large, public project I could try this out on, I will report the numbers for them as well or try to tune the cache behavior.

## Trying this out

In third-party projects, these optimizations can be tried by adding this to the pubspec:

```yaml
dependency_overrides:
  build_resolvers:
    git:
      url: https://github.com/simolus3/build.git
      path: build_resolvers
      ref: persistent-analyzer-byte-store
  build_runner_core:
    git:
      url: https://github.com/simolus3/build.git
      path: build_runner_core
      ref: persistent-analyzer-byte-store
```

This effect is most visible when running a full build with the analyzer cache:

1. Apply these changes to your pubspec
2. Run `rm -r .dart_tool/`
3. Run `dart run build_runner build --delete-conflicting-outputs`
4. Run `rm -r .dart_tool/build/`. This deletes the build cache but keeps the analyzer cache alive.
5. Re-run `dart run build_runner build --delete-conflicting-outputs`. The time spent in the build (shown at the end as "succeeded after x seconds") should ideally be lower now.

### DIFF
--- a/build_resolvers/lib/src/analysis_driver.dart
+++ b/build_resolvers/lib/src/analysis_driver.dart
@@ -22,6 +22,7 @@ Future<AnalysisDriverForPackageBuild> analysisDriver(
   AnalysisOptions analysisOptions,
   String sdkSummaryPath,
   PackageConfig packageConfig,
+  ByteStore? byteStore,
 ) async {
   final cachePath = p.join('.dart_tool', 'build_resolvers', 'cache');
   await Directory(cachePath).create(recursive: true);
@@ -34,11 +35,7 @@ Future<AnalysisDriverForPackageBuild> analysisDriver(
     ),
     resourceProvider: buildAssetUriResolver.resourceProvider,
     sdkSummaryBytes: File(sdkSummaryPath).readAsBytesSync(),
-    byteStore: createByteStore(
-      directory: cachePath,
-      maxFileSize: 1024 * 1024 * 1024, // 1 GiB
-      memoryCacheSize: 1024 * 1024 * 128, // 128 MiB
-    ),
+    byteStore: byteStore,
     uriResolvers: [
       buildAssetUriResolver,
     ],
--- a/build_resolvers/lib/src/analysis_driver.dart
+++ b/build_resolvers/lib/src/analysis_driver.dart
@@ -17,16 +17,13 @@ import 'build_asset_uri_resolver.dart';
 /// Builds an [AnalysisDriverForPackageBuild] backed by a summary SDK.
 ///
 /// Any code must be resolvable through [buildAssetUriResolver].
-Future<AnalysisDriverForPackageBuild> analysisDriver(
+AnalysisDriverForPackageBuild analysisDriver(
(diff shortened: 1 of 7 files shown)


build_resolvers/lib/src/analysis_driver.dart
```diff
@@ -31,6 +34,11 @@ Future<AnalysisDriverForPackageBuild> analysisDriver(
     ),
     resourceProvider: buildAssetUriResolver.resourceProvider,
     sdkSummaryBytes: File(sdkSummaryPath).readAsBytesSync(),
+    byteStore: createByteStore(
+      directory: cachePath,
+      maxFileSize: 1024 * 1024 * 1024, // 1 GiB
```
jakemac53, line 39: this could likely be increased some, although I don't have a sense of how large this cache would get on big projects

build_resolvers/lib/src/analysis_driver.dart
```diff
@@ -23,6 +23,9 @@ Future<AnalysisDriverForPackageBuild> analysisDriver(
   String sdkSummaryPath,
   PackageConfig packageConfig,
 ) async {
+  final cachePath = p.join('.dart_tool', 'build_resolvers', 'cache');
```
jakemac53, line 26: maybe add an `analyzer` dir in here as well just to be a bit more specific

jakemac53:
Anything we can do to help analyzer performance sgtm

build_resolvers/lib/src/analysis_driver.dart
simolus3: The analysis server uses a 1GiB cache as well. That is for pretty much everything the analyzer has ever seen, so I think the limitation is already fairly large here.

Since we previously used a complete in-memory byte store, I have now chosen an approach that doesn't have a limitation on how much we keep in memory as well. We also write everything into a single file in the end instead of doing a write for every cache change. The analyzer also spawns an isolate doing regular cleanup work for each byte store, so I think a first-party implementation in `build_resolvers` that knows when to persist data is better.

jakemac53:
We can work on fixing things up for the 3.0 dev release this week, probably should just hold on this until we resolve those issues

mosuem:
Closing this, as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #3411, opened 2022-11-21
Report and shutdown after file watch errors

grouma:
Internally we are seeing reliability issues with our file watcher. This should help diagnose the root cause.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.1.1-dev
+
+- Report file watching errors and stop the daemon.
+
+
 ## 3.1.0
 
 - Add `BuildResults.changedAssets` containing asset URIs changed during a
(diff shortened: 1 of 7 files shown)


build_daemon/lib/constants.dart
```diff
@@ -12,6 +12,12 @@ const optionsSkew = 'DIFFERENT OPTIONS';
 
 const buildModeFlag = 'build-mode';
 
+/// Daemon shuts down after this timeout if there is no active connection.
+const defaultIdleTimeout = Duration(seconds: 30);
+
+const fileChangeEventErrorCode = 1;
```
jakemac53, line 18: We might want a less common error code for these

grouma: Done.

build_daemon/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 3.1.1-dev
```
jakemac53, line 1: we don't have other changes coming here, do you want to just release this right away? Feel free to drop the -dev if so.

jakemac53 (approved):

PR #3416, opened 2022-11-30
use TypeError instead of NullThrownError

jakemac53:
This fixes the dev bot failures

### DIFF
--- a/_test/test/common/utils.dart
+++ b/_test/test/common/utils.dart
@@ -68,6 +68,9 @@ Future<void> _startServer(String command, List<String> buildArgs,
   }
 
   final proc = _process = await Process.start(command, buildArgs);
+  unawaited(proc.exitCode.then((_) {
+    if (_process != null) _process = null;
+  }));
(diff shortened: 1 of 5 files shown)


jakemac53:
Looks like some additional failures due to unexpected output on stderr when running `dart run`, because it now does a `pub get` as a part of that by default?

I will try to figure out the best way to resolve that.

jakemac53:
Filed https://github.com/dart-lang/sdk/issues/50592 on the sdk, skipping these expectations for now.

kevmoo (approved):

natebosch (approved):

_test/test/common/utils.dart
```diff
@@ -68,6 +68,9 @@ Future<void> _startServer(String command, List<String> buildArgs,
   }
 
   final proc = _process = await Process.start(command, buildArgs);
+  unawaited(proc.exitCode.then((_) {
```
natebosch, line 71: How did this come up? Was there a problem when the process died earlier than expected?

Issue #3321, opened 2022-06-10
Proposal: Add a mode where all actual file system interactions are delayed until the build completes 


jakemac53:
# Problem

Today we eagerly delete all invalidated outputs, and then incrementally re-output them during the build.

This causes problems for the analyzer (or any other tool watching the file system), because of the large number of file updates over a long period of time. It will end up re-analyzing the same files multiple times as their dependencies are written, and it also causes spurious errors when the build starts, until it completes.

This leads to a bad IDE experience for users, especially in large projects with long builds.

# Solution

The general idea is to delay all file system interactions until a build is complete, and then do the minimal amount of file system interactions possible. If a file was invalidated but re-written to contain the same content, we shouldn't touch it at all. If it changed, we should overwrite directly the existing file, etc.
 
# High level design

Add a new lifecycle method to `RunnerAssetWriter`, `onBuildComplete` or similar, which is called when a build finishes. Most implementations will just delegate the call to any other writer they wrap, or possibly call super if they extend another writer. 

Add a new `DelayedAssetWriter` which only records actions. It wraps another writer, but does not delegate calls to it. This will capture both the pending deletes as well as pending writes as they happen. It will need to cache the bytes of the writes. It will implement `onBuildComplete` and perform all the actual file system interactions. This class will also need to take an AssetReader in its constructor, if we want it to be able to compare digests and be smart about not doing unnecessary writes.

We will also need a new `AssetReader` implementation, which has a field which is this `DelayedAssetWriter`, and it will need to use that cache to read files from. Some more design will need to be done to figure out exactly the best place to slot that reader in.

# Risks

## Cache Size

This strategy requires all files written in a given build to fit in memory together, with an unbounded cache size. Some potential options we could consider:

- Have a bounded sized cache, and write files in batches if the cache starts getting full.
- Only use this strategy for Dart files, and not other files. Some other tools might not get the benefit, but it would still fix the analyzer problem.
- Only use this strategy for files under a certain size.

## Could be breaking for some builders

If a builder today calls out to an external process which reads from the file system, it won't see updated files

- These builders are already encouraged to use `package:scratch_space`, which won't have this problem. All our own builders do this.

jakemac53:
cc @scheglov @devoncarew @jacob314 @natebosch

natebosch:
LGTM.

Do we already have a way that written assets are communicated through to the `AssetReader`, or are we going through the files system always today?

[...]

PR #3418, opened 2022-12-03
Add option to delay file system writes

simolus3:
This adds a `--delay-writes` option to the `build`, `watch`, `serve` and `daemon` commands. When enabled, file system writes will only happen once at the end of a build.

We can determine the impact of this option on larger builds (in terms of how the analyzer reacts, memory used by the build system, ...). If it turns out that some refinements are needed (like only doing this for Dart files or source assets), we can improve the option in subsequent releases. Once everything works, we might want to make this the default mode (at least if not using the low resource mode).

This probably needs more integration tests, but I'm not sure what the place best to add them would be.

Closes https://github.com/dart-lang/build/issues/3321


jakemac53:
Oh wait nvm, the linked issue, apparently filed by me, describes the reasoning, LOL.

build_runner/lib/src/entrypoint/base_command.dart
```diff
@@ -97,7 +97,12 @@ abstract class BuildRunnerCommand extends Command<int> {
               'If multiple filters are applied then outputs matching any filter '
               'will be built (they do not need to match all filters).')
       ..addMultiOption(enableExperimentOption,
-          help: 'A list of dart language experiments to enable.');
+          help: 'A list of dart language experiments to enable.')
+      ..addFlag(
```
jakemac53, line 101: Should we consider defaulting this to on? If we think it is better, then I think it probably makes sense to do? It isn't likely a lot of people will enable it otherwise.

simolus3: I have enabled the option by default if the low-resources option is _not_ enabled. I think it's a better default behavior, but the in-memory overlay used by the delayed writes contradicts the intention of using less memory in low-resources.

build_runner/lib/src/generate/environment.dart
```diff
@@ -0,0 +1,36 @@
+import 'package:build_runner_core/build_runner_core.dart';
+import 'package:logging/logging.dart';
+
+import '../logging/std_io_logging.dart';
+
+BuildEnvironment createEnvironment({
```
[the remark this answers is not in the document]
simolus3: I've renamed it to `createDefaultEnvironment` :+1:

build_runner_core/lib/src/asset/writer.dart
```diff
@@ -3,12 +3,123 @@
...
 
 @Deprecated('No longer used')
 typedef OnDelete = void Function(AssetId id);
 
 abstract class RunnerAssetWriter implements AssetWriter {
   Future delete(AssetId id);
+
+  /// Potentially runs pending work that needs to happen just before the build
+  /// completes, after all other interactions with this [AssetWriter] have
+  /// already been completed.
```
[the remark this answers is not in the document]
simolus3: We happen to exclusively implement the interface and use delegation to wrap other writers. I've added a `@mustCallSuper` annotation, I think that makes the intention clear if we ever end up extending existing writers.

build_runner_core/lib/src/asset/writer.dart
```diff
@@ -3,12 +3,123 @@
...
+
+/// A [RunnerAssetWriter] implementation that performs all writes at once in
+/// [onBuildComplete].
+///
+/// This write behavior may reduce load on external tools doing work on each
+/// file system change (like the analysis server).
+class DelayedAssetWriter implements RunnerAssetWriter {
+  final RunnerAssetWriter _delegate;
+
+  final Map<AssetId, List<int>?> overlay = {};
```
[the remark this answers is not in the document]
simolus3: I use `null` to express the state that an asset has been deleted (via `delete()`)  without the delayed writer being flushed (via `onBuildComplete`).

I've added a comment explaining the structure. Should I store information about deleted files separately to avoid the nullable type instead?

build_runner_core/lib/src/asset/writer.dart
```diff
@@ -3,12 +3,123 @@
...
+  /// Obtain a [RunnerAssetReader] capable of reading pending writes of this
+  /// writer (before they are persisted with [onBuildComplete]).
+  RunnerAssetReader reader(RunnerAssetReader delegate, String rootPackage) {
+    return _DelayAwareReader(delegate, this, rootPackage);
+  }
+
+  @override
+  Future delete(AssetId id) async {
+    overlay[id] = null;
+    _work.add(() => _delegate.delete(id));
```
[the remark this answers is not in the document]
simolus3: Absolutely, we can get all of the information needed to do the write out of the `overlay` field as well.

build_runner/lib/src/entrypoint/base_command.dart
jakemac53: SGTM

build_runner/lib/src/entrypoint/base_command.dart
```diff
@@ -102,6 +102,7 @@ abstract class BuildRunnerCommand extends Command<int> {
         delayWritesOption,
         help: 'Delays all file system interactions until the end of a build, '
             'potentially reducing load on tools like the analysis server.',
+        defaultsTo: null,
```
jakemac53, line 105: Interesting I didn't know null was allowed for flags :). It feels weird, but does make sense here I think. Easier than manually checking `wasParsed` just to set the value back to null haha.

build_runner_core/lib/src/asset/writer.dart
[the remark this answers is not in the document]
jakemac53: That seems fine to me, it also handles cleanly the case where a file is deleted at the start of a build, but then re-written later.

jakemac53:
cc @natebosch did you want to review this as well?

build_runner/lib/src/entrypoint/options.dart
```diff
@@ -100,7 +104,11 @@ class SharedOptions {
     required this.isReleaseBuild,
     required this.logPerformanceDir,
     required this.enableExperiments,
-  });
+    bool? delayAssetWrites,
+  }) :
+        // Delayed asset writes should be enabled by default if we're not in the
```
natebosch, line 109: How confident are we that this is the right choice for the default?

build_runner_core/lib/src/asset/writer.dart
```diff
@@ -3,12 +3,133 @@
...
+    await for (final assetId in _delegate.findAssets(glob, package: package)) {
+      if (!_delayed.overlay.containsKey(assetId)) {
+        // Overlay ids are reported afterwards
+        yield assetId;
+      }
+    }
+
+    for (final entry in _delayed.overlay.entries) {
+      if (entry.key.package == package && entry.value != null) {
+        yield entry.key;
```
natebosch, line 112: I think we need to check against the `glob` here.

example/pubspec.yaml
```diff
@@ -13,3 +13,9 @@ dev_dependencies:
   build_runner: ^2.0.0
   build_web_compilers: ^3.0.0
   lints: '>=1.0.0 <3.0.0'
+
+dependency_overrides:
+  build_runner:
+    path: ../build_runner
+  build_runner_core:
+    path: ../build_runner_core
```
natebosch, line 21: Please add a trailing newline

natebosch (approved):
I was initially worried, but I don't think it should be possible for this cache to have the same content as the CachingAssetReader - we should always invalidate an asset before it could be possible to write it, and we'd never read an asset from disk before writing it during the build.

build_runner_core/lib/src/asset/writer.dart
simolus3: Oh right, of course :facepalm: Good catch!

jakemac53:
new failure here looks potentially related

simolus3:
I think that particular failure is from the `InMemoryRunnerAssetWriter` though: When it writes a `makeAssetId('a|foo/bar.dart')` on Windows, it notifies fake watchers about a write in `D:\a\foo/bar.dart` because it just takes the path of the asset id without modification. I think `D:\a\foo\bar.dart` would be the correct path in real builds.

natebosch:
Does the test pass if we revert the last commit "Fix symlink test" and merge master? I'd prefer to avoid the `is` check.

simolus3:
Were you referring to b93d5343486874400d227692daf4d927320a85b3? Unfortunately, `createMergedOutputDirectories` needs a path-providing reader when it wants to support symlinks.

I agree that the `is` isn't great, should I move the interface of `PathProvidingAssetReader` into `RunnerAssetReader` (perhaps by encouraging throwing an `UnsupportedError` in most implementations)? That would make it a lot easier to delegate.

natebosch:
Sorry for the slow response.

Yes, I think merging the interface of `PathProvidingAssetReader` into `RunnerAssetReader` is sensible.

I'd guess few implementation won't be able to implement it, but I haven't looked closely.

mosuem:
@simolus3 is this still WIP?

Closing this for now as it is stale and the files are out of date. Feel free to reopen in case you still want to land it!

PR #3424, opened 2022-12-13
Drop support for unsound null safety in Dart 3

jakemac53:
**Note**: This will be broken until the ddc resources move https://github.com/dart-lang/sdk/issues/50700

Remove entirely the build_vm_compilers package, this should be marked discontinued on pub.

Drop all code relating to sound/unsound null safety.

Update build_web_compilers, build_vm_compilers, and build_runner to only
support 3.x sdks.

Point at the new DDC resources in the SDK based on https://github.com/dart-lang/sdk/issues/50700.


jakemac53:
cc @nshahan

jakemac53:
cc @simolus3 it looks like drift does still use build_vm_compilers. We are looking at probably discontinuing that package, would that be an issue for you, or is it more legacy? What are you using it for?

simolus3:
I don't think drift is using `build_vm_compilers` at all - I also couldn't find a dependency to it, I'll take another look tomorrow.

But I think the community tests are failing because they're running on Dart 2.18.5, which can't run `build_runner` because it now has a`@dart=3.0` language version passed down from its updated pubspec.

jakemac53:
Ah yeah I think it was just the way the tool works it would end up adding in that dependency even if it wasn't previously there. But it no longer overrides that package.

jakemac53:
Sending out for review, DDC changes should land soonish.

jakemac53:
@simolus3 it looks like you have a test which isn't opted in yet in drift (test/engines/delegated_database_test.dart). Is there a preferred solution on your end for getting our bots here green? I think we could just skip it (remove it from the preset maybe), or it could be migrated, or open to other suggestions also.

build_modules/pubspec.yaml
```diff
@@ -35,3 +35,7 @@ dev_dependencies:
     path: test/fixtures/a
   b:
     path: test/fixtures/b
+
+dependency_overrides:
+  build_runner:
+    path: ../build_runner
```
natebosch, line 41: [nit] please add trailing newline

build_runner/pubspec.yaml
```diff
@@ -49,11 +49,17 @@ dependencies:
...
   test_descriptor: ^2.0.0
   test_process: ^2.0.0
   _test_common:
     path: ../_test_common
+
+dependency_overrides:
+  build_modules:
+    path: ../build_modules
+  build_web_compilers:
+    path: ../build_web_compilers
```
natebosch, line 65: [nit] add a trailing newline

build_web_compilers/pubspec.yaml
```diff
@@ -33,3 +33,7 @@ dev_dependencies:
     path: test/fixtures/a
   b:
     path: test/fixtures/b
+
+dependency_overrides:
+  build_modules:
+    path: ../build_modules
```
natebosch, line 39: [nit] add a trailing newline

natebosch (approved):

simolus3:
> there a preferred solution on your end for getting our bots here green?

Sorry for the delayed response, I've missed the comment. As far as I'm concerned, we can temporarily skip the community tests. I will test drift with the latest build runner on a 3.0 dev SDK as soon as this PR is merged.  As soon as everything works (it may very well be that drift tests fail in 3.0 for reasons unrelated to `build`), I can follow up with a PR to re-enable the tests.

I have migrated all tests to null-safety now, but not on the branch used by the community tests. You could try changing the `ref` to `develop` here, but I wouldn't waste any time on it if that doesn't fix the problem. I can take the time later to test with a dev SDK and follow up with a fix.

https://github.com/dart-lang/build/blob/bbbe12c61bebadb1e4060d767ef818ca406893e7/.github/workflows/community.yml#L21

jakemac53:
@simolus3 thanks, I tried that branch but got errors in the dependency solve, so just commenting out the community jobs for now, happy to restore them once they are working again

PR #3426, opened 2022-12-15
Check packageGraph when validating build definition

BenVercammen:
In my setup, I've got something like this:

- `my_codegen` package
  - contains generator classes
- `my_core` package
  - `build.yaml` containing builders from `my_codegen` package, used for generating core code
  - `pubspec.yaml` 
    - has a "dev_dependency" on `my_codegen`
- `my_other_package`
  - `pubspec.yaml` 
    - has a dependency on `my_core`
    - has a dependency on `isar_generator`

Now the problem is that when I run `dart run build_runner build` in `my_other_package` (for generating `isar` code), I get the following logs:
```console
Building package executable...
Built build_runner:build_runner.
[INFO] Entrypoint:Generating build script...
Unhandled exception:
PackageNotFoundException: my_codegen
``` 

Debugging the `_generateBuildScript()` method, I couldn't figure out why the `my_codegen` build configuration is being included. I also tried to add a `build.yaml` file in `my_other_package` containing `enabled: false` for all previously defined builders, but to no avail.

At first I added `on PackageNotFoundException` to `_allMigratedToNullSafety()`, but it still gave issues down the road.
So eventually, the solution to my problem was to "invalidate" the build definition, because the imported package wasn't present in the `packageGraph`. I'm not totally satisfied with this solution, but right now it does the trick for me. And it might just be a nice extra validation step anyway.

Still, if there is some obvious configuration thing I'm missing out on, I'm more than happy to hear this.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.3.4-dev
+
+- Warn if a `package:` builder import cannot be resolved and skip it,
+  instead of creating an invalid build script or failing in other obscure ways.
+
 ## 2.3.3
 
 - Remove references to `NullThrownError`.
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -127,7 +127,8 @@ Future<BuildScriptInfo> findBuildScriptOptions({
       if (packageGraph?.allPackages.containsKey(pkg) ?? true) {
         return true;
       }
-      _log.warning('Could not load imported package "$pkg" for definition "${definition.key}".');
+      _log.warning(
+          'Could not load imported package "$pkg" for definition "${definition.key}".');
       return false;
     }
     return definition.package == packageGraph!.root.name;
--- a/build_runner/lib/src/build_script_generate/build_script_generate.dart
+++ b/build_runner/lib/src/build_script_generate/build_script_generate.dart
@@ -120,7 +120,16 @@ Future<BuildScriptInfo> findBuildScriptOptions({
   bool _isValidDefinition(dynamic definition) {
     // Filter out builderDefinitions with relative imports that aren't
     // from the root package, because they will never work.
-    if (definition.import.startsWith('package:') as bool) return true;
+    if (definition.import.startsWith('package:') as bool) {
+      // Make sure package is known in packageGraph (when present),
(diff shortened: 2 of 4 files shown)


BenVercammen:
Yes, the `build.yml` file in `my_core` looks like this:
```yml
targets:
  $default:
    builders:
      build_test|test_bootstrap:
        enabled: false
    auto_apply_builders: false

builders:
  my_code_builder:
    import: "package:my_codegen/my_codegen.dart"
    builder_factories: ["myCodeBuilder"]
    build_extensions: {".dart": [".h.dart"]}
    auto_apply: none
    build_to: source
  ...
```

There are actually 4 builder definitions in there, and all 4 of them get picked up.

jakemac53:
Yeah it makes sense to leave it as a dev dependency,  as long as you _only_ want that builder definition to be usable from the `my_core` package itself. So it is essentially a "private" builder.

I do wonder a bit if we should be exposing that more directly as a feature (builder definitions which are ignored entirely outside of the defining packages build). Then you could solve it by just adding `private: true` or something instead.

BenVercammen:
Well, the thing is that I do use (some of) those builders in other packages that depend on `my_core`. So I wouldn't go as far as making them "private". (but I see you just figured that out by looking at the `build.yaml` more closely)

I might be overlooking some basic configuration thing, because something else I'm doing (and I don't really like) is adding `one_of_my_generators.build.yaml` files with the following content:
```yaml
---
# disable all inherited builders...
targets:
  $default:
    auto_apply_builders: false
```

This also works to keep the builder from being applied in `my_other_package`'s `build.dart` file, but the `PackageNotFoundException` is still being thrown as long as it's a `dev_dependency`.

I'll probably have to move the `build.yml` definition file back the `my_codegen` package, can't remember why I put it in `my_core` in the first place... I'll see if that works and will let you know.

BenVercammen:
Okay, I've moved the `build.yaml` file back to the `my_codegen` package, and everything seems to be working just fine.

So I guess the lesson learned here is to keep the `builders:` definitions in `build.yaml` in the same package that contains the actual builder. 

Still, the additional check might be useful, with some informative logging in place?

BenVercammen:
I've added a test case in `build_script_generate_test.dart`.

jakemac53 (approved):
LGTM, I pushed a changelog/pubspec update also

jakemac53:
I went ahead and ran the formatter and pushed the change

PR #3427, opened 2022-12-16
prep to publish build_runner

jakemac53:
This contains the fix to stop referencing `NullThrownError`, so that we are compatible with Dart 3.0


vsmenon (approved):
lgtm

PR #3445, opened 2023-01-27
Enable drift community test again

simolus3:
With the additional dependency overrides, tests in drift should hopefully pass with Dart 3 and the latest `build` packages.


jakemac53 (approved):
We can go ahead and merge this on red if we need to - and address the unrelated failures separately

PR #3446, opened 2023-01-27
fix test for reading from the sdk

jakemac53:
This now checks if it can read `bin/dart` which is far less likely to move in the future :)


natebosch (approved):

jakemac53:
ah crap that file has a different name on windows I guess lol, will search for a different file, maybe `lib/libraries.json`?

PR #3367, opened 2022-09-12
Update pub.dev README to include Build-Filters

chillbrodev:
This feature was not displayed in the official README featured on pub.dev. It was hidden in the CHANGELOG. I essentially copied the CHANGELOG info and moved it to the README.

By including this information on the main README it will allow other package users to understand how to configure this package further.

PR #3451, opened 2023-02-01
mention the --build-filters flag in the command line options section

jakemac53:
Replacement PR for https://github.com/dart-lang/build/pull/3367. We now just link the partial builds doc which explains the feature more thoroughly.

### DIFF
--- a/build_runner/README.md
+++ b/build_runner/README.md
@@ -87,7 +87,7 @@ All the above commands support the following arguments:
   logged. By default this is false.
 - `--build-filter`: Build filters allow you to choose explicitly which files to
   build instead of building entire directories. See further documentation on
-  this feature [here](../docs/partial_builds.md).
+  this feature [here](#partial_builds).
 
 Some commands also have additional options:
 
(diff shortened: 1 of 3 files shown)


build_runner/README.md
```diff
@@ -85,6 +85,9 @@ All the above commands support the following arguments:
   be provided.
 - `--[no-]fail-on-severe`: Whether to consider the build a failure on an error
   logged. By default this is false.
+- `--build-filter`: Build filters allow you to choose explicitly which files to
+  build instead of building entire directories. See further documentation on
+  this feature [here](../docs/partial_builds.md).
```
natebosch, line 90: Will this link work when the readme is on pub?

natebosch (approved):

build_runner/README.md
jakemac53: Huh, good question. cc @jonasfj

natebosch: We might want to use an absolute URL here to be safe.

jakemac53: Yeah ok I will just make it absolute, the relative url works better with IDEs (navigates to the actual file in your IDE), but the absolute uri still takes you to a good place.

PR #3458, opened 2023-02-16
dont pass an explicit --native-null-assertion option if none was provided

jakemac53:
Related to https://github.com/dart-lang/sdk/issues/50710, this will allow the optimization level in dart2js to control the default setting for this flag. It can still be overridden with explicit config.

### DIFF
--- a/build_web_compilers/lib/src/dart2js_bootstrap.dart
+++ b/build_web_compilers/lib/src/dart2js_bootstrap.dart
@@ -24,7 +24,7 @@ import 'web_entrypoint_builder.dart';
 Future<void> bootstrapDart2Js(
   BuildStep buildStep,
   List<String> dart2JsArgs, {
-  required bool nativeNullAssertions,
+  required bool? nativeNullAssertions,
 }) =>
     _resourcePool.withResource(() => _bootstrapDart2Js(buildStep, dart2JsArgs,
         nativeNullAssertions: nativeNullAssertions));
@@ -32,7 +32,7 @@ Future<void> bootstrapDart2Js(
 Future<void> _bootstrapDart2Js(
   BuildStep buildStep,
   List<String> dart2JsArgs, {
-  required bool nativeNullAssertions,
+  required bool? nativeNullAssertions,
 }) async {
   var dartEntrypointId = buildStep.inputId;
   var moduleId =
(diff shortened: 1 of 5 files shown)


build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -193,12 +193,15 @@ String _appBootstrap({
   required String moduleScope,
   required String entrypointLibraryName,
   required String oldModuleScope,
-  required bool nativeNullAssertions,
-}) =>
-    '''
+  required bool? nativeNullAssertions,
+}) {
+  var nativeAssertsCode = nativeNullAssertions == null
```
jakemac53, line 198: cc @nshahan for this part in particular, I presume it is fine to simply omit this line?

srujzs: It looks like we default to false in DDC (we still check in sound mode though) if we never call the `nativeNonNullAsserts` function, which is consistent with dart2js, so this looks good.

srujzs (approved):

build_web_compilers/lib/src/dart2js_bootstrap.dart
```diff
@@ -24,15 +24,15 @@ import 'web_entrypoint_builder.dart';
 Future<void> bootstrapDart2Js(
   BuildStep buildStep,
   List<String> dart2JsArgs, {
-  required bool nativeNullAssertions,
+  required bool? nativeNullAssertions,
```
natebosch, line 27: [optional] Doesn't need to be required anymore

natebosch (approved):

build_web_compilers/lib/src/dart2js_bootstrap.dart
jakemac53: I like making it required to be explicit even though it's nullable

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
jakemac53: Ok, that does sound like a behavior change (we defaulted it on before), but regardless it seems best to leave this up to the compilers.

PR #3463, opened 2023-02-20
Don't mixin non-mixins.

lrhn:
The comparable interface is not a mixin and is not intended as a mixin. It has no implementation to inherit.

Also, it won't work in Dart 3.0.

Broke when we enabled class modifiers in platform libraries.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 3.1.1-dev
 
 - Report file watching errors and stop the daemon.
-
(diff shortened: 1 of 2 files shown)


kevmoo:
@jakemac53 – we should probably get this ready to roll as a "dot" release, right?

lrhn:
We need this to get into our internal repository before we can roll the SDK in there with experiments enabled. Preferably ASAP.

If the SDK or other internal tools also depend on this package, we probably need to update it there too.

devoncarew:
Ah, I didn't realize Jake was OOO this week.

devoncarew (approved):

Issue #3466, opened 2023-02-23
Can't use enable-experiment with web projects


jodinathan:
We are testing `records` and `patterns` in our web app (AngularDart) using `dart run build_runner serve --enable-experiment="records,patterns"` and it works when you use the new features in the project but claims that `records` isn't enabled for files outside the project that use those features.

Digging the source code it only worked when we added `enabledExperiments: ['records', 'patterns']` to [FrontendServerClient.start](https://github.com/dart-lang/build/blob/1faecee3b71e54c8ec3b11dc46a2d0ec374aa1ff/build_runner/lib/src/build_script_generate/bootstrap.dart#L156)

I guess this is a bug?

Dart SDK version: 3.0.0-263.0.dev (dev) (Tue Feb 21 17:59:38 2023 -0800) on "macos_x64"

jakemac53:
Ah so this is for the actual build script it looks like? What if you do `dart run --enable-experiment=records,patterns build_runner build`?

jodinathan:
nope

```log
[INFO] Entrypoint:Generating build script...
[INFO] Entrypoint:Generating build script completed, took 790ms

[INFO] Bootstrap:Precompiling build script......
[WARNING] Bootstrap:
../../../oni-group/server/oni_server/lib/src/storage/storage.dart:15:46: Error: This requires the experimental 'records' language feature to be enabled.
Try passing the '--enable-experiment=records' command line option.
      Uint8List bytes) contentTypeFinder) => (
```

jakemac53:
You can try https://github.com/dart-lang/build/pull/3467, but I am pretty confident it works so you could also just wait

PR #3467, opened 2023-02-28
Support experiments in the build script

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3466

Also preps for publishing build_runner

### DIFF
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -95,8 +95,7 @@ Future<void> main(List<String> args) async {
   if (localCommandNames.contains(commandName)) {
     exitCode = await commandRunner.runCommand(parsedArgs) ?? 1;
   } else {
-    var experiments =
-        parsedArgs.command!['enable-experiment'] as List<String>?;
+    var experiments = parsedArgs.command!['enable-experiment'] as List<String>?;
     while ((exitCode = await generateAndRun(args, experiments: experiments)) ==
         ExitCode.tempFail.code) {}
   }
--- a/build_runner/bin/build_runner.dart
+++ b/build_runner/bin/build_runner.dart
@@ -96,7 +96,7 @@ Future<void> main(List<String> args) async {
     exitCode = await commandRunner.runCommand(parsedArgs) ?? 1;
   } else {
     var experiments =
-        (parsedArgs.command!)['enable-experiment'] as List<String>?;
+        parsedArgs.command!['enable-experiment'] as List<String>?;
     while ((exitCode = await generateAndRun(args, experiments: experiments)) ==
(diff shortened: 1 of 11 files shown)


build_runner/bin/build_runner.dart
```diff
@@ -95,7 +95,10 @@ Future<void> main(List<String> args) async {
   if (localCommandNames.contains(commandName)) {
     exitCode = await commandRunner.runCommand(parsedArgs) ?? 1;
   } else {
-    while ((exitCode = await generateAndRun(args)) == ExitCode.tempFail.code) {}
+    var experiments =
+        (parsedArgs.command!)['enable-experiment'] as List<String>?;
```
natebosch, line 99: Are these parens needed?
```suggestion
        parsedArgs.command!['enable-experiment'] as List<String>?;
```

build_runner/test/build_script_generate/experiments_test.dart
```diff
@@ -0,0 +1,34 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+@Tags(['experiments'])
+@TestOn('vm')
+import 'package:build_runner/build_script_generate.dart';
+import 'package:logging/logging.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('build scripts can use experiments', () async {
```
natebosch, line 12: yet another good reason for us to have some kind of fake experiment...

natebosch (approved):

build_runner/test/build_script_generate/experiments_test.dart
jakemac53: yeah it would be better, but at least pinning to a specific SDK here should make this be valid for quite some time.

jakemac53:
@natebosch see some changes I had to make, we now fail earlier if you provide an invalid experiment as a result of this change. Unfortunately we don't get any information about the failure, so its a worse experience as well.

jakemac53:
Running into lots of weird failures on the daemon tests.... I don't see how this change could have affected it but it seems pretty reliably broken :(. I will have to circle back to this later when I have more time to debug it.

natebosch (approved):

jakemac53:
It looks like these tests are also timing out on master, not sure if that makes me more or less sad 🤣

natebosch (approved):

jakemac53:
Ok, tests should pass now since I finally got the other failures resolved. Will merge this internally then release if all is good.

jakemac53:
> Will merge this internally then release if all is good.

I forgot this package doesn't have to be rolled internally, will wait until monday to publish though to avoid publishing friday afternoon.

PR #3470, opened 2023-03-10
Build daemon ChangeProvider cleanup

jakemac53:
Removes all methods from ChangeProvider, and extract only the relevant ones into separate subclasses which are now public (AutoChangeProvider vs ManualChangeProvider).

Previously these modes were conflated and the different impls provided essentially no-op implementations of the non-applicable apis.

This should fix the failures we are seeing on the bots, because the daemon was shutting down early due to an empty stream given by the manual change provider. This was broken in https://github.com/dart-lang/build/pull/3411, but we only recently published that change, and weren't using dependency overrides for the tests that were actually broken, so we only saw it after the publish.

### DIFF
--- a/build_daemon/lib/daemon.dart
+++ b/build_daemon/lib/daemon.dart
@@ -101,8 +101,10 @@ class Daemon {
     if (!_doneCompleter.isCompleted) _doneCompleter.complete(exitCode);
   }
 
-  void _createPortFile(int port) =>
-      File(portFilePath(_workingDirectory)).writeAsStringSync('$port');
+  void _createPortFile(int port) {
+    var path = portFilePath(_workingDirectory);
+    File(path).writeAsStringSync('$port');
+  }
 
   void _createVersionFile() => File(versionFilePath(_workingDirectory))
       .writeAsStringSync(currentVersion);
--- a/build_daemon/lib/daemon.dart
+++ b/build_daemon/lib/daemon.dart
@@ -101,10 +101,8 @@ class Daemon {
     if (!_doneCompleter.isCompleted) _doneCompleter.complete(exitCode);
   }
 
-  void _createPortFile(int port) {
-    var path = portFilePath(_workingDirectory);
-    File(path).writeAsStringSync('$port');
-  }
+  void _createPortFile(int port) =>
+      File(portFilePath(_workingDirectory)).writeAsStringSync('$port');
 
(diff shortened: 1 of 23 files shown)


jakemac53:
Note that there will be some minor breakage in the internal roll, should be easy to update atomically though with the roll CL.

jakemac53:
Also open to discussing whether we should explore a different, non-breaking change here. But I don't think there are a lot of clients of this package so the breaking change is likely ok?

build_runner/lib/src/daemon/change_providers.dart
```diff
@@ -11,33 +13,25 @@ import 'package:watcher/watcher.dart' show WatchEvent;
 
 /// Continually updates the [changes] stream as watch events are seen on the
 /// input stream.
-///
-/// The [collectChanges] method is a no-op for this implementation.
-class AutoChangeProvider implements ChangeProvider {
+class AutoChangeProviderImpl implements AutoChangeProvider {
```
natebosch, line 16: Ugh, I wish I had a better suggestion than "Impl" 😢

natebosch (approved):

build_daemon/lib/daemon.dart
```diff
@@ -101,8 +101,10 @@ class Daemon {
     if (!_doneCompleter.isCompleted) _doneCompleter.complete(exitCode);
   }
 
-  void _createPortFile(int port) =>
-      File(portFilePath(_workingDirectory)).writeAsStringSync('$port');
+  void _createPortFile(int port) {
+    var path = portFilePath(_workingDirectory);
+    File(path).writeAsStringSync('$port');
```
grouma, line 106: Seems like an unnecessary change?

grouma (approved):

build_daemon/lib/daemon.dart
jakemac53: ah yeah I had extracted it to a variable for debugging, but don't need it any more

grouma:
> Also open to discussing whether we should explore a different, non-breaking change here. But I don't think there are a lot of clients of this package so the breaking change is likely ok?

I assume you will handle pulling this into Google3? There's only one internal client which is used by a couple tools.

jakemac53:
> I assume you will handle pulling this into Google3? There's only one internal client which is used by a couple tools.

Yes, its just the one client I could see that will need an update but it is very minor

PR #3472, opened 2023-03-13
release build_runner and build_daemon

jakemac53:


.github/workflows/publish.yaml
```diff
@@ -5,6 +5,7 @@ name: Publish
 on:
   pull_request:
     branches: [ master ]
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
natebosch, line 8: Can you mention the reason for this change in the commit message.

natebosch (approved):

.github/workflows/publish.yaml
jakemac53: done

PR #3474, opened 2023-03-20
Bump the _test SDK dep

natebosch:
The SDK is not reporting API use from newer SDKs than the lower bound. We don't test on an SDK that old anymore, so there is no problem bumping the minimum bound.

### DIFF
--- a/_test/pubspec.yaml
+++ b/_test/pubspec.yaml
@@ -2,7 +2,7 @@ name: _test
 publish_to: none
 
 environment:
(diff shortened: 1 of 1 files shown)


jakemac53 (approved):

bkonyi (approved):
Is it worth bumping the upper bound to 4.0.0 at this point?

natebosch:
> Is it worth bumping the upper bound to 4.0.0 at this point?

I don't think we've started doing that on any packages yet.

PR #3476, opened 2023-03-27
Fix the build

jakemac53:
## build_daemon

We are using `unawaited` which wasn't made available until 2.15, but we can't get a valid solve until `2.18` so I am going to require that.

This also regenerates some built_value code so that it won't trigger lints.

## build_modules

Delete the old dart2js  worker code which has been deprecated for a long time. This is technically breaking but in practice should not be, and I haven't marked it as a breaking change. Feel free to push back.


jakemac53:
cc @simolus3 regarding the drift failure, looks like just a general change in built output due to a later version of mockito. Will ignore it here.

build_daemon/pubspec.yaml
```diff
@@ -1,10 +1,10 @@
 name: build_daemon
-version:  4.0.0
+version:  4.0.1-dev
 description: A daemon for running Dart builds.
 repository: https://github.com/dart-lang/build/tree/master/build_daemon
 
 environment:
-  sdk: ">=2.14.0 <3.0.0"
```
natebosch, line 7: Should we consider removing the use of `unawaited`, publishing, then adding it bak with the constraint bump?

There could be folks with a 2.14.0 SDK who are broken right?

kevmoo: If we don't get reports of an issue, I'd say leave it. All of our stats show folks move pretty quickly to our latest SDKs.

jakemac53: Realistically they couldn't get a version solve under 2.17 anyways, so it doesn't matter.

jakemac53:
The scope of this expanded a bit, PTAL. In particular the choice to delete the deprecated apis from build_modules without a breaking change is worth reviewing. Note that this is not risky for our own packages any longer, as they haven't used this API for quite some time. It could only break some other usage. If we did break somebody we could always restore it and then publish it as breaking, but I think that is extremely unlikely.

PR #3495, opened 2023-05-05
Add package config to build step

### LINKED ISSUE #3492: API to resolve language version of a package
I have a builder that generates Dart files without having Dart sources as its input. These generated sources should respect the language version of the package for which they're generated (for instance, I can generate record types if a user depends on `>=3.0.0-0` and fall back to classes otherwise).
When there is a primary Dart input file, we can use the analyzer API to resolve its language version. When code is generated for other assets though, there is no easy way to determine a package's language version. (Using build APIs to read a pubspec works, but requires builder authors to replicate a lot of logic already handled by first-party Dart tools. `Isolate.packageConfig` also works, but might break hermetic builds)

I would like to have a method on `Resolver` that returns the language version of a package, e.g. 

```dart
  /// Returns the default Dart language version for sources in the given
  /// [package] that don't have an explicit [LanguageVersionToken] override.
  ///
  /// Throws a [PackageNotFoundException] if the package is not part of the
  /// build.
  Future<Version> languageVersionForPackage(String package);
```

I'm happy to help implement this of course (and it seems like we have all the necessary information in `build_resolvers`), but it would be a breaking the `Resolver` interface and it doesn't seem like this is worth a build 3? Previously the `build_resolvers` package would depend on minor versions of the `build` package, but that seems to have changed. So if this API proposal is accepted, I could use some advice on how to do it with minimum breakage (maybe put it on `BuildStep` instead?).

simolus3:
This adds a `packageConfig` getter to `BuildStep`, which can be used to read the language version of packages involved in the build. The package config exposed in the build step hides the actual files of packages on the file system. Instead, 

By default, the package config will be resolved from the current isolate in `runBuilder`. However, `build_runner_core` also constructs a config based on the `PackageGraph` which avoids loading the raw package config multiple times.

Closes #3492.

### DIFF
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -80,8 +80,8 @@ class BuildStepImpl implements BuildStep {
 
   @override
   Future<PackageConfig> get packageConfig async {
-    final resolved = _resolvedPackageConfig ??=
-        Result.capture(Future.sync(_resolvePackageConfig));
+    final resolved =
+        _resolvedPackageConfig ??= Result.capture(_resolvePackageConfig());
 
     return (await resolved).asFuture;
   }
--- a/build/lib/src/builder/build_step_impl.dart
+++ b/build/lib/src/builder/build_step_impl.dart
@@ -10,6 +10,7 @@ import 'package:analyzer/dart/element/element.dart';
 import 'package:async/async.dart';
 import 'package:crypto/crypto.dart';
 import 'package:glob/glob.dart';
+import 'package:package_config/package_config_types.dart';
 
 import '../analyzer/resolver.dart';
 import '../asset/exceptions.dart';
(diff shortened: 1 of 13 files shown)


build/lib/src/generate/run_builder.dart
```diff
@@ -62,3 +93,26 @@ Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
...
 }
+
+extension on Package {
+  Package transformToAssetUris() {
+    return Package(
+      name,
+      Uri(scheme: 'asset', pathSegments: [name, '']),
+      packageUriRoot: Uri(scheme: 'asset', pathSegments: [name, 'lib', '']),
+      relativeRoot: false,
+      extraData: extraData,
```
jakemac53, line 104:
```suggestion
      languageVersion: languageVersion,
```

build/lib/src/generate/run_builder.dart
```diff
@@ -62,3 +93,26 @@ Future<void> runBuilder(Builder builder, Iterable<AssetId> inputs,
...
   }
 }
+
+extension on Package {
+  Package transformToAssetUris() {
+    return Package(
+      name,
+      Uri(scheme: 'asset', pathSegments: [name, '']),
+      packageUriRoot: Uri(scheme: 'asset', pathSegments: [name, 'lib', '']),
+      relativeRoot: false,
```
jakemac53, line 103: This could be a bit more general?

```suggestion
      packageUriRoot: packageUriRoot,
      relativeRoot: relativeRoot,
```

build/lib/src/builder/build_step_impl.dart
```diff
@@ -60,14 +61,31 @@ class BuildStepImpl implements BuildStep {
...
       {StageTracker? stageTracker,
       void Function(Iterable<AssetId>)? reportUnusedAssets})
       : allowedOutputs = UnmodifiableSetView(expectedOutputs.toSet()),
         _stageTracker = stageTracker ?? NoOpStageTracker.instance,
         _reportUnusedAssets = reportUnusedAssets;
 
+  @override
+  Future<PackageConfig> get packageConfig async {
+    final resolved = _resolvedPackageConfig ??=
+        Result.capture(Future.sync(_resolvePackageConfig));
```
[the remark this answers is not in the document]
simolus3: It would guard against synchronous exceptions thrown by the callback, but that's not relevant here since we're passing an `async` function in `runBuilder`. I'll just invoke it directly.

build/lib/src/generate/run_builder.dart
simolus3: Ouch, missed the most important piece of information :D I've also expanded the test to ensure we get the language version.

Should we still remove the `extraData` field?

jakemac53: Oh sorry I didn't intend to suggest removing that, just a goof up

jakemac53 (approved):

PR #3500, opened 2023-05-09
Update pubspec.yaml

devoncarew:
- add topics to the pubspec

---

#### Contribution guidelines:

- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/wiki/External-Package-Maintenance#making-a-change).

Many Dart repos have a weekly cadence for reviewing PRs - please allow for a week or two of latency for initial review feedback.

---

### DIFF
--- a/scratch_space/pubspec.yaml
+++ b/scratch_space/pubspec.yaml
@@ -1,7 +1,10 @@
 name: scratch_space
-version: 1.0.2-dev
-description: A tool to manage running external executables within package:build
+version: 1.0.2
+description: A tool to manage running external executables within package:build.
 repository: https://github.com/dart-lang/build/tree/master/scratch_space
+
+topics:
+  - build-runner
 
 environment:
   sdk: ">=2.19.0 <3.0.0"


devoncarew:
The failure is in the community tests - package:drift related - and is due to a recent analyzer publish, and a visitor no longer implementing all the required methods.

scratch_space/pubspec.yaml
```diff
@@ -1,8 +1,11 @@
 name: scratch_space
-version: 1.0.2-dev
-description: A tool to manage running external executables within package:build
+version: 1.0.2
+description: A tool to manage running external executables within package:build.
 repository: https://github.com/dart-lang/build/tree/master/scratch_space
 
+topics:
+  - build-runner
```
jakemac53, line 7: I think we have used "code generation" for some other packages, so it might be worth having a bit more general topic here also.

It also feels like maybe this should be `build_runner` since its the name of the package?

parlough: Topics are supposed to only contain dashes or alphanumeric characters:

https://github.com/dart-lang/site-www/pull/4807/files#diff-96d69c1c40466944f504bd8a24f1f4741020e348d7a360115acaeaa3039708b7R436-R437

simolus3:
The drift failure related to the analyzer has since been fixed, the tests should pass after restarting them.

scratch_space/pubspec.yaml
devoncarew: This package is used by codegen packages, but is it itself codegen related?

jakemac53: > Topics are supposed to only contain dashes or alphanumeric characters:

Does that imply we shouldn't be using other packages as topics, given they use underscores?

> This package is used by codegen packages, but is it itself codegen related?

I am not really sure how to draw the line here... yes this code does not itself generate code but it is intended for use in code generators.

devoncarew: Hmm, well, this is something we can always change w/ a commit and a publish. I added `'codegen'` to the list of package topics (in addition to the existing `build-runner`) w/ the assumption that this package is often associated w/ those two topics.

jakemac53 (approved):

jakemac53:
Going to go ahead and merge/publish on red - the failures are a separate issue related to the latest package:graphs release

PR #3507, opened 2023-05-12
Use topologicalSort with a secondary sort

natebosch:
When sorting the builder definitions in topological order based on their
builder dependency conversation, also apply a secondary sort based on
the alpha sort of their keys. This turns what would be an arbitrary and
potentially changing ordering into an arbitrary, but consistent
ordering.

See discussion at https://github.com/dart-lang/graphs/issues/86

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.4.4-wip
 
-- Use a stable order for builders without and order defined by dependencies.
+- Use a stable order for builders without an order defined by dependencies.
 
 ## 2.4.3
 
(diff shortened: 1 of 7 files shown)


build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.4.4-wip
+
+- Use a stable order for builders without and order defined by dependencies.
```
jakemac53, line 3:
```suggestion
- Use a stable order for builders without an order defined by dependencies.
```

jakemac53 (approved):
Lets go ahead and publish this too? Does it fix some of the known issues people have had?

jakemac53 (approved):
Lets go ahead and publish this too? Does it fix some of the known issues people have had?

natebosch:
> Does it fix some of the known issues people have had?

That's hard to say - it probably has an equal chance of fixing issues, or causing entirely new ones.
What it will do is nail down the ordering so that this is the last time this can come up - and ones folks fix ordering issues with this release we can be confident they won't have to do it again.

PR #3508, opened 2023-05-15
blast_repo fixes

devoncarew:
dependabot

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,7 +3,9 @@
 version: 2
 
 updates:
-  - package-ecosystem: "github-actions"
-    directory: "/"
+  - package-ecosystem: github-actions
+    directory: /
     schedule:
-      interval: "monthly"
+      interval: monthly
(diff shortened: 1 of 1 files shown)


jakemac53:
I assume if this fails (for most updates it probably will since dependabot and mono_repo aren't friends), then it just won't submit?

devoncarew:
Yeah, it won't autosubmit until the checks have run and passed.

> for most updates it probably will since dependabot and mono_repo aren't friends

This would be good to address (I expect w/ some work on the mono_repo side) as otherwise the mono_repo maintained repos will need manual work w/ each dep bump.

jakemac53:
> This would be good to address (I expect w/ some work on the mono_repo side) as otherwise the mono_repo maintained repos will need manual work w/ each dep bump.

yeah we definitely do want to address it, I can't remember where we landed on this last, cc @kevmoo

jakemac53 (approved):

Issue #3519, opened 2023-05-24
Can't run build_runner because BuildForInputLogger


water-user:
I'm using objectbox package, i'm trying to run build_runner and i'm getting error:
```
../../../../AppData/Local/Pub/Cache/hosted/pub.dev/build_runner_core-7.2.7/lib/src/logging/build_for_input_logger.dart:13:7: Error: The non-abstract class 'BuildForInputLogger' is missing implementations for these members:
 - Logger.onLevelChanged
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

class BuildForInputLogger implements Logger {
      ^^^^^^^^^^^^^^^^^^^
../../../../AppData/Local/Pub/Cache/hosted/pub.dev/logging-1.2.0/lib/src/logger.dart:162:22: Context: 'Logger.onLevelChanged' is defined here.
  Stream<Level?> get onLevelChanged {
                     ^^^^^^^^^^^^^^
pub finished with exit code 1
PS C:\Users\Daniel\AndroidStudioProjects\water-io-flutter\projects\water-io-hydration-flutter>
```
In order to route, prioritize, and act on this as soon as possible please include:

* Dart SDK Version: **Dart SDK version: 2.19.2 (stable) (Tue Feb 7 18:37:17 2023 +0000) on "windows_x64"**
* What package(s) from this repo you are using: **build_runner 2.3.3**
* I'm running `flutter pub run build_runner build` to generate .g.dart for objectbox package. 
* Windows
* Chrome

In other computers of my team it's working without any problem, only in my computer it's getting this error. how can i fix it? Thanks!

jakemac53:
Please run a `dart pub upgrade`, that should give you a newer version of build_runner_core with the fix

fehmidevoteam:
I have the same error 
pub-cache/hosted/pub.dev/build_runner_core-7.2.7/lib/src/logging/build_for_input_logger.dart:13:7: Error: The non-abstract class 'BuildForInputLogger' is missing implementations for these members:
 - Logger.onLevelChanged

fehmidevoteam:
@jakemac53 actually we couldn't migrate to 3.10.1. I don't think that is related to flutter version. Because it was fine before. and we are also using a specific versions of so i didn't think it is a compatibility versions problem:
freezed_annotation: 2.1.0
freezed: 2.1.0+1
build_runner: 2.2.1

[...]

PR #3521, opened 2023-05-24
Build runner core backport

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3519

Currently, users on the previous flutter SDK (3.7, <6 months old), can't get the latest package:build_runner_core but can get the latest package:logging. This causes compile errors due to a member not being implemented.

This backports the fix to build_runner_core 7.2.7+1, which should be resolvable with build_runner version 2.3.3, which was the last version to allow a Dart version <3.0.0 (it supports 2.14 and later).

Once this is approved, I will tag and release, and then resolve merge conflicts with master so there is a changelog entry.

### DIFF
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 7.2.7+1
+
+- Backport the fix for package:logging version 1.2.0.
+
 ## 7.2.7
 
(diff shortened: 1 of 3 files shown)


adar2378:
Thanks for the quick PR @jakemac53

natebosch (approved):

Issue #3535, opened 2023-06-23
Support passing arguments to ddc


annagrin:
Apologies if I missed it, but is there a way to pass arguments to ddc in build_web_compilers, for example, `--canary` flag? 

Related: https://github.com/dart-lang/webdev/issues/1949

natebosch:
We do not currently support passing arbitrary arguments to DDC.
Is this something that we expect end users to be able to do, or is this limited to Dart team use for testing purposes?

annagrin:
> Is this something that we expect end users to be able to do, or is this limited to Dart team use for testing purposes?

This is only for testing features under development in DDC (the testing would be done in webdev)

natebosch:
In the very short term I would override to a local copy of `build_web_compilers` and hardcode the argument in the list.
https://github.com/dart-lang/build/blob/ac65c072611ea105fe8ad9367c7938d743c82d22/build_web_compilers/lib/src/dev_compiler_builder.dart#L185

If you'd like to open a PR we could make `--canary` an explicitly supported flag similar to how we have one for `--emit-debug-symbols`. https://github.com/search?q=repo%3Adart-lang%2Fbuild+emitDebugSymbols&type=code

annagrin:
Need a longer term, will open a PR for canary flag

PR #3536, opened 2023-06-26
Add --canary option to build web compilers

annagrin:
Add `canary` option for `build_web_compilers:ddc` and `build_web_compilers:sdk_js` builders to enable canary features in DDC.

This setting is disabled by default but can be enabled by setting it to `true` globally:

```yaml
global_options:
  build_web_compilers:ddc:
    options:
      canary: true
  build_web_compilers:sdk_js:
    options:
      canary: true
```

Closes: https://github.com/dart-lang/build/issues/3535

### DIFF
--- a/build_web_compilers/web/stack_trace_mapper.dart
+++ b/build_web_compilers/web/stack_trace_mapper.dart
@@ -81,6 +81,7 @@ class LazyMapping extends Mapping {
 
   LazyMapping(this._provider);
 
+  // ignore: unreachable_from_main
   List toJson() => _bundle.toJson();
 
   @override
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.0.4-wip
+## 4.0.4
 
 - Add the `canary` option for `build_web_compilers:ddc` and
   `build_web_compilers:sdk_js`builders to enable canary features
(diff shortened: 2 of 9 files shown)


annagrin:
@natebosch @jakemac53 I had to add --canary option to ddc and sdk_js builders in web_dev_compilers, as they have to match. Is there a way to set this option in a better way - i.e for the whole set of web_dev_compilers builders?

build_web_compilers/web/stack_trace_mapper.dart
```diff
@@ -81,6 +81,7 @@ class LazyMapping extends Mapping {
 
   LazyMapping(this._provider);
 
+  // ignore: unreachable_from_main
```
annagrin, line 84: Looks like a new warning, was't sure this API is used at runtime somewhere so I hesitated to remove it:)

natebosch: Ignoring it is the correct thing to do for now. It looks to me like a bug in the lint.

https://github.com/dart-lang/linter/issues/4495

jakemac53:
> @natebosch @jakemac53 I had to add --canary option to ddc and sdk_js builders in web_dev_compilers, as they have to match. Is there a way to set this option in a better way - i.e for the whole set of web_dev_compilers builders?

You have to do them separately like you have here

jakemac53 (approved):
If you need this published soon go ahead and drop the `-wip`

annagrin:
> If you need this published soon go ahead and drop the `-wip`

Done!

Issue #3550, opened 2023-08-05
Get error on flutter pub run build_runner build --delete-conflicting-outputs


vadimlukutin:
Null check operator used on a null value
#0      PackageGraph.forPath.<anonymous closure> (package:build_runner_core/src/package_graph/package_graph.dart:110:72)
#1      MappedListIterable.elementAt (dart:_internal/iterable.dart:415:31)
#2      ListIterator.moveNext (dart:_internal/iterable.dart:344:26)
#3      List.addAll (dart:core-patch/growable_array.dart:325:17)
#4      PackageGraph.forPath (package:build_runner_core/src/package_graph/package_graph.dart:110:12)
<asynchronous suspension>
#5      main (file:///home/vadimlukutin/.pub-cache/hosted/pub.dev/build_runner-2.4.6/bin/build_runner.dart:27:30)
<asynchronous suspension>

vadimlukutin:
helped run command
> flutter gen-l10n

jakemac53:
Can you try running a `flutter pub get` and then re-running the same command? I think your package config and pubspec might be out of sync.

tayadepritam0:
hello my friend 

try adding packages first
json_annotation 
json_serializable
build_runner

click on pub get 
then open terminal and run command
flutter pub run build_runner build --delete-conflicting-outputs

PR #3552, opened 2023-08-07
add better error message when the package config is out of date

jakemac53:
See https://github.com/dart-lang/build/issues/3550 - previously we would just assert that all dependencies of all packages were in the package config. This will give a better error, indicating which dependency (and of which package) is missing, as well as suggesting that you fetch dependencies.

### DIFF
--- a/build_runner_core/lib/src/package_graph/package_graph.dart
+++ b/build_runner_core/lib/src/package_graph/package_graph.dart
@@ -72,13 +72,17 @@ class PackageGraph {
           'This program must be ran from the root directory of your package.');
     }
     final rootPubspec = _pubspecForPath(packagePath);
-    final rootPackageName = rootPubspec['name'] as String;
+    final rootPackageName = rootPubspec['name'] as String?;
+    if (rootPackageName == null) {
+      throw StateError('The current package has no name, please add one to the '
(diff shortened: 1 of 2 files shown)


build_runner_core/lib/src/package_graph/package_graph.dart
```diff
@@ -72,13 +72,17 @@ class PackageGraph {
           'This program must be ran from the root directory of your package.');
     }
     final rootPubspec = _pubspecForPath(packagePath);
-    final rootPackageName = rootPubspec['name'] as String;
+    final rootPackageName = rootPubspec['name'] as String?;
+    if (rootPackageName == null) {
+      throw StateError('The current package has no name, please add one to the '
```
natebosch, line 77: would it be reasonable for us to call it `$root` or something along those lines and handle this case silently?

natebosch (approved):

build_runner_core/lib/src/package_graph/package_graph.dart
natebosch: I double checked and you're correct, `dart pub get` refuses to run without a package name. Throwing a StateError SGTM.

Issue #3555, opened 2023-08-08
Consider a simpler build mode which does not do incremental builds


jakemac53:
We get a lot of complaints about build_runner not scaling to large projects. This is generally thought to be a result of how the analyzer is used in the package - basically any time a builder runs on a file it establishes a dependency on all files transitively imported by that file, which is O(N), and there is probably a fairly high constant multiplier as well. So you end up with O(N^2) complexity (where N is the # of libraries) just for the tracking of deps.

If we instead abandoned incrementality entirely, we wouldn't have this cost. You would regenerate all files every time you do a build, but I believe that overall it would be significantly faster, at least for large projects.

We could also stop computing hashes of files which does also take some time.

cc @natebosch wdyt? worth exploring or no?

natebosch:
> just does the minimal work to ensure the necessary files are loaded.

It would still need to crawl transitive imports right?

> Then generate a package with some large number of files that all import each other, and run the script on it to see how it does?

I'm worried that indexing against a generated project could be misleading. I can imagine that different types of projects might have dependency graphs and library graphs with very different characteristics which might sway performance.

I wonder if we could get any analytics about real world asset graphs, file changes, and build performance.

natebosch:
We could also consider drawing a harder distinction between local and pub packages, and set a boundary where we cache some work for pub packages, but use pessimistic invalidation over tracking for local code.

jakemac53:
> It would still need to crawl transitive imports right?

Only once (possibly per phase) for the entire app.  Possibly could be optimized to add new sources early as well (as soon as they are written) and avoid doing it once per phase.

> I'm worried that indexing against a generated project could be misleading. I can imagine that different types of projects might have dependency graphs and library graphs with very different characteristics which might sway performance.

Yeah this would be a starting point, see how the worst case scenario stacks up. If it shows a significant improvement that is a sign its worth continuing to investigate (reach out to some people who have filed issues and have them try it on their real projects).

jakemac53:
> We could also consider drawing a harder distinction between local and pub packages, and set a boundary where we cache some work for pub packages, but use pessimistic invalidation over tracking for local code.

Maybe, but I don't think this will help the common case. People run into this when their applications get large, and even if they split it up into multiple packages they are probably path dependencies usually.

jakemac53:
We could optimize any package whose transitive deps had no codegen in the current build - but I think that still misses the use case of a single large application package, which I suspect is common practice.

[...]

PR #3556, opened 2023-08-09
Experimental: Compute and cache transitive digests as files in build_resolvers

jakemac53:
Spawned from discussion [here](https://github.com/dart-lang/build/issues/3555#issuecomment-1671572726).

- Adds a builder in build_resolvers which emits a file containing the transitive digest of the library and all its imports.
  - When computing the transitive digest, we always first look for transitive digest files next to all of our immediate library deps, if they exist we merge that file into our digest and don't crawl any deeper.
    - If we can't read any transitive dependency for some reason we emit a warning and just give up in the builder, which means we will never emit a transitive digest file for that library (which is fine, just not optimal).
- The resolver reads these files to establish dependencies - for each library dependency we first check for a transitive digest file, if it exists we just read that and don't crawl that libraries deps. If it doesn't exist then we do the old behavior.
  - There is actually an additional step here that will continue the recursive crawl for libraries that have not yet been seen in order to load them into the resource provider for the analyzer.
- I also exposed a shared resource to avoid duplicate parsing of libraries between the resolver itself and the transitive digest builder. We might eventually want to migrate this to a builder as well, and serialize that information to disk instead, which would avoid some unnecessary re-parsing of libraries on rebuilds.

### DIFF
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -1,0 +1,45 @@
+import 'dart:async';
+import 'dart:collection';
+
+import 'package:build/build.dart';
+import 'package:convert/convert.dart';
+import 'package:crypto/crypto.dart';
+import 'src/build_asset_uri_resolver.dart';
+
+Builder transitiveDigestsBuilder(_) => _TransitiveDigestsBuilder();
+
+class _TransitiveDigestsBuilder extends Builder {
+  @override
+  Future<void> build(BuildStep buildStep) async {
+    final seen = <AssetId>{buildStep.inputId};
+    final queue = Queue<AssetId>.from(seen);
+    final digestSink = AccumulatorSink<Digest>();
+    final byteSink = md5.startChunkedConversion(digestSink);
+    while (queue.isNotEmpty) {
+      final next = queue.removeLast();
+      final transitiveDigestId = next.addExtension(transitiveDigestExtension);
+      if (await buildStep.canRead(transitiveDigestId)) {
+        byteSink.add(await buildStep.readAsBytes(transitiveDigestId));
+        continue;
+      }
+      byteSink.add(await buildStep.readAsBytes(next));
+      final deps = parseDependencies(await buildStep.readAsString(next), next);
+      for (final dep in deps) {
+        if (!seen.add(dep)) continue;
+        queue.add(dep);
+      }
+    }
+    byteSink.close();
+    final digest = digestSink.events.single;
+    await buildStep.writeAsBytes(
+        buildStep.inputId.addExtension(transitiveDigestExtension),
+        digest.bytes);
+  }
+
+  @override
+  Map<String, List<String>> get buildExtensions => const {
+        '.dart': ['.dart$transitiveDigestExtension'],
+      };
+}
+
+const transitiveDigestExtension = '.transitive_digest';
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -1,12 +1,23 @@
+// Copyright (c) 2023, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
 import 'dart:async';
 
 import 'package:build/build.dart';
 import 'package:convert/convert.dart';
 import 'package:crypto/crypto.dart';
+
 import 'src/build_asset_uri_resolver.dart';
 
 Builder transitiveDigestsBuilder(_) => _TransitiveDigestsBuilder();
 
+/// Computes a digest comprised of the current libraries digest as well as its
+/// transitive dependency digests, and writes it to a file next to the library.
+///
+/// For any dependency that has a transitive digest already written, we just use
+/// that and don't crawl its transitive deps, as the transitive digest includes
+/// all the information we need.
 class _TransitiveDigestsBuilder extends Builder {
   @override
   Future<void> build(BuildStep buildStep) async {
@@ -14,26 +25,33 @@ class _TransitiveDigestsBuilder extends Builder {
     final queue = [...seen];
     final digestSink = AccumulatorSink<Digest>();
     final byteSink = md5.startChunkedConversion(digestSink);
+
+    /// This gives us access to the resolvers cache of library dependencies,
+    /// which helps us to avoid duplicate parsing of libraries.
     final dependenciesOf = await buildStep.fetchResource(dependenciesResource);
 
     while (queue.isNotEmpty) {
       final next = queue.removeLast();
 
-      // If we have a transitive digest ID available, just add that digest.
+      // If we have a transitive digest ID available, just add that digest and
+      // continue.
       final transitiveDigestId = next.addExtension(transitiveDigestExtension);
       if (await buildStep.canRead(transitiveDigestId)) {
         byteSink.add(await buildStep.readAsBytes(transitiveDigestId));
         continue;
       }
+
+      // Otherwise, add its digest and queue all its dependencies to crawl.
       byteSink.add((await buildStep.digest(next)).bytes);
       final deps = await dependenciesOf(next, buildStep);
       if (deps == null) {
         throw StateError(
-            'Unable to read asset: $next, could not compute transitive deps');
+            'Unable to read asset, could not compute transitive deps: $next');
       }
+
+      // Add all previously unseen deps to the queue.
       for (final dep in deps) {
-        if (!seen.add(dep)) continue;
-        queue.add(dep);
+        if (seen.add(dep)) queue.add(dep);
       }
     }
     byteSink.close();
@@ -47,5 +65,3 @@ class _TransitiveDigestsBuilder extends Builder {
         '.dart': ['.dart$transitiveDigestExtension'],
       };
 }
-
-const transitiveDigestExtension = '.transitive_digest';
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -17,13 +17,16 @@ class _TransitiveDigestsBuilder extends Builder {
     final byteSink = md5.startChunkedConversion(digestSink);
     while (queue.isNotEmpty) {
       final next = queue.removeLast();
+
+      // If we have a transitive digest ID available, just add that digest.
       final transitiveDigestId = next.addExtension(transitiveDigestExtension);
       if (await buildStep.canRead(transitiveDigestId)) {
-        byteSink.add(await buildStep.readAsBytes(transitiveDigestId));
+        byteSink.add((await buildStep.digest(transitiveDigestId)).bytes);
         continue;
       }
-      byteSink.add(await buildStep.readAsBytes(next));
-      final deps = parseDependencies(await buildStep.readAsString(next), next);
+      byteSink.add((await buildStep.digest(next)).bytes);
+      final deps = await (await buildStep.fetchResource(depsResource))
+          .readDeps(buildStep, next);
       for (final dep in deps) {
         if (!seen.add(dep)) continue;
         queue.add(dep);
@@ -43,3 +46,15 @@ class _TransitiveDigestsBuilder extends Builder {
 }
 
 const transitiveDigestExtension = '.transitive_digest';
+
+final depsResource = Resource<_ParsedDepsCache>(_ParsedDepsCache.new,
+    dispose: (cache) => cache._cachedDeps.clear());
+
+class _ParsedDepsCache {
+  final _cachedDeps = <AssetId, Set<AssetId>>{};
+
+  Future<Set<AssetId>> readDeps(BuildStep buildStep, AssetId id) async {
+    return _cachedDeps[id] ??=
+        parseDependencies(await buildStep.readAsString(id), id);
+  }
+}
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -1,5 +1,4 @@
 import 'dart:async';
-import 'dart:collection';
 
 import 'package:build/build.dart';
 import 'package:convert/convert.dart';
@@ -12,31 +11,35 @@ class _TransitiveDigestsBuilder extends Builder {
   @override
   Future<void> build(BuildStep buildStep) async {
     final seen = <AssetId>{buildStep.inputId};
-    final queue = Queue<AssetId>.from(seen);
+    final queue = [...seen];
     final digestSink = AccumulatorSink<Digest>();
     final byteSink = md5.startChunkedConversion(digestSink);
+    final dependenciesOf = await buildStep.fetchResource(dependenciesResource);
+
     while (queue.isNotEmpty) {
       final next = queue.removeLast();
 
       // If we have a transitive digest ID available, just add that digest.
       final transitiveDigestId = next.addExtension(transitiveDigestExtension);
       if (await buildStep.canRead(transitiveDigestId)) {
-        byteSink.add((await buildStep.digest(transitiveDigestId)).bytes);
+        byteSink.add(await buildStep.readAsBytes(transitiveDigestId));
         continue;
       }
       byteSink.add((await buildStep.digest(next)).bytes);
-      final deps = await (await buildStep.fetchResource(depsResource))
-          .readDeps(buildStep, next);
+      final deps = await dependenciesOf(next, buildStep);
+      if (deps == null) {
+        throw StateError(
+            'Unable to read asset: $next, could not compute transitive deps');
+      }
       for (final dep in deps) {
         if (!seen.add(dep)) continue;
         queue.add(dep);
       }
     }
     byteSink.close();
-    final digest = digestSink.events.single;
     await buildStep.writeAsBytes(
         buildStep.inputId.addExtension(transitiveDigestExtension),
-        digest.bytes);
+        digestSink.events.single.bytes);
   }
 
   @override
@@ -46,15 +49,3 @@ class _TransitiveDigestsBuilder extends Builder {
 }
 
 const transitiveDigestExtension = '.transitive_digest';
-
-final depsResource = Resource<_ParsedDepsCache>(_ParsedDepsCache.new,
-    dispose: (cache) => cache._cachedDeps.clear());
-
-class _ParsedDepsCache {
-  final _cachedDeps = <AssetId, Set<AssetId>>{};
-
-  Future<Set<AssetId>> readDeps(BuildStep buildStep, AssetId id) async {
-    return _cachedDeps[id] ??=
-        parseDependencies(await buildStep.readAsString(id), id);
-  }
-}
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -45,8 +45,13 @@ class _TransitiveDigestsBuilder extends Builder {
       byteSink.add((await buildStep.digest(next)).bytes);
       final deps = await dependenciesOf(next, buildStep);
       if (deps == null) {
-        throw StateError(
-            'Unable to read asset, could not compute transitive deps: $next');
+        // We warn here but do not fail, the downside is slower builds.
+        log.warning('''
+Unable to read asset, could not compute transitive deps: $next
+
+This may cause less efficient builds, see the following doc for help:
+https://github.com/dart-lang/build/blob/master/docs/faq.md#unable-to-read-asset-could-not-compute-transitive-deps''');
+        return;
       }
 
       // Add all previously unseen deps to the queue.
--- a/build_resolvers/lib/builder.dart
+++ b/build_resolvers/lib/builder.dart
@@ -26,10 +26,6 @@ class _TransitiveDigestsBuilder extends Builder {
     final digestSink = AccumulatorSink<Digest>();
     final byteSink = md5.startChunkedConversion(digestSink);
 
-    /// This gives us access to the resolvers cache of library dependencies,
-    /// which helps us to avoid duplicate parsing of libraries.
-    final dependenciesOf = await buildStep.fetchResource(dependenciesResource);
-
     while (queue.isNotEmpty) {
       final next = queue.removeLast();
 
--- a/build_resolvers/lib/src/build_asset_uri_resolver.dart
+++ b/build_resolvers/lib/src/build_asset_uri_resolver.dart
@@ -18,9 +18,9 @@ import 'package:graphs/graphs.dart';
 import 'package:path/path.dart' as p;
 import 'package:stream_transform/stream_transform.dart';
 
-import '../builder.dart';
-
 const _ignoredSchemes = ['dart', 'dart-ext'];
+
+const transitiveDigestExtension = '.transitive_digest';
 
 class BuildAssetUriResolver extends UriResolver {
   /// A cache of the directives for each Dart library.
@@ -52,7 +52,7 @@ class BuildAssetUriResolver extends UriResolver {
 
   /// The assets which have been resolved from a [BuildStep], either as an
   /// input, subsequent calls to a resolver, or a transitive import thereof.
-  final _buildStepTransitivelyResolvedAssets = Expando<HashSet<AssetId>>();
+  final _buildStepTransitivelyResolvedAssets = <BuildStep, HashSet<AssetId>>{};
 
   /// Use the [instance] getter to get an instance.
   BuildAssetUriResolver._();
@@ -71,8 +71,8 @@ class BuildAssetUriResolver extends UriResolver {
               FutureOr<void> Function(AnalysisDriverForPackageBuild))
           withDriverResource,
       {required bool transitive}) async {
-    final transitivelyResolved =
-        _buildStepTransitivelyResolvedAssets[buildStep] ??= HashSet();
+    final transitivelyResolved = _buildStepTransitivelyResolvedAssets
+        .putIfAbsent(buildStep, HashSet.new);
     bool notCrawled(AssetId asset) => !transitivelyResolved.contains(asset);
 
     final uncrawledIds = entryPoints.where(notCrawled);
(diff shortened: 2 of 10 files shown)


jakemac53:
I did quickly run this on the `build_modules` package which uses json_serializable and it was slower than the current build, at least initially. But that isn't a very representative example.

I would love to get some initial community feedback on it, you should be able to try it by just adding the following to your pubspec:

```
dependency_overrides:
  build_resolvers:
    git:
      ref: transitive_deps
```

If you do try this out, please tell me how it works for you as well as your project structure. Do you have a single large app package or multiple sub-packages? Roughly how many files do you have per package?

cc @simolus3 @rrousselGit not sure if you have some more complex examples worth running this on

natebosch:
> it was slower than the current build

Where would the extra time be going? Did you compare to a full build after a clean, or to an incremental build?

jakemac53:
> Where would the extra time be going? Did you compare to a full build after a clean, or to an incremental build?

It was in spent creating these files - I only did clean builds because I was editing build_modules where incremental builds are impossible.

jakemac53:
I tried the example project in json_serializable and build times are just slightly longer on the intial build and basically identical for other builds - but that is also not a good example because there are very few deps.

I think we need some data from real world projects here, or we could make some example setups to see how things perform in the worst and best case scenarios.

simolus3:
Would it still be sound if `PerActionResolver` also called `reportUnusedAssets()` on all assets read for the analyzer when the transitive digest is available? Perhaps that would be a more aggressive optimization to reduce dependecies in the asset graph.

I'm happy to run this on the examples in the drift repository, but I suppose they aren't a good indicator either. Applying the optimization to [this larger project](https://github.com/syphon-org/syphon/) (my change to test incremental rebuilds was to edit a comment in `lib/domain/events/messages/schema.dart`) was inconclusive too.

ikbendewilliam:
In the project I'm currently working on, we have 968 dart files in lib and 207 in test. It's one large app package. We have 6 generator packages (I think 8 generators). 
The command I used: `fvm flutter packages pub run build_runner build --delete-conflicting-outputs`
(I do need to optimise our generate-for)

Initial build without this experimental PR: `Succeeded after 6m 17s with 562 outputs (10219 actions)`
Initial build with it: `Succeeded after 7m 19s with 3752 outputs (13409 actions)`

Adding a single argument in a viewmodel results in the following builds:
Without: `Succeeded after 2m 13s with 90 outputs (625 actions)`
With: `Succeeded after 3m 7s with 90 outputs (625 actions)`

Removing the same argument to see it wasn't a fluke:
Without: `Succeeded after 2m 5s with 90 outputs (625 actions)`
With: `Succeeded after 2m 55s with 90 outputs (625 actions)`

Hope this helps, if you want I have more projects to run on and can do different changes, but this one is the only one where it takes longer.

jakemac53:
> Would it still be sound if `PerActionResolver` also called `reportUnusedAssets()` on all assets read for the analyzer when the transitive digest is available? Perhaps that would be a more aggressive optimization to reduce dependecies in the asset graph.

If you mean the files read because they haven't yet been loaded into the analyzer, then yes it would be sound to call `reportUnusedAssets` but it wouldn't make much of a difference other than in asset graph size. It shouldn't change the actual invalidation behavior to have the extra dependencies since the transitive digest file would capture the same information.

jakemac53:
> In the project I'm currently working on, we have 968 dart files in lib and 207 in test. It's one large app package. We have 6 generator packages (I think 8 generators).

Thanks for the information and trying it out. It isn't too surprising then that it isn't helping based on that structure. The optimization only really applies to your package dependencies by default, you would have to split the app package up into smaller packages or try adding multiple targets in your build.yaml, which can be a hard thing to get right (and keep right) depending on your structure of your app. If you are interested in getting help doing that I could help you, but if not I totally understand.

I think the extra async work we are doing to check if these files are readable outweighing any advantages you are getting currently.

I am going to see if I can write some tests to validate that the optimization is in fact actually working and I didn't just screw up the implementation somehow, because I am surprised overall by the behavior we are all experiencing.

jakemac53:
Ok, I have identified some problem areas in the implementation as well as some optimization opportunities. I don't have a _ton_ of time to spend on this but given the potential impact for people I think its worth spending a bit more effort here. I will follow up in a bit and probably ask people to try again :).

jakemac53:
ok @ikbendewilliam can you try again with the latest? I think you will need to run another `[dart|flutter] pub get` to get the latest version of the branch.

In my testing for devtools, incremental builds are about twice as fast now, although I did add two custom targets there to realize that benefit, but even without that it was about 30% faster than before and initial builds were also about 10% faster.

jakemac53:
Ok, I built a fake app, just to try and get a sense of how this can scale. The structure isn't likely very representative, but there are a large number of files (>10k), enough that I knew the old approach would not scale well. Any app _could_ be structured in roughly this way and actually this structure is somewhat intentionally bad (for the original approach at least). We know we have apps that have roughly this number of files (or even greater) in the wild. Although they likely aren't doing analyzer based codegen on every single library like this.

Specifically the structure is 100 packages with 101 files each. Each package has 1 entrypoint in lib which imports all the other 100 files. Those files are under `lib/src` with numbers like `0.dart` etc. Those all import all previous ones, so `2.dart` imports `1.dart` and `0.dart`.

Each file (the entrypoint as well as lib/src files)  also imports the entrypoint of each previous app, so files in `pkg5` import `package:pkg4/pkg4.dart` through `package:pkg0/pkg0.dart`.

I then have a root package which is hand written and defines a single builder, which applies to all packages, and all dart files. It resolves each file, and outputs a simple text file with the list of top level declarations in each file (there are actually none, so it doesn't really do anything other than resolve each library).

That root package also depends on pkg99, so that codegen will run on all those packages.

With this change, I got the following results:

1m 49s fresh (11601 outputs, 11669 actions)
1m 48s editing pkg50/pkg50.dart (4999 outputs, 4999 actions) - **Observation**: why is this almost as long as a fresh build?
56.5s editing pkg75/pkg75.dart (2449 outputs, 2449 actions)
22.6s editing pkg90/pkg90.dart (919 outputs,919  actions)
10.6s editing pkg95/pkg95.dart (409 outputs, 409 actions)
627ms editing pkg99/pkg99.dart (1 outputs, 1  actions)

Without this change, I get:

26m 14s fresh (11361 outputs, 11429 actions)
 editing pkg50/pkg50.dart ( outputs,  actions)
 editing pkg75/pkg75.dart ( outputs,  actions)
 editing pkg90/pkg90.dart ( outputs,  actions)
 editing pkg95/pkg95.dart ( outputs,  actions)
 editing pkg99/pkg99.dart ( outputs,  actions)
 
 (will fill in above results as they complete, its a long wait LOL)

ikbendewilliam:
I've done a similar test as last time, here are the results:

The initial build FAILED with following error:
```dart
[SEVERE] drift_dev on test/database/people/people_dao_storage_test.dart:

Invalid argument(s): Missing library: package:flutter_test/flutter_test.dart
```
(with a long list of packages)
Without the PR it does work!

and another one I saw:
```dart
[SEVERE] drift_dev on lib/some/viewmodel/some_viewmodel.dart:

LateInitializationError: Field 'library' has not been initialized.
```

Initial build: `4m 7s, 10607 actions` (without PR `7m 40s, 562 outputs & 10219 actions`)
Adding an argument in a VM: `1m 18s-1m 30s, 1257 actions` (without PR `2m 20s, 90 outputs & 625 actions`)

So definetly faster this time! But it seems there is an issue that breaks drift_dev

If you want I can reach out for a project with multiple packages.

jakemac53:
Yeah we have some drift tests that run on our PRs that are failing too, I can look into those. Something to do with the `libraries` getter which isn't that surprising so ill check it out.

jakemac53:
cc @simolus3 do you know if drift always uses the `libraries` getter or only for certain builders (of which there are hopefully few?). That call will effectively have to be O(transitive imports) and will basically invalidate any optimizations here for any files on which it happens.

jakemac53:
> If you want I can reach out for a project with multiple packages.

If you know somebody who has a project with multiple packages they can try this on that would be great. (not sure if that is what you meant haha)

ikbendewilliam:
> If you know somebody who has a project with multiple packages they can try this on that would be great. (not sure if that is what you meant haha)

Yes that's what I meant 😉 I'll see what I can do

simolus3:
I think drift should actually not be using the `Resolver.libraries` API at all.

@ikbendewilliam, could you re-run the failed build with `--verbose` and post the stack trace?

ikbendewilliam:
The whole log file is 10MB 😅

This is the first instance where it happens:
```
...
[FINE] json_serializable on lib/vendor/keyboard_avoider/keyboard_avoider.dart:Running JsonLiteralGenerator - 2 of 2
[SEVERE] drift_dev on lib/...dart:

Invalid argument(s): Missing library: package:firebase_core/firebase_core.dart
Libraries: [dart:async, dart:collection, ... (this line allone is 140kb)]
Root children: [dart:async, dart:collection, ... (3124 total)]
Readers: [dart:async, dart:collection, ... (1244 total)]
Log: [load][targetLibrary: package:...dart = /apppackage/...dart]
[load][targetLibrary: package:...dart = /path_provider/lib/path_provider.dart]
[load][targetLibrary: package:...dart = /apppackage/...dart]
[load][targetLibrary: package:...dart = /apppackage/...dart]
[load][targetLibrary: package:json_annotation/json_annotation.dart = /json_annotation/lib/json_annotation.dart]
[load][targetLibrary: package:...dart = /apppackage/...dart]
[load][targetLibrary: package:...dart = /apppackage/...dart]
[load][targetLibrary: package:firebase_remote_config/firebase_remote_config.dart = /firebase_remote_config/lib/firebase_remote_config.dart]
[load][targetLibrary: package:flutter/foundation.dart = /flutter/lib/foundation.dart]
[removeLibraries][uriSet: {package:plugin_platform_interface/plugin_platform_interface.dart, ... 45.1kb}]
#1      LibraryContext.remove (package:analyzer/src/dart/analysis/library_context.dart:265:20)
#2      AnalysisDriver._removePotentiallyAffectedLibraries (package:analyzer/src/dart/analysis/driver.dart:1777:22)
#3      AnalysisDriver._applyPendingFileChanges (package:analyzer/src/dart/analysis/driver.dart:1243:7)
#4      AnalysisDriverScheduler._run (package:analyzer/src/dart/analysis/driver.dart:2152:18)
<asynchronous suspension>
]

package:analyzer/src/summary2/linked_element_factory.dart 130:7   LinkedElementFactory.createLibraryElementForReading
package:analyzer/src/summary2/linked_element_factory.dart 194:14  LinkedElementFactory.elementOfReference
package:analyzer/src/summary2/linked_element_factory.dart 225:12  LinkedElementFactory.libraryOfUri
package:analyzer/src/summary2/linked_element_factory.dart 229:19  LinkedElementFactory.libraryOfUri2
package:analyzer/src/summary2/bundle_reader.dart 1620:28          ResolutionReader.libraryOfUri
package:analyzer/src/summary2/bundle_reader.dart 432:30           LibraryElementLinkedData._readLibraryOrAugmentation
package:analyzer/src/summary2/bundle_reader.dart 406:5            LibraryElementLinkedData._read
package:analyzer/src/summary2/bundle_reader.dart 199:5            ElementLinkedData.read
package:analyzer/src/summary2/bundle_reader.dart 395:13           LibraryElementLinkedData.read
package:analyzer/src/dart/element/element.dart 4498:17            LibraryElementImpl._readLinkedData
package:analyzer/src/dart/element/element.dart 4242:5             LibraryElementImpl.libraryExports
package:analyzer/src/dart/element/element.dart 4683:12            LibraryOrAugmentationElementImpl.children
package:analyzer/src/dart/element/element.dart 4103:18            LibraryElementImpl.children
package:analyzer/src/dart/element/element.dart 2880:27            ElementImpl.visitChildren
package:analyzer/dart/element/visitor.dart 299:13                 RecursiveElementVisitor.visitLibraryElement
package:drift_dev/src/analysis/resolver/discover.dart 177:5       _FindDartElements.find
package:drift_dev/src/analysis/resolver/discover.dart 77:22       DiscoverStep.discover
package:drift_dev/src/analysis/driver/driver.dart 168:7           DriftAnalysisDriver.prepareFileForAnalysis
package:drift_dev/src/analysis/driver/driver.dart 198:11          DriftAnalysisDriver.prepareFileForAnalysis
package:drift_dev/src/analysis/driver/driver.dart 198:11          DriftAnalysisDriver.prepareFileForAnalysis
package:drift_dev/src/analysis/driver/driver.dart 198:11          DriftAnalysisDriver.prepareFileForAnalysis
package:drift_dev/src/analysis/driver/driver.dart 198:11          DriftAnalysisDriver.prepareFileForAnalysis
package:drift_dev/src/analysis/driver/driver.dart 251:5           DriftAnalysisDriver.resolveElements
package:drift_dev/src/backends/build/analyzer.dart 37:21          DriftAnalyzer.build

[SEVERE] drift_dev on lib/....dart:

Invalid argument(s): Missing library: package:flutter/material.dart
```

It just spews out very similar errors for every import it seems.

jakemac53:
Ah so this is using `await _buildStep.resolver.findLibraryByName(name);`. That API actually uses `libraries`,  unfortunately.

I don't actually see why `libraries` wouldn't work though. It should be doing its own separate crawl.

jakemac53:
cc @robrbecker I am not sure if you all still use any codegen that uses the Resolver or just rely on parsing nowadays. If you do though, you should give PR a try and see how it does!

jakemac53:
@jodinathan you might also be interested in trying this?

simolus3:
> Ah so this is using `await _buildStep.resolver.findLibraryByName(name);`. That API actually uses `libraries`, unfortunately.

We're only calling `findLibraryByName` to resolve Dart SDK libraries (e.g. when we need a `LibraryElement` for `dart:core`, we'd call `findLibraryByName('dart.core')` since we can't use `libraryFor`). In my understanding of the implementation this shouldn't start crawling entrypoints if the SDK library exists.

I also don't see `findLibraryByName` or `.libraries` in the stack trace, or is that implied by `AnalysisDriver._applyPendingFileChanges` being called?

greglittlefield-wf:
> cc @robrbecker I am not sure if you all still use any codegen that uses the Resolver or just rely on parsing nowadays. If you do though, you should give PR a try and see how it does!

We do still have a few builders that use Resolver, such as built_value and json_serializable, so we gave these changes a spin. Thanks for putting this experiment together and letting us know about it! 

I wasn't able to pull in this branch specifically, but I attempted to backpatch the changes to our version build_resolvers 1.5.4 ([backpatch branch](https://github.com/greglittlefield-wf/build/compare/build_resolvers-v1.5.4...greglittlefield-wf:build:transitive-digests--build_resolvers-v1.5.4?expand=1)) and that seemed to build alright, and did end up impacting performance, so that's what we tested with.

**tl;dr** Generally, we didn't see huge improvements except for when I removed the existing `generate_for` configurations that restricted the files that resolved-AST builders run on, except for one case where we saw a modest speedup.

<details>
<summary><em>Expand for more details</em></summary>

#### Large application

I first tried testing this change on our largest application, which has little code of its own and many dependencies.

Locally, on an M1 mac, for a fresh `pub run build_runner build`:

- Before: 5m 3s
- With this build_resolvers patch: 4m 43s

So, a speedup of 20s, taking as long as 93.4% of the original build time.

In our CI environment, however, I saw a slight slowdown:
- Without patch: 9m 51s, 9m 45s
- With patch: 10m 0s, 10m 8s

A slowdown of 16.0s, or 103% of the original time.

#### Medium-sized package

I tested in one of the application's dependencies that had more of its own code, and didn't see a significant difference in build times. 

However, I remembered that this package, along with most of our packages, have `generate_for` filters for its resolved-AST builders so they only resolve the files they needed to, as a performance optimization.

So, I tried taking those out and seeing the difference the patch made.

Before taking those filters out, locally I was seeing:
-  2m 43s for a fresh build
-  8.9s for a rebuild while running `pub run build_runner watch`, after changing a specific file that utilized a resolved-AST builder and was imported in a lot of other places

Without `generate_for` filters, locally I was seeing:
- Without build_resolvers patch: 
    - 6m 34s initial build
    - 2m 43s `watch` rebuild after changing the same file as above
- With build_resolvers patch: 
    - 4m 27s Initial build
    - 52.1s `pbr watch` rebuild after changing the same file as above

So, the patch sped up:
- initial builds up by 2m 7s, taking as long as 67.8% the time without the patch
- rebuilds by 1m 50.9s , taking as long as 32.0% the time without the patch

However, things were still much slower than when we had those `generate_for`s in place.

#### Other packages

We tried the same in a couple other packages, but generally saw about the same performance with/without the patch.

However, there was a case of a larger package where we saw the initial build go from 9m 5s to 7m 12s locally.

</details>

PieterAelse:
I tried it on our quite big codebase (details below) and only saw a small performance improvement (sub 10%). 
Total time went down from 47.5 minutes to 44 minutes 🤔 

Codebase details:
- ~130 packages (dart/Flutter)
- 98 of which do build_runner stuff
- Total of 7369 dart files (in the ~130 packages)

jakemac53:
Back from vacation now, sorry for the slow response!

@greglittlefield-wf :

For the larger package, did you try watch mode? Does that package itself do much Dart related codegen using a Resolver? It might be worth clarifying that these changes will not affect any Dart -> JS compilation, which might be a larger part of that build? If you could try running in `serve` mode with `--track-performance` and then navigating to the `/$perf` page on one of the ports that is served, you might be able to figure out where the majority of time is spent, and possibly filter out just the codegen aspects?

For the medium package, it seems to have helped a lot but not as much as I might have hoped still (in particular once you removed the generate_for).

It would also be interesting to compare asset graph sizes (that should help narrow down whether this is having the intended affect or something is defeating the optimization). Just looking at the size of the `.dart_tool/build/<some-hash>/asset_graph.json` file in either mode should be sufficient.

@PieterAelse :

Was this for a single build in the "application" package or are you summing up build times across several packages?

Or is there a `build_to: cache` builder which runs on all those deps from the "application" package?

PieterAelse:
@jakemac53 It's the time it took to run codegeneration on all packages (we have tooling in place to trigger codegen for the whole codebase). So basically the `build_runner` time of all packages.

jakemac53:
> @jakemac53 It's the time it took to run codegeneration on all packages (we have tooling in place to trigger codegen for the whole codebase). So basically the `build_runner` time of all packages. _(I did not try watch mode, not something I use normally in my workflow)_

Ah ok, so is that 98 separate invocations of `build_runner`? If your builds are on average <30 seconds, this change is unlikely to help much. It might help more for the particular packages which have more dependencies though. So you could try just building one of those which might filter out some of the noise?

If the builds above were "fresh" builds in particular, that involves potentially a lot of fixed overhead to compile the build script itself, transitively resolve all files (this does still have to happen at least once per build), and also potentially create the SDK summary. That might be the majority of the time in your case if its happening 98 times.

greglittlefield-wf:
@jakemac53 No worries, thanks for the reply!

> For the larger package, did you try watch mode? Does that package itself do much Dart related codegen using a Resolver? 

No, it doesn't do any codegen itself. I just now realized that virtually all of the codegen done by its dependencies is `build_to:source`, meaning it wouldn't get run during that large package's build, right? 🤦 Sorry, I don't know why I was thinking otherwise. So these changes not really affecting that package's build times makes sense to me.

It does seem like JS compilation is the bottleneck in the large and medium packages (except for when `generate_for` is missing), based on `/$perf`, which lines up with what I was expecting.

For the large package, the size of asset_graph.json is 226.2 MB. For the medium package, 58.5 MB.

jakemac53:
@greglittlefield-wf are you able to do a build without JS for the medium sized app so I can isolate the changes more?

Specifically I would like it if you could do the following with and without this change:

- remove generate_for (most people don't have this and ideally nobody would need it)
- run build_runner in watch mode with `--build-filter "**/*.dart"`
- edit some library that is imported by a lot of other libraries

If you could give me the asset graph size with and without this change, as well as timing info, that would be awesome :).

greglittlefield-wf:
@jakemac53 Here's that info!

Medium sized app, after removing generate_for, and running `dart run build_runner watch --build-filter "**/*.dart”`:

| | Without this change | With this change |
| --- | --- | --- | 
| Asset graph size | 153MiB | 102MiB |
| Initial build time | 4m 22s | 2m 29.5s |
| Build time while watching,<br>after editing library | 2m 7s | 41.2s

Times are averaged over two runs.

greglittlefield-wf:
@jakemac53 You're welcome, thanks for taking a look!

Oh interesting, you mean further benefit with these build_resolvers changes and when not using `generate_for`? Or benefit on top our current `generate_for` setup? If the latter, we'd definitely appreciate more info on that!

jakemac53:
@simolus3 I looked into the customer tests failure - interestingly when I force `drift_dev` to be a path dependency I get a different error, which is that `test/generated/custom_tables.g.dart` does not exist.

This is a result of how the targets in `drift` are currently set up. Adding a dependency from the `$default` target to the `custom` target does resolve the issue, but it does highlight that this is a breaking change for people with custom targets.

greglittlefield-wf:
@jakemac53 Gotcha. We try to limit wherever possible the number of libraries imported from files the generators run on, since we've noticed that helps with rebuild performance. 

Without doing more in-depth analysis I can't say for sure how many are typically transitively imported from the same package, but from manually inspecting some of them, it seems like most files import 1-2 other libraries, with a couple few importing 5-10. If more precise data there would be helpful, I could try to get that.

Those globs match a small proportion of files, in this case 22 out of 920 files.

jakemac53:
@greglittlefield-wf given that info I would be surprised if using targets would be worth the maintenance burden of multiple targets for your case.

jakemac53:
Looking a bit more into the error that shows up when running drift codegen, I noticed in the logs for the error there is a separate section for logs, one of which looks like `[removeLibraries][uriSet: {...}]`.

Whenever this error occurs, it looks like the library that caused the error does show up in that set. I tried removing all our calls to `resourceProvider.deleteFile` but that didn't help, so I don't know what is causing the libraries to be removed.

cc @scheglov would you be able to help point me in the right direction here? If a library is invalidated, does it get removed?

scheglov:
The reason we have these `[removeLibraries][uriSet: {...}]` messages is that we had this issue in production, and I failed to reproduce and fix it. My current hypothesis is that this is an async race condition. One flow of execution ensure that `a.dart` is loaded, and started linking other library, or resolve something against it; and then another flow of execution said "change file" that remove `a.dart` library element from the context; so when the first flow attempt to use it... boom!

scheglov:
I don't know. To know this, I need to understand the problem fully, which I don't.

A possible solution could be to never have concurrent read/write invocations to the analyzer - if you have "read" invocations -  asked for a library element, or a resolved unit, never call `changeFile` ("write" invocation) until these "read" invocations return the result.

jakemac53:
> A possible solution could be to never have concurrent read/write invocations to the analyzer - if you have "read" invocations - asked for a library element, or a resolved unit, never call `changeFile` ("write" invocation) until these "read" invocations return the result.

Unfortunately this isn't even possible for us to do - resolving one library can actually require running another builder, which might need to also resolve some of its own libraries.

jakemac53:
Actually, I was wrong about that - we can't have a lock while we are preparing the files (updating the resource provider and calling changeFile) - but we can (and actually already do) grab a lock when asking for a library element.

However, while we are asking for that library element some other thing could be separating doing the crawl since that doesn't have any sort of lock on it. I will try adding a new lock that will prevent such crawls while we are asking for a library element, and see what happens.

jakemac53:
That did in fact appear to fix the issue, thanks @scheglov for pointing me in the right direction.

I will have to evaluate how this affects performance but I think it should be OK. It could be optimized more by allowing multiple concurrent reads or writes also if necessary although I don't know if we have a shared implementation of that.

build_resolvers/lib/builder.dart
```diff
@@ -0,0 +1,72 @@
...
+class _TransitiveDigestsBuilder extends Builder {
+  @override
+  Future<void> build(BuildStep buildStep) async {
+    final seen = <AssetId>{buildStep.inputId};
+    final queue = [...seen];
+    final digestSink = AccumulatorSink<Digest>();
+    final byteSink = md5.startChunkedConversion(digestSink);
+
+    /// This gives us access to the resolvers cache of library dependencies,
+    /// which helps us to avoid duplicate parsing of libraries.
```
natebosch, line 30: [nit] 
```suggestion
    // This gives us access to the resolvers cache of library dependencies,
    // which helps us to avoid duplicate parsing of libraries.
```

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -226,7 +243,7 @@ Future<String> packagePath(String package) async {
 
 /// Returns all the directives from a Dart library that can be resolved to an
 /// [AssetId].
-Set<AssetId> _parseDirectives(String content, AssetId from) => HashSet.of(
+Set<AssetId> parseDependencies(String content, AssetId from) => HashSet.of(
```
natebosch, line 246: [nit] can this stay private? I don't see any usage in the PR outside this library.

build_resolvers/lib/src/build_asset_uri_resolver.dart
```diff
@@ -240,9 +257,17 @@ Set<AssetId> _parseDirectives(String content, AssetId from) => HashSet.of(
           .map((content) => AssetId.resolve(Uri.parse(content), from: from)),
     );
 
+/// A resource which gives access to the cached dependencies of assets based on
+/// parsing the directives.
+final dependenciesResource = Resource<
+        Future<Iterable<AssetId>?> Function(AssetId id, BuildStep buildStep)>(
+    () => (AssetId id, BuildStep buildStep) => BuildAssetUriResolver.instance
```
natebosch, line 264: Since this touches global state and doesn't use `dispose` or `beforeExit` I don't think we are gaining anything from `Resource`.

Would it make sense to replace the usage with a direct definition?

```dart
Future<Iterable<AssetId>?> dependenciesOf(
        AssetId id, BuildStep buildStep) async =>
    (await BuildAssetUriResolver.instance
            ._updateCachedAssetState(id, buildStep))
        ?.dependencies;
```

natebosch (approved):

build_resolvers/lib/src/build_asset_uri_resolver.dart
jakemac53: Yep just forgot to swap it back after experimenting (initially I just made this public and used it directly).

build_resolvers/lib/src/build_asset_uri_resolver.dart
jakemac53: Good point, done!

PR #3580, opened 2023-09-20
Fix an issue where multiple AnalyzerResolvers instances lead to incorrect behavior

jakemac53:
Deprecate the unnamed AnalyzerResolvers constructor, replace it with a static instance getter and `AnalyzerResolvers.custom` factory constructor.

These new APIs guarantee a 1:1 relationship between the BuildAssetUriResolver and an analysis engine, which is necessary to ensure `changeFile` is always called on each analysis engine when needed.

Note that the deprecated constructor does not uphold this invariant, it remains backwards compatible with its old implementation, and uses the same BuildAssetUriResolver as the shared resolver.

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.3.3-wip
+## 2.4.0-wip
 
 - Deprecate the unnamed `AnalyzerResolvers` constructor, an replace it with the
   `AnalyzerResolvers.sharedInstance` getter and `AnalyzerResolvers.custom`
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 2.3.3-wip
+
+- Deprecate the unnamed `AnalyzerResolvers` constructor, an replace it with the
+  `AnalyzerResolvers.sharedInstance` getter and `AnalyzerResolvers.custom`
+  constructor. These new apis enforce a 1:1 relationship between the
+  `BuildAssetUriResolver` instances and `AnalyzerResolvers` instances, which is
+  necessary to ensure correct builds in the presence of multiple resolvers.
+
 ## 2.3.2
 
 - Skip file delete for SDK summary and deps file. This will only impact behavior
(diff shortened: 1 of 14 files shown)


build_resolvers/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 2.3.3-wip
```
natebosch, line 1:
```suggestion
## 2.4.0-wip
```

build_resolvers/lib/src/resolver.dart
```diff
@@ -402,30 +403,65 @@ class AnalyzerResolvers implements Resolvers {
...
+  }) =>
+      AnalyzerResolvers._(
+          analysisOptions: analysisOptions,
+          sdkSummaryGenerator: sdkSummaryGenerator,
+          packageConfig: packageConfig,
+          // Custom resolvers get their own asset uri resolver, as there should
+          // always be a 1:1 relationship between them.
+          uriResolver: BuildAssetUriResolver());
+
+  /// See [AnalyzerResolvers.custom] for docs.
```
natebosch, line 419: [optional] Could use a `{@template analyzer_resolvers_constructor}` and `{@macro analyzer_resolvers_constructor}` to copy the doc to the other constructors.

build_resolvers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_resolvers
-version: 2.3.2
+version: 2.3.3-wip
```
natebosch, line 2:
```suggestion
version: 2.4.0-wip
```

build_runner_core/pubspec.yaml
```diff
@@ -15,7 +15,7 @@ dependencies:
   async: ^2.5.0
   build: ^2.4.0
   build_config: ^1.0.0
-  build_resolvers: ^2.0.0
+  build_resolvers: ^2.3.3
```
natebosch, line 18:
```suggestion
  build_resolvers: ^2.4.0
```

build_test/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
   async: ^2.5.0
   build: ^2.1.0
   build_config: ^1.0.0
-  build_resolvers: ^2.0.0
+  build_resolvers: ^2.3.3
```
natebosch, line 13:
```suggestion
  build_resolvers: ^2.4.0
```

natebosch (approved):

build_resolvers/lib/src/resolver.dart
jakemac53: Looks like you lose navigation from within the IDE though so I think that is ultimately worse?

PR #3582, opened 2023-09-21
update to the latest dart_flutter_team_lints

jakemac53:


jakemac53:
Ah, I will also have to update the min SDK to 3.0.0 across all these packages. I think that is fine though.

natebosch (approved):

PR #3594, opened 2023-10-06
Remove blacklistPatterns

natebosch:
It is unused in the places we use `DefaultBuildTarget`.


davidmorgan:

Issue #3607, opened 2023-11-08
[webdev server BUILD ERROR] -> OS Error: File name too long, errno  = 36 


Parsaloi:
* Dart SDK Version (`dart --version`) => `Dart SDK version: 3.1.5 (stable) (Tue Oct 24 04:57:17 2023 +0000) on "linux_x64"`
 => `build_runner: ^2.4.0`, `build_web_compilers: ^4.0.0`

 => Web dev builder, trying to compile web app artifacts
=> Linux
=> Chrome, Firefox

* ERROR LOGS are as follows:

Missing some or all of the above might make the issue take longer or be impossible to act on.
```bash
graphical_user_interface_displaying_messages_using_dart_services_web_appV2_messages' (OS Error: File name too long, errno = 36)
[SEVERE] package:build_daemon/daemon.dart 142:5                 _createDaemonWorkspace
[SEVERE] package:build_daemon/daemon.dart 128:5                 _tryGetLock
[SEVERE] package:build_daemon/daemon.dart 35:17                 new Daemon
[SEVERE] package:build_runner/src/entrypoint/daemon.dart 57:18  DaemonCommand._run
[SEVERE] package:build_runner/src/entrypoint/daemon.dart 52:15  DaemonCommand.run.<fn>
[SEVERE] dart:async                                             runZoned
[SEVERE] package:build/src/experiments.dart 18:5                withEnabledExperiments
[SEVERE] package:build_runner/src/entrypoint/daemon.dart 51:12  DaemonCommand.run
[SEVERE] package:args/command_runner.dart 212:27                CommandRunner.runCommand
[SEVERE] package:args/command_runner.dart 122:25                CommandRunner.run.<fn>
[SEVERE] dart:async                                             new Future.sync
[SEVERE] package:args/command_runner.dart 122:14                CommandRunner.run
[SEVERE] package:build_runner/src/entrypoint/run.dart 26:31     run
[SEVERE] .dart_tool/build/entrypoint/build.dart 126:16          main
[SEVERE] 
Unhandled exception:
Bad state: Unable to start build daemon.
#0      _handleDaemonStartup.<anonymous closure> (package:build_daemon/client.dart:79:21)
#1      _runUserCode (dart:async/stream_pipe.dart:11:23)
#2      Stream.firstWhere.<anonymous closure> (dart:async/stream.dart:1700:9)
#3      _RootZone.runGuarded (dart:async/zone.dart:1582:10)
[...]

PR #3608, opened 2023-11-08
use base64 encoded md5 hash of working dir to build the unique workspace dir

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3607

@Parsaloi you could try this out if you would like by adding the following to your pubspec.yaml:

```yaml
dependency_overrides:
  build_daemon:
    git:
      ref: shorter-file-names
      path: build_daemon
```

### DIFF
--- a/build_daemon/pubspec.yaml
+++ b/build_daemon/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_daemon
-version:  4.0.1-wip
+version:  4.0.1
 description: A daemon for running Dart builds.
 repository: https://github.com/dart-lang/build/tree/master/build_daemon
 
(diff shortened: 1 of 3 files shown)


build_daemon/pubspec.yaml
```diff
@@ -9,6 +9,7 @@ environment:
 dependencies:
   built_collection: ^5.0.0
   built_value: ^8.1.0
+  crypto: ^3.0.3
```
natebosch, line 12: Is this a new transitive dep when you exclude dev dependencies? Consider mentioning either way in the commit message.

natebosch (approved):

build_daemon/pubspec.yaml
jakemac53: It is new  yes, I can mention it.

Issue #3615, opened 2023-11-21
Why `build_config` only support `.yaml` extension not both `.yaml` and `.yml`?


schaipipat:
Not sure if this is the right place to ask or not, but my background from using many of YAML file, it suppose to recognize that both `.yaml` and `.yml` is the extension for the same file type.

schaipipat:
May be a good thing to have for people who just come from another languages/tools like me. I'm happy to contribute if you're busy. Thanks a lot. 🙇‍♂️

jakemac53:
I do know of one other person who also made this mistake so you definitely aren't alone :)

jakemac53:
I do feel quite comfortable adding a check in for that file and emitting a warning. We can still keep a consistent convention that way and it doesn't promise any kind of long term support or maintenance burden, it would be a nice isolated piece of code.

PR #3616, opened 2023-11-21
add a warning if a build.yml file is found in the doctor command

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3615

Note that this would only give the warning if running the `doctor` command, which people may not be in the habit of running, but is typically where we put these kinds of diagnostics.

@schaipipat out of curiosity were you aware of the doctor command? Would this have helped you or not?

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,7 +1,9 @@
-## 2.4.7-wip
+## 2.4.7
 
 - Fix broken link in README.md.
 - Bump the min sdk to 3.0.0.
+- Add a warning to the `doctor` command if a `build.yml` file is found (instead
+  of a `build.yaml` file).
 
 ## 2.4.6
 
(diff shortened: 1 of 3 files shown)


natebosch (approved):
This LGTM. We could also consider a verbose log like "No build.yaml configuration file, using defaults." and a different verbose log for when the file is read. Anyone running with  `-v` could see this without having to run a separate `doctor` command.

jakemac53:
I tried originally putting this into the code that actually looks for these files, but it ended up getting complicated because those APIs are actually called for all transitive packages deps, and don't have access to a logging API. So this was the simplest approach.

schaipipat:
@jakemac53 Not really. All the tools that using `build_runner` never mention about `doctor` command so I haven't run it when I start facing the issue.

jakemac53:
I am going to go ahead and land this because it is better than not having anything - longer term I think we need to figure out how to promote the `doctor` command more I think.

PR #3630, opened 2023-12-27
log error stack

natebosch:
- Always log stack traces for Error objects
- Fix json serializable annotation for ignored field


natebosch:
I was not on a sensible branch when I opened this... I'l try again.

PR #3640, opened 2024-01-11
Support latest pkg:js in pkg:build_runner and pkg:build_web_compilers

kevmoo:
Prepare both for publish

### DIFF
--- a/build_web_compilers/lib/src/platforms.dart
+++ b/build_web_compilers/lib/src/platforms.dart
@@ -6,6 +6,7 @@ import 'package:build_modules/build_modules.dart';
 
 const _libraries = [
   '_internal',
+  '_js_annotations',
   'async',
   'collection',
   'convert',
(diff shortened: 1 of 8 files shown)


jakemac53 (approved):

build_web_compilers/lib/src/platforms.dart
```diff
@@ -6,6 +6,7 @@ import 'package:build_modules/build_modules.dart';
 
 const _libraries = [
```
kevmoo, line 7: My brain always goes to "this should be a `Set`" but no biggy...

jakemac53: I mean, you aren't wrong, but it also doesn't matter here in practice :)

kevmoo: Oh, I know. 😉

PR #3651, opened 2024-02-20
Ignore stack overflow URLs in link checker

jakemac53:
We get a 403 (forbidden) for these currently. Just ignore them.

I also tried adding a user agent but that didn't seem to help.


natebosch (approved):
SGTM. We shouldn't need to worry about a stack overflow link moving.

PR #3656, opened 2024-02-29
Move test skips before the test body

natebosch:
Style improves in many cases when the block structured argument is last.
The skip also is likely to get lost visually when it follows a long test
body, it is metadata that makes sense to put early in the test
definition.


kevmoo (approved):
LGTM

jakemac53 (approved):

PR #3685, opened 2024-05-13
fixes for the latest dev build and latest analyzer

jakemac53:
We do now require the latest analyzer since the `withNullability` parameter was previously required, we can't cleanly remove it without also increasing the dependency.

We also seem to be getting a few more lints triggered on the latest dev, which this should fix.

Also updates the min SDK to ~3.3.0~ 3.4.0 across the board (analyzer 6.5.0 requires this).


jakemac53:
This won't be able to land until 3.4.0 is published, but then it should be good.

PR #3724, opened 2024-07-01
allow 3.6.x sdks

jakemac53:
unblock the version bump in the SDK

### DIFF
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_modules
-version: 5.0.9-wip
+version: 5.0.9
 description: >-
   Builders to analyze and split Dart code into individually compilable modules
   based on imports.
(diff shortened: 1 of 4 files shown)


build_modules/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
+version: 5.0.9
 description: >-
   Builders to analyze and split Dart code into individually compilable modules
   based on imports.
 repository: https://github.com/dart-lang/build/tree/master/build_modules
 resolution: workspace
 
 environment:
-  sdk: '>=3.5.0-259.0.dev <3.6.0'
+  sdk: '>=3.5.0-259.0.dev <3.7.0'
```
natebosch, line 10: Is there any reason not to change the lower bound to `>=3.5.0`?

natebosch (approved):

build_modules/pubspec.yaml
jakemac53: Well, 3.5 isn't actually out yet.

jakemac53: At a minimum our "pubspec" SDK jobs would fail.

natebosch: Ah, surprising that we're prepping for 3.6 without launching 3.5...

LGTM.

jakemac53: Yeah it happens when we cut the branch that will be 3.5 stable. That way we don't have misleading -dev tags that make people put in bad constraints.

Issue #3730, opened 2024-08-01
Multi compiler support in `dart_web_compilers`


simolus3:
In https://github.com/dart-lang/build/pull/3727, there was a discussion on how to best support scenarios where we'd compile Dart apps with dart2js and dart2wasm, using a browser compatibility check at runtime to select the right application. I wanted to extract this into a separate issue so that it doesn't get lost.

For the feature detection in an entrypoint script, we probably want to clone what Flutter is doing. A question is how this would be enabled (or if it should be the default), possible options are:

1. Keeping the current `compiler` builder option and adding a new boolean flag controlling whether `dart2wasm` should be running too. A downside is that users then won't be able to use dart2wasm exclusively (if that's what they want).
2. Turning the `compiler` option into an array (which would be a singleton if a string is given). The compiler could then default to `[dart2js, dart2wasm]` and the entrypoint builder would take care of emitting an entrypoint script if multiple compilers are enabled. A question is what we'd do if both `dartdevc` and `dart2js` are listed - we could simply reject those configurations, or only consider `dart2js` in the entrypoint script while still emitting `dartdevc` modules (can be useful for tests where one wants to test compatibility with all three compilers).

A couple of follow-up issues before we call this closed.

- [x] https://github.com/dart-lang/build/issues/3747
- [x] https://github.com/dart-lang/build/issues/3746

jakemac53:
There is also the option of a `dart2wasm+dart2js` named compiler or something. I think I like that more than an array of compilers where we have to check the valid combinations.

simolus3:
Is there anything left I can help with to get this released?

kevmoo:
I *think* we're done, modulo bugs. The question: do we release before Dart 3.6 stable is out.

We could do a beta release, if that's useful for you.

simolus3:
Well I'd definitely prefer a beta release over having to add git dependency overrides everywhere :) I've opened a PR for that, but if it's too much work to get this released it's not critical either.

Issue #3652, opened 2024-02-24
Preamble for Customized Web Builders


pattobrien:
## Context

i've been experimenting with the latest `js_interop` features, by building a node.js interop package (see: https://github.com/pattobrien/dart3_node_interop).

Since node.js is designed to run outside of the browser, there have been a number of unique aspects to compiling the js code from Dart source - notably, there's some code that must be initialized to add globals like `require` and `process` to the global scope. A node.js specific preamble can be found [here](https://github.com/pattobrien/dart3_node_interop/blob/main/build/node_preamble.js).

## Issue Description

It should be possible for devs to define some `js` code to be inserted along side the generated Dart code. This would make it a lot easier for developers to stay within the dart ecosystem, even when interop-ing with various configurations of js code.

Not being able to do so has caused interop package authors to fork the entire `build_web_compilers` package in order to work around the limitations and apply a few basic customizations. Such packages are:

- [jaspr](https://github.com/schultek/build/tree/master/jaspr_web_compilers)
- [node_interop](https://github.com/pulyaevskiy/node-interop) - now deprecated
- [firebase_node](https://pub.dev/packages/firebase_node) - now deprecated

I'm curious if such functionality already exists, or if it's something that would need to be added to the builders.

### Additional Info

* Dart SDK version: 3.3.0 (stable) (Tue Feb 13 10:25:19 2024 +0000) on "macos_arm64"
* `build_web_compilers v4.0.9`

jakemac53:
This package does not support compiling to node. Ultimately, nobody on the team has any expertise in node or even uses it at all. So we can't reasonably support it.

Previously, there was a community `build_node_compilers` package but it looks to be quite out of date now.

If we could resolve this through some general mechanism to include a pre-amble as you suggest, that would be interesting though. I don't know exactly what it would look like, but if you wanted to work on a contribution we could review that.

pattobrien:
>  if you wanted to work on a contribution we could review that.

Absolutely! I'd love to take a swing at it.

I have a working Dart VSCode extension prototype, which runs in a Node runtime - so I can confirm its not only possible, but was generally surprised by how interoperable it is considering its a completely different environment. The implementation for preamble insertion is somewhat dependent on [module-scoped exports](https://github.com/dart-lang/sdk/issues/55029), so I'm going to gather some feedback there and will open a PR once I have a good direction to propose.

PR #3737, opened 2024-08-21
Compile with dart2js and dart2wasm

simolus3:
This adds support for compiling with multiple compilers, typically with dart2js and dart2wasm. The [updated readme](https://github.com/simolus3/build/blob/multi-compilation/build_web_compilers/README.md#configuration) explains how this works, but this essentially boils down to:

```yaml
targets:
  $default:
    builders:
      build_web_compilers:entrypoint:
        options:
          compilers:
            dart2js:
              args:
                - O2
            dart2wasm:
              args:
                - O3
```

This doesn't change the defaults yet, we continue to use ddc for dev builds and dart2js for release builds.

Closes https://github.com/dart-lang/build/issues/3730.


build_web_compilers/lib/src/loader.js
```diff
@@ -0,0 +1,40 @@
...
+
+(async () => {
+  if (supportsWasmGC()) {
+    let { instantiate, invoke } = await import("./$basename.mjs");
+
+    let modulePromise = WebAssembly.compileStreaming(fetch("$basename.wasm"));
+    let instantiated = await instantiate(modulePromise, {});
+    invoke(instantiated, []);
+  } else {
+    const scriptTag = document.createElement("script");
```
simolus3, line 35: I think we can use `import()` here as well if we wanted to. Should be worried about [not supporting these browsers](https://caniuse.com/?search=import()) at all if we switch to `import`?

pattobrien:
Any chance that this work also would/could allow for custom entrypoints / preambles, e.g. for node js runtimes? 

https://github.com/dart-lang/build/issues/3652

simolus3:
I agree that we should support custom entrypoints, I'm not sure about the best way to allow that though. I've added a simple template as a builder option in aac8b6a7d6ad333799d21f7144e0648c295d020b, not sure if that's the best approach though.

build_web_compilers/build.yaml
```diff
@@ -138,7 +138,7 @@ builders:
         dart2js_args:
           - --enable-asserts
       release_options:
-        compiler: dart2js
+        compiler: both
```
jakemac53, line 141: I think we want something more specific for this name, which might end up being pretty verbose, but it is what it is. Possibly, `dart2wasm+dart2js` or something.

build_web_compilers/lib/src/dart2js_bootstrap.dart
```diff
@@ -21,23 +21,37 @@ import 'platforms.dart';
...
+/// When [generateEntrypoint] is `false` (it defaults to `true`), the generated
+/// asset is named `.dart.bootstrap.js` instead of `.dart.js`. This is used by
+/// the entrypoint builder to invoke both the dart2js and the dart2wasm
+/// compilers and generate a custom entrypoint script detecting supported
+/// browser features to load the relevant script.
 Future<void> bootstrapDart2Js(
   BuildStep buildStep,
   List<String> dart2JsArgs, {
   required bool? nativeNullAssertions,
+  bool generateEntrypoint = true,
```
jakemac53, line 34: nit: maybe just have this be a more general `String outputExtension = '.dart.js'`. I would also probably make the fallback extension be `.dart.app.js` or something, and not re-use the `.dart.bootstrap.js` extension - it isn't really a bootstrap it is the actual compiled app 🤷‍♂️.

Another idea would be to allow the user to control the extension per compiler entirely. So, maybe the  `compiler` option starts allowing a map of compiler name to output extension, for the "bootstrap" js file for that compiler? If only a name of a compiler is given, then it defaults to `{ <compiler>: '.dart.js' }`.

This leaves the runtime logic of how to choose which version of the app to use more up to the user, which some users might need. And we could provide a generic JS file which can be included otherwise for trivial setups?

build_web_compilers/lib/src/loader.js
jakemac53: cc @sigmundch do you know what our story is on browser compatibility nowadays?

build_web_compilers/lib/src/loader.js
```diff
@@ -0,0 +1,40 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+// For apps compiled both with dart2js and dart2wasm, this entrypoint script
+// is responsible for detecting browser features and then loading either the
+// wasm or the JS bundle depending on which features are supported.
+//
+// Note: build_web_compiler loads a minified version of this file. To re-generate
+// it, run:
+//
+// esbuild lib/src/loader.js --bundle "--external:*.mjs" --format=esm --minify > lib/src/loader.min.js
```
jakemac53, line 12: I don't know what `esbuild` is or how to set it up :). Ideally I would like to avoid adding dependencies like this into the build process.  This file is only 40 lines so it is probably fine to just include it unminified?

build_web_compilers/lib/src/dart2js_bootstrap.dart
simolus3: I like the option of being able to customize the extension on a per-compiler basis. Making the generic JS loader script optional is also nice, since it allows for more flexibility instead of the static builder option template.

> And we could provide a generic JS file which can be included otherwise for trivial setups?

Should we enable this with the option as well? So that users that want our default loader would do:

```
  compilers:
    dart2js: '.dart2js.js'
    dart2wasm: '.dart2wasm.mjs'
    loader: '.dart.js'
```

And users who need their own entrypoint would simply remove the `loader` entry from `compilers`? This is pushing the `compilers` option a little, but if we get the default right most users hopefully don't have to worry about it too much?

jakemac53: I think I would prefer to move the `loader` config out to a separate option, and not have it be a part of `compilers`? And then it can have its own sane default depending on the `compilers` configuration:

- **dart2js**: none
- **dart2wasm**: '.dart.js'
- **dart2js** _and_ **dart2wasm**: '.dart.js'

simolus3: Should we move the `dart2js_args` and `dart2wasm_args` options into the compilers map as well? (while still supporting the old mode as a fallback of course)? Then we could have:

```yaml
compilers:
  dart2js:
    extension: ".dart2js.js"
    args: ['-O4']
  dart2wasm:
    extension: ".dart.mjs"
```

I'm asking because then we should probably design values in the `compilers` map to be objects instead of strings right away, which may also be useful if we come up with other compiler-specific options later.

jakemac53: That seems pretty reasonable to me

build_web_compilers/lib/src/loader.js
sigmundch: No concerns in terms of browser compatibility. In fact, this same loader.js file uses an `import` a few lines above to load dart2wasm's `.mjs` file. That said, the file that is being loaded is not an ES6 module. Dart2js is not producing code in that module format. As such, I'd be concerned that it may not behave as expected if it were loaded via an ES6 module `import`, instead of adding a script tag to the page.

build_web_compilers/build.yaml
```diff
@@ -132,16 +132,19 @@ builders:
...
-        dart2wasm_args:
-          - --enable-asserts
-        dart2js_args:
-          - --enable-asserts
+        compilers:
+          dartdevc:
+            extension: '.dart.js'
       release_options:
-        compiler: dart2wasm+dart2js
+        compilers:
```
jakemac53, line 140: @sigmundch @kevmoo not sure who to talk to about this, but what do we want the defaults to be here? This controls the default release build configuration for all dart web apps.

sigmundch: Just to clarify, do you mean the default output extensions for each compilation? default flags? both?

build_web_compilers/lib/src/loader.js
sigmundch: agree, probably not worth minifying at this time (esbuild is popular and fast JS-to-JS compiler/minifier, see https://esbuild.github.io/)

build_web_compilers/build.yaml
simolus3: > Just to clarify, do you mean the default output extensions for each compilation? default flags? both?

Both - we'd compile with `dart2js` (which we instruct to emit a `.dart2js.js` file here, we don't rename additional files for deferred imports), `dart2wasm` (outputs `.wasm`, `.dart2wasm.mjs`) and a custom entrypoint (`.dart.js`) running feature detection for WasmGC and then loading either the `.dart2js.js` script or the `.dart2wasm.mjs` module.

> This controls the default release build configuration for all dart web apps.

Although I suspect most users do override the default here - `build_web_compilers` never passed any optimization flags to `dart2js` by default.

build_web_compilers/lib/src/loader.js
simolus3: > In fact, this same loader.js file uses an `import` a few lines above to load dart2wasm's `.mjs` file

It only does that if WasmGC is supported, so since no browsers support WasmGC but not `import`, it would just look like a function call to them I guess?

> As such, I'd be concerned that it may not behave as expected if it were loaded via an ES6 module `import`, instead of adding a script tag to the page.

That makes sense, thanks! I'll leave the script tag generation then.

build_web_compilers/build.yaml
simolus3: A downside of dart2wasm by default is that existing apps importing `dart:html` will see failed builds, so maybe we should use `dart2js` for now and enable both compilers in a future release after the legacy web interop libraries have been removed from the SDK?

sigmundch: Yeah, I share that concern - unless we can determine upfront that the app has no uses of `dart:html`, then we could have an adaptive default.

That said, do we also offer a mode that is dart2wasm only? or only the combo wasm+js? Given that browser support is not there yet for wasm-gc, I'd prefer that we only offer the combo, so JS can be used as fallback for those browsers.

simolus3: > That said, do we also offer a mode that is dart2wasm only? 

We do as of https://github.com/dart-lang/build/pull/3727. This PR adds support for enabling multiple compilers.

> I'd prefer that we only offer the combo, so JS can be used as fallback for those browsers.

Personally I would like to be able to compile with dart2wasm only for testing purposes (e.g. in integration tests that should run with dart2js once and then with dart2wasm). These tests failing when running on unsupported browsers is better than testing dart2js twice. Perhaps we can emit a warning in the builder when only dart2wasm is enabled?

jakemac53: Ok, lets use dart2js by default, and keep the other options as opt in

_test/test/dart2wasm_integration_test.dart
```diff
@@ -59,6 +59,19 @@ void main() {
...
+      await expectTestsPass(
+        usePrecompiled: true,
+        buildArgs: [
+          '--release',
+          '--config=both',
+          '--output=${d.sandbox}',
+        ],
+        testArgs: _testArgs,
+      );
+      await _expectWasCompiledWithDart2Wasm();
```
jakemac53, line 73: This should probably also check that the dart2js file was created

build_web_compilers/README.md
```diff
@@ -120,16 +265,13 @@ targets:
...
-          - -DANOTHER_VAR=true
+          - -O2
 ```
 
-These may also be specified on the command line with a `--define` argument.
-
-```sh
-webdev serve -- '--define=build_web_compilers:ddc=environment={"SOME_VAR":"changed"}'
-```
+Note that the `compiler` option takes precedence over the new `compilers` and
```
jakemac53, line 273: We should probably make this an error during option parsing

build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -26,16 +28,33 @@ const mergedMetadataExtension = '.dart.ddc_merged_metadata';
...
+  /// The [EnabledEntrypointCompiler.extension] to use by default when multiple
+  /// compilers are enabled.
+  ///
+  /// This can't be `.dart.js` for multiple web compilers because we'd then run
+  /// into duplicate outputs being generated.
+  final String entrypointExtension;
+
+  const WebCompiler(
+    this.optionName,
+    this.entrypointExtensionWhenOnlyCompiler,
```
jakemac53, line 55: I would make these named, it isn't clear looking at a constructor invocation which is which, especially since they are the same type

build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -89,69 +95,152 @@ class WebEntrypointBuilder implements Builder {
...
+        ));
+
+        hasDart2Wasm |= compiler == WebCompiler.Dart2Wasm;
+      }
+
+      if (hasDart2Wasm) {
+        defaultLoaderOption = '.dart.js';
+      }
+    } else {
+      var compilerName = config[compilerOption] as String? ?? 'dartdevc';
```
jakemac53, line 184: I am a bit surprised that we have a default of `dartdevc` here, we shouldn't need that I don't think? Can we ever get here without any compiler being set, given the default options?

simolus3: This caught be by surprise as well, but the `?? 'dartdevc'` has been there before and there is no `compiler: dartdevc` in `build.yaml` under `dev_options`.

Should I remove the fallback from here and add a default to `build.yaml`?

build_web_compilers/README.md
simolus3: The readme is outdated now, I've changed it so that `compilers` takes precedence.

This is a bit unfortunate, but we can't make this an error because the keys from the user's `build.yaml` are merged with the defaults in `build_web_compilers` - so we can't assume that something's off just because both are set.
The current approach is to use the existing `compiler` option in the defaults (so that `compilers` never appears in the map unless explicitly enabled) and then have `compilers` take precedence.

I wonder if we should unify these into a single option, where `compiler` is either a string (old) or a map (new)? That makes it harder to be in an ambiguous state.

build_web_compilers/lib/src/web_entrypoint_builder.dart
jakemac53: > and there is no `compiler: dartdevc` in `build.yaml` under `dev_options`.

Oh ok, not sure why I did it that way 🤣

> Should I remove the fallback from here and add a default to build.yaml?

Lets just leave it as is, I don't want to accidentally break something.

jakemac53 (approved):

build_web_compilers/README.md
jakemac53: I think how you have it is fine. The plural reads nicer now that multiple can be configured, and it is a cleaner break.

jakemac53:
@simolus3 let me know when this is ready to merge from your perspective and I will push the button!

simolus3:
Nothing comes to mind, this is ready from my side :+1:

kevmoo:
I couldn't wait!
Thanks SO MUCH!

kevmoo:
Gah! I shouldn't have merged before trying it

1. There is a cast error w/ Yaml I'm hitting. I have a fix.
2. Loading JS fails in Safari with (wasm loads fine in Chrome and Safari Tech preview)

```
[Error] Unhandled Promise Rejection: ReferenceError: Can't find variable: joinPathSegments
	resolveUrlWithSegments (main.dart.js:3)
	(anonymous function) (main.dart.js:25)
	(anonymous function) (main.dart.js:1)
	Global Code (main.dart.js:29)
```

kevmoo:
See https://github.com/dart-lang/build/issues/3746 @simolus3 !

kevmoo:
Also https://github.com/dart-lang/build/issues/3747 @simolus3

kevmoo:
We're close!!!

PR #3740, opened 2024-08-23
Write SDK summary to a temp file and rename

natebosch:
Avoid writing partial output to the file that could be concurrently read
by another isolate in the same process. Create a temp directory to
incrementally write the file, then use `rename` to atomically move it to
the location where other isolates may try to read it.

It's best to still avoid using the SDK summary from and analysis driver
in multiple isolates, but this should reduce flakiness in some cases
where they do run concurrently.


kevmoo (approved):
Thanks!!

jakemac53 (approved):

PR #3741, opened 2024-08-23
Limit which health checks run

natebosch:
The `breaking` and `leaking` health checks are not currently compatible
with repos using pub workspaces.

Keep `coverage` for now since it's not obvious it's a problem. We may
want to revisit duplicate testing between coverage and the other test
configurations.

Run `version`, `changelog`, and `do-not-submit` checks which are
sensible for every package.


kevmoo (approved):

kevmoo:
I'm going to land this and then rebase the wasm PR over it!

PR #3742, opened 2024-08-23
Account for the snapshot of dartdevc in the Dart SDK changing from a JIT snapshot to an AOT snapshot

a-siva:
Dart SDK is in the process of changing snapshots in the SDK from JIT snapshots to AOT snapshots.
This change ensures that the package will continue to work when that change happens.

Note: Having the package reaching into the snapshot using the Dart SDK layout makes it brittle as things will break if the Dart SDK changes the SDK layout, instead we need to add a Dart CLI command that will invoke the dartdevc compiler and have this package use the Dart CLI for invoking dartdevc.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 4.0.12-wip
 
 - Bump the min sdk to 3.5.0.
+- Account for dartdevc snapshot in the Dart SDK changing to an AOT snapshot.
 
 ## 4.0.11
 
--- a/build_web_compilers/lib/src/sdk_js_compile_builder.dart
+++ b/build_web_compilers/lib/src/sdk_js_compile_builder.dart
@@ -86,8 +86,17 @@ Future<void> _createDevCompilerModule(
   try {
     // Use standalone process instead of the worker due to
     // https://github.com/dart-lang/sdk/issues/49441
-    result = await Process.run(p.join(sdkDir, 'bin', 'dart'), [
-      p.join(sdkDir, 'bin', 'snapshots', 'dartdevc.dart.snapshot'),
+    var snapshotPath = p.join(
+        sdkDir, 'bin', 'snapshots', 'dartdevc_aot.dart.snapshot');
+    var execSuffix = Platform.isWindows ? '.exe' : '';
+    var dartPath = p.join(sdkDir, 'bin', 'dartaotruntime$execSuffix');
+    if (!File(snapshotPath).existsSync()) {
+      snapshotPath = p.join(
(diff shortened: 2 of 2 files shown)


jakemac53:
> Note: Having the package reaching into the snapshot using the Dart SDK layout makes it brittle as things will break if the Dart SDK changes the SDK layout, instead we need to add a Dart CLI command that will invoke the dartdevc compiler and have this package use the Dart CLI for invoking dartdevc.

I agree - and same applies for frontend_server and the modular kernel compiler (I can't recall which snapshot actually does that). Having a supported entry point to these services would be great.

jakemac53:
@a-siva note that our current SDK constraint is `>=3.5.0 <3.7.0`, which would allow `3.6.0`. Is this change intended to land in 3.6.0? Or more specifically, will the old snapshot be present in 3.6.0?

If the old snapshot is removed in 3.6.0, then the current version will be selectable but won't function. We might want to mention that in the SDK Changelog or something, to try and reduce the number of issues filed (people should only need to `dart pub upgrade` to be fixed). The main risk is if they are stuck on the old version for some reason, and can't upgrade.

jakemac53:
Fwiw, if we are willing to commit to a non-breaking API surface for the command line arguments, we could possibly also drop the tight SDK pinning here.

a-siva:
> @a-siva note that our current SDK constraint is `>=3.5.0 <3.7.0`, which would allow `3.6.0`. Is this change intended to land in 3.6.0? Or more specifically, will the old snapshot be present in 3.6.0?
> 
> If the old snapshot is removed in 3.6.0, then the current version will be selectable but won't function. We might want to mention that in the SDK Changelog or something, to try and reduce the number of issues filed (people should only need to `dart pub upgrade` to be fixed). The main risk is if they are stuck on the old version for some reason, and can't upgrade.
> 
My intent was to delete the old snapshot in the next beta release of 3.6, I see that this could be an issue for folks who are 
stuck on an older version of the package and can't upgrade.

> Or, we keep both snapshots around in 3.6, and delete the old one in 3.7.
This is a possibility, increases the size of the sdk but I guess that is a trade off we pay for ensuring older packages like this continue to work.

The ideal solution for this would be to add a command
`'dart compile devjs .....' 
`which would remove such dependencies on the sdk.

jakemac53:
> My intent was to delete the old snapshot in the next beta release of 3.6, I see that this could be an issue for folks who are
> stuck on an older version of the package and can't upgrade.

Just to be clear I would not necessarily be opposed to doing this anyways. I just wanted to make you aware of the breakage, so you can take it into consideration when deciding when to remove it.

jakemac53 (approved):

PR #3744, opened 2024-08-27
Remove pubspec.lock files

natebosch:
I don't think these are read by any tool. We have checked in
`package_config.json` files instead.


natebosch:
I was wrong, we do still use them. It's how we tell what type of dependency a package was.

Issue #3747, opened 2024-09-10
New wasm compile fails with JS fallback


kevmoo:
At https://github.com/dart-lang/build/commit/49d83f489236073212ce4c7ba8c6c19ec1a1037b

with this build.yaml
```yaml
# See https://github.com/dart-lang/build/tree/master/build_web_compilers#configuration
targets:
  $default:
    builders:
      build_web_compilers|dart2js_archive_extractor:
        options:
          filter_outputs: false
      build_web_compilers|entrypoint:
        generate_for:
          - web/**.dart
        options:
          compilers:
            dart2wasm:
              args:
                - -O4
                - --no-strip-wasm
            dart2js:
              args:
                - --dump-info
                - --no-source-maps
                - --show-package-warnings
                - -O4
```

(and a fix for https://github.com/dart-lang/build/issues/3746)

We see a runtime crash in safari (JS mode)

```
[Error] Unhandled Promise Rejection: ReferenceError: Can't find variable: joinPathSegments
	resolveUrlWithSegments (main.dart.js:3)
	(anonymous function) (main.dart.js:25)
	(anonymous function) (main.dart.js:1)
	Global Code (main.dart.js:29)
```

Issue #3746, opened 2024-09-10
Wasm compile crashes as build time


kevmoo:
At https://github.com/dart-lang/build/commit/49d83f489236073212ce4c7ba8c6c19ec1a1037b

with this build.yaml
```yaml
# See https://github.com/dart-lang/build/tree/master/build_web_compilers#configuration
targets:
  $default:
    builders:
      build_web_compilers|dart2js_archive_extractor:
        options:
          filter_outputs: false
      build_web_compilers|entrypoint:
        generate_for:
          - web/**.dart
        options:
          compilers:
            dart2wasm:
              args:
                - -O4
                - --no-strip-wasm
            dart2js:
              args:
                - --dump-info
                - --no-source-maps
                - --show-package-warnings
                - -O4
```

get this error

```
type 'YamlMap' is not a subtype of type 'Map<String, Object?>?' in type cast
dart:_internal                                                      MappedIterator.moveNext
package:build_web_compilers/src/web_entrypoint_builder.dart 171:62  new EntrypointBuilderOptions.fromOptions
package:build_web_compilers/src/web_entrypoint_builder.dart 279:58  new WebEntrypointBuilder.fromOptions
package:build_web_compilers/builders.dart 16:26                     webEntrypointBuilder
package:build_runner                                                BuildCommand._run
package:args/command_runner.dart 212:13                             CommandRunner.runCommand
package:build_runner                                                run
.dart_tool/build/entrypoint/build.dart 145:16                       main
```

CC @jakemac53

kevmoo:
Here's a patch with the fix. Of course it'll need a test

```diff
diff --git a/build_web_compilers/lib/src/web_entrypoint_builder.dart b/build_web_compilers/lib/src/web_entrypoint_builder.dart
index b3bad5a4..a880d5c6 100644
--- a/build_web_compilers/lib/src/web_entrypoint_builder.dart
+++ b/build_web_compilers/lib/src/web_entrypoint_builder.dart
@@ -10,6 +10,7 @@ import 'package:build/build.dart';
 import 'package:build_modules/build_modules.dart';
 import 'package:collection/collection.dart';
 import 'package:path/path.dart' as p;
+import 'package:yaml/yaml.dart';
 
 import 'common.dart';
 import 'dart2js_bootstrap.dart';
@@ -163,17 +164,17 @@ final class EntrypointBuilderOptions {
     // dart2js + dart2wasm). Since the default builder configuration doesn't
[...]

PR #3748, opened 2024-09-10
Fix build web compilers wasm issue

simolus3:
This fixes three issues from the dart2js+dart2wasm PR:

1. Since `YamlMap` is not a `Map<String, dynamic>`, `.cast()` doesn't work. Tests didn't catch this because we weren't reading from yaml.
2. The loader script had multiple issues - it referenced an undefined function and assumes that `WebAssembly` is defined. Tests didn't catch this because this was only exercised in the fallback path, and we were only testing on Chrome. I've fixed the loader script and added a custom browser platform that disables WebAssembly in Chrome to make sure the fallback path is working.
3. We've stopped emitting dart2js archives because we expected them to look like `main.dart.js*`. When using a custom build extension for dart2js, they are actually named `main.dart2js.js*` - I've fixed that, this is tested by the deferred imports hello world test in `_test`.

Closes https://github.com/dart-lang/build/issues/3746.
Closes https://github.com/dart-lang/build/issues/3747.

### DIFF
--- a/_test/dart_test.yaml
+++ b/_test/dart_test.yaml
@@ -9,6 +9,13 @@ tags:
   # moment, but want to avoid warnings from the test runner about using undefined
   # targets.
 
+define_platforms:
+  chrome_without_wasm:
+    name: chrome_without_wasm
+    extends: chrome
+    settings:
+      arguments: "--js-flags=--noexpose_wasm"
+
 override_platforms:
   chrome:
     settings:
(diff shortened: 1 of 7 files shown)


kevmoo:
This seems to work everywhere: https://j832-com--wasm-check-wzo4ep4u.web.app/

jakemac53 (approved):

Issue #3751, opened 2024-09-14
[build_web_compilers] wasm compilation bug with non-root entrypoint


schultek:
## Description

Having a Dart web project with an entrypoint that is in a subfolder (e.g. '/web/app/main.dart') instead of the root ('/web/main.dart') fails with the new wasm compiler.

The issue is that the `main.dart.js` always fetches the `.wasm` file as if it would exist in the root:

<img width="662" alt="Bildschirmfoto 2024-09-14 um 14 06 13" src="https://github.com/user-attachments/assets/addd0766-3c68-40de-a5c0-d61c4094a00b">

Which results in: 

<img width="850" alt="Bildschirmfoto 2024-09-14 um 14 07 04" src="https://github.com/user-attachments/assets/262eedaf-e22c-4626-b33c-823b42b01480">

- The `main.dart.js` and `main.mjs` are correctly fetched from their sub-path.
- The `.wasm` file is correctly generated at `/app/main.wasm`, but tried to be fetched from the root.

## Steps to reproduce

1. Create new web project with `dart create -t web test`
2. Update depencency to `build_web_compilers: ^4.1.0-beta.0`
3. Add `build.yaml` with 
  ```yaml
  global_options:
  build_web_compilers:entrypoint:
    options:
      compiler: dart2wasm
  ```
4. Move `web/main.dart` to `web/app/main.dart` and adjust the script path in `index.html` accordingly
5. Run `dart run build_runner serve` and open `localhost:8080`  

## Context:

- Dart SDK version: 3.6.0-218.0.dev (dev) (Thu Sep 5 05:03:17 2024 -0700) on "macos_arm64"
- Package: build_web_compilers: ^4.1.0-beta.0

Tagging:

@kevmoo since this could also be a general wasm compilation issue
@simolus3 since you worked on this

PR #3752, opened 2024-09-15
Fix relative imports in wasm loader

simolus3:
Currently, the loader script emitted by `build_web_compilers` resolves URLs against the base-URL of the active document. This breaks when the entrypoint script is not in the same directory as the page, e.g. in

```
web/
  index.html
  src/
    app.dart
```

To fix this, this PR changes the logic to resolve URLs against the loader script (which would be `/src/app.dart.js` in the example, properly resolving the other dart2wasm/dart2js components).
I've also added an integration test using a `<script>` tag pointing to a subdirectory to make sure that this is working now.

Closes https://github.com/dart-lang/build/issues/3751


jakemac53:
Will take a look at this early next week, we had an off-site this week

build_web_compilers/pubspec.yaml
```diff
@@ -1,5 +1,5 @@
 name: build_web_compilers
-version: 4.1.0-beta.0
+version: 4.1.0-dev
```
kevmoo, line 2: We should go back to using `-wip` instead of `-dev`

simolus3:
With the last commit I'm convinced the test failure is probably a flake or something else not caused by me.

```
test\resolve_source_test.dart: resolveSource can resolve with specified language versions from a PackageConfig (failed)
  PathAccessException: Cannot open file, path = 'D:\a\build\build\build_test\.dart_tool\build_resolvers\sdk.sum' (OS Error: The process cannot access the file because it is being used by another process.
  , errno = 32)
  dart:io                                                 _File.readAsBytesSync
  package:build_resolvers/src/analysis_driver.dart 33:43  analysisDriver
  package:build_resolvers/src/resolver.dart 460:26
```

jakemac53:
> With the last commit I'm convinced the test failure is probably a flake or something else not caused by me.

That looks familiar, I believe that when build_resolvers is used by multiple tests it can happen flakily. cc @natebosch do you remember if we had a solution here? I thought we did something

natebosch:
> cc @natebosch do you remember if we had a solution here? I thought we did something

In #3740 we made a change where the file gets written to a separate path and then renamed. This should prevent any situation where the file gets read when it is partially written. It should also prevent any situation where we are trying to read that file when any outside process has it open, which is what this error looks like. The only contention can be from other isolates, or other `build_runner` instances, and even there the `rename` should be synchronous.

I don't know if I've seen an error from that particular stack trace before. This might be one we want to keep an eye on and investigate if it's failing frequently.

simolus3:
> some tests with `@TestOn('browser')` need to exclude wasm it looks like. Or, be re-written to be wasm compatible.

Only the new `subdir_source_test.dart` was failing. I think it's because it includes multiple entrypoints (where one is from a subdirectory relative to the test directory, the thing we want to test). But DDC (our probably rather our loader for DDC?) doesn't support multiple entrypoints attached to the same website. We need multiple entrypoints because one is the test (always in the same directory) and another one is in a subdirectory.

simolus3:
@jakemac53 Wasm tests on `master` got broken by https://github.com/dart-lang/sdk/commit/f98a84a3b13cf008c3c636a477b5d7f834d2c9db, I've pushed a fix in d3246e2f0091b168c7d1c389f6589b2f9f601b8d.

build_runner_core/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 7.3.3-wip
 
 - Bump the min sdk to 3.5.0.
+- Fix internal analysis warnings.
```
kevmoo, line 4:
```suggestion
```

Not needed. Only need to document user-visible changes.

kevmoo:
Holy cow! 🐮  we're green! 💚

jakemac53 (approved):

PR #3760, opened 2024-10-02
add the breaking API change health check

jakemac53:
Checking to see if this still is broken for workspace repos


davidmorgan:

PR #3762, opened 2024-10-08
Constrain build_web_compilers upper sdk bound

jakemac53:
This was accidentally removed (but luckily not published)


jakemac53:
Actually just going to lump this with another PR

PR #3763, opened 2024-10-08
Fix CI jobs

jakemac53:
This is an absolute mess now, our workspace is super fragmented at this point, but in general this:

- Fixes new analyzer deprecations
- Requires the new analyzer
  -  Which requires a new SDK
    -  Which requires moving more packages out of the workspace
- Fixes the jobs that run against `main` by allowing 3.7.0 dev SDKs. These versions will be broken at some point because the DDC binary is going to move, so we are not allowing 3.7.0 stable.

### DIFF
--- a/build/test/builder/build_step_impl_test.dart
+++ b/build/test/builder/build_step_impl_test.dart
@@ -131,7 +131,7 @@ void main() {
         expect(aLib.definingCompilationUnit.libraryImports.length, 2);
         expect(
             aLib.definingCompilationUnit.libraryImports
-                .any((import) => import.library.name == 'b'),
+                .any((import) => import.importedLibrary!.name == 'b'),
             isTrue);
 
         var bLib = await resolver.findLibraryByName('b');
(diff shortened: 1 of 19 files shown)


jakemac53:
cc @scheglov do you know what min analyzer version I need for these new apis (if any)?

scheglov:
> cc @scheglov do you know what min analyzer version I need for these new apis (if any)?

I think `6.9.0` should be enough, older than that not sure.

scheglov (approved):

build/test/builder/build_step_impl_test.dart
```diff
@@ -128,13 +128,15 @@ void main() {
 
         var aLib = await resolver.libraryFor(primary);
         expect(aLib.name, 'a');
-        expect(aLib.importedLibraries.length, 2);
-        expect(aLib.importedLibraries.any((library) => library.name == 'b'),
+        expect(aLib.definingCompilationUnit.libraryImports.length, 2);
+        expect(
+            aLib.definingCompilationUnit.libraryImports
+                .any((import) => import.library.name == 'b'),
```
jakemac53, line 134: @scheglov This test is failing now, and if I do `print(aLib.definingCompilationUnit.libraryImports.map((l) => l.name).toList());` I get `[a, a]`. Any ideas?

jakemac53: Nvm figured it out, I need to use `importedLibrary` not `library`.

PR #3768, opened 2024-10-30
Use the 'dart compile js-dev' command for invoking DDC …

a-siva:
Use the 'dart compile js-dev' command for invoking DDC instead of reaching into the dart-sdk and using the snapshot which is an internal implementation detail of the dart sdk.

### DIFF
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -7,7 +7,7 @@ repository: https://github.com/dart-lang/build/tree/master/build_modules
 resolution: workspace
 
 environment:
-  sdk: '>=3.6.0 <3.8.0-z'
+  sdk: '>=3.7.0-82.0.dev <3.8.0-z'
 
 dependencies:
   analyzer: '>=5.1.0 <8.0.0'
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -7,7 +7,7 @@ repository: https://github.com/dart-lang/build/tree/master/build_modules
 resolution: workspace
 
 environment:
-  sdk: '>=3.7.0-81.0.dev <3.8.0'
+  sdk: '>=3.7.0-81.0.dev <3.8.0-z'
 
 dependencies:
   analyzer: '>=5.1.0 <8.0.0'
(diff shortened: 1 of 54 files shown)


build_modules/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ repository: https://github.com/dart-lang/build/tree/master/build_modules
 resolution: workspace
 
 environment:
-  sdk: '>=3.5.0 <3.7.0-z'
+  sdk: '>=3.7.0-82.0.dev <3.8.0-z'
```
jakemac53, line 10: We can only depend on released dev versions so we might need to wait until the next release (3.7.0-81.0.dev is the current most recent one)

jakemac53, line 10: Fwiw I think we can drop the `-z` off the `<3.8.0-z` now as well. That was a short term thing to allow 3.7 dev releases but not stable.

jakemac53:
Once we have a released dev SDK to depend on, I can update the CI jobs etc and push that here.

jakemac53:
cc @a-siva @nshahan something seems to be broken with the `dart compile js-dev` command on windows? I am not getting a lot of useful information from the GitHub logs though.

a-siva:
> cc @a-siva @nshahan something seems to be broken with the `dart compile js-dev` command on windows? I am not getting a lot of useful information from the GitHub logs though.

Will investigate, looks like tests are waiting for temp directories to get deleted but they are in use, could be a test issue.

jakemac53:
> Will investigate, looks like tests are waiting for temp directories to get deleted but they are in use, could be a test issue.

I think that possibly the worker isn't shutting down like it normally would.

jakemac53:
Note that we just send it a sigterm by calling `Process.kill`. I don't know how the `dart` tool deals with that, and if it can handle shutting down its sub-processes etc.

a-siva:
> Note that we just send it a sigterm by calling `Process.kill`. I don't know how the `dart` tool deals with that, and if it can handle shutting down its sub-processes etc.

It does not handle shutting down sub-processes, if that is the issue we should also see errors on other platforms (linux, mac)

jakemac53:
> It does not handle shutting down sub-processes, if that is the issue we should also see errors on other platforms (linux, mac)

Well, only windows won't allow deleting a directory because some sub-process has file handles in that directory still open. So it is possible that is the only reason we don't see the error on other platforms. If the process is left dangling though that is of course bad.

a-siva:
Closing this PR in favor of a new PR which would be done post the reformat PR.

PR #3771, opened 2024-11-15
Batch writes

simolus3:
Another attempt for https://github.com/dart-lang/build/issues/3321, my [previous PR](https://github.com/dart-lang/build/pull/3418) for this got stale, this one is a bit simpler.

This attempts to solve the issue of `build_runner` and external tools watching for file system events (like the analysis server) not playing too well together. At the moment, a common issue is that, as a user runs `build_runner build` and the generated assets are written as they are generated, the analysis server will see many file changes and keep re-analyzing sources in an incomplete state.
By making `build_runner` cache pending writes in memory and then flushing them after a completed build, we'll likely get all changes to disk before they're picked up by external tools, reducing friction.

I've implemented this at a fairly low level (the raw file readers and writers making up the default `IOEnvironment`). They have to opt-in to batches by checking for the presence of a zone key when reading or writing assets. We support batches by default when not in low-resources mode. A batch runs in a zone wrapping a whole build run.

A todo is that I need to write integration tests for this, but it would be good to get a review for the overall approach before completing this.

Closes https://github.com/dart-lang/build/issues/3321.

### DIFF
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -106,19 +106,17 @@ Future<BuildResult> build(List<BuilderApplication> builders,
   );
   var terminator = Terminator(terminateEventStream);
   try {
-    return await runWithFileSystemBatch(() async {
-      var build = await BuildRunner.create(
-        options,
-        environment,
-        builders,
-        builderConfigOverrides!,
-        isReleaseBuild: isReleaseBuild!,
-      );
-      var result = await build
-          .run({}, buildDirs: buildDirs!, buildFilters: buildFilters!);
-      await build.beforeExit();
-      return result;
-    }, environment.writer);
+    var build = await BuildRunner.create(
+      options,
+      environment,
+      builders,
+      builderConfigOverrides,
+      isReleaseBuild: isReleaseBuild,
+    );
+    var result =
+        await build.run({}, buildDirs: buildDirs, buildFilters: buildFilters);
+    await build.beforeExit();
+    return result;
   } finally {
     await terminator.cancel();
     await options.logListener.cancel();
--- a/build_runner/lib/src/generate/build.dart
+++ b/build_runner/lib/src/generate/build.dart
@@ -106,17 +106,19 @@ Future<BuildResult> build(List<BuilderApplication> builders,
   );
   var terminator = Terminator(terminateEventStream);
   try {
-    var build = await BuildRunner.create(
-      options,
-      environment,
-      builders,
-      builderConfigOverrides,
-      isReleaseBuild: isReleaseBuild,
-    );
-    var result =
-        await build.run({}, buildDirs: buildDirs, buildFilters: buildFilters);
-    await build.beforeExit();
-    return result;
+    return await runWithFileSystemBatch(() async {
+      var build = await BuildRunner.create(
+        options,
+        environment,
+        builders,
+        builderConfigOverrides!,
+        isReleaseBuild: isReleaseBuild!,
+      );
+      var result = await build
+          .run({}, buildDirs: buildDirs!, buildFilters: buildFilters!);
+      await build.beforeExit();
+      return result;
+    }, environment.writer);
   } finally {
     await terminator.cancel();
     await options.logListener.cancel();
--- a/build_runner_core/lib/src/asset/batch.dart
+++ b/build_runner_core/lib/src/asset/batch.dart
@@ -1,0 +1,191 @@
+// Copyright (c) 2024, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:build/build.dart';
+import 'package:glob/glob.dart';
+import 'package:meta/meta.dart';
+
+import '../environment/io_environment.dart';
+import 'reader.dart';
+import 'writer.dart';
+
+/// Runs [inner] in a zone that indicates to compatible [RunnerAssetWriter]s
+/// that written assets should not be written to the underlying file system
+/// directly, but rather in a batch after [inner] has completed.
+///
+/// This solves a common issue when `build_runner` and other tools such as an
+/// analysis server are running concurrently: During a build, generated assets
+/// are written one-by-one. The analyzer will discover these changes while the
+/// build is running and report lints for a partial state of the workspace where
+/// not all generated files are present, leading to a subpar experience.
+///
+/// [runWithFileSystemBatch] is intended to wrap a single build run. Instead of
+/// writing assets directly, they are first cached in memory and then written at
+/// once after the build has completed. This causes fewer invalidations for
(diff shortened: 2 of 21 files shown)


build_runner_core/lib/src/asset/batch.dart
```diff
@@ -0,0 +1,191 @@
...
+/// The default [IOEnvironment] uses readers and writes that are batch-aware
+/// outside of low-memory mode.
+@internal
+final class FileSystemWriteBatch {
+  final Map<AssetId, _PendingFileState> _pendingWrites = {};
+
+  FileSystemWriteBatch._();
+
+  Future<void> _completeWrites(RunnerAssetWriter writer) async {
+    await Future.wait(_pendingWrites.entries.map((e) async {
```
jakemac53, line 65: It likely doesn't matter, but I recently did some benchmarking related to `.entries` and it is definitely slower than iterating the keys and then grabbing out the value manually.

I will leave it up to you, either is fine, I am sure the file I/O dwarfs this cost anyways.

build_runner_core/lib/src/asset/batch.dart
```diff
@@ -0,0 +1,191 @@
...
+        return state.content!;
+      }
+    } else {
+      return await _inner.readAsBytes(id);
+    }
+  }
+
+  @override
+  Future<String> readAsString(AssetId id, {Encoding encoding = utf8}) async {
+    final bytes = await readAsBytes(id);
```
jakemac53, line 143: It is probably better to copy the state checking logic and call the inner `readAsString`, in case the inner reader is optimizing for that case.

build_runner_core/lib/src/asset/batch.dart
```diff
@@ -0,0 +1,191 @@
...
+@internal
+final class BatchAwareWriter extends RunnerAssetWriter {
+  final RunnerAssetWriter _inner;
+
+  BatchAwareWriter(this._inner);
+
+  FileSystemWriteBatch? _batchFor(AssetId id) {
+    // Don't batch writes for hidden files unlikely to be consumed by other
+    // tools.
+    if (id.pathSegments case ['.dart_tool', ...]) {
```
jakemac53, line 157: `id.pathSegments` is fairly expensive to do a lot, I would just check for a string prefix.

```suggestion
    if (id.path.startsWith('.dart_tool/') {
```

However, I am curious what prompted you to add this? It feels like it is hard coding in an assumption  that might not hold later on.

build_runner_core/lib/src/asset/batch.dart
```diff
@@ -0,0 +1,191 @@
...
+  }
+}
+
+@internal
+final class BatchAwareWriter extends RunnerAssetWriter {
+  final RunnerAssetWriter _inner;
+
+  BatchAwareWriter(this._inner);
+
+  FileSystemWriteBatch? _batchFor(AssetId id) {
```
jakemac53, line 154: This function feels like it is named improperly - it is always returning the same (current) batch.

Really, it is just checking if the path should be batched at all. Maybe make it return a `bool` instead and just use `FileSystemWriteBatch.current` explicitly after the check?

build_runner/lib/src/generate/build.dart
```diff
@@ -106,17 +106,19 @@ Future<BuildResult> build(List<BuilderApplication> builders,
...
-      environment,
-      builders,
-      builderConfigOverrides,
-      isReleaseBuild: isReleaseBuild,
-    );
-    var result =
-        await build.run({}, buildDirs: buildDirs, buildFilters: buildFilters);
-    await build.beforeExit();
-    return result;
+    return await runWithFileSystemBatch(() async {
```
jakemac53, line 109: I think I would prefer to add a new method to `RunnerAssetWriter` which gets invoked at the end of each build. Then we can avoid the zones and just use normal asset reader/writer overriding.

It would be breaking but I think worth it, breaking changes to `build_runner_core` are not really very painful.

build_runner_core/lib/src/asset/batch.dart
simolus3: >However, I am curious what prompted you to add this? It feels like it is hard coding in an assumption that might not hold later on.

My idea was to reduce the size of the in-memory batch by not including files that are unlikely to be observed by other tools directly (like hidden assets in `.dart_tool/build/generated`).
I can remove that if we want that behavior, or otherwise add a comment explaining the motivation for that check.

jakemac53: I think it probably just isn't worth doing - generally these build-to-cache files are going to eventually get read back by something else (and subsequently cached) anyways, so there won't actually be any extra memory used, and we can avoid re-reading the file this way.

Also, I believe the analyzer _does_ look in the build cache for files.

build_runner_core/lib/src/asset/batch.dart
simolus3: Counterintuitive, but I've changed it to iterate over keys with a separate lookup.

build_runner_core/lib/src/asset/batch.dart
```diff
@@ -0,0 +1,164 @@
...
+/// During a typical build run emitting generated files one-by-one, it's
+/// possible that other running tools such as an analysis server will have to
+/// re-analyze incomplete states multiple times.
+/// By storing pending outputs in memory first and then committing them at the
+/// end of the build, we have a better view over that needs to happen.
+///
+/// The default [IOEnvironment] uses readers and writes that are batch-aware
+/// outside of low-memory mode.
+@internal
+final class FileSystemWriteBatch {
```
jakemac53, line 28: It looks like this can just be a private class instead of using the `@internal` annotation

jakemac53 (approved):

jakemac53:
@simolus3 looks like some test utilites also need to implement the new method :)

simolus3:
I'll need someone to approve the workflow again :)

jakemac53:
Given this is effectively tested by the regular integration tests, I am fine with landing this now if you are. Ready for me to merge?

simolus3:
I'm fine with the current state too :+1: I mainly had a "no files are visible before the build has completed" idea for integration tests, but since we have unit tests for the functionality and integration tests ensuring this didn't break anything, I think this is ready.

build_runner_core/lib/src/environment/io_environment.dart
```diff
@@ -34,17 +35,33 @@ class IOEnvironment implements BuildEnvironment {
...
-        writer = FileBasedAssetWriter(_packageGraph) {
-    if (_outputSymlinksOnly && Platform.isWindows) {
+  IOEnvironment._(this.reader, this.writer, this._isInteractive,
+      this._outputSymlinksOnly, this._packageGraph);
+
+  factory IOEnvironment(
+    PackageGraph packageGraph, {
+    bool? assumeTty,
+    bool outputSymlinksOnly = false,
+    bool lowResourcesMode = false,
```
davidmorgan, line 45: FYI

I noticed that this param is never passed, so batching is always enabled.

Nothing to do, I'll figure out what the right fix is--I'm refactoring the "readers" anyway, that's why I noticed :)

Thanks.

PR #3774, opened 2024-12-04
Bump to latest analyzer

kevmoo:


build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -47,7 +47,8 @@ Future<String> _generateBuildScript() async {
       ..writeln('// ignore_for_file: directives_ordering')
       ..writeln(library.accept(emitter));
 
-    return DartFormatter().format(content.toString());
+    return DartFormatter(languageVersion: DartFormatter.latestLanguageVersion)
```
jakemac53: Really this should be the language version of the containing package.

Or, we could put a language version comment in this file and then hard code that version?

kevmoo: See my latest commit. We'd have to create a `Library` element or something for the current package using the analyzer API. I wish there was a cleaner way to do this...

jakemac53: We probably should be adding a language version comment anyways, we are generating code that we only know will work as of some language version anyways.

kevmoo: done

kevmoo:
@jakemac53 – so adding the language bits break the community tests. Thoughts?

jakemac53:
> @jakemac53 – so adding the language bits break the community tests. Thoughts?

This could happen if a builder factory string is configured using some syntax that isn't available in 3.6? cc @simolus3

jakemac53:
It does mean this is technically a breaking change unfortunately...

simolus3:
I'll take a look at the tests - but I think the build script should still run with an older language version due to the `environment.sdk` entry of the package where `build_runner` runs, right?

build_runner/lib/src/build_script_generate/build_script_generate.dart
```diff
@@ -25,13 +26,18 @@ const scriptKernelCachedSuffix = '.cached';
...
 Future<String> generateBuildScript() =>
     logTimedAsync(_log, 'Generating build script', _generateBuildScript);
 
 Future<String> _generateBuildScript() async {
   final info = await findBuildScriptOptions();
   final builders = info.builderApplications;
   final library = Library((b) => b.body.addAll([
+        Code(
+          '// @dart=${_lastShortFormatDartVersion.major}.${_lastShortFormatDartVersion.minor}',
+        ),
```
simolus3, line 40: This breaks all builds, the community tests just ran first :P It looks like `package:code_builder` does not insert a newline after the comment, so subsequent definitions got commented out. This fixes it:

```suggestion
        Code(
          '// @dart=${_lastShortFormatDartVersion.major}.${_lastShortFormatDartVersion.minor}\n',
        ),
```

kevmoo:
thanks!

kevmoo:
Can we run w/ this @jakemac53 ?

natebosch:
I don't understand how this is breaking, I _think_ the generated build script should be fine to hardcode to any language version.

Why do we need both the comment in the file and the argument to dart_style? Can we pick one or the other?

jakemac53:
@kevmoo once https://github.com/dart-lang/build/pull/3783 lands we should get this landed, happy to help with that. I think the merge conflicts will go away.

jakemac53:
We do need to update some changelogs though

jakemac53:
Skipping the changelog check for build_daemon, only dev_dependencies was touched

jakemac53 (approved):

PR #3783, opened 2024-12-16
Revert "Add APIs for using analyzer element model 2. (#3775)"

jakemac53:
This reverts commit fbdfb677c0f0dd188b38a9119be06a15c4043b1b.

I have forked the repo prior to this revert here https://github.com/dart-lang/build/tree/resolver-2-methods.

The plan is to use that fork internally until all builders are updated to the new APIs. Then we will do a breaking change in this repo to migrate to all the new APIs. We do not plan to publish ever the *2 APIs on pub.

In the short term, we will update our analyzer upper constraints across all packages but not add these new methods or expose the new APIs. Those versions will be published (soon ish) to unblock other work (such as updating to the latest dart_style).

### DIFF
--- a/build/CHANGELOG.md
+++ b/build/CHANGELOG.md
@@ -3,7 +3,7 @@
 - Bump the min sdk to 3.6.0-228.0.dev.
 - Remove some unnecessary casts and non-null assertions now that we have private
   field promotion.
-- Require analyzer ^7.0.0.
+- Require analyzer ^6.9.0.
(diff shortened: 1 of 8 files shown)


jakemac53:
cc @kevmoo we should land your PR after this since it was a simpler upgrade that allows the previous version and doesn't expose the new APIs

scheglov (approved):

PR #3784, opened 2024-12-16
Prep to release packages

jakemac53:
Releases support for analyzer 7.x and dart_style 3.x.


jakemac53:
Ignoring warnings about test fixtures package configs which are git ignored, they should not actually be ignored afaict, and don't need to ship with the package in any case.

kevmoo (approved):

PR #3787, opened 2024-12-18
build_daemon: drop unneeded built_value_generator dep

kevmoo:


kevmoo:
@jakemac53 – ug. Looks like drift would need to bump to support latest mockito?

davidmorgan:

PR #3790, opened 2025-01-02
fix strict top level inference lints

jakemac53:
These are all just `_` params that were falling back on dynamic.

Technically these are all breaking changes but none should be in practice.


jakemac53:
Ignoring the dart:html and dart:js warnings for now, separate fix is out for those here https://github.com/dart-lang/build/pull/3793

natebosch (approved):

PR #3793, opened 2025-01-06
migrate to package:web and dart:js_interop

jakemac53:
Resolves some deprecation warnings about dart:html

### DIFF
--- a/build_runner/web/graph_viz_main.dart
+++ b/build_runner/web/graph_viz_main.dart
@@ -2,32 +2,40 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
-import 'dart:async';
 import 'dart:convert';
-import 'dart:html';
-import 'dart:js' as js;
+import 'dart:js_interop';
 
-final _graphReference = js.context[r'$build']! as js.JsObject;
-final _details = document.getElementById('details')!;
+import 'package:http/http.dart' as http;
+import 'package:web/web.dart';
+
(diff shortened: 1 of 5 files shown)


build_runner/web/graph_viz_main.dart
```diff
@@ -2,32 +2,40 @@
...
 
 void _error(String message, [Object? error, StackTrace? stack]) {
   var msg = [message, error, stack].where((e) => e != null).join('\n');
-  _details.innerHtml = '<pre>$msg</pre>';
+  _details.innerHTML = '<pre>$msg</pre>'.toJS;
+  console.error(msg.toJS);
 }
 
-Future _focus(String query, {String? filter}) async {
+void _focus(String query, [String? filter]) async {
```
srujzs, line 38: Is the intention for this method to never be awaited (it looks like that's the case today already)? If so, it's fine to not do anything with the `Future`, but I can't tell.

srujzs (approved):

build_runner/web/graph_viz_main.dart
jakemac53: Yes, the return value is not used. This function is used in js interop which was complaining about the Future return type - but since I don't actually need the future this was the easiest solution :).

jakemac53: (and to be clear the JS side also doesn't use the return value, its just a callback for an event)

srujzs: Sounds good, the switch to `void` makes sense then.

natebosch (approved):

PR #3794, opened 2025-01-06
Add top level argument type in generated file

natebosch:
Clean up the last instance of `^Builder.*(_)` after the same pattern was
changed everywhere else to adhere to the strict top level inference
lint.

Add the `BuilderOptions` annotation in a string being written as a
top level function in a generated file.

There do not appear to be instances of this pattern in documentation.

### DIFF
--- a/build_runner/test/integration_tests/utils/build_descriptor.dart
+++ b/build_runner/test/integration_tests/utils/build_descriptor.dart
@@ -161,7 +161,7 @@ ${builders.map(_builderFactory).join('\n')}
 ''';
 
 String _builderFactory(TestBuilderDefinition builder) =>
-    'Builder ${builder.key}Factory(_) => ${builder.key};';
+    'Builder ${builder.key}Factory(BuilderOptions _) => ${builder.key};';
 
 String _buildToolFile(
         Iterable<TestBuilderDefinition> builders, Uri callingScript) =>


build_runner/test/integration_tests/utils/build_descriptor.dart
```diff
@@ -161,7 +161,7 @@ ${builders.map(_builderFactory).join('\n')}
 ''';
 
 String _builderFactory(TestBuilderDefinition builder) =>
-    'Builder ${builder.key}Factory(_) => ${builder.key};';
+    'Builder ${builder.key}Factory(BuilderOptions _) => ${builder.key};';
```
jakemac53, line 164: I don't think there is any value in this case?

natebosch: Yeah it's really low priority, I can drop the PR. I only changed it because it showed up in a regex search and if I were writing this line today I'd do it this way to match our other code.

natebosch: I was mainly looking for this pattern in user facing docs because those would be worth fixing, but it doesn't look like we have any.

jakemac53: hmm ok, I sort of prefer it the other way even though it ends up as dynamic... the whole point of the `_`  pattern is to reduce boilerplate and the type isn't useful information here imo.

I am definitely on board with changing it if we do want to add types, I just don't think I agree that's a pattern I like?

jakemac53: It would be sort of interesting if for wildcards the type defaulted to `void` or something like that. I am not sure what it actually defaults to (probably still dynamic?).

natebosch: I'm not sure if I'm 100% sold on the pattern, but it's the style we landed on to apply across the Dart and Flutter teams when we implemented the strict top level inference lint. It would apply here if this were handwritten and analyzed code.

jakemac53: > but it's the style we landed on to apply across the Dart and Flutter teams when we implemented the strict top level inference lint

But the lint explicitly doesn't cover local functions like this - it seems to only look at public API. I am curious what the purpose of it is.

natebosch: > But the lint explicitly doesn't cover local functions like this

This is writing a top level function into a generated file.

jakemac53: Ohhhhh, I see this is actually a string which is being emitted at the top level, so it is not a local function. Nevermind, LGTM.

jakemac53 (approved):

PR #3795, opened 2025-01-17
support 3.8 pre-release sdks

jakemac53:
@nshahan I added you as a reviewer to confirm that we don't expect any breaking changes to the dev compiler command line in 3.8?

cc @jyameo

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 5.0.11-wip
+## 5.0.11
+
+- Support 3.8.0 pre-release sdks.
 
 ## 5.0.10
 
--- a/build_modules/pubspec.yaml
+++ b/build_modules/pubspec.yaml
@@ -1,5 +1,5 @@
 name: build_modules
-version: 5.0.11-wip
+version: 5.0.11
 description: >-
   Builders to analyze and split Dart code into individually compilable modules
   based on imports.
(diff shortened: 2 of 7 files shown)


nshahan:
Because of the way these packages are so tightly coupled to the SDK versions I'm not sure what counts as a breaking change in this context. If we were to want to build web compilers to pass different arguments in 3.8 when invoking DDC does that count as a breaking change? I'm also planning on removing the `--sound-null-safety` at some point, possibly in 3.8 if we can ship 3.7 with it printing a deprecated warning.

jakemac53:
> If we were to want to build web compilers to pass different arguments in 3.8 when invoking DDC does that count as a breaking change?

Yes, this is exactly the kind of change that would break build_web_compilers. I think in this case though, we only support sound null safety anyways I believe, so that should be OK (I don't see us passing that flag anywhere).

natebosch (approved):

jakemac53:
Actually I just remembered we are guarding stable releases anyways because of the `-z`, so if build_web_compilers does get broken in some 3.8 dev release, we can just release a fix and it will be OK, it can't break 3.8 stable because we aren't declaring support yet for that.

This will in effect declare support for 3.7 stable but that should be fine, I assume the branch is now cut and any breaking changes there already would be in (modulo cherry picks which shouldn't have changes like this).

PR #3797, opened 2025-01-24
Fix crash on assets ending in a dot

simolus3:
We create a logger for each builder containing the asset id it's running on. Loggers aren't allowed to end with a dot because that character is used to build hierarchies.

In the somewhat absurd case where users have files ending with a dot in their project and also somehow managed to end up with builders actually matching an empty extension, `build_runner` would crash. I think this might be causing https://github.com/simolus3/drift/issues/3435.


jakemac53 (approved):
🤣 🚀

PR #3813, opened 2025-02-01
Refactor: split `BuildAssetUriResolver` into `AnalysisDriverModel` andd `AnalysisDriverModelUriResolver`.

davidmorgan:
`BuildAssetUriResolver` is critical to performance scalability, it's where the import graph is walked and turned into lists of inputs on each `BuildStep`.

It's very complicated :)

So my plan is to create a new implementation alongside the existing one to work on, and to increase test coverage; I have a "simplest possible implementation" ready in the next PR.

Before then, there is an improvement possible here. This PR splits one thing `BuildAssetUriResolver` does out, and introduce an interface `AnalysisDriverModel` for the part I want to re-implement.

`AnalysisDriverModel` is `build_runner`'s state related to build steps that use analysis, including an in-memory filesystem for use by the analyzer.

`UriResolver` is the analyzer's view of that state.


build_resolvers/lib/src/analysis_driver_model_uri_resolver.dart
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  @override
+  Uri pathToUri(String path) {
+    var pathSegments = p.posix.split(path);
+    var packageName = pathSegments[1];
+    if (pathSegments[2] == 'lib') {
+      return Uri(
+        scheme: 'package',
+        pathSegments: [packageName].followedBy(pathSegments.skip(3)),
```
jensjoha, line 36: Again I realize this is just copied from other places, so you don't have to change it here.
I'll just say that we, at least in my opinion, are not doing ourselves any favors by using `followedBy` and `skip`. We're creating a list and two iterables.

jensjoha (approved):

build_resolvers/lib/src/analysis_driver_model_uri_resolver.dart
```diff
@@ -0,0 +1,66 @@
...
+    final assetId = parseAsset(uri);
+    if (assetId == null) return null;
+
+    var file = analysisDriverModel.resourceProvider.getFile(assetPath(assetId));
+    return FileSource(file, assetId.uri);
+  }
+
+  @override
+  Uri pathToUri(String path) {
+    var pathSegments = p.posix.split(path);
```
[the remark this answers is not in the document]
davidmorgan: `UriResolver` is an interface owned by the analyzer that abstracts over a filesystem, it gives it a way to turn URIs into source+paths and paths into URIs.

This class is the view `build_runner` gives the analyzer of the files it wants to be analyzer.

The paths can be anything as long as they map 1:1 to URIs, so the exact path format is an implementation detail here.

Expanded the class comment to mention some of this:

/// A [UriResolver] on top of [AnalysisDriverModel]'s in-memory filesystem
///
/// This is the analyzer's view of the current build: the files that
/// `build_runner` wants the analyzer to analyze.
///
/// The in-memory filesystem uses POSIX-style paths with `package:foo/bar'
/// mapping to `/foo/bar`.

build_resolvers/lib/src/analysis_driver_model_uri_resolver.dart
davidmorgan: Thanks.

The analyzer only calls this O(N) times, where N is the number of files; so that won't matter until we've fixed the O(N^2) behaviour. I look forward to getting to that point ;)

davidmorgan (commented):
Thanks.

Issue #3811, opened 2025-02-01
Scalable builds


davidmorgan:
The benchmark I'm adding in #3802 shows the problem pretty clearly,

json_serializable |   |   |  
-- | -- | -- | --
libraries | clean/ms | no changes/ms | incremental/ms
1 | 21573 | 2840 | 4278
100 | 23040 | 2974 | 6327
250 | 28098 | 3285 | 12578
500 | 42940 | 4288 | 35061
750 | 69483 | 6561 | 67670
1000 | 115941 | 9178 | 119308

notice the incremental build time for 500 libraries and 1000 libraries: it increases from 35s to 119s, an increase of x3.4. For double the number of libraries the time increase should be x2 :)

The benchmark also runs for `built_value`, `freezed` and `mockito`, numbers on the PR, the story is pretty similar, except:

 - `built_value` and `json_serializable` are about 2x as slow as `freezed` on the big build, it turns out this is purely because they use shared parts, filed #3803 for that sub-problem
 -  `mockito` is about 2x as fast as `freezed`, I believe that is because mockito adds a part file to the test where it does not become a dep of all the other generators; conversely `built_value`, `json_serializable` and `freezed` all add part files in the app code, doubling the number of deps of all the generators

Apart from these multipliers the numbers are very similar, I think it does not matter much what the generators are actually doing.

amrgetment:
@davidmorgan I tried `cached_build_runner` and here are the results cc @tenhobi 

`dart run build_runner build --delete-conflicting-outputs`

The result:
<img width="434" alt="Image" src="https://github.com/user-attachments/assets/42b956f6-0f2b-4dbc-9396-d5490146318b" />

`dart run cached_build_runner build`

The result:
<img width="438" alt="Image" src="https://github.com/user-attachments/assets/cb003349-6f18-4958-a277-abfcfbcc41d2" />

```
dev_dependencies:
  build_runner: ^2.4.14
  cached_build_runner:
      path: /Users/amr/Desktop/cached_build_runner/

```

davidmorgan:
@amrgetment Thanks, but caching is not relevant to this issue: it just hides the problem.

amrgetment:
@davidmorgan 
Caching is still part of the solution, but faster builds alone won’t fully solve the problem.

In a real use case, the user might add changes, such as a login feature. However, they typically wouldn’t add multiple features—like reset password, change password, logout, and registration—all at once before running a new build_runner command.

[...]

PR #3814, opened 2025-02-01
Add minimal `AnalysisDriverModel` implementation

davidmorgan:
For #3811, review and merge #3813 first.

Make `resolver_test.dart` test `AnalysisDriverModel` exactly like it tests `BuildAssetUriResolver`.
    
Add minimal implementation to `AnalysisDriverModel` to make the tests pass.
    
Add TODOs where there is functionality not covered by tests that is therefore not added.

This implementation does work for real builds, a local edit to point `BuildAssetUriResolver.shared` to `AnalysisDriverModel` is enough to make the benchmarks run. Because there is no caching in the new implementation at all, it's slower :) which gives us a good starting point to add caching.

### DIFF
--- a/build_resolvers/lib/src/analysis_driver_model.dart
+++ b/build_resolvers/lib/src/analysis_driver_model.dart
@@ -118,19 +118,10 @@ class AnalysisDriverModel {
     });
   }
 
-  /// Walks the import graph from [ids], returns full transitive deps.s
+  /// Walks the import graph from [ids], returns full transitive deps.
   Future<Set<AssetId>> _expandToTransitive(
       AssetReader reader, Iterable<AssetId> ids) async {
     final result = <AssetId>{};
-    await __expandToTransitive(reader, ids, result);
-    return result;
-  }
-
-  /// Walks the import graph from [ids], ignoring nodes already in [result].
-  ///
-  /// Call with [result] empty to add full transitive deps to [result].
-  Future<void> __expandToTransitive(
-      AssetReader reader, Iterable<AssetId> ids, Set<AssetId> result) async {
     final nextIds = Queue.of(ids);
     while (nextIds.isNotEmpty) {
       final nextId = nextIds.removeFirst();
@@ -145,6 +136,7 @@ class AnalysisDriverModel {
       final deps = _parseDependencies(content, nextId);
       nextIds.addAll(deps.where((id) => !result.contains(id)));
     }
+    return result;
   }
 }
 
--- a/build_resolvers/lib/src/analysis_driver_model.dart
+++ b/build_resolvers/lib/src/analysis_driver_model.dart
@@ -121,20 +121,25 @@ class AnalysisDriverModel {
   /// Walks the import graph from [ids], returns full transitive deps.
   Future<Set<AssetId>> _expandToTransitive(
       AssetReader reader, Iterable<AssetId> ids) async {
-    final result = <AssetId>{};
+    final result = ids.toSet();
     final nextIds = Queue.of(ids);
     while (nextIds.isNotEmpty) {
       final nextId = nextIds.removeFirst();
 
-      // Skip if already seen.
-      if (!result.add(nextId)) continue;
-
-      // Skip if not readable.
+      // Skip if not readable. Note that calling `canRead` still makes it a
+      // dependency of the `BuildStep`.
       if (!await reader.canRead(nextId)) continue;
 
       final content = await reader.readAsString(nextId);
       final deps = _parseDependencies(content, nextId);
-      nextIds.addAll(deps.where((id) => !result.contains(id)));
+
+      // For each dep, if it's not in `result` yet, it's newly-discovered:
+      // add it to `nextIds`.
+      for (final dep in deps) {
+        if (result.add(dep)) {
+          nextIds.add(dep);
+        }
+      }
     }
     return result;
   }
--- a/build_resolvers/lib/src/analysis_driver_model.dart
+++ b/build_resolvers/lib/src/analysis_driver_model.dart
@@ -3,11 +3,17 @@
 // BSD-style license that can be found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:collection';
 
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:analyzer/dart/ast/ast.dart';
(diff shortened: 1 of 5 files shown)


build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,148 @@ import 'package:build/build.dart';
...
+      for (final id in ids) {
+        // TODO(davidmorgan): add test coverage for over-notification of
+        // changes, fix the implementaion here.
+        driver.changeFile(id.asPath);
+      }
+      await driver.applyPendingFileChanges();
+    });
+  }
+
+  /// Walks the import graph from [ids], returns full transitive deps.s
```
jensjoha, line 121: Delete the `.s`?

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,148 @@ import 'package:build/build.dart';
...
+      AssetReader reader, Iterable<AssetId> ids) async {
+    final result = <AssetId>{};
+    await __expandToTransitive(reader, ids, result);
+    return result;
+  }
+
+  /// Walks the import graph from [ids], ignoring nodes already in [result].
+  ///
+  /// Call with [result] empty to add full transitive deps to [result].
+  Future<void> __expandToTransitive(
```
jensjoha, line 132: Not a big fan of the double underscore naming.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,148 @@ import 'package:build/build.dart';
...
+    }
+  }
+}
+
+const _ignoredSchemes = ['dart', 'dart-ext'];
+
+/// Parses Dart source in [content], returns all depedencies: all assets
+/// mentioned in directives, excluding `dart:` and `dart-ext` schemes.
+List<AssetId> _parseDependencies(String content, AssetId from) =>
+    parseString(content: content, throwIfDiagnostics: false)
```
jensjoha, line 156: Stuff like this is completely unreadable to me.

jensjoha (approved):

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Done.

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Oh. Apparently I thought this was going to be needed for a recursive call, but actually it's iterative. So no need for this method at all, merged into the one above. Thanks!

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Noted, thanks. This is direct from the current implementation, so I'll leave as-is for now.

davidmorgan (commented):
Thanks.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,140 @@ import 'package:build/build.dart';
...
-      {required bool transitive});
+      {required bool transitive}) async {
+    /// TODO(davidmorgan): add test coverage for whether transitive
+    /// sources are read when [transitive] is false, fix the implementation
+    /// here.
+    /// TODO(davidmorgan): add test coverage for whether
+    /// `.transitive_deps` files cut off the reporting of deps to the
+    /// [buildStep], fix the implementation here.
+
+    // Find transitive deps, this also informs [buildStep] of all inputs).
```
pedromassango, line 92:
```suggestion
    // Find transitive deps, this also informs [buildStep] of all inputs.
```

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,140 @@ import 'package:build/build.dart';
...
+    parseString(content: content, throwIfDiagnostics: false)
+        .unit
+        .directives
+        .whereType<UriBasedDirective>()
+        .map((directive) => directive.uri.stringValue)
+        // Uri.stringValue can be null for strings that use interpolation.
+        .nonNulls
+        .where(
+          (uriContent) => !_ignoredSchemes.any(Uri.parse(uriContent).isScheme),
+        )
```
pedromassango, line 157: Can we merge this so we avoid the one full interaction that simply filters out nullable values?

```suggestion
        .where(
          (uriContent) => uriContent != null && !_ignoredSchemes.any(Uri.parse(uriContent).isScheme),
        )
```

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Thanks, done.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,140 @@ import 'package:build/build.dart';
...
+  /// Walks the import graph from [ids], returns full transitive deps.
+  Future<Set<AssetId>> _expandToTransitive(
+      AssetReader reader, Iterable<AssetId> ids) async {
+    final result = <AssetId>{};
+    final nextIds = Queue.of(ids);
+    while (nextIds.isNotEmpty) {
+      final nextId = nextIds.removeFirst();
+
+      // Skip if already seen.
+      if (!result.add(nextId)) continue;
```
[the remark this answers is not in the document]
davidmorgan: Done.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -19,36 +25,140 @@ import 'package:build/build.dart';
...
+
+      // Skip if not readable.
+      if (!await reader.canRead(nextId)) continue;
```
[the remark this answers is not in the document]
davidmorgan: Done.

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Thanks--this is just copied and pasted from existing code, I'd rather look at improving it as a separate piece of work rather than tweak it while copying.

davidmorgan (commented):
Thanks.

PR #3821, opened 2025-02-03
[Do not submit] test everything on beta

jakemac53:
cc @davidmorgan for fyi this is how I typically test the beta channel before release, just set the tested channel to beta for all the packages, regenerate the configs, and send a PR to run the bots. The PR will just be closed once we get the all green.


jakemac53:
It is also best to check at least on of the jobs to ensure it got the build you expected, see for instance here https://github.com/dart-lang/build/actions/runs/13123745031/job/36615579655?pr=3821#step:3:7.

PR #3822, opened 2025-02-04
Add to `resolver_test`, add to `AnalysisDriverModel` as needed to satisfy the tests.

davidmorgan:
For #3811 

Run the tests with both "shared" instances, re-used between tests (as before) and "new" instances.

The new implementation is still not used, and is not finished: it's a further step towards adding tests / simplifying to understand what's going on and what can be done to improve performance.

This version does already save some time on the big builds :)

skip-changelog-check

### DIFF
--- a/build_resolvers/lib/src/analysis_driver_model.dart
+++ b/build_resolvers/lib/src/analysis_driver_model.dart
@@ -34,16 +34,24 @@ class AnalysisDriverModel {
   final MemoryResourceProvider resourceProvider =
       MemoryResourceProvider(context: p.posix);
 
+  /// The import graph of all sources needed for analysis.
+  final _graph = _Graph();
+
+  /// Assets that have been synced into the in-memory filesystem
+  /// [resourceProvider].
+  final _syncedOntoResourceProvider = <AssetId>{};
+
   /// Notifies that [step] has completed.
   ///
   /// All build steps must complete before [reset] is called.
   void notifyComplete(BuildStep step) {
-    // TODO(davidmorgan): add test coverage, fix implementation.
+    // This implementation doesn't keep state per `BuildStep`, nothing to do.
   }
 
   /// Clear cached information specific to an individual build.
   void reset() {
-    // TODO(davidmorgan): add test coverage, fix implementation.
+    _graph.clear();
+    _syncedOntoResourceProvider.clear();
   }
 
   /// Attempts to parse [uri] into an [AssetId] and returns it if it is cached.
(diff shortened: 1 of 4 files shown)


build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -82,66 +90,84 @@ class AnalysisDriverModel {
...
+    }
 
-      // For each dep, if it's not in `result` yet, it's newly-discovered:
-      // add it to `nextIds`.
-      for (final dep in deps) {
-        if (result.add(dep)) {
-          nextIds.add(dep);
+    // Notify [buildStep] of its inputs.
+    for (final id in inputIds) {
+      await buildStep.canRead(id);
```
jensjoha, line 133: Does "canRead" say "this is what you should process" or does it say "this is what you are allowed to read" or how am I supposed to read the code in combination with the comment?

jensjoha: Also possibly some were already "canRead" above (in `if (transitive)`). Maybe this could be something like

```dart
if (transitive) {
  // bla bla
  for (final id in inputIds) {
    // Notify [buildStep] of its inputs. And also other stuff.
    if (await buildStep.canRead(id) && !id.path.endsWith(_transitiveDigestExtension) && _graph.nodes[id]!.isMissing) {
      // Some good comment here.
      idsToSyncOntoResourceProvider.add(id);
      _syncedOntoResourceProvider.remove(id);
    }
  }
} else {
  // Notify [buildStep] of its inputs.
  for (final id in inputIds) {
    await buildStep.canRead(id);
  }
}

File: build_resolvers/lib/src/analysis_driver_model.dart [hunk 2]
```diff
@@ -167,3 +193,108 @@ extension _AssetIdExtensions on AssetId {
...
+/// The directive graph of all known sources.
+///
+/// Also tracks whether there is a `.transitive_digest` file next to each source
+/// asset, and tracks missing files.
+class _Graph {
+  final Map<AssetId, _Node> nodes = {};
+
+  /// Walks the import graph from [ids] loading into [nodes].
+  Future<void> load(AssetReader reader, Iterable<AssetId> ids) async {
+    final nextIds = Queue.of(ids);
```
jensjoha, line 206: I would have commented about this in one of the previous reviews as well but forgot.
Is there any reason this is a Queue instead of a List? A List should do `removeLast` instead - so ordering would change -but otherwise I'd assume it was similar and probably perform better because it doesn't have to do casts and head and tail book keeping etc.

jensjoha (approved):

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: `canRead` asks the reader if the file is readable (exists + is allowed to be read), _and_ marks the file as an input to the `buildStep`. I need to untangle those two soon to make progress.

 - notifying about inputs needs to be separate from reading; we can't efficiently track inputs while passing them around one by one
 - there should be no need to eagerly check whether a file was written, `build_runner` can track which files get written
 
so I'll leave it separate as after those changes the reading and notifying will actually be separate :)

Thanks.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -82,66 +90,84 @@ class AnalysisDriverModel {
...
-      if (!await reader.canRead(nextId)) continue;
+    // If requested, find transitive imports.
+    if (transitive) {
+      await _graph.load(buildStep, entryPoints);
+      idsToSyncOntoResourceProvider = _graph.nodes.keys.toList();
+      inputIds = _graph.inputsFor(entryPoints);
 
-      final content = await reader.readAsString(nextId);
-      final deps = _parseDependencies(content, nextId);
+      // Check for missing inputs that were written during the build.
```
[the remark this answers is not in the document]
davidmorgan: Updated to

```
// Check for inputs that were missing when the directive graph was read
// but have since been written by another build action.
```

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: No particular reason; but any cost here should be `O(n)`, I want to solve the `O(n^2)` issues before getting onto constant factor improvements ... added a TODO. Thanks.

davidmorgan (commented):
Thanks.

build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -167,3 +194,109 @@ extension _AssetIdExtensions on AssetId {
...
+
+  /// Walks the import graph from [ids] loading into [nodes].
+  Future<void> load(AssetReader reader, Iterable<AssetId> ids) async {
+    // TODO(davidmorgan): check if List is faster.
+    final nextIds = Queue.of(ids);
+    while (nextIds.isNotEmpty) {
+      final nextId = nextIds.removeFirst();
+
+      // Skip if already seen.
+      if (nodes.containsKey(nextId)) continue;
```
jakemac53, line 213: As far as I can tell you will run into trouble here - you may have previously put a `_Node.missing` in for this node, and will not be able to resolve it later on.

This is a key driver of the performance issues - the fact that we have to be able to adapt to new files being added means we cannot (safely) cache globally the transitive deps.

jakemac53: Note that even within a single build step you will run into this problem potentially, they are allowed to add new files which are imported by something they just resolved, and then resolve those files.

davidmorgan: Yes, probably new work is needed for files created during the build.

Missing files are currently checked again, but only to see if they now exist, the graph is not recomputed.

It passes all the tests, though :) so more tests are needed, too.

One question, is it the case that all files that are output by generators during the build are written by an `AssetWriter`? Then, that should be a way to catch new files efficiently, rather than checking explicitly every time.

Thanks.

jakemac53: Yes, all files will be written by an `AssetWriter`. So, you may be able to add in some extra lifecycle hooks, but it starts getting into pretty dubious territory.

There is lots of weirdness because builders do not run in order most of the time, so you cannot rely on any phased ordering (instead they run when their output is requested). So, files go in and out of being "visible" depending on which builder happens to be running.

In order for a given input to have a _consistent_ set of inputs, you have to do the crawl each time from each builder.  We could possibly do something though where we accept that the inputs are over declared sometimes (a build step might pick up the transitive deps of a file it can't actually read). 

Also, once you have been able to read a library you can cache its direct dependencies, which I believe the old system did do. So, you may still be able to do some of this.

davidmorgan: Makes sense, thanks. Anyway, next step, test coverage :)

PR #3823, opened 2025-02-04
Simpler in memory filesystem

davidmorgan:
For #3811.

Switch from using the analyzer's in memory filesystem implementation, which does a lot of things that are not needed, to a local implementation that does only what is needed.

Improvements:

 - more efficient mutations to state, for example "write and check if changed" in one operation
 - tests can take advantage of checking the state (although this is not used yet)
 - reveals that currently all source passed to the analyzer does an unnecessary utf8 encode+decode and additional md5sum, this can be fixed (which I will follow up on)


davidmorgan:
CI found some test failures, please hold off on reviewing while I take a look :)

davidmorgan:
One of the "unused" methods was used ;) now implemented, PTAL.

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  /// Writes [content] to [path].
+  ///
+  /// Records the change in [changedPaths], only if the content actually
+  /// changed.
+  bool writeFile(String path, String content) {
+    final oldContent = _data[path];
+    _data[path] = content;
+    if (content != oldContent) _changedPaths.add(path);
```
jensjoha, line 49: Let's only do this comparison once.

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,175 @@
...
+      p.posix.join('/${assetId.package}', assetId.path);
+
+  /// Attempts to parse [uri] into an [AssetId].
+  ///
+  /// Handles 'package:' or 'asset:' URIs, as well as 'file:' URIs that have the
+  /// same pattern used by [assetPath].
+  ///
+  /// Returns null if the Uri cannot be parsed.
+  static AssetId? parseAsset(Uri uri) {
+    if (const ['dart', 'dart-ext'].any(uri.isScheme)) return null;
```
jensjoha, line 99: Can't we write this similar to the line below instead?

jensjoha (approved):

davidmorgan (commented):
Thanks :)

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,173 @@
...
+  /// call to [clearChangedPaths].
+  Iterable<String> get changedPaths => _changedPaths;
+
+  /// Clears [changedPaths].
+  void clearChangedPaths() => _changedPaths.clear();
+
+  // `UriResolver` methods.
+
+  @override
+  Uri pathToUri(String path) {
```
jakemac53, line 62: What is the first segment of the path that we are always skipping?

It might be good to put a comment with the expected path format here.

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,173 @@
...
+    final assetId = parseAsset(uri);
+    if (assetId == null) return null;
+
+    var file = getFile(assetPath(assetId));
+    return FileSource(file, assetId.uri);
+  }
+
+  /// Path of [assetId] for the in-memory filesystem.
+  static String assetPath(AssetId assetId) =>
+      p.posix.join('/${assetId.package}', assetId.path);
```
jakemac53, line 89: I wouldn't bother using package:path to do this, especially since there is already a hard coded leading `/`, might as well just do a string interpolation to add the `/` between these.

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,173 @@
...
+      final parts = p.split(uri.path);
+      return AssetId(parts[1], p.posix.joinAll(parts.skip(2)));
+    }
+    return null;
+  }
+
+  // `ResourceProvider` methods.
+
+  @override
+  p.Context get pathContext => p.posix;
```
jakemac53, line 111: Should probably go ahead and use this getter where we are using package:path

jakemac53 (approved):

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,173 @@
...
+import 'package:build/build.dart' hide Resource;
+import 'package:path/path.dart' as p;
+
+/// The in-memory filesystem that is the analyzer's view of the build.
+///
+/// Tracks modified paths, which should be passed to
+/// `AnalysisDriver.changeFile` to update the analyzer state.
+class AnalysisDriverFilesystem implements UriResolver, ResourceProvider {
+  final Map<String, String> _data = {};
+  final Set<String> _changedPaths = {};
```
kevmoo, line 21:
```suggestion
  final _data = <String, String>{};
  final _changedPaths = <String>{};
```

Does anyone else prefer this style ☝️

parlough: Personally not for fields. Declaring types on the left allow them to be consistent with getters and instance methods, and require me to consciously specify types for public members. These aren't public but it would be strange to be inconsistent between public and private fields.

build_resolvers/lib/src/analysis_driver_filesystem.dart
```diff
@@ -0,0 +1,191 @@
...
+// ignore: implementation_imports
+import 'package:analyzer/src/clients/build_resolvers/build_resolvers.dart';
+import 'package:build/build.dart' hide Resource;
+import 'package:path/path.dart' as p;
+
+/// The in-memory filesystem that is the analyzer's view of the build.
+///
+/// Tracks modified paths, which should be passed to
+/// `AnalysisDriver.changeFile` to update the analyzer state.
+class AnalysisDriverFilesystem implements UriResolver, ResourceProvider {
```
scheglov, line 19: Ouch. I did not expect that clients will start implementing `ResourceProvider`.
This will potentially complicate making any change to this interface.

PR #3825, opened 2025-02-05
AnalysisDriverModel: update the directive graph on newly-discovered inputs.

davidmorgan:
For #3811

### DIFF
--- a/build_resolvers/lib/src/analysis_driver_model.dart
+++ b/build_resolvers/lib/src/analysis_driver_model.dart
@@ -204,31 +204,31 @@ class _Graph {
     final processed = <AssetId>{};
     final previouslyMissingFiles = <AssetId>{};
     while (nextIds.isNotEmpty) {
-      final nextId = nextIds.removeFirst();
+      final id = nextIds.removeFirst();
 
-      if (!processed.add(nextId)) continue;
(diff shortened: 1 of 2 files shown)


build_resolvers/lib/src/analysis_driver_model.dart
```diff
@@ -203,33 +192,50 @@ class _Graph {
...
+  ///
+  /// Returns the set of files that were in the graph as missing and have now
+  /// been loaded.
+  Future<Set<AssetId>> load(AssetReader reader, Iterable<AssetId> ids) async {
     // TODO(davidmorgan): check if List is faster.
     final nextIds = Queue.of(ids);
+    final processed = <AssetId>{};
+    final previouslyMissingFiles = <AssetId>{};
     while (nextIds.isNotEmpty) {
       final nextId = nextIds.removeFirst();
```
jensjoha, line 207: `nextId` doesn't read well... Maybe just `id`?

jensjoha (approved):

build_resolvers/lib/src/analysis_driver_model.dart
davidmorgan: Done.

Issue #3827, opened 2025-02-05
comment with `{@code}` causes `build_runner` to crash 


dickermoshe:
There seems to be a bug in `build_runner` that any comment which contains this throws an error:

```dart
/// {@}
class A{}

/// {@hello}
class A{}

/// {@hello hi}
class A{}
```

Any comment with this will cause `build_runner` to give up:

```
[SEVERE] embed on lib/src/tz/android/bindings.dart:

This builder requires Dart inputs without syntax errors.
However, package:dateutil/src/tz/android/bindings.dart (or an existing part) contains the following errors.

Try fixing the errors and re-running the build.

[INFO] Running build completed, took 2.0s
[INFO] Caching finalized dependency graph completed, took 34ms
```

This is particularly troubling because `jnigen` put's many comments like this in the code because that's what java code has.

The error which is emitted is 
`/dateutil/lib/src/tz/android/bindings.dart(6..10): Doc directive 'hello' is unknown.`

* Dart SDK version: 3.6.1 (stable) (Tue Jan 7 09:50:00 2025 -0800) on "windows_x64"
* build_runner 2.4.14
* Issue exists on Linux an Windows

pubspec.yaml
```
name: dateutil
description: A starting point for Dart libraries or applications.
version: 1.0.0
# repository: https://github.com/my_org/my_repo

environment:
  sdk: ^3.6.1

dependencies:
  b: ^0.2.0
  collection: ^1.19.1
  embed_annotation: ^1.2.1
  equatable: ^2.0.7
  ffi: ^2.1.3
  jni: ^0.14.0
  meta: ^1.16.0
  path: ^1.9.1

dev_dependencies:
  build_runner: ^2.4.14
  embed: ^1.4.0
  ffigen: ^16.1.0
  jnigen: ^0.14.0
  lints: ^5.0.0
  melos: ^6.3.2
  test: ^1.24.0
```

PR #3828, opened 2025-02-06
Resolver should only throw for severe errors

dickermoshe:
Closes https://github.com/dart-lang/build/issues/3827

While using a code generator package I ran into an issue where I was being told that a file had a syntax error.
There was no error, analyzer reported no issues.

After further debugging, it turned out to be a `Doc directive 'hello' is unknown.` error.
Because this error is not severe, it should be ignored.

This PR makes build_resolver only throw on severe syntax errors

---

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Refactor `BuildAssetUriResolver` into `AnalysisDriverModel` and
   `AnalysisDriverModelUriResolver`. Add new implementation of
   `AnalysisDriverModel`.
+- Make resolver only throw `SyntaxErrorInAssetException` on severe syntax errors
 
 ## 2.4.3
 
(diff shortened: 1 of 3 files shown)


davidmorgan:
Thanks! I ran into the exact same issue with `@code` a few months ago, I had to go digging, turns out I "fixed" it for only one codepath, the analyzer's "sort directives":

https://dart-review.googlesource.com/c/sdk/+/391740/3/pkg/analysis_server/lib/src/g3/utilities.dart

The parsed results "errors" do indeed include warnings and infos, and it's the correct thing to ignore them here.

Could you please add a unit test? In `build_resolvers/test/resolver_test.dart` it should work to copy the `are reported` test, add the `@code` case instead of the syntax error, and check that it doesn't throw.

davidmorgan (approved):
Looks good, thanks! Let's wait for @jakemac53 to take a look also, this is my first external PR review for `build_runner`.

davidmorgan:
Now merged--will aim to release this week. Thanks!

PR #3831, opened 2025-02-06
Add --use-experimental-resolver flag.

davidmorgan:
This makes it easier for me to compare, and for external users to try the new implementation before it's live.

@jakemac53 could you please help me figure out what I need to do with `pubspec.yaml` here?

It introduces new API in `build_resolvers` that is needed in `build_runner`, I tried bumping the dep but that obviously doesn't work because `pub get` fails on CI. Should I change to a path dependency until everything is released? Is there anything else I need to do / be aware of?

Thanks :)

### DIFF
--- a/build_resolvers/CHANGELOG.md
+++ b/build_resolvers/CHANGELOG.md
@@ -4,6 +4,8 @@
   `AnalysisDriverFilesystem`. Add new implementation of
   `AnalysisDriverModel`.
 - Make resolver only throw `SyntaxErrorInAssetException` on severe syntax errors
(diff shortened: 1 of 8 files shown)


jakemac53:
> Should I change to a path dependency until everything is released? Is there anything else I need to do / be aware of?

I _think_ you can just change the dependency to be `^2.4.4-wip` so it matches the actual version in the pubspec. But then you need to remember to update that to `^2.4.4`.

If this is for a package not in the pub workspace, then you will want a dependency override.

davidmorgan:
@jakemac53 done, PTAL :) thanks.

PR #3844, opened 2025-02-11
Fix order-dependent test: test the correct instance(s).

davidmorgan:


build_resolvers/test/resolver_test.dart
```diff
@@ -738,36 +738,37 @@ int? get x => 1;
              }
           ''',
         }, (resolver) async {
```
jensjoha, line 740: Does this still have to be async? Or should the calls actually be `await resolver.libraryFor` as below?

build_resolvers/test/resolver_test.dart
```diff
@@ -738,36 +738,37 @@ int? get x => 1;
...
         });
       });
 
-      test('are reported for part files with errors', () {
-        return resolveSources({
+      test('are reported for part files with errors', () async {
+        await resolveSources({
           'a|lib.dart': '''
             library a_library;
             part 'errors.dart';
```
jensjoha, line 774: I can seemingly not put comments on lines further down --- but why does these in particular have to change when stuff further down looks to be done the same way as these were before?

jensjoha (approved):

build_resolvers/test/resolver_test.dart
davidmorgan: Actually, reverted that attempt at a fix.

I realized both these tests were missing `resolvers: createResolvers()`, which means all four test variants were testing the same shared instance. So, that's the correct fix--done. Thanks!

PR #3845, opened 2025-02-11
Use `dart2wasm` support-detection scripts

simolus3:
`dart2wasm` started emitting a `.support.js` file which contains a JavaScript expression evaluating to `true` if the browser supports the WebAssembly module emitted by `dart2wasm`: https://github.com/dart-lang/sdk/issues/59951.

With this PR, we start reading that file and use that as a feature detection if available.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.1.2-dev
+
+- Use support-detection scripts emitted by `dart2wasm`.
+
 ## 4.1.1
 
 - Support 3.8.0 pre-release sdks.
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 4.1.2-dev
+## 4.1.2-wip
 
 - Use support-detection scripts emitted by `dart2wasm`.
 
--- a/build_web_compilers/lib/src/dart2wasm_bootstrap.dart
+++ b/build_web_compilers/lib/src/dart2wasm_bootstrap.dart
@@ -18,18 +18,32 @@ import 'web_entrypoint_builder.dart';
 final _resourcePool = Pool(maxWorkersPerTask);
 
 /// Result of invoking the `dart2wasm` compiler.
-///
-/// The `supportExpression` is an optional JavaScript expressions emitted by
-/// newer versions of the compiler. It can be evaluated as part of a loader
-/// script to determine whether the current JavaScript environment supports
-/// `dart2wasm`.
-typedef Dart2WasmBootstrapResult = ({String? supportExpression});
+final class Dart2WasmBootstrapResult {
+  /// Whether `dart2wasm` did compile the Dart program.
+  ///
+  /// This is typically false when the program transitively imports an
+  /// unsupported library, or if the compiler fails for another reason.
+  final bool didCompile;
+
+  /// An optional JavaScript expression that might be emitted by `dart2wasm`.
+  /// The expression evaluates to `true` if the current JavaScript environment
+  /// supports all the features expected by the generated WebAssembly module.
+  final String? supportExpression;
+
+  const Dart2WasmBootstrapResult.didNotCompile()
+      : didCompile = false,
+        supportExpression = null;
+
+  Dart2WasmBootstrapResult({
+    required this.supportExpression,
+  }) : didCompile = true;
+}
 
 /// Invokes `dart compile wasm` to compile the primary input of [buildStep].
 ///
 /// This only emits the `.wasm` and `.mjs` files produced by `dart2wasm`. An
 /// entrypoint loader needs to be emitted separately.
-Future<Dart2WasmBootstrapResult?> bootstrapDart2Wasm(
+Future<Dart2WasmBootstrapResult> bootstrapDart2Wasm(
   BuildStep buildStep,
   List<String> additionalArguments,
   String javaScriptModuleExtension,
@@ -38,7 +52,7 @@ Future<Dart2WasmBootstrapResult?> bootstrapDart2Wasm(
       buildStep, additionalArguments, javaScriptModuleExtension));
 }
 
-Future<Dart2WasmBootstrapResult?> _bootstrapDart2Wasm(
+Future<Dart2WasmBootstrapResult> _bootstrapDart2Wasm(
   BuildStep buildStep,
   List<String> additionalArguments,
   String javaScriptModuleExtension,
@@ -69,7 +83,7 @@ $librariesString
 
 https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-skipped-compiling-warnings
 ''');
-      return null;
+      return const Dart2WasmBootstrapResult.didNotCompile();
     }
 
     var scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
@@ -131,6 +145,7 @@ https://github.com/dart-lang/build/blob/master/docs/faq.md#how-can-i-resolve-ski
   } else {
     log.severe('ExitCode:${result.exitCode}\nStdOut:\n${result.stdout}\n'
         'StdErr:\n${result.stderr}');
+    return const Dart2WasmBootstrapResult.didNotCompile();
   }
 
   var supportFile =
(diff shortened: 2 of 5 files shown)


build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.2-dev
```
davidmorgan, line 1: I'm new to this, the docs mention `-wip` for unpublished

https://github.com/dart-lang/ecosystem/wiki/Publishing-automation

or did you want to publish this as a dev release?

build_web_compilers/lib/src/dart2wasm_bootstrap.dart
```diff
@@ -17,20 +17,28 @@ import 'web_entrypoint_builder.dart';
 
 final _resourcePool = Pool(maxWorkersPerTask);
 
+/// Result of invoking the `dart2wasm` compiler.
+///
+/// The `supportExpression` is an optional JavaScript expressions emitted by
+/// newer versions of the compiler. It can be evaluated as part of a loader
+/// script to determine whether the current JavaScript environment supports
+/// `dart2wasm`.
+typedef Dart2WasmBootstrapResult = ({String? supportExpression});
```
davidmorgan, line 26: Please make it a class rather than a typedef+record, it's more maintainable. Once `supportExpression` is a field the docs for it can go on the field, which will help, it's a lot to explain in one comment :)

I suggest returning non-null `Dart2WasmBootstrapResult` everywhere and giving it a boolean to say whether it succeeded, rather than using null for failure.

build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -311,20 +312,26 @@ class WebEntrypointBuilder implements Builder {
...
 
-  (AssetId, String)? _generateLoader(AssetId input) {
+  (AssetId, String)? _generateLoader(
+      AssetId input, Dart2WasmBootstrapResult? dart2WasmResult) {
     var loaderExtension = options.loaderExtension;
     var wasmCompiler = options.optionsFor(WebCompiler.Dart2Wasm);
-    if (loaderExtension == null || wasmCompiler == null) {
+    if (loaderExtension == null ||
+        wasmCompiler == null ||
+        dart2WasmResult == null) {
```
davidmorgan, line 334: This looks like it will now do nothing in cases where previously it might have done something? When `dart2WasmResult` is null and the others are not null?

build_web_compilers/test/dart2wasm_bootstrap_test.dart
```diff
@@ -69,9 +69,11 @@ void main() {
           'a|web/index.dart.js': decodedMatches(
             stringContainsInOrder(
               [
-                'if (supportsWasmGC())',
+                'if',
```
davidmorgan, line 72: Are there more interesting cases to cover now?

And: please mention in a comment where this string is now coming from, i.e. that it's via the `supportExpression` in `Dart2WasmBootstrapResult`.

davidmorgan (commented):
Thanks! A few comments.

I'll be the main person working on `build_runner` for a while, I'm focused on performance first but will also need to learn all the other things :) ... hopefully my comments make some kind of sense ;)

build_web_compilers/CHANGELOG.md
simolus3: I think this project has used `-dev` before the automation established `-wip`, and I still haven't learned... It should be `-wip`, fixed!

build_web_compilers/test/dart2wasm_bootstrap_test.dart
simolus3: The challenge for the test is that we no longer now that the contents of the `if` statement here are, because they depend on what `dart2wasm` emits. So I had to generalize the test and added a comment explaining the reasons.

davidmorgan: Thanks.

build_web_compilers/lib/src/web_entrypoint_builder.dart
davidmorgan: Makes sense, thanks :)

davidmorgan (commented):
All looks good, thanks.

Will wait to see if anyone from the US side wants to review as well.

Unrelated, we have a [discussion section](https://github.com/dart-lang/build/discussions) now, it might be useful for more general discussion about the work / direction.

davidmorgan:
There is some issue with CI that I'm trying to fix, as part of it it seems there might already be a failure in this code on Windows, any ideas please?

https://github.com/dart-lang/build/actions/runs/13284718347/job/37090779528?pr=3851

simolus3:
Was that the first time it has failed or has this been broken for a while? It looks like `dart2wasm` is no longer honoring the `multi-root-scheme` option, but I'm also not sure why (and also confused about this only happening on Windows).

davidmorgan:
It's the first time I've seen it.

Possibly it's caused by an SDK change; there was a new "unused element" warning for `dart:js` that I hadn't seen before either.

But it could also be that the CI has not been working smoothly.

I think the next thing to try is to repro with different SDK versions, I'll try to get to that today/tomorrow.

davidmorgan:
Dusting off my Windows VM now, let's see if I can figure it out :)

davidmorgan:
Yes it's due to an SDK change: I can repro on `3.8.0-edge.66fa0e2a491371b098323a0311a0e400ad1fad65 (main) (Thu Feb 13 01:49:09 2025 -0800)` but not on stable or on the also very recent `3.8.0-91.0.dev (dev) (Tue Feb 11 16:06:19 2025 -0800)`. Lemme track down the commit...

davidmorgan:
Looks like https://dart-review.googlesource.com/c/sdk/+/397721 is probably the cause as it's the only dart2wasm commit in the range; I'll try to verify.

davidmorgan:
Hmm actually let me pin the SDK version so we can have working CI and get unstuck #3856

kevmoo:
How are we handling older SDKs that DON'T have the feature? We're not bumping the min SDK

davidmorgan:
> How are we handling older SDKs that DON'T have the feature? We're not bumping the min SDK

It just falls back to the existing behaviour.

davidmorgan:
@simolus3 it turns out I can merge in the CI fix :) and it passes. So I'll merge this now.

I wanted to get it in before my #3854 because that's relatively disruptive.

I have no idea how this usually works--does it make sense for me to release `build_web_compilers` with this change, or wait for some more to accumulate? Thanks!

davidmorgan (approved):

kevmoo:
FYI @mkustermann 😄

davidmorgan:
> Thanks for your help getting this merged!
> 
> > I have no idea how this usually works--does it make sense for me to release `build_web_compilers` with this change, or wait for some more to accumulate? Thanks!
> 
> I actually don't know what the typical workflow is, but since the SDK change is very recent and won't land in stable for a while, I think it's fine to not bother with a release right now.

Noted, thanks. Feel free to ping me any time you feel a release is overdue.

simolus3:
> Feel free to ping me any time you feel a release is overdue.

Are the `build_test` changes mentioned in the `build_web_compilers` changelog ready for release? It sounds like we'd have to release the two packages together to adopt the breaking changes.
Given that `dart2wasm` has started emiting code for which feature detection is helpful (https://github.com/dart-lang/sdk/issues/59951#issuecomment-2687345312), I think it may be a good time to get this out.

davidmorgan:
Hmmm yes it does need the `build_test` breaking changes.

`build_test` is starting to settle down but I don't think it's quite ready for release yet, I think probably `StubWriter` and `WrittenAssetWriter` will also go away to finish the reader/writer refactoring. Actively working on that :)

Thanks.

davidmorgan:
Oh ... but actually it's only a dev dependency, I think it should work fine to release. Let me try it :)

davidmorgan:
https://github.com/dart-lang/build/pull/3895

davidmorgan:
@simolus3 I think that worked :)

simolus3:
Unfortunately it looks like [the publishing action might be broken](https://github.com/dart-lang/build/actions/runs/13630285940/job/38096347715) :( I don't see the version on pub.dev either.

davidmorgan:
Coincidentally I checked and someone helped me make it actually publish, so, yup, _now_ I think it worked ;) thanks!

PR #3849, opened 2025-02-12
Move AssetReader input tracking to new class InputTracker.

davidmorgan:
The `AssetReader` API is user-facing, or rather generator-facing: it provides generators a way to read files, and `build_runner` then records that those files are inputs of the action.

But for `build_runner` itself it's useful to separate reading from tracking. So, this PR starts splitting out reader state and making it accessible to `build_runner` internal code.

### DIFF
--- a/build/lib/internal.dart
+++ b/build/lib/internal.dart
@@ -1,0 +1,12 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Internal build state for `build_resolvers`, `build_runner_core` and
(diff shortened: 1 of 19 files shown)


build/lib/internal.dart
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+/// Internal build state for `build_resolvers`, `build_runner_core` and
```
jensjoha, line 5: Any particular reason the two exported files aren't just imported where you need it, or that this isn't inside "src" or something, to discourage more than the documentation on "library"?

jensjoha (approved):

PR #3854, opened 2025-02-13
Combine `InMemoryAssetReader` and `InMemoryAssetWriter` into `InMemoryAssetReaderWriter`

davidmorgan:
For #3811 

For my next PR the plan was to give access to the filesystem backing a `Reader`.

This turned out to be hard to do because of how the tests are set up: the reader and writer are often entirely disconnected. So there is no backing filesystem, even an in-memory one.

This PR refactors to fix that in a lot of cases: so reads and writes in tests are backed by one `InMemoryAssetReaderWriter` instance.


_test_common/lib/test_phases.dart
```diff
@@ -70,12 +68,11 @@ void _printOnFailure(LogRecord record) {
 Future<BuildResult> testBuilders(
   List<BuilderApplication> builders,
   Map<String, /*String|List<int>*/ Object> inputs, {
+  BuildResult? resumeFrom,
```
jensjoha, line 71: Why do it this way? The whole Expando thing seems like somewhat of a hack, and unless there's a good reason I'd prefer not to use it.

jensjoha (approved):

_test_common/lib/test_phases.dart
davidmorgan: Thanks, done.

I was trying to avoid changing the return type to avoid such a big breaking change to tests ... but I was confused, I thought this was in `build_test` which is published. But fortunately it's in `_test_common` which we can just change. So, changed the return type :)

davidmorgan (commented):
Thanks, done.

build_runner/test/generate/watch_test.dart
```diff
@@ -311,38 +338,46 @@ void main() {
           rootPackage('a', path: path.absolute('a')): ['b'],
           package('b', path: path.absolute('a', 'b')): []
         });
+        readerWriter = InMemoryRunnerAssetReaderWriter(
```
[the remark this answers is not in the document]
davidmorgan: Thanks, good point: the root package is always `a` so there is no need to create a new one, removed in both places.

Issue #3855, opened 2025-02-13
Breakage due to SDK change


davidmorgan:
https://github.com/dart-lang/build/actions/runs/13306891674/job/37159883946

I can repro on 3.8.0-edge.66fa0e2a491371b098323a0311a0e400ad1fad65 (main) (Thu Feb 13 01:49:09 2025 -0800) but not on stable or on the also very recent 3.8.0-91.0.dev (dev) (Tue Feb 11 16:06:19 2025 -0800). 

My guess is it's due to https://dart-review.googlesource.com/c/sdk/+/397721

PR #3856, opened 2025-02-13
Pin SDK version due to https://github.com/dart-lang/build/issues/3855.

davidmorgan:
So we have working CI again.


jensjoha (approved):

jakemac53:
Fwiw I would just skip the failing tests instead of not running any tests on newer SDKs

davidmorgan:
> Fwiw I would just skip the failing tests instead of not running any tests on newer SDKs

Thanks--I did consider that but there's an open PR for dart2wasm support so I'd like to have the coverage.

Changed the PR from pinning all packages except `build_web_compilers` to pinning just `build_web_compilers`, which seems much better :)

PR #3857, opened 2025-02-13
Remove multi reader

davidmorgan:
For #3811 

Further simplification of test readers in preparation for refactoring the real ones.


build_test/lib/src/resolve_source.dart
```diff
@@ -194,11 +205,18 @@ Future<T> _resolveAssets<T>(
     }
     inputAssets[assetId] = assetValue;
   }));
-  final inMemory = InMemoryAssetReaderWriter(
+  final readerWriter = InMemoryAssetReaderWriter(
     sourceAssets: inputAssets,
     rootPackage: rootPackage,
   );
 
+  // Then, read any additionally requested files from the filesystem. These are
```
jensjoha, line 213: "These are" ?

jensjoha (approved):

build_test/lib/src/resolve_source.dart
davidmorgan: Thanks, done.

PR #3860, opened 2025-02-14
Add `Filesystem` underneath `Reader`, simplify tests further.

davidmorgan:
For #3811.


build_runner_core/lib/src/asset/file_based.dart
```diff
@@ -61,81 +66,50 @@ class FileBasedAssetReader extends AssetReader implements AssetReaderState {
...
   Future writeAsString(AssetId id, String contents,
-      {Encoding encoding = utf8}) async {
-    var file = _fileFor(id, packageGraph);
-    await _descriptorPool.withResource(() async {
-      await file.create(recursive: true);
-      await file.writeAsString(contents, encoding: encoding);
-    });
-  }
+          {Encoding encoding = utf8}) =>
+      filesystem.writeAsString(id, contents);
```
jensjoha, line 102: Shouldn't you pass along the encoding as well?
Or not have options for enconding and always use utf8...

build_test/lib/src/in_memory_reader.dart
```diff
@@ -13,57 +14,35 @@ abstract class RecordingAssetReader implements AssetReader {
...
 
-/// An implementation of [AssetWriter] that records outputs to [assets].
-abstract class RecordingAssetWriter implements AssetWriter {
-  Map<AssetId, List<int>> get assets;
-}
-
 /// An implementation of [AssetReader] and [AssetWriter] with primed in-memory
 /// assets.
+///
+/// TODO(davidmorgan): merge into `FileBasedReader` and `FileBasedWriter`.
```
jensjoha, line 20: Why? Memory based and file based seems at odds.

build_test/lib/src/in_memory_reader.dart
```diff
@@ -100,65 +79,38 @@ class InMemoryAssetReaderWriter extends AssetReader
...
-  @override
-  void cacheStringAsset(AssetId id, String contents, {Encoding? encoding}) {
-    encoding ??= utf8;
-    assets[id] = encoding.encode(contents);
-  }
-
   @override
   Future writeAsBytes(AssetId id, List<int> bytes) async {
-    assets[id] = bytes;
+    assets[id] = Uint8List.fromList(bytes);
```
jensjoha, line 84: maybe `bytes is Uint8List ? bytes : Uint8List.fromList(bytes);`?
In general I think the interface should be changed, but probably not in this cl...

jensjoha (approved):

build_runner_core/lib/src/asset/file_based.dart
davidmorgan: Done, thanks.

This is from the generator-facing `AssetWriter` interface, so no choice about the signature :)

build_test/lib/src/in_memory_reader.dart
davidmorgan: I think it will end up as one class, with a different `Filesystem` implementation for in-memory and file based.

Once all the optional functionality is refactored from inheritance hierarchy to members like `AssetFinder`, it should be possible to get rid of `AssetReaderState`, get rid of delegation, and have one concrete `AssetReaderWriter` with all the interesting/optional functionality provided via those members.

The key thing allowing this refactoring is that the inheritance hierarchy / delegates are always built so they run in the same order--so for example caching always happens underneath per-generator visibility restriction. So actually that can just be "do I have visibility restriction, if so apply it, do I have a cache, if so use it" and so on :)

build_test/lib/src/in_memory_reader.dart
davidmorgan: Done, thanks.

This is generator-facing API so we can't easily change it.

davidmorgan (commented):
Thanks :)

Some quite detailed responses about where this is headed--hopefully they make sense, happy to elaborate / discuss as needed. Thanks!

PR #3872, opened 2025-02-20
use 'compile js-dev' command for compilation

a-siva:
Use the command 'compile js-dev'  for compilation instead of directly invoking the snapshot which uses internal knowledge
about the location of the snapshot.

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Bump the min SDK to 3.7.0.
 - Use `build_test` 3.0.0.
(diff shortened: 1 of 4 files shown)


davidmorgan (approved):
Looks good, thanks.

How soon would you like this to be released? I'm currently mid-refactoring, which makes it awkward to release immediately, is 1-2 weeks okay?

Thanks :)

a-siva:
> How soon would you like this to be released? I'm currently mid-refactoring, which makes it awkward to release immediately, is 1-2 weeks okay?

Sounds good.

PR #3874, opened 2025-02-21
Refactor `BuildCacheReader` to `BuildCacheAssetPathProvider`.

davidmorgan:
For #3811.

Move handling of generated asset paths into the base `Reader`.

Disable `BatchReader`/`BatchWriter` for now, write caching should go next to `Filesystem`, that's easier to redo after the refactor.

PR #3876, opened 2025-02-24
Refactor `FileBasedAssetReader` and `FileBasedAssetWriter` to `ReaderWriter`.

davidmorgan:
For #3811, review and merge #3874 first.


build_test/lib/src/in_memory_reader_writer.dart
```diff
@@ -15,102 +17,90 @@ import 'test_reader_writer.dart';
...
+    );
+  }
+
+  InMemoryAssetReaderWriter.using({
+    required super.rootPackage,
+    required super.assetFinder,
+    required super.assetPathProvider,
+    required super.filesystem,
+    required super.cache,
+    required super.inputTracker,
```
jensjoha, line 46: You assume below this is non-null. Shouldn't you mark it as such?

jensjoha: (and possibly overwrite the inputTracker getter to non-null to avoid some `!`s below)

jensjoha (approved):

build_test/lib/src/in_memory_reader_writer.dart
davidmorgan: That's awkward because of the `extends` relationship, the super class `copyWith` wants it to be nullable. It could be `covariant`.

Really the awkward part is the test and non-test code doing different things, making it nullable for one and not for the other.

Fortunately the next PR addresses exactly this, unifying test and non-test use and moving `InputTracker` out of `AssetReader` entirely to the one place it's actually used :)

Thanks.

jensjoha: I don't think I understand... Why wouldn't this work:

```diff
diff --git a/build_test/lib/src/in_memory_reader_writer.dart b/build_test/lib/src/in_memory_reader_writer.dart
index f4ca7bdd..66e010f4 100644
--- a/build_test/lib/src/in_memory_reader_writer.dart
+++ b/build_test/lib/src/in_memory_reader_writer.dart
@@ -43,9 +43,12 @@ class InMemoryAssetReaderWriter extends ReaderWriter
     required super.assetPathProvider,
     required super.filesystem,
     required super.cache,
-    required super.inputTracker,
+    required InputTracker super.inputTracker,
   }) : super.using();
 
+  @override
+  InputTracker get inputTracker => super.inputTracker!;
+
   @override
   InMemoryAssetReaderWriter copyWith({
     AssetPathProvider? assetPathProvider,
```

I mean, with it going away it doesn't matter, but I don't understand the argument.

davidmorgan: Ah, you are correct.

The problem I mentioned only happens if `inputTracker` is a parameter to `copyWith`, but it no longer is. Sorry for the confusion, I was repeating compiler complaints from a stale state :)

PR #3878, opened 2025-02-24
Revert "use 'compile js-dev' command for compilation"

davidmorgan:
Reverts dart-lang/build#3872

Checking if this is the cause of timeouts on Windows https://github.com/dart-lang/build/actions/runs/13501493388

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 - Bump the min SDK to 3.7.0.
 - Use `build_test` 3.0.0.
-- Use the command 'compile js-dev' for compilation instead of invoking
-  dart with the snapshot name (uses internal implementation detail of
-  the dart sdk which could change).
 
 ## 5.0.11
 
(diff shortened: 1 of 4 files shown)


davidmorgan:
Hi @a-siva,

There seems to be some issue with this change on Windows, some of the e2e tests time out, here is a run from a clean branch

https://github.com/dart-lang/build/actions/runs/13501493388

I'm not sure what's going on, I tried to repro the timeouts on my Windows machine and couldn't. The only clue I saw in the  failures is something about not being able to delete the temporary "scratch" folder.

It does seem to fail reliably on the CI, and it passes with this revert, so I suggest we revert until we can figure it out, to unblock other PRs :)

a-siva:
> I'm not sure what's going on, I tried to repro the timeouts on my Windows machine and couldn't. The only clue I saw in the failures is something about not being able to delete the temporary "scratch" folder.
> 
This might have something to do with child processes not getting killed correctly. Not sure why it would happen only on the CI and not on local runs.

Is there a command line I can use to run the e2e test locally and see if processes are not getting killed.

davidmorgan:
Hmm that's worth a try--I'll try it today, let's coordinate on an issue https://github.com/dart-lang/build/issues/3880

I'd like to merge this in the meantime since other PRs aren't running CI fully.

sigurdm (approved):

Issue #3875, opened 2025-02-22
Bug: build_runner fails to cleanup outdated outputs


schultek:
I am investigating [this](https://stackoverflow.com/questions/79456073/why-do-i-seemingly-need-to-add-client-on-most-jaspr-components-dart) build_runner error when using Jaspr. After some digging I found a reproducible setup and think its a general build_runner bug. It boils down to the following:

I have two builders:
- Builder A reads in `.a` and generates `.b` (e.g. `main.a` -> `main.b`)
- Builder B reads in `.b` and generates `.c` (e.g. `main.b` -> `main.c`)

Builder A has some logic that it only **conditionally** generates `.b` files and that condition might change while watching the files (`dart run build_runner watch`).

When that happens, Builder A no longer outputs `main.b`. At that point I expect build_runner to also cleanup `main.c` since Builder B is also no longer executed since it doesn't have any input. However `main.c` still persists with now outdated contents.

To reproduce do the following steps:

1. Create a blank dart project.
2. Add `build` as a dep and `build_runner` as a dev dep.
3. Add the following files:

<details>
<summary>build.yaml</summary>

```dart
builders:
  a:
    import: "package:buildtest/builders.dart"
    builder_factories:
      - aBuilder
    build_extensions:
      .a:
        - .b
    auto_apply: all_packages
    build_to: source
    required_inputs: [".a"]
  b:
    import: "package:buildtest/builders.dart"
    builder_factories:
      - bBuilder
    build_extensions:
      .b:
        - .c
    auto_apply: all_packages
    build_to: source
    required_inputs: [".b"]
```

</details>

<details>
<summary>lib/builders.dart</summary>

```dart
import 'dart:async';

import 'package:build/build.dart';

Builder aBuilder(BuilderOptions options) => ABuilder();
Builder bBuilder(BuilderOptions options) => BBuilder();

class ABuilder extends Builder {
  @override
  FutureOr<void> build(BuildStep buildStep) async {
    var content = await buildStep.readAsString(buildStep.inputId);
    if (content.contains('on')) {
      await buildStep.writeAsString(buildStep.inputId.changeExtension('.b'), content);
    }
  }

  @override
  Map<String, List<String>> get buildExtensions => {
    '.a': ['.b'],
  };
}

class BBuilder extends Builder {
  @override
  FutureOr<void> build(BuildStep buildStep) async {
    var content = await buildStep.readAsString(buildStep.inputId);
[...]

PR #3879, opened 2025-02-24
Fix bug where outputs of outputs wouldn't get deleted

jakemac53:
Fixes https://github.com/dart-lang/build/issues/3875

### DIFF
--- a/build_runner/test/generate/watch_test.dart
+++ b/build_runner/test/generate/watch_test.dart
@@ -22,7 +22,10 @@ import 'package:watcher/watcher.dart';
 void main() {
   /// Basic phases/phase groups which get used in many tests
   final copyABuildApplication = applyToRoot(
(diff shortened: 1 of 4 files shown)


jakemac53:
Hmm, a previous regression test is failing here for another scenario. Not sure I will have the time to continue diagnosing this soon, so @davidmorgan if you want to pick it up from here that might be good.

Otherwise, I can come back to it again when I have some free time at the end of a day.

davidmorgan:
The timeouts on Windows should be addressed now https://github.com/dart-lang/build/issues/3880 is that the failure you meant? Thanks.

jakemac53:
> The timeouts on Windows should be addressed now #3880 is that the failure you meant? Thanks.

There is a different test which is testing something related to failed actions, and the transitive outputs of those.

davidmorgan:
Closing as this is fixed post-refactor in https://github.com/dart-lang/build/pull/4130 :)

PR #3881, opened 2025-02-25
One `InputTracker` per build

davidmorgan:
For #3811, review and merge #3876 first.

Unify test and non-test input tracking by having one input tracker per build, using the current primary input ID as a key.


build/lib/src/state/input_tracker.dart
```diff
@@ -6,7 +6,30 @@ import 'dart:collection';
 
 import '../asset/id.dart';
 
-/// Records inputs for one generator run.
+/// Records inputs for each generator.
 class InputTracker {
-  final assetsRead = HashSet<AssetId>();
+  final _inputs = HashMap<AssetId, HashSet<AssetId>>();
```
jakemac53, line 11: I don't believe this is safe to do this way, if I understand it correctly. The keys in this map would need to be the output ids, instead of the primary input id.

The reason being that an asset can be a primary input to any number of different actions.

I would try adding a test where there are two builders with the same input extension, but which read different additional files during the build, and make sure they don't pick up each other's dependencies.

jakemac53: Another consideration here is this will retain these sets much longer in memory (for an entire build), instead of releasing them as soon as a build completes.

davidmorgan: I think there is a much simpler answer here, if the state is per `BuildStep`, it should be stored per `BuildStep` :) ... I think I can get that working.

davidmorgan:
Closing in favour of #3884

PR #3892, opened 2025-02-27
Merge `AssetWriterSpy` functionality into `SingleStepReader`.

davidmorgan:
For #3811.

`SingleStepReader` becomes `SingleStepReaderWriter`. Use it explicitly where it makes sense, add internal-only `AssetReaderWriter` and use that where it makes sense. Convert from `AssetReader`+`AssetWriter` in public APIs to one `AssetReaderWriter`/`ReaderWriter`/`SingleStepReaderWriter` in internal APIs.

Codepath used for testing that partly simulates a build is now called out explicitly.


build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -645,36 +647,40 @@ class _SingleBuild {
...
+      runningBuild: RunningBuild(
+        packageGraph: _packageGraph,
+        targetGraph: _targetGraph,
+        assetGraph: _assetGraph,
+        nodeBuilder: _buildAsset,
+        globNodeBuilder: _buildGlobNode,
+      ),
+      runningBuildStep: RunningBuildStep(
+        phaseNumber: phaseNumber,
+
```
jensjoha, line 667: the blank line is weird.

jensjoha (approved):

build_runner_core/lib/src/generate/build_impl.dart
davidmorgan: Thanks! Done.

PR #3896, opened 2025-03-03
Bump actions/cache to v4.

davidmorgan:


sigurdm (commented):
This file seems to be generated - perhaps better to update some mono-repo file somewhere?

davidmorgan:
> This file seems to be generated - perhaps better to update some mono-repo file somewhere?

Hmm as far as I can see the actions/cache version is only specified here; I reran the generation and it didn't change anything or complain. (There's a presubmit that checks that the generated config is up to date, too).

mosuem:
It's recommended to pin hashes instead of tags - see https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

davidmorgan:
> It's recommended to pin hashes instead of tags - see https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

Thanks--done.

sigurdm (approved):

PR #3897, opened 2025-03-03
Stop using transitive digest files for resolvers.

davidmorgan:
For #3811.

Transitive digests files are used to cut off the "combined digests" calculation, by having one file stand in for many files in the "transitive inputs digest" calculation. And, they reduce the size of the build graph.

However they don't work within the same package, and so we anyway need a fix for these that works without them.

They are marked as an "optional" build step, which means that they are only built if some later generator requests them. I guess this is to help them work as an optimization. But, this makes the build significantly more complicated: it means that one build "phase" can cause a previous build "phase" to run by asking for an optional output from it. I think that with transitive digest files removed we can also just build everything eagerly, run the phases in order, and make everything much simpler. That's what the next PR does :)

For the benchmark we have, this already improves performance a lot, since the benchmark we have only has nodes in the same package that do not benefit from transitive digest files--but it does have to handle twice as many nodes because of _checking_ for the files

```
before this PR:
loop,1,23553,4599,5876
loop,100,25142,4772,8619
loop,250,31910,5048,15655
loop,500,51946,6339,39503
loop,750,84186,9139,79632
loop,1000,129335,10977,143517

after:
loop,1,24385,4365,5483
loop,100,26646,4417,7557
loop,250,30159,4582,12666
loop,500,41991,5000,25656
loop,750,59208,6361,43598
loop,1000,78480,8256,69272
```


jensjoha (approved):

jakemac53:
@davidmorgan does the example in your case use flutter? Did we try this both ways on any real apps or just a synthetic benchmark? It should be beneficial for small/medium sized flutter apps as implemented, because it will cut off the (transitive) flutter dependencies.

This optimization does come with a cost to be sure, but it also at least provides you a path towards getting good performance today, if you structure your app into packages or targets.

It might be nice if we kept this as an option at least for those who it does help out.

davidmorgan:
Thanks @jakemac53, yes, this very likely makes some cases slower where there are large deps.

What I currently believe/hope ;) is that the solution for those cases is the same as the solution for large library cycles: to handle the extra nodes in the build graph efficiently.

So my plan is to implement that, then check back as to whether there is anything left to do for the cases where `transitive_deps` files are currently helping.

I have various experimental branches after this one, including one which I did try on the large built_value-using app, which saw about 2x speedup.

I also need to worry about the web compile case :) and will make sure to check that, too.

Lots to do ;)

PR #3899, opened 2025-03-03
Run phases sequentially

davidmorgan:
For #3811, review and merge #3897 first.

With the removal of transitive digest files, "optional" generators are no longer heavily relied on for performance.

So, just run all the generators eagerly.

This means each phase can complete sequentially, so drop support for running optional actions from earlier phases. And, run the build actions one by one, removing all use of `Future.wait` to run things "in parallel".

This all has no big effect on performance

```
before this PR:

loop,1,24385,4365,5483
loop,100,26646,4417,7557
loop,250,30159,4582,12666
loop,500,41991,5000,25656
loop,750,59208,6361,43598
loop,1000,78480,8256,69272

after:

loop,1,23703,4171,5535
loop,100,26758,4142,8487
loop,250,31544,4645,12700
loop,500,40896,4984,23992
loop,750,56848,6023,41003
loop,1000,74130,7545,66228
```

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -13,7 +13,6 @@ import 'package:build/src/internal.dart';
 import 'package:crypto/crypto.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
-import 'package:pool/pool.dart';
 import 'package:watcher/watcher.dart';
 
 import '../asset/finalized_reader.dart';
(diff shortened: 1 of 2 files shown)


davidmorgan:
It turns out `build_web_compilers` also uses optional actions, I'll have to take a closer look :) closing for now.

jakemac53:
Yeah I do think build_web_compilers and others really need something like this... or they would need a significant redesign.

They will end up trying to compile non-web code without it today.

davidmorgan:
Yep that's what broke :)

Optional build does make a lot of sense for cross platform things.

It might be possible to figure out up front which optional actions are needed.

But I think it's also possible to keep the out of order phases and optimize with sufficiently careful state management, let's see :)

jakemac53:
> It might be possible to figure out up front which optional actions are needed.

Fwiw we do already know whether a given "module" can be compiled to web or not.

We could just emit an error file for those modules instead, but not actually report those errors unless a web entry point builder depends on that module

PR #3901, opened 2025-03-04
Stop using Future.wait in `build_impl`.

davidmorgan:
It only increases performance in some situations with an I/O bottleneck, which is not obviously the case; for now let's remove as it makes it clearer what's going on. Just do tasks one by one.

It's still the case that there is some "parallelism" because a build action can require an input from an action in an earlier phase that didn't run yet. So the max running actions is one per phase.

Speeds things up a bit.

```
before this PR:

loop,1,24385,4365,5483
loop,100,26646,4417,7557
loop,250,30159,4582,12666
loop,500,41991,5000,25656
loop,750,59208,6361,43598
loop,1000,78480,8256,69272

after:
loop,1,22517,4053,5170
loop,100,25499,4039,7974
loop,250,29008,4456,12994
loop,500,39135,5091,24792
loop,750,54465,6100,38380
loop,1000,72807,7287,59647
```

### DIFF
--- a/build_runner_core/lib/src/generate/build_impl.dart
+++ b/build_runner_core/lib/src/generate/build_impl.dart
@@ -13,7 +13,6 @@ import 'package:build/src/internal.dart';
 import 'package:crypto/crypto.dart';
 import 'package:logging/logging.dart';
 import 'package:path/path.dart' as p;
-import 'package:pool/pool.dart';
 import 'package:watcher/watcher.dart';
 
 import '../asset/finalized_reader.dart';
@@ -154,7 +153,6 @@ class _SingleBuild {
   final AssetGraph _assetGraph;
   final Set<BuildFilter> _buildFilters;
   final List<BuildPhase> _buildPhases;
-  final List<Pool> _buildPhasePool;
   final BuildEnvironment _environment;
   final _lazyPhases = <String, Future<Iterable<AssetId>>>{};
   final _lazyGlobs = <AssetId, Future<void>>{};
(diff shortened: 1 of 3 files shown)


gmpassos:
NOTE: CI servers are usually slower in terms of disk performance.

davidmorgan:
> NOTE: CI servers are usually slower in terms of disk performance.

That should still be fine--`build_runner` spends almost no time on I/O. If it does turn out to slow things down, hopefully we notice, as that gives something to fix :) thanks!

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -616,34 +613,27 @@ class _SingleBuild {
...
-          anchorNode,
-        ),
-      ),
-    );
-    return outputLists.fold<List<AssetId>>(
-      <AssetId>[],
-      (combined, next) => combined..addAll(next),
-    );
+    final outputs = <AssetId>[];
+    for (final node in _assetGraph.packageNodes(action.package).toList()) {
```
jensjoha, line 617: if the `.toList()` is needed you should probably pass `growable: false`.

build_runner_core/lib/src/generate/build_impl.dart
```diff
@@ -185,7 +185,7 @@ class _SingleBuild {
       _buildPhases = buildImpl._buildPhases,
       _buildPhasePool = List.generate(
         buildImpl._buildPhases.length,
-        (_) => Pool(buildPhasePoolSize),
+        (_) => Pool(1),
```
jensjoha, line 188: does a pool of 1 make sense?

jensjoha (approved):

build_runner_core/lib/src/generate/build_impl.dart
davidmorgan: Done.

build_runner_core/lib/src/generate/build_impl.dart
davidmorgan: Good question! Looking at where the pool is used, there are no concurrent calls possible any more. So, removed. Thanks.

PR #3903, opened 2025-03-05
Use AOT snapshot for kernel_worker

a-siva:
Use AOT snapshot for kernel_worker


davidmorgan (approved):

isra-khan:
HI A-siva I need your help with something if we can get connect? 
https://stackoverflow.com/questions/79497480/flutter-socketexception-the-semaphore-timeout-period-has-expired-when-connect

mraleph:
@isra-khan see https://github.com/dart-lang/sdk/issues/52919#issuecomment-1634117628, also don't comment on random issues on random repos.

PR #3907, opened 2025-03-05
Make  mutations to `AssetNode` explicit.

davidmorgan:
I think `AssetGraph` can be used to figure out what changed so that it's not necessary to compute transitive digests.

But, that requires clarity on exactly how `AssetGraph`/`AssetNode` state gets updated :) which this starts to introduce.

I guess where this is probably headed is that `AssetGraph` owns most of the updates, rather than users making them directly on the nodes.

### DIFF
--- a/_test_common/lib/assets.dart
+++ b/_test_common/lib/assets.dart
@@ -13,14 +13,10 @@ AssetNode makeAssetNode([
   Digest? lastKnownDigest,
 ]) {
   var id = makeAssetId(assetIdString);
-  if (outputs == null) {
-    return AssetNode.source(id, lastKnownDigest: lastKnownDigest);
-  } else {
-    return AssetNode.source(
-      id,
-      lastKnownDigest: lastKnownDigest,
-      outputs: outputs,
-      primaryOutputs: outputs,
-    );
-  }
+  return AssetNode.source(
+    id,
+    lastKnownDigest: lastKnownDigest,
+    outputs: outputs,
+    primaryOutputs: outputs,
+  );
 }
--- a/_test_common/lib/assets.dart
+++ b/_test_common/lib/assets.dart
@@ -13,10 +13,14 @@ AssetNode makeAssetNode([
   Digest? lastKnownDigest,
 ]) {
   var id = makeAssetId(assetIdString);
-  var node = AssetNode.source(id, lastKnownDigest: lastKnownDigest);
-  if (outputs != null) {
-    node.outputs.addAll(outputs);
-    node.primaryOutputs.addAll(outputs);
+  if (outputs == null) {
+    return AssetNode.source(id, lastKnownDigest: lastKnownDigest);
+  } else {
+    return AssetNode.source(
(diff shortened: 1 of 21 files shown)


_test_common/lib/assets.dart
```diff
@@ -13,10 +13,14 @@ AssetNode makeAssetNode([
   Digest? lastKnownDigest,
 ]) {
   var id = makeAssetId(assetIdString);
-  var node = AssetNode.source(id, lastKnownDigest: lastKnownDigest);
-  if (outputs != null) {
-    node.outputs.addAll(outputs);
-    node.primaryOutputs.addAll(outputs);
+  if (outputs == null) {
```
jensjoha, line 16: Isn't this if unnecessary?
You return the same thing and the constructor takes null for the parameters you pass `outputs` in to below.

build_runner_core/lib/src/asset_graph/node.dart
```diff
@@ -30,25 +30,30 @@ class AssetNode {
 
   /// The assets that any [Builder] in the build graph declares it may output
   /// when run on this asset.
-  final Set<AssetId> primaryOutputs = <AssetId>{};
+  final Set<AssetId> _primaryOutputs;
+  Iterable<AssetId> get primaryOutputs => _primaryOutputs;
```
jensjoha, line 34: If you are supposed to go via the `inspect` below, why have these?

build_runner_core/lib/src/asset_graph/node.dart
```diff
@@ -79,47 +84,106 @@ class AssetNode {
...
+
+  /// Write access to collections in the node.
+  AssetNodeMutator get mutate => AssetNodeMutator(this);
+
+  /// Access to unmodifable views on collections in the node.
+  AssetNodeInspector get inspect => AssetNodeInspector(this);
+}
+
+/// Write access to collections in the node.
+extension type AssetNodeMutator(AssetNode node) {
```
jensjoha, line 172: Is this supposed to do something additional in the future?
As this is now I don't think it actually makes any change, or am I missing something?

_test_common/lib/assets.dart
davidmorgan: Thanks, done.

build_runner_core/lib/src/asset_graph/node.dart
davidmorgan: You only need to go via `inspect` if the method you want is on `Set`; almost all are already on `Iterable`.

build_runner_core/lib/src/asset_graph/node.dart
davidmorgan: Right, no actual change.

Rather a dev usability change, I can now search for uses of `mutate` :)

jakemac53: I was also wondering what the point was - maybe put that in a comment  on the extension type

PR #3917, opened 2025-03-11
Add script to run `dart run build_runner build` using deps from pub.

davidmorgan:
This is already useful for the handful of places in the repo that use codegen; but it becomes essential with the PRs adding use of `built_value` to `build_runner_core`, since you can't run `build_runner` using local deps at all if that package is broken.


tool/build_runner_build.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash --
```
jensjoha, line 1: This leaves Windows people out.
Why not write it as a dart script?

davidmorgan: Done.

It's still not going to work on Windows because it needs the pub cache location, but it's a lot closer to working on Windows :)

PTAL.

tool/build_runner_build.dart
```diff
@@ -0,0 +1,79 @@
...
+  File.fromUri(tempDirectory.uri.resolve('pubspec.yaml')).writeAsStringSync('''
+name: none
+environment:
+  sdk: ^3.7.0
+dependencies:
+  build_runner: '2.4.15'
+  build_test: any
+''');
+  Process.runSync('dart', ['pub', 'get'], workingDirectory: tempDirectory.path);
+  final pubConfigLines =
```
jensjoha, line 33: Could you explain what's going on here?
I don't understand why - after the `dart pub get` call - the newly created `.dart_tool/package_config.json` isn't fine as-is?

Also, when reading the file anyway why not use the path specified for `build_runner` to get the path instead of making hardcoding it to '$home/.pub-cache/hosted/pub.dev/build_runner-2.4.15' (making it probably not work in Windows?)

davidmorgan: Done.

Normally when you run `build_runner` it uses the pubspec of the package you're running it in, so if you are working on package `foo` and run `build_runner` then it runs using the full pubspec of `foo`.

This is needed because `build_runner` will compile generators referenced in the `foo` pubspec. So a pubspec that only knows about `build_runner` isn't sufficient to run `build_runner`.

When the package you are working on is `build_runner_core` this gets more complicated, because the pubspec refers to local versions of `build_runner` packages that may be broken as you work on them. But it _also_ refers to generators that `build_runner` needs to know about.

So: merge a published-packages version of `build` dependencies into the local pubspec, and this is what `build_runner` needs to run on `build_runner_core` even if it's broken.

tool/build_runner_build.dart
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
jensjoha: Isn't this a new file?

tool/build_runner_build.dart
jensjoha: Ok - maybe some comments in the code explaining that - and maybe adding comments in general regexps and weird replaces etc isn't exactly super readable.

tool/build_runner_build.dart
```diff
@@ -0,0 +1,82 @@
...
+    mergedConfig = mergedConfig.replaceAll(localConfigPattern, line);
+  }
+  final mergedConfigFile = File.fromUri(
+    tempDirectory.uri.resolve('package_config.json'),
+  );
+  mergedConfigFile.writeAsStringSync(mergedConfig);
+
+  final buildRunnerPath = pubConfigLines
+      .singleWhere((l) => l.contains('/build_runner-2.4.15'))
+      .replaceAll('      "rootUri": "', '')
```
jensjoha, line 65: This relies on the formatting of the json file? Maybe it should just be read as a json file instead...

tool/build_runner_build.dart
davidmorgan: Thanks, added a comment and got rid of the regexps :)

jensjoha:
👍

PR #3933, opened 2025-03-18
Updates handling

davidmorgan:
Pass updates from loading the graph to `Build`, so it always has updates unless it's a new build.

### DIFF
--- a/build_runner_core/lib/src/generate/build_definition.dart
+++ b/build_runner_core/lib/src/generate/build_definition.dart
@@ -34,7 +34,14 @@ class BuildDefinition {
   final AssetGraph assetGraph;
   final BuildScriptUpdates? buildScriptUpdates;
 
-  BuildDefinition._(this.assetGraph, this.buildScriptUpdates);
+  /// When reusing serialized state from a previous build: the file updates
+  /// since that build.
+  ///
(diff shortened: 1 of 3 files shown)


build_runner_core/lib/src/generate/build_definition.dart
```diff
@@ -197,15 +204,6 @@ class _Loader {
       assetGraph,
     );
     updates.addAll(_computeBuilderOptionsUpdates(assetGraph, buildPhases));
-    await assetGraph.updateAndInvalidate(
```
jensjoha, line 200: It is non-obvious to me why this is not needed any more?

jensjoha: (answering my own question: digging around I found that the updates are passed as `updatesFromLoad` for `BuildSeries` and that on `BuildSeries.run`s first run these are passed to Build.run, which runs _safeBuild with them that call _updateAssetGraph with them which calls assetGraph.updateAndInvalidate...)

jensjoha (approved):

build_runner_core/lib/src/generate/build_definition.dart
davidmorgan: Right :)

PR #3937, opened 2025-03-20
Add what changed and what is built to verbose logging.

davidmorgan:
Example output with `<--` marking new lines.

```
[FINE] Updates: added lib/lib2.dart, lib/lib3.dart, lib/lib4.dart, (1 more); modified lib/lib1.dart.  <--
[INFO] Updating asset graph...
[INFO] Updating asset graph completed, took 6ms

[INFO] Running build...
[FINE] Build lib/lib1.dart->.json_serializable.g.part because inputs digest changed from a7b9298 to d12df74. <--
[FINE] json_serializable on lib/lib1.dart:Running JsonSerializableGenerator - 1 of 2
[FINE] json_serializable on lib/lib1.dart:Running JsonLiteralGenerator - 2 of 2
[FINE] Build lib/lib2.dart->.json_serializable.g.part because lib/lib2.json_serializable.g.part was marked for build. <--
[FINE] json_serializable on lib/lib2.dart:Running JsonSerializableGenerator - 1 of 2

...

[FINE] Build lib/lib4.dart->.g.dart because lib/lib4.g.dart was marked for build. <--
[FINE] Build lib/lib5.dart->.g.dart because lib/lib5.g.dart was marked for build. <--
[INFO] Running build completed, took 1.1s
```

### DIFF
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -66,7 +66,7 @@ void _stdIOLogListener(LogRecord record, {required bool verbose}) =>
 /// splits the header for levels >= WARNING.
 String _recordHeader(LogRecord record, bool verbose) {
   var maybeSplit = record.level >= Level.WARNING ? '\n' : '';
-  return verbose || record.loggerName.contains(' ')
+  return record.loggerName.contains(' ')
       ? '${record.loggerName}:$maybeSplit'
       : '';
 }
--- a/build_runner/lib/src/logging/std_io_logging.dart
+++ b/build_runner/lib/src/logging/std_io_logging.dart
@@ -26,7 +26,7 @@ StringBuffer colorLog(LogRecord record, {required bool verbose}) {
   final level = color.wrap('[${record.level}]');
   final eraseLine = ansiOutputEnabled && !verbose ? '\x1b[2K\r' : '';
   var lines = <Object>[
-    '$eraseLine$level ${_recordHeader(record, verbose)}${record.message}',
+    '$eraseLine$level ${_recordHeader(record)}${record.message}',
   ];
 
   if (record.error != null) {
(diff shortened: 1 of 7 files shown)


build_runner/lib/src/logging/std_io_logging.dart
```diff
@@ -66,7 +66,7 @@ void _stdIOLogListener(LogRecord record, {required bool verbose}) =>
 /// splits the header for levels >= WARNING.
 String _recordHeader(LogRecord record, bool verbose) {
```
jensjoha, line 67: verbose is no longer used?

build_runner_core/lib/src/logging/log_renderer.dart
```diff
@@ -0,0 +1,83 @@
...
+  /// it matches.
+  String trimmedIds(Iterable<AssetId> ids, {String? trimPrefix}) {
+    final firstThree = ids
+        .map(id)
+        .take(3)
+        .map((s) => _trim(s, trimPrefix: trimPrefix));
+    if (firstThree.length == ids.length) {
+      return firstThree.join(', ');
+    } else {
+      return [...firstThree, '(${ids.length - 3} more)'].join(', ');
```
jensjoha, line 61: Doesn't have to be in this PR (or at all) but I always dislike seeing stuff like "a, b, c and 1 more" --- it would have been shorter and more informative to just say "a, b, c, d".

jensjoha (approved):

build_runner_core/lib/src/logging/log_renderer.dart
davidmorgan: Me too :)

I _think_ this is a good tradeoff, happy to tweak in future if it turns out not to be :)

Build outputs: mostly a build action has one output per input, so it won't hit the limit. This is logged a _lot_ of times, often multiple times per file in the main package, so I like having the limit just in case someone manages to end up with lots of outputs.

Files that changed: usually it will be exactly one file that the user just added / modified / changed, which won't hit the limit. But there's also the case that they changed git branch or did a "pub get" where there may be a large list of files; I wanted to avoid logging the whole list for that case as it's probably not interesting.

PR #3938, opened 2025-03-20
Change detection by comparing changed input lists

davidmorgan:
For #3811.

Determine when to rerun by comparing what changed to the list of inputs, instead of by computing transitive digests.

Anyone who has been watching my draft PRs will know this has taken a lot of attempts to get right :) ... I got it right in the end using [this experimental PR](https://github.com/dart-lang/build/pull/3936) which runs both old and new algorithms side by side and throws if there is any difference. So, this is 100% equivalent over all test cases including e2e tests.

This removes one of the bigger `O(n^2)` algorithms, the transitive digest computation when there is a library cycle. So, there is some fairly nice speedup.

```
Before this PR:
json_serializable
loop,1,22902,3935,5084
loop,100,24758,4193,8395
loop,250,30815,4838,12957
loop,500,56025,6941,27749
loop,750,102795,12461,45304
loop,1000,211007,18776,77577 <--

with this PR:
json_serializable
json_serializable
loop,1,23662,3833,5028
loop,100,25251,4051,7370
loop,250,30820,4974,12545
loop,500,49110,7378,24480
loop,750,101279,12873,37691
loop,1000,159219,17996,56268 <--
```

### DIFF
--- a/_test_common/lib/matchers.dart
+++ b/_test_common/lib/matchers.dart
@@ -45,11 +45,6 @@ class _AssetGraphMatcher extends Matcher {
     final result = <AssetNode>[];
     for (var node in graph.allNodes) {
       if (!checkPreviousInputsDigest) {
-        if (node.type == NodeType.generated) {
-          node = node.rebuild(
-            (b) => b.generatedNodeState.previousInputsDigest = null,
-          );
-        }
         if (node.type == NodeType.postProcessAnchor) {
           node = node.rebuild(
             (b) => b..postProcessAnchorNodeState.previousInputsDigest = null,
--- a/build_runner_core/CHANGELOG.md
+++ b/build_runner_core/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Add `NodeType` to `AssetNode`, remove subtypes. Make mutations explicit.
 - Use `built_value` for `AssetNode` and related types, and for serialization.
 - Add details of what changed and what is built to `--verbose` logging.
+- New change detection algorithm.
 
 ## 8.0.0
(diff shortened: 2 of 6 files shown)


jensjoha (approved):

davidmorgan:
> Any advice on how to try/run this improvement?

You can try it with dependency overrides and the commit has as [described here](https://github.com/dart-lang/build/discussions/3834), the commit hash in 01832d1e800375fe4ed1fb3695c1b3fb5e9c778b.

I don't think it's an improvement right now though :) along the way I made some simplifications that slow things down, I have one new optimization to land and a few old ones to reland before it'll be in a good state.

gmpassos:
> > Any advice on how to try/run this improvement?
> 
> You can try it with dependency overrides and the commit has as [described here](https://github.com/dart-lang/build/discussions/3834), the commit hash in [01832d1](https://github.com/dart-lang/build/commit/01832d1e800375fe4ed1fb3695c1b3fb5e9c778b).
> 

Should I use the parameter `--use-experimental-resolver`?

davidmorgan:
> > > Any advice on how to try/run this improvement?
> > 
> > 
> > You can try it with dependency overrides and the commit has as [described here](https://github.com/dart-lang/build/discussions/3834), the commit hash in [01832d1](https://github.com/dart-lang/build/commit/01832d1e800375fe4ed1fb3695c1b3fb5e9c778b).
> 
> Should I use the parameter `--use-experimental-resolver`?

Most of the changes apply both with and without that flag. So, either way is good :)

PR #3962, opened 2025-04-07
Switch to new resolver, remove old resolver.

davidmorgan:
For #3811.

Switch to using `AnalysisDriverModel` always, remove `BuildAssetUriResolver` and `--use-experimental-resolver` flag.

Clean up use of a few utility methods that were in the old file, `assetPath` and `packagePath`. Move deps parsing now used only by the "transitiveDigestsBuilder" into that file; note that the transitive digests builder is no longer enabled by default, it's not clear if anyone still uses it.

As a result of the switch, `AnalysisDriverModel` is now used in more tests, make it work :)

 - allow concurrent calls to `performResolve`, which happen when a read during resolve triggers an earlier phase generated file to be built
 - track file syncing by phase, add code to check whether something might have changed between the current phase and the last phase at which it was synced
 - make `readPhased` in `SingleStepReaderWriter` trigger builds of generated assets in earlier phases


build_runner_core/lib/src/generate/single_step_reader_writer.dart
```diff
@@ -444,14 +444,41 @@ class SingleStepReaderWriter extends AssetReader
...
+    );
+  }
+
+  @override
+  bool hasChanged(AssetId id, {required int comparedToPhase}) {
+    if (comparedToPhase == phase) return false;
+    if (_runningBuild == null) return false;
+
+    final node = _runningBuild.assetGraph.get(id);
+    if (node == null) return false;
```
jensjoha, line 474: Previously in this CL I saw that we deleted a file if reading it returned null or something.
Why does a null node here indicate a non-change?

jensjoha (approved):

build_runner_core/lib/src/generate/single_step_reader_writer.dart
davidmorgan: Null here means it's not in the asset graph, which means it's not a generated file; if you read it once, you'll get the same value by reading it again. So it only needs syncing once.

The dartdoc for the method now says Returns `false` for all other types of asset, including unknown assets." :)

Thanks.

PR #3966, opened 2025-04-08
Library cycle graph loader recursion support

davidmorgan:
For #3811.

Trying to use this on real builds quickly shows that it does not quite work yet, it needs to support reads triggering builds of earlier phase assets.

So, do that :) ... the main trick being to change the local "next IDs" list into a class-level list by phase, so recursive runs at earlier phases can finish the pending work that they need.

In doing so, simplify a bit: remove `_newAssets`, use `_graphsToComputeAtPhase` for phase 0 instead.

And expand test coverage to include recursive loads.

### DIFF
--- a/build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
+++ b/build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
@@ -2,6 +2,7 @@
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
+import 'dart:collection';
 import 'dart:math';
 
 import 'package:graphs/graphs.dart';
@@ -39,7 +40,19 @@ import 'phased_value.dart';
 /// Secondly, because the loader is for use _during_ the build, it might be that
 /// not all files have been generated yet. So, results must be returned based on
 /// incomplete data, as needed.
+///
+/// There can be multiple concurrent computations running on top of the same
+/// state, as noted in the implementation. It is allowed to call the methods
+/// `libraryCycleOf`, `libraryCycleGraphOf` and `transitiveDepsOf` while a prior
+/// call to any of the methods is still running, provided each newer call is at
+/// an earlier phase. This happens when a load does a read that triggers a build
+/// of a generated file in an earlier phase.
 class LibraryCycleGraphLoader {
+  /// The phases at which evaluation is currently running.
+  ///
+  /// Used to check that recursive loads are always to an earlier phase.
+  final List<int> _runningAtPhases = [];
+
   /// The dependencies of loaded assets, as far as is known.
   ///
   /// Source files do not change during the build, so as soon as loaded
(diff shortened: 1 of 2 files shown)


build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
```diff
@@ -340,20 +408,36 @@ class LibraryCycleGraphLoader {
...
+        'already running at : $_runningAtPhases',
+      );
+    }
+    _runningAtPhases.add(assetDepsLoader.phase);
+    try {
+      await _load(assetDepsLoader, id);
+      _buildCycles(assetDepsLoader.phase);
+      return _cycles[id]!;
+    } finally {
+      _runningAtPhases.removeLast();
```
jensjoha, line 430: I assume this is assumed to be the same as the one added above..? Maybe assert that?

build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
```diff
@@ -39,27 +40,39 @@ import 'phased_value.dart';
...
   final Map<AssetId, PhasedValue<AssetDeps>> _assetDeps = {};
 
-  /// Generated assets that were loaded before they were generated.
+  /// Assets to load.
   ///
-  /// The `key` is the phase at which they have been generated and can be read.
-  final Map<int, Set<AssetId>> _assetDepsToLoadByPhase = {};
+  /// The `key` is the phase to load them at or after. A [SplayTreeMap] is used
+  /// to keep the keys sorted so earlier phases are processed first.
+  final SplayTreeMap<int, List<AssetId>> _idsToLoad = SplayTreeMap();
```
jensjoha, line 73: Where is the sorted order used? I can't seem to find it.

jensjoha (approved):

build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
davidmorgan: It's implicit in `_nextIdToLoad`, tweaked the comment to mention it.

build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
davidmorgan: Good idea, thanks, done.

Removed the try/finally since the linter doesn't like me throwing inside a "finally", the code isn't actually designed to work when there are exceptions anyway.

build/lib/src/library_cycle_graph/library_cycle_graph_loader.dart
jensjoha: Yeah, it does `_idsToLoad.entries.where((entry) => entry.key <= upToPhase).first;` --- though wouldn't that just be a slow way to do `_idsToLoad.entries.first` then (well, and a check that i is in fact `<= upToPhase`)?

Issue #3977, opened 2025-04-16
Element2 migration


davidmorgan:
With analyzer 7.4.0 deprecation warnings have started appearing, so I'm filing this to reference when suppressing those.

Need to migrate to the element2 model and release, soon.

davidmorgan:
This is done :)

PR #3978, opened 2025-04-16
Ignore element2 deprecation warnings.

davidmorgan:
The warnings break the CI, so we need to ignore them for now.

Filed #3977 for the migration.


davidmorgan:
Going to pin the analyzer instead because we also hit a json_serializable issue with analyzer 7.4.0.

PR #3984, opened 2025-04-24
Do invalidation in `Build` instead of the asset graph

davidmorgan:
For #3811.

Prior to this PR, invalidation uses a combination of: 1) walk rdeps to mark nodes as "needs building" or "needs checking" in the asset graph; and 2) propagate failures via rdeps in `Build` and 3) check deps in `Build`.

With this PR, rdeps are not used: invalidation happens only "in order". State that doesn't need serializing, the progress of the checking the current build, is moved out of the asset graph into the `Build` class.

Changes to `AssetNode`:
 - Remove `pendingBuildAction`.
 - Rename `lastKnownDigest` -> `digest` and document how it's used.
 - Remove `isFailure`, add `result` which is `null` for not run, `false` for failed, `true` for succeeded
 - Remove `wasOutput` boolean: presence of `digest` indicates that it was output, because if there is an output, we immediately compute the digest; add getter checking `digest` instead
 - Tests pass `Digest([])` where a digest is needed to indicate `wasOutput` but the content doesn't matter.


build_runner_core/lib/src/generate/build.dart
```diff
@@ -753,115 +798,102 @@ class Build {
...
 
-    // Early bail out condition, this is a forced update.
-    if (firstNodeState.pendingBuildAction == PendingBuildAction.build) {
+    if (firstOutputState.result == null) {
       if (logFine) {
         _logger.fine(
-          'Build ${renderer.build(forInput, outputs)} because '
-          '${renderer.id(firstNode.id)} was marked for build.',
+          'Build ${renderer.build(primaryInput, outputs)} the outputs are '
+          'optional outputs than have become needed.',
```
jensjoha, line 839: that? Or maybe rephrase entirely?

jensjoha (approved):

build_runner_core/lib/src/generate/build.dart
davidmorgan: Thanks, done.

PR #3985, opened 2025-04-28
Add unit test coverage for incremental build invalidation.

davidmorgan:
For #3811.

Add a new type of unit test covering invalidation logic: build, modify inputs, build again, verify what gets written and/or deleted.

There is a lot to cover here:

The test cases added in this PR only have generated assets with inputs that are also primary inputs. Already we encounter one existing issue #3875 to do with not deleting assets that have become stale. For now I will preserve rather than fix this, aiming for a release that matches current behaviour before fixing :)

Follow-up PRs will build on this to add test coverage for more complex cases: 1) when generators have more single-asset inputs than just the primary input; and 2) when generators have inputs from resolving source, i.e. full transitive deps inputs.

Mostly this coverage is already present in existing unit tests and e2e tests, but I wanted to add smaller+faster+clearer coverage before the next PRs changing the invalidation logic.

### DIFF
--- a/build_runner_core/test/invalidation/invalidation_tester.dart
+++ b/build_runner_core/test/invalidation/invalidation_tester.dart
@@ -1,0 +1,306 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:convert';
+
+import 'package:build/build.dart';
+import 'package:build_test/build_test.dart';
+import 'package:built_collection/built_collection.dart';
+import 'package:crypto/crypto.dart';
+import 'package:logging/logging.dart';
+import 'package:test/test.dart';
+
+/// Test case which build steps are triggered by changes to files.
+///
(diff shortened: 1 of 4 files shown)


build_runner_core/test/invalidation/invalidation_tester.dart
```diff
@@ -0,0 +1,306 @@
...
+    _sourceAssets.clear();
+    for (final name in names) {
+      _sourceAssets.add(name.assetId);
+    }
+  }
+
+  /// Adds a builder to the test.
+  ///
+  /// [from] and [to] are the input and output extension of the builder,
+  /// without ".dart". So `from: '', to: '.g.dart'` is a builder that takes
```
jensjoha, line 54: From

```
  TestBuilder(this._tester, String from, Iterable<String> to, this.isOptional)
    : buildExtensions = {'$from.dart': to.map((to) => '$to.dart').toList()};
```

it seems that the "to" here should just be `.g`?

jensjoha (approved):

build_runner_core/test/invalidation/invalidation_tester.dart
davidmorgan: Good catch, thanks. Done.

PR #3987, opened 2025-04-29
More inputs invalidation test; fix for deletions

davidmorgan:
For #3811.

More test coverage. I found another bug which _is_ a regression from published behaviour, so the second commit fixes it. 

The bug is about a change of input from present->deleted not triggering generation.

The problem is that the graph update code removes inputs if the input was removed, so by the time `_buildShouldRun` is reached there is no more record of the dependency. There must have been some code accounting for that case that got simplified away :) ... fix it by keeping the inputs for deleted files, changing the deleted nodes to `missingSource` nodes.

### DIFF
--- a/build_runner_core/test/asset_graph/graph_test.dart
+++ b/build_runner_core/test/asset_graph/graph_test.dart
@@ -71,8 +71,8 @@ void main() {
           nodes.add(testAddNode());
         }
         graph
-          ..remove(nodes[1].id)
-          ..remove(nodes[4].id);
+          ..removeForTest(nodes[1].id)
+          ..removeForTest(nodes[4].id);
 
         expectNodeExists(nodes[0]);
         expectNodeDoesNotExist(nodes[1]);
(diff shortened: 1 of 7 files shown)


build_runner_core/test/asset_graph/graph_test.dart
```diff
@@ -513,14 +513,6 @@ void main() {
             expect(globNode.globNodeState!.inputs, contains(coolAssetId));
             expect(globNode.globNodeState!.results, contains(coolAssetId));
             await checkChangeType(ChangeType.REMOVE);
-            expect(
```
jensjoha, line 516: I don't understand why is this removed.

jensjoha (approved):

build_runner_core/test/asset_graph/graph_test.dart
davidmorgan: It used to be that the asset graph update would invalidate the glob node _and_ modify the result to remove the deleted node; now it just invalidates the glob node, so in this test it's not updated yet.

I have a work-in-progress PR that refactors away from using the graph to track invalidation, at which point this test goes away altogether.

Thanks.

PR #3995, opened 2025-05-05
Add back write caching, simplify caching, use sync I/O

davidmorgan:
For #3811.

This is more about adding back functionality removed during the refactoring than performance. The reason write caching was added was to prevent files appearing during the build from causing reanalysis.

But it does seem to be about 10% faster.

 - Add write caching to the existing `FilesystemCache`
 - Simplify: make the cache synchronous, use sync I/O, remove LruCache
 - Because caching is now either passthrough or combined with write caching, `invalidate` on writing outputs is no longer needed
 - Remove a few more `Future.wait` in favour of for loops

Possibly some cache size limit is still a good idea, although for most projects the total source size seen should be small, and there is always the "low resource mode" flag to turn it off.


build/lib/src/state/filesystem_cache.dart
```diff
@@ -16,144 +14,233 @@ abstract interface class FilesystemCache {
...
-    (value) => value.lengthInBytes,
-  );
-
-  /// Pending [readAsBytes] operations.
-  final _pendingBytesContentCache = <AssetId, Future<Uint8List>>{};
+  final _bytesContentCache = <AssetId, Uint8List>{};
 
   /// Cached results of [exists].
   ///
   /// Don't bother using an LRU cache for this since it's just booleans.
```
jensjoha, line 106: This comment no longer makes sense.

davidmorgan:
There is a possible issue with memory usage _before_ this PR https://github.com/dart-lang/build/issues/3999

So I think it is not good to remove the lru cache and complicate things further :)

The write part of the cache needs to not do discards, so that needs a bit of separate logic if keeping the lru cache; it should be a lot simpler now that everything is sync. I'll update the PR to do that then send for review again. Sorry for the noise!

build/lib/src/state/filesystem_cache.dart
```diff
@@ -16,144 +14,233 @@ abstract interface class FilesystemCache {
...
-    return _pendingBytesContentCache.putIfAbsent(id, () async {
-      final result = await ifAbsent();
-      _bytesContentCache[id] = result;
-      unawaited(_pendingBytesContentCache.remove(id));
-      return result;
-    });
+  bool exists(AssetId id, {required bool Function() ifAbsent}) {
+    final maybeResult = _canReadCache[id];
+    if (maybeResult != null) return maybeResult;
+    return _canReadCache[id] = ifAbsent();
```
jensjoha, line 146: I got confused about the `_canReadCache` name. Maybe `_existsCache` would make more sense?

jensjoha (commented):
Leaving the comments and posting with your updated comment.

jensjoha:
One thing though, with you changing this anyway: If not too much trouble, could you try doing the sync change as a separate PR? It would be interesting to know how much that changes on its own.

davidmorgan:
Closing in favour of https://github.com/dart-lang/build/pull/4001

PR #4000, opened 2025-05-08
Use sync I/O.

davidmorgan:
For #3811.

Use sync I/O for file operations, simplify caching.

Also: when reading as string, check the bytes cache before reading; in some cases this saves a read.

Split out from https://github.com/dart-lang/build/pull/3995. It looks like the 10% speedup was due to the sync change (+maybe saving one read) and not the write cache, which makes sense: we could only see a performance benefit of write caching if one generator read the output of another, which is not the case in those benchmarks.

### DIFF
--- a/build/lib/src/state/filesystem.dart
+++ b/build/lib/src/state/filesystem.dart
@@ -6,35 +6,19 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:typed_data';
 
-import 'package:pool/pool.dart';
-
 /// The filesystem the build is running on.
 ///
 /// Methods behave as the `dart:io` methods with the same names, with some
 /// exceptions noted in the docs.
 abstract interface class Filesystem {
   /// Whether the file exists.
-  Future<bool> exists(String path);
-
-  /// Whether the file exists.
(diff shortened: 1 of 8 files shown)


gmpassos:
Interesting change.

IMHO, this needs to be benchmarked on different hardware, since SSD speed will affect the results (at least it will be interesting to see how). Note that a large project won’t load too much data, as it’s only Dart source code, but it will need to resolve and access hundreds of different files.

(I’m totally ignoring HDDs for real-world cases.)

davidmorgan:
> Interesting change.
> 
> IMHO, this needs to be benchmarked on different hardware, since SSD speed will affect the results (at least it will be interesting to see how). Note that a large project won’t load too much data, as it’s only Dart source code, but it will need to resolve and access hundreds of different files.
> 
> (I’m totally ignoring HDDs for real-world cases.)

Sync reads are pretty consistently faster than async reads in Dart; but if we do find a case where async is needed, I suspect the way to do it is to add a "batch read" that uses async rather than this file-by-file API.

For now this is more about reducing complexity of the change to put back write caching :)

PR #4001, opened 2025-05-08
Add back write caching.

davidmorgan:
Retry of #3995.

Switch to sync I/O landed already in #4000.

Differences to #3995:

 - don't remove LruCache; this means writes need caching separately so they don't get evicted, cache in `_Write` class
 - noticed and fixed that deletes were being done to a writer with no cache; this seems to have sped things up a little bit (maybe a few percent), generated files are now written as new versions instead of deleted then written later

Fix `copyWith` for `onCanReadController` in the test ReaderWriter, as this now matters for one test.


build/lib/src/state/filesystem_cache.dart
```diff
@@ -34,6 +49,25 @@ abstract interface class FilesystemCache {
...
+  void writeAsString(
+    AssetId id,
+    String contents, {
+    Encoding encoding = utf8,
+    required void Function() writer,
+  });
+
+  /// Deletes [id].
+  ///
+  /// [deleter] is a function that does the actual write. If this cache does
```
jensjoha, line 67: actual write? Is it actual delete?

build/lib/src/state/filesystem_cache.dart
```diff
@@ -106,6 +185,24 @@ class InMemoryFilesystemCache implements FilesystemCache {
     return result;
   }
 
+  @override
```
jensjoha, line 188: twice?

build/lib/src/state/filesystem_cache.dart
```diff
@@ -129,4 +236,40 @@ class InMemoryFilesystemCache implements FilesystemCache {
...
   }
+
+  @override
+  void writeAsString(
+    AssetId id,
+    String contents, {
+    Encoding encoding = utf8,
+    required void Function() writer,
+  }) {
+    _stringContentCache[id] = contents;
```
jensjoha, line 247: should there be some handling here of non-utf8 encoding? (I seem to recall a comment from some point about only saving utf8-encoded strings, also above we only read from _stringContentCache if using utf8)

build_runner_core/test/generate/build_test.dart
```diff
@@ -600,10 +600,20 @@ void main() {
           outputs: {'a|web/a.txt.copy': 'a'},
         );
 
+        var copyId = makeAssetId('a|web/a.txt.copy');
+
+        var readCompleter = Completer<bool>();
```
jensjoha, line 605: maybe `canReadCompleter`? Or `canReadInBuild`?

build_test/lib/src/test_builder.dart
```diff
@@ -196,8 +200,16 @@ Future<TestBuilderResult> testBuilder(
 /// builder to run for all inputs. To use the default builder config instead,
 /// set [testingBuilderConfig] to `false`.
 ///
-/// Returns a [TestBuilderResult] with the [TestReaderWriter] used for
-/// the build, which can be used for further checks.
+/// Optionally pass [readerWriter] to pass in the filesystem that will be
```
jensjoha, line 203: The two `pass`es makes it a little hard to read. Maybe, as above, something like `Optionally pass [ReaderWriter] to set the filesystem...`

jensjoha (approved):

build/lib/src/state/filesystem_cache.dart
davidmorgan: Yes, this was incorrect; fixed and added test coverage.

Could you take another look, please?

build_runner_core/test/generate/build_test.dart
davidmorgan: Done.

build_test/lib/src/test_builder.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks!

Could you please take another look for the non-UTF-8 question? Now with test coverage, which I should have added in the first place.

build/lib/src/state/filesystem_cache.dart
```diff
@@ -209,12 +208,14 @@ class InMemoryFilesystemCache implements FilesystemCache {
...
   }) {
+    // Encodings other than utf8 do not use `_stringContentCache`. Read as
+    // bytes then convert, instead.
     if (encoding != utf8) {
       final bytes = readAsBytes(id, ifAbsent: ifAbsent);
       return encoding.decode(bytes);
     }
 
-    // Check _stringContentCache first to use it as a cache for conversion of
+    // Check `_stringContentCache` first to use it as a cache for conversion of
```
jensjoha, line 218: I can't seem to put a comment at the right place, but line 223-235 could basically be replaced by a call like `final bytes = readAsBytes(id, ifAbsent: ifAbsent);` (as above).

build/lib/src/state/filesystem_cache.dart
```diff
@@ -244,15 +245,18 @@ class InMemoryFilesystemCache implements FilesystemCache {
...
-    _stringContentCache[id] = contents;
-    _bytesContentCache.remove(id);
-    _existsCache[id] = true;
-
+    // Encodings other than utf8 do not use `_stringContentCache`.
+    if (encoding == utf8) {
+      _stringContentCache[id] = contents;
+    } else {
+      _stringContentCache.remove(id);
+    }
```
jensjoha, line 253: and this stuff could (almost) be replaced by a call to `writeAsBytes`.

build/lib/src/state/filesystem_cache.dart
davidmorgan: So it can ... thanks! Done.

build/lib/src/state/filesystem_cache.dart
davidmorgan: Thanks; extracted common `_writeBytes`.

Issue #3999, opened 2025-05-08
OOM with prerelease version


davidmorgan:
Discovered here

https://github.com/dart-lang/build/discussions/3834#discussioncomment-13058030

davidmorgan:
@amrgetment thanks for running with the extra logging.

I was hoping to see some output like "Asset graph size: <number of bytes>", but I don't see that in the output you just posted?

"Failed after 23m" ... 23m is huge, is that normal? i.e. did it take that long before?

A sample repro would be fantastic.

I'll anyway see if I can figure something out, and update here if I do.

Thanks a lot for your help :)

amrgetment:
23m is too much, it should be like minutes or so
I will prepare a sample repo over the weekend

davidmorgan:
@amrgetment I tried a bunch of big (1000 file) examples using `retrofit_generator`, getting generation time up to about 1m20s, but wasn't able to cause anything obviously bad with memory.

Do you have any other generators I should throw into the mix?

Another random question: is your code in one package, or does it split across multiple packages?

I'll keep trying things: I'm close to releasing the new version so any random testing I can do is useful.

Looking forward to that repro :)

amrgetment:
I use a lot of generators:
```
  # Generators
  retrofit_generator: ^9.1.9
  build_runner: ^2.4.15
  auto_route_generator: ^10.0.1
  json_serializable: ^6.9.5
  assets_scanner_plus: ^2.0.0
  injectable_generator: ^2.7.0

```
My code in a single package

amrgetment:
My project almost covered all generators 😅

davidmorgan:
Thanks! I'll take a look at them all. So far I added auto_route_generator, nothing suspicious yet.

I uploaded new version https://github.com/dart-lang/build/pull/3998/commits/15e775f898ca1a5952cfc89b43e0b1f1c992e30d which logs which generator is running, that might give us a clue :)

davidmorgan:
@amrgetment Great news! I have a repro :)

No idea yet what triggered it, but I managed to hit a state where it gets stuck with RAM usage climbing forever until it fails with OOM.

So hopefully, no need for you to spend time on a repro now.

I will update here once I either have a fix, or have failed to find one :)

amrgetment:
So it is not related to a specific generator, great 👍 , Good Luck

amrgetment:
I think those logs need to be a part of `--verbose` for future needs

davidmorgan:
[...]

PR #4010, opened 2025-05-14
Add invalidation stress test.

davidmorgan:
For #3811.

Add invalidation stress test, add smaller test cases for and fix two bugs that it found. It also would have caught #3999.

`PhasedAssetDeps` bug: this is related to but distinct from #3999, the loaded asset deps data was reporting its phase as the first phase with incomplete data, causing data at later phases to be not readable. In fact, it can have incomplete data for arbitrary phases, if the load was scheduled for load at a later phase but no later phase picked up the work because there was no more resolution done. Fix by just treating the data as complete for this case: it won't be used, because the loaded data is only used to reconstruct the import graphs exactly as they were in the previous build.

`LibraryCycleGraphLoader` bug: the comment states that `_idsToLoad` is either empty or unchanged; this is not quite true: a recursive call can add work at a later phase; it's possible for a recursive call to add work at the current phase, breaking that invariant. This is unlikely to actually happen in practice, so just fall back to a slower remove.

### DIFF
--- a/build_runner_core/test/invalidation/invalidation_tester.dart
+++ b/build_runner_core/test/invalidation/invalidation_tester.dart
@@ -4,7 +4,9 @@
 
 import 'dart:async';
 import 'dart:convert';
+import 'dart:math';
 
+import 'package:analyzer/dart/element/element.dart';
 import 'package:build/build.dart';
 import 'package:build_runner_core/src/util/constants.dart';
 import 'package:build_test/build_test.dart';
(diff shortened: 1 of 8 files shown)


build_runner_core/test/invalidation/invalidation_tester.dart
```diff
@@ -220,7 +262,11 @@ enum OutputStrategy {
   /// Output with fixed content.
   fixed,
 
-  /// Output with digest of all files that were read.
+  /// Output with a list of inputs: each line `$name,$hash` for inputs
+  /// that could be read, or `$name` if not.
+  ///
+  /// If a builder resolves files, this includes all transitively resolved
+  /// files and their hashes.s
```
jensjoha, line 269: delete trailing `s`

jensjoha (approved):

build_runner_core/test/invalidation/invalidation_tester.dart
davidmorgan: Done; thanks.

PR #4028, opened 2025-06-05
Use a synchronous StreamController for daemon logs

nshahan:
The logging from build_runner when running builds through a daemon was asynchronous and was causing some `webdev build` commands to missing the final status message.

Fixes: https://github.com/dart-lang/webdev/issues/2489

### DIFF
--- a/build_runner/lib/src/daemon/daemon_builder.dart
+++ b/build_runner/lib/src/daemon/daemon_builder.dart
@@ -231,7 +231,7 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
     DaemonOptions daemonOptions,
   ) async {
     var expectedDeletes = <AssetId>{};
-    var outputStreamController = StreamController<ServerLog>();
+    var outputStreamController = StreamController<ServerLog>(sync: true);
 
     var environment = BuildEnvironment(
       packageGraph,


nshahan:
It's totally fine if you want to include this into your upcoming changes too.

davidmorgan:
Thanks.

Do you want to release this now? There don't seem to be any other changes to build_daemon, so it should just work, and it will avoid getting tangled with the much larger pending release.

If so please also

 - update the CHANGELOG to rename `4.0.5-wip` to `4.0.5` and to add e.g. `Fix missing last log message in webdev.'
 - update the pubspec.yaml to version `4.0.5`

then I'll merge and publish.

Thanks :)

nshahan:
It would be nice to get the fix published but it is not urgent. We have worked around the missing message in the output to get our tests passing. 

Just to clarify, the touched file in this PR is actually part of package:build_runner. Is that package otherwise untouched and safe to be published?

davidmorgan:
Oh. Not sure why I thought it was build_daemon :)

Then, it'll go out with the main release after all.

It would be relatively easy to backport to an old version if needed, so please do let me know if you get tired of waiting for the release :) thanks.

davidmorgan (approved):

PR #4031, opened 2025-06-13
Simplify and fix clean command.

davidmorgan:
The previous PR broke the `clean` command because I removed a timing block and changed the meaning of the early `return` statements.

On closer inspection, `clean` can't actually find the asset graph in normal use because the `assetGraphPath` changes based on the script location, and `clean` runs from the main script while all the `build` commands run from the generated build script. So I think `clean` never actually deleted "generated source files".

So remove that code, including the early return that I broke, and just delete the whole cache folder.

### DIFF
--- a/build_runner/bin/src/commands/clean.dart
+++ b/build_runner/bin/src/commands/clean.dart
@@ -6,10 +6,8 @@ import 'dart:async';
 
 import 'package:args/args.dart';
 import 'package:args/command_runner.dart';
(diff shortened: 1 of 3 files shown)


gmpassos:
Just confirming that this has fixed the issue. 👍

Tested with commit:
https://github.com/dart-lang/build/commit/94d20ffcfee847dc7cf2503c6edefe3a0d106cf9

davidmorgan:
Thanks :)

Issue #4043, opened 2025-06-17
`build_test` feature to load transitive dep sources


davidmorgan:
Per https://github.com/dart-lang/build/discussions/4036#discussioncomment-13487669.

davidmorgan:
Now released in `3.2.0`.

PR #4048, opened 2025-06-17
Support for loading all sources for tests.

davidmorgan:
For #4043.


davidmorgan:
@simolus3 how about this?

I realized loading imported files is probably unnecessary: it's much simpler to just load everything, and seems fast enough.

simolus3:
`loadIsolateSources` is pretty close to what I'm doing today (except that I'm hardcoding packages, but that doesn't save much so I'm keen to switch to this API once it's available). Thanks!

jensjoha (approved):

Issue #4083, opened 2025-07-08
Faster skip when annotation is not imported


davidmorgan:
If you apply an annotation-triggered builder using source_gen to a codebase that mostly doesn't use it, every file gets analyzed to check for the annotation. _Edit: correction; files with _no_ annotations are skipped in some circumstances, see [this comment](https://github.com/dart-lang/build/issues/4083#issuecomment-3049523704)._

It seems like a good idea to bail out earlier based on a faster check. For example: check if the package that contains the annotation is imported.

davidmorgan:
Skip based on the presence of `package:built_value/`, benchmark modified so most files do not use the builder.

Without skip:

```
dart run _benchmark --generator=built_value --build-repo-path=$PWD/.. benchmark --size=2000 --mostly-no-codegen
built_value
shape,libraries,clean/ms,no changes/ms,incremental/ms
loop,2000,32434,8522,15341
forwards,2000,32051,8433,16447
backwards,2000,39872,9340,21883
```

With skip:

```
built_value
shape,libraries,clean/ms,no changes/ms,incremental/ms
loop,2000,27649,8904,11821
forwards,2000,28140,8786,11576
backwards,2000,24954,8866,9272
```

rrousselGit:
By that standard, we should be able to skip files that do import the annotation but never apply it within the library.

davidmorgan:
> By that standard, we should be able to skip files that do import the annotation but never apply it within the library.

I was going to ask for your feedback :) ... to help end on a good API. See #4084 for the change I used for benchmarking.

What I currently have is something declarative in the yaml:

```
builders:
  built_value:
    defaults:
      options:
        build_runner:
          build_if:
            primary_input_contains: "package:built_value/"
```

It would be possible to do exactly the same in the builder code; if you're using source_gen then that would need a change to source_gen. But I like the idea of starting to add declarative features for builders: declarative is much more amenable to optimization :)

`primary_input_contains` with a single value is the simplest thing I could think of, but probably insufficient.

We could also do

 - check multiple patterns, combine with AND or OR; so you could check import+annotation
 - regexp matchers
 - check the entire library, i.e. find all parts then match

I expect most builders can just use the single string match on the primary input, examples of more complex cases would be interesting.

rrousselGit:
I guess the question is: What exactly are we skipping here?

Writing to a file? Resolving Elements for a library?

___

That "build_if" idea sounds like it'd break re-exports of annotations.

[...]

PR #4084, opened 2025-07-08
Add option to skip based on primary input content.

davidmorgan:
For #4083


build_runner_core/lib/src/package_graph/build_triggers.dart
```diff
@@ -0,0 +1,201 @@
...
+  static final RegExp _regexp = RegExp(r'^[a-z][a-z0-9_/.]*$');
+  String get import;
+
+  ImportBuildTrigger._();
+  factory ImportBuildTrigger(String import) =>
+      _$ImportBuildTrigger._(import: import);
+
+  @override
+  bool triggersOnPrimaryInput(String source) {
+    return source.contains('package:$import');
```
jensjoha: Technically an import string can be split into multiple lines, e.g.

```dart
import 'd' /* hello */ 'a'
//there
'r'
// I'm
't'
// a
':'
// perfectly
'i'
// valid
'o'
// import
''
// !
;
```

(and I have seen than --- not as absurdly as I did above but as I recall to make it fit on lines or something)

build_runner_core/lib/src/package_graph/build_triggers.dart
```diff
@@ -0,0 +1,201 @@
...
+  static final RegExp _regexp = RegExp(r'^[a-zA-Z_][a-zA-Z0-9]*$');
+  String get annotation;
+
+  AnnotationBuildTrigger._();
+  factory AnnotationBuildTrigger(String annotation) =>
+      _$AnnotationBuildTrigger._(annotation: annotation);
+
+  @override
+  bool triggersOnPrimaryInput(String source) {
+    return source.contains('@$annotation');
```
jensjoha: I believe `@ myAnnotation` is valid too.

build_runner_core/lib/src/package_graph/build_triggers.dart
```diff
@@ -0,0 +1,201 @@
...
+    final buildTriggers = <String, Set<BuildTrigger>>{};
+    final warningsByPackage = <String, List<String>>{};
+    for (final entry in configByPackage.entries) {
+      final packageName = entry.key;
+      final config = entry.value;
+      final triggersByBuilder = config.triggersByBuilder;
+      for (final entry in triggersByBuilder.entries) {
+        final builderName = entry.key;
+
+        if (!builderName.contains(_builderNamePattern)) {
```
jensjoha, line 54: Won't this be found below?

build_runner_core/lib/src/package_graph/build_triggers.dart
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/// A build trigger: a heuristic that possibly skips running a build step based
+/// on the primary input source.
+abstract class BuildTrigger {
+  /// Parses a trigger, returning `null` on failure, optionally with a warning
+  /// message.
+  ///
+  /// The only supported trigger is [ImportBuildTrigger].
+  static (BuildTrigger?, String?) tryParse(String trigger) {
```
jensjoha: Could this be private?
If it has to be public I think I'd prefer a "proper return type" rather than a record.

jensjoha (commented):
This seems like an awful lot of manual config that's bound to go wrong at some point.
We can't do something similar without manual config?

build_runner_core/lib/src/package_graph/build_triggers.dart
davidmorgan: No, the checks below are just on the triggers, not the builder names.

build_runner_core/lib/src/package_graph/build_triggers.dart
davidmorgan: Thanks; made private.

build_runner_core/lib/src/package_graph/build_triggers.dart
davidmorgan: I was hesitant to do full parsing simply because there is no convenient place to dedupe the parsing work; but I do think it's better and the way to go in the end.

Filed https://github.com/dart-lang/build/issues/4128 to figure out a way to save parsing multiple times, and changed the code to do parsing so annotation + import checks are both (hopefully) now correct.

build_runner_core/lib/src/package_graph/build_triggers.dart
davidmorgan: Addressed by switch to parsing.

Also, I realized that annotations should be checked for in all compilation units, it's not like imports where they can only be in the main library. So, I added finding and loading all the parts for the check, and test coverage for that.

build_runner_core/lib/src/package_graph/build_triggers.dart
```diff
@@ -190,8 +204,27 @@ abstract class AnnotationBuildTrigger
...
+  bool triggersOn(List<CompilationUnit> compilationUnits) {
+    for (final compilationUnit in compilationUnits) {
+      for (final declaration in compilationUnit.declarations) {
+        for (final metadata in declaration.metadata) {
+          var name = metadata.name.name;
+          if (metadata.constructorName != null) {
+            name = '$name.${metadata.constructorName}';
+          }
+          if (annotation == name) return true;
+          final periodIndex = name.indexOf('.');
```
jensjoha, line 234: Do we end up comparing to `metadata.constructorName` or could `name` (before line 216) have dots in it where we want to remove the first part? Could there be a comment explaining? (and either compare directly to `metadata.constructorName` or make sure we don't depending on what we're actually trying to achieve)

jensjoha (approved):

build_runner_core/lib/src/package_graph/build_triggers.dart
davidmorgan: Thanks, added comment + a bit of test coverage on the awkward case. (Import prefixes and class names can't actually be distinguished).

PR #4095, opened 2025-07-12
Clean up test helpers.

davidmorgan:
Remove unused test helpers, inline test helpers only used once or that don't add enough.


_test_common/lib/builders.dart
```diff
@@ -48,3 +36,40 @@ class DelegatingBuilder implements Builder {
   @override
   Future build(BuildStep buildStep) async => delegate.build(buildStep);
 }
+
+class PlaceholderBuilder extends Builder {
+  final String inputExtension;
+  final Map<String, String> outputExtensionsToContent;
```
lrhn: The "plural-to-singular" name confuses me.
(Also, what is "Content"? More docs could make this readable to someone who doesn't know the rest of the code. :wink: )

_test_common/lib/builders.dart
```diff
@@ -48,3 +36,40 @@ class DelegatingBuilder implements Builder {
...
+
+  @override
+  Map<String, List<String>> get buildExtensions => {
+    inputExtension: outputExtensionsToContent.keys.toList(),
+  };
+
+  PlaceholderBuilder(
+    this.outputExtensionsToContent, {
```
lrhn: It takes a map as argument. Is it OK if that map changes afterwards? (If not, document that it shouldn't. If so, ... document that anyway.)

_test_common/lib/builders.dart
```diff
@@ -48,3 +36,40 @@ class DelegatingBuilder implements Builder {
...
+    this.inputExtension = r'$lib$',
+  });
+
+  @override
+  Future build(BuildStep buildStep) async {
+    outputExtensionsToContent.forEach((extension, content) {
+      buildStep.writeAsString(
```
lrhn: Function is async, doesn't use `await`, calls an async `writeAsString` in a synchronous loop and doesn't await the result of that. Waiting for the returned future does not wait for the writing, which is *probably* intended for this API.

Either:
```dart
  for (final MapEntry(key: extension, value: content)) {
    await buildStep.writeAsString(
      _outputId(buildStep.inputId, inputExtension, extensions)),
      content,
    );
  }
```
or 
```dart
  await [for (final MapEntry(key: extension, value: content))
    buildStep.writeAsString(
      _outputId(buildStep.inputId, inputExtension, extensions)),
      content,
    )
  ].wait;
```

_test_common/lib/builders.dart
```diff
@@ -48,3 +36,40 @@ class DelegatingBuilder implements Builder {
   @override
   Future build(BuildStep buildStep) async => delegate.build(buildStep);
 }
+
+class PlaceholderBuilder extends Builder {
+  final String inputExtension;
```
lrhn: The extension always includes the leading `.`, right?

_test_common/lib/common.dart
```diff
@@ -25,40 +23,3 @@ Digest computeDigest(AssetId id, String contents) {
...
-    inputExtension: outputExtensionsToContent.keys.toList(),
-  };
-
-  PlaceholderBuilder(
-    this.outputExtensionsToContent, {
-    this.inputExtension = r'$lib$',
-  });
-
-  @override
-  Future build(BuildStep buildStep) async {
```
lrhn, line 44: Ah, so you only moved the code, you dind't write it. (So technically OK to keep as-is). It's still suspicious hat it doen't await anything.

build_runner_core/test/environment/create_merged_dir_test.dart
```diff
@@ -613,3 +612,71 @@ void _expectAllFiles(Directory dir) {
...
+class TestBuildEnvironment implements BuildEnvironment {
+  final TestReaderWriter _readerWriter;
+
+  @override
+  TestReaderWriter get reader => _readerWriter;
+
+  @override
+  TestReaderWriter get writer => _readerWriter;
+
+  /// If true, this will throw a [NonInteractiveBuildException] for all calls to
```
lrhn:
```dart
  /// Whether to throw ...`
```

(Otherwise you have technically not said what `false` means :stuck_out_tongue:).

build_runner_core/test/environment/create_merged_dir_test.dart
```diff
@@ -613,3 +612,71 @@ void _expectAllFiles(Directory dir) {
...
+  /// [prompt].
+  final bool throwOnPrompt;
+
+  /// The next response for calls to [prompt]. Must be set before calling
```
lrhn: New paragraph after first sentence.

build_runner_core/test/environment/create_merged_dir_test.dart
```diff
@@ -613,3 +612,71 @@ void _expectAllFiles(Directory dir) {
...
+  }
+
+  int? _nextPromptResponse;
+
+  TestBuildEnvironment({
+    TestReaderWriter? readerWriter,
+    this.throwOnPrompt = false,
+  }) : _readerWriter = readerWriter ?? TestReaderWriter();
+
+  /// Prompt the user for input.
```
lrhn: "Promts ...."

build_runner_core/test/environment/create_merged_dir_test.dart
```diff
@@ -613,3 +612,71 @@ void _expectAllFiles(Directory dir) {
...
+  /// chosen option is returned.
+  ///
+  /// If this environmment is non-interactive (such as when running in a test)
+  /// this method should throw [NonInteractiveBuildException].
+  @override
+  Future<int> prompt(String message, List<String> choices) {
+    if (throwOnPrompt) throw NonInteractiveBuildException();
+
+    assert(_nextPromptResponse != null);
+    return Future.value(_nextPromptResponse);
```
lrhn: Consider writing `_nextPromptResponse!`.

The `Future.value` technically accepts `int?` as argument, but will throw if the value is `null`. Which it probably isn't if asserts are enabled.
I really, really want to disallow passing `null` as an argument when `T` is not nullable, so start preparing for this line to become a type error.

_test_common/lib/builders.dart
davidmorgan: Please see next comment :)

_test_common/lib/builders.dart
davidmorgan: This whole class is confusing: builder APIs talk in terms of file extensions, but "placeholders" are a special case, and this builder is just about placeholders.

So: renamed the variables to say what they actually do, then added a comment in the place where they are called "extensions" to explain that they're a special case of that API.

Thanks :)

_test_common/lib/builders.dart
davidmorgan: Made it a BuiltMap, I like those ;)

_test_common/lib/builders.dart
davidmorgan: Oh, I was thinking of getting rid of the `forEach`, a lint would have noticed that if I had.

I like the first one--done--thanks :)

_test_common/lib/common.dart
davidmorgan: Yup, I guess it just happened to work. It's actually only used in one test ... I guess it's not super well tested ;)

Anyway it's much clearer now, thanks. Placeholders are a bit of a corner case that people may not be aware of, so it's nice to explain clearly where we test it.

build_runner_core/test/environment/create_merged_dir_test.dart
davidmorgan: Yup yup, done. Thanks.

build_runner_core/test/environment/create_merged_dir_test.dart
davidmorgan: Done.

build_runner_core/test/environment/create_merged_dir_test.dart
davidmorgan: Done.

build_runner_core/test/environment/create_merged_dir_test.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks :)

PR #4096, opened 2025-07-12
Read builder config from AssetReader.

davidmorgan:
Use `AssetReader` when reading `build.yaml`.

Remove old test support: just write `build.yaml` to the test filesystem.

Inline one test support method now only used in one place.


build_runner/test/build_script_generate/builder_ordering_test.dart
```diff
@@ -222,3 +221,10 @@ void main() {
     });
   });
 }
+
+Map<String, BuildConfig> parseBuildConfigs(
+  Map<String, Map<String, dynamic>> configs,
+) => Map<String, BuildConfig>.fromIterable(
+  configs.keys,
+  value: (key) => BuildConfig.fromMap(key as String, [], configs[key]!),
+);
```
lrhn: Or don't use `fromIterable` (ever!):
```dart
 => {for (var key in config.keys) key: BuildConfig.fromMap(key, [], configs[key]!)};
```

No `as String` needed, that's just because `fromIterable` sucks at typing. Among other things.
(The only `.from` method you should use today is *possibly* `.fromIterables`. It has adequate typing, no casting to dynamic, and parallel iteration isn't that easy in a map literal. For everything else, use literals.)

build_runner_core/lib/src/package_graph/target_graph.dart
```diff
@@ -219,25 +224,38 @@ class TargetNode {
   String toString() => target.key;
 }
 
-Future<BuildConfig> _packageBuildConfig(PackageNode package) async {
-  final dependencyNames = package.dependencies.map((n) => n.name);
+Future<BuildConfig> _packageBuildConfig(
+  AssetReader? reader,
+  PackageNode package,
+) async {
+  final dependencies = package.dependencies.map((n) => n.name).toList();
```
lrhn: (I just like literals.
```dart
   final dependencies = [for (var n in package.dependencies) n.name];
```
I think they're easier to read, and I know up-front that it's building a list,
I don't have to reach the end before I see `.toList()`.)

build_runner/test/build_script_generate/builder_ordering_test.dart
davidmorgan: Done.

build_runner_core/lib/src/package_graph/target_graph.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks :)

PR #4102, opened 2025-07-18
Support testing builder factories.

davidmorgan:
Improvements to `build_test` so I can test #4084, and a few bug fixes for minor issues discovered along the way.

I plan to replace most use of the private `testPhases` with the methods from `build_test`, improving them as needed. There are a lot of tests to update, though, so in this PR just update 1-2 tests to cover the `build_test` fixes and new functionality.


davidmorgan:
Landed as https://github.com/dart-lang/build/pull/4187

PR #4118, opened 2025-08-06
Allow analyzer 8.

davidmorgan:
And prepare to release.


jensjoha (approved):
If it actually works, LGTM.

davidmorgan:
Thanks!

I thought this PR was going to version solve to analyzer 8, but it doesn't because of dev dependencies on builders that will need dart_style+source_gen to allow version 8.

So, tested with a version override here https://github.com/dart-lang/build/pull/4120 and pulled in a few fixes and `ignore` comments from there to this PR.

The e2e Windows failures are annoying but I think are unrelated to this PR, they've been a problem for a while ... I've kicked off the process of getting a Windows machine so I can look into it.

Issue #4127, opened 2025-08-11
Exceptions are hard to spot in the logs of build_runner


rrousselGit:
Hello!

Let's play a came of "spot the error message":

```
dart run build_runner watch -d
Resolving dependencies in `/Users/remirousselet/dev/rrousselGit/riverpod`... 
Downloading packages... 
Got dependencies in `/Users/remirousselet/dev/rrousselGit/riverpod`.
Resolving dependencies in `/Users/remirousselet/dev/rrousselGit/riverpod`... 
Downloading packages... 
Got dependencies in `/Users/remirousselet/dev/rrousselGit/riverpod`.
4s riverpod_generator on 17 inputs: 14 output, 3 no-op; spent 3s sdk,             
   1s analyzing                                                                   
0s source_gen:combining_builder on 46 inputs: 14 same, 32 no-op                   
4s mockito:mockBuilder on 68 inputs: 20 skipped, 48 no-op; spent 4s analyzing     
1s freezed on 160 inputs: 100 skipped, 1 output, 59 no-op                         
                                                                                  
Failed to build with build_runner in 11s with warnings; wrote 29 outputs.         
                                                                                  
log output for riverpod_generator on test/integration/generated.dart              
E InvalidTypeException: The type is invalid and cannot be converted to code.      
log output for build_runner                                                       
W Root package config not readable, manual restarts will be needed after running  
  `pub upgrade`.                                                                  
  Deleting 15 declared outputs which already existed on disk.                     
```

In case you missed it, that the
> E InvalidTypeException: The type is invalid and cannot be converted to code.      

It's pretty hard to spot one line among many others.

Some ideas to improve this are:
- Use a different color to log user errors
- Format them with a header corresponding to the generator name. For example:
   `[my_generator]: InvalidTypeException: The type is invalid and cannot be converted to code.`
- List errors as the very last line of the logs, with a blank line separating errors from other logs

davidmorgan:
Thanks for filing, happy to tweak this a bit.

Generally I don't like build warnings, one way to make errors more visible is to not have any warnings :) ... but that's more of a long term aspirational goal than a fix.

We can certainly add color, everything is in place to use ANSI color codes, I just thought I'd keep it as simple as possible to start with. The `E` could be red ... not sure how visible that would be, making the whole error red is probably too much.

Output is per generator and then one output for build_runner, an option would be to move all blocks with errors to the end:

```
log output for build_runner                                                       
[...]

PR #4147, opened 2025-08-15
Tweak warnings errors

davidmorgan:
For #4127 

Changes to make failures more prominent/actionable:

 - move blocks of log output that do _not_ contain errors before the status line; only errors go after the status line
 - remove "log output for" and "on" from the block header, move the filename first so it's "<filename> <builder>", and make the filename bold red if the block contains any error
 - make the "E" prefix that indicates an error bold red

And, remove the "deleting declared outputs" warning, it's not actionable.


davidmorgan:
Will land as https://github.com/dart-lang/build/pull/4190

davidmorgan:
Landing as https://github.com/dart-lang/build/pull/4190

Issue #4154, opened 2025-08-18
Build Runner (2.7.0) sometimes fails to watch Files


HeftyCoder:
**EDIT**: This issue generally seems to appear after building the application at least once (either in debug or release).

Upgraded Flutter from 3.32.8 to 3.35.1 and build_runner from 2.5.x to 2.7.0. Watching files with `dart run build_runner watch` failed multiple times:

```
Failed to watch files in package:<package_name>.
PathNotFoundException: Directory listing failed, path = <relative_path>my_app\build\flutter_plugin_android_lifecycle\.transforms\8afebd18f8d6cab29a4e7940b859a844\transformed\bundleLibRuntimeToDirDebug\bundleLibRuntimeToDirDebug_dex\io\flutter\embedding\engine\plugins\lifecycle\*' (OS      

  , errno = 3)
```

I tried `flutter clean`, and then re-running the build_runner. It then returned the below error once, before returning the same error as above after retrying.

```
Failed to watch files in package:my_app.
FileSystemException: Directory watcher closed unexpectedly, path = '<relative_path>_my_app
```
I ran `flutter clean` for a second time, restarted the PC and attempted to watch again. Got the second error once more. Finally, after re-running the watcher, things seem to work fine until we build the app.

On a side-note, while this behavior persisted, I also noticed that running `dart run build_runner build` would not output the results directly to the files. I'm not entirely sure what it was, but I either had to wait some time or re-ran the command until the changes were visible.

----------------------------------------
The attempted to watch directory is the main app directory, which contains `pubspec.yaml`, `build.yaml` etc.
 
* The Dart SDK Version: 3.9.0 (stable)
* Flutter Version: Flutter 3.35.1
* Development Platform: Windows
* Build Target: Android
* Excluded the dart analysis server as well as the project from Windows Defender.
* Packages include `drift_builder`, `json_serializable`, `freezed`, `go_router_builder`, and `riverpod_generator`.

build.yaml, for completeness:
```
targets:
  $default:
    builders:
      drift_dev:
        options:
          store_date_time_values_as_text: true
          apply_converters_on_variables: true
          named_parameters: true
          write_to_columns_mixins: true
```

davidmorgan:
Thanks for filing.

My guess is it's due to Flutter writing temporary files to `build` at the same time as `build_runner` is trying to watch it. I'll dig a little and see if I can figure something out.

reidbaker:
[...]

PR #4160, opened 2025-08-25
Recover from watcher restart.

davidmorgan:
For #4154, WIP


davidmorgan:
No longer needed with watcher 1.2.0.

PR #4163, opened 2025-08-26
Remove package:test version check.

davidmorgan:
The version check is no longer needed, anything newer than 7 years is new enough :)


jensjoha (approved):
I don't understand how the title and the content match, but other than that I'm fine with it.

davidmorgan:
Thanks!

> I don't understand how the title and the content match, but other than that I'm fine with it.

It was reading the pubspec of the package you launch `build_runner` in to check the version of the `test` dependency, the version check is no longer needed, so, removed. (Mostly since it requires reading the pubspec earlier than I'd like).

PR #4165, opened 2025-08-28
Refactor build_runner command line

davidmorgan:
Refactor to cleanly separate

 - the command line
 - the parsed command line
 - configuration deduced from command line + files
 - configuration that is only used for testing

And, use immutable collections for options/configuration.
 
There are almost no functional changes here; just one: I tweaked the "help" messages for the commands, I couldn't resist :) ... the parsing / command implementations are mostly just moved around unchanged.

There is some obvious next refactoring to do to use the new bundled of options wherever it makes sense, but this seemed a good place to finish for now. Added some TODOs.

All this is to help with two things:

 - when the file watcher loses events, we want to go back to just after the command line was parsed and redo everything that depends on files; so there needs to be a clear separation at that point
 - rewriting the bootstrap so we can AOT compile similarly benefits from everything being clearly separated


build_runner/lib/src/build_runner.dart
```diff
@@ -0,0 +1,201 @@
...
+  /// The nested `build_runner` invocation reaches [run] with [builders] set,
+  /// so it runs the command instead of bootstrapping.
+  Future<int> _runWithBuilders() async {
+    buildLog.configuration = buildLog.configuration.rebuild((b) {
+      b.mode = commandLine.type.buildLogMode;
+    });
+
+    // Build and run, retrying if the nested `build_runner` invocation exits
+    // with exit code `tempFail`.
+    while (true) {
```
jensjoha, line 169: I don't have an overview over when "tempFail" is returned --- but should there be a limit to the number of times this is retried?

build_runner/lib/src/build_runner_command_line.dart
```diff
@@ -0,0 +1,435 @@
...
+  }
+
+  /// The number of arguments in [rest] that are from before any `--` on the
+  /// command line.
+  int get restArgumentsBeforeSeparator {
+    final restArgumentsAfterSeparator =
+        arguments.contains('--')
+            ? arguments.skipWhile((a) => a != '--').skip(1).length
+            : 0;
+    return rest.length - restArgumentsAfterSeparator;
```
jensjoha:
```
  final index = arguments.indexOf('--') + 1;
  final numArgumentsAfter = (index > 0) ? arguments.length - index : 0;
  return rest.length - numArgumentsAfter;
```

seems more readable to me.

jensjoha (approved):

build_runner/lib/src/build_runner.dart
davidmorgan: This matches current behaviour

https://github.com/dart-lang/build/blob/c63229d95b7d63470a4464a32810dd7f715fd2ad/build_runner/bin/build_runner.dart#L97

--I'll have a new and hopefully better version soon as I refactor the bootstrap code for AOT compile.

build_runner/lib/src/build_runner_command_line.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks :)

PR #4166, opened 2025-08-29
Refactor build state

davidmorgan:
Refactor `BulidEnvironment` and `BuildConfiguration` into one `BuildPlan` that computes most state that can be computed before the build starts.

The one big thing _not_ pulled in (yet) is `AssetGraph`.

Moves without changes: move `BuildFilter` next to the new `BuildPlan`, move `_finalizeBuild` and `_convertToFailure` into `build.dart`, moved constants related to `TargetGraph` into `target_graph.dart`, move `findBuildConfigOverrides` into `target_graph.dart`.
Cleanup: remove unused `expectedGeneratedDir`, `requiredSourcePaths`, `requiredRootSourcePaths`, `TestBuildEnvironment`.


build_runner/lib/src/daemon/daemon_builder.dart
```diff
@@ -229,88 +216,60 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
+            // Assume we will create an outputDir.
+            true,
+            expectedDeletes,
+            buildPlan.reader,
+          ),
+        )
+        .map((data) => WatchEvent(data.type, '${data.id}'))
+        .debounceBuffer(
+          buildPlan.testingOverrides.debounceDelay ??
+              const Duration(milliseconds: 250),
```
jensjoha, line 259: As I recall there as also a hardcoded 250 ms something something in the previous PR --- and two (I think) in this one. Maybe we should have a constant field somewhere instead.
Doesn't have to be this PR though.

jensjoha (approved):

build_runner/lib/src/daemon/daemon_builder.dart
davidmorgan: Thanks. Yes, there's one in `watch`, one in `daemon`; I left them separate because there isn't really a strong reason for them to be the same, they are both about debouncing file events but the best latency might be different as they are quite different use cases.

PR #4190, opened 2025-09-05
Better warnings and errors logging.

davidmorgan:
For https://github.com/dart-lang/build/issues/4127

Changes to make failures more prominent/actionable:

- move blocks of log output that do not contain errors before the status line; only errors go after the status line
- remove "log output for" and "on" from the block header, move the filename first so it's " ", and make the filename bold red if the block contains any error
- make the "E" prefix that indicates an error bold red

And, remove the "deleting declared outputs" warning, it's not actionable.

### DIFF
--- a/build_runner/lib/src/logging/build_log_messages.dart
+++ b/build_runner/lib/src/logging/build_log_messages.dart
@@ -100,12 +100,13 @@ class BuildLogMessages {
 
     for (final message in messages) {
       var first = true;
+      final isError = message.severity == Severity.error;
       for (final line in message.text.split('\n')) {
         result.add(
           AnsiBufferLine([
-            if (message.severity == Severity.error) AnsiBuffer.boldRed,
+            if (isError) AnsiBuffer.boldRed,
             first ? message.severity.prefix : '  ',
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/logging/build_log_messages.dart
```diff
@@ -72,40 +72,50 @@ class BuildLogMessages {
...
       ]),
     );
 
-    for (final message in _messageByCategory[category]!) {
+    for (final message in messages) {
       var first = true;
       for (final line in message.text.split('\n')) {
         result.add(
           AnsiBufferLine([
+            if (message.severity == Severity.error) AnsiBuffer.boldRed,
```
jensjoha: (maybe mostly because of the github interface, I'm not sure): This would read better if `message.severity == Severity.error` was up with `var first`, e.g. `final isError = message.severity == Severity.error;` and then here `if (isError) AnsiBuffer.boldRead`.

davidmorgan: Works for me. Thanks :)

Issue #2730, opened 2020-06-16
Discontinue the `timing` package and implement in `build_runner_core` instead


natebosch:
Trying to prune packages published by the Dart team.

ricardodalarme:
hey @davidmorgan I suppose we can close this one

davidmorgan:
Indeed. Thanks!

PR #4192, opened 2025-09-05
Fold `timing` into `build_runner`.

davidmorgan:
As suggested in https://github.com/dart-lang/build/issues/2730 the `timing` package isn't used outside `build_runner`, inline it so it can be discontinued.


build_runner/lib/src/generate/timing.dart
```diff
@@ -0,0 +1,391 @@
+// Copyright (c) 2018, the Dart project authors.  Please see the AUTHORS file
```
jensjoha, line 1: How much of this is actually used?

davidmorgan: All of it, I trimmed a few small bits that weren't used.

PR #4201, opened 2025-09-08
Separate asset graph loading

davidmorgan:
Separate loading the previous asset graph from creating the new one.

- separate deciding what to do (restart, delete) from doing it; stop using exceptions for the restart
- move some test cases from `build_definition_test` into new `build_plan_test`, turn some into smaller tests in new `BuildPhases` digest test

Restarts triggered later are still using exceptions, the plan is to do a similar refactoring of the new asset graph creating in `build_definition.dart`.


build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -26,6 +33,23 @@ class BuildPlan {
...
+  /// or before doing another build.
+  ///
+  /// It includes generated outputs of the previous build, which might not be
+  /// findable later because they might not be outputs of the current build due
+  /// to changes to configuration.
+  ///
+  /// If the old asset graph needs deleting prior to a restart it's also
+  /// included here.
+  ///
+  /// Call[deletePreviousBuildOutputs] to delete them.
```
jensjoha: missing space

jensjoha (approved):

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Done, thanks.

PR #4203, opened 2025-09-09
Refactor asset graph creation into `BuildPlan`

davidmorgan:
Separate reading state from updating state; in particular, accumulate a list of files+folders to delete rather than deleting immediately.

Stop using exceptions for restart control flow.

Tests that were in `build_definition_test.dart`: some become equivalent tests in `build_plan_test.dart`, some are actually `TargetGraph` tests so they move to `target_graph_test.dart`, some become smaller tests in `build_phases_test.dart`, some were redundant with other tests and have been removed.

### DIFF
--- a/build_runner/lib/src/build_plan/build_plan.dart
+++ b/build_runner/lib/src/build_plan/build_plan.dart
@@ -161,7 +161,7 @@ class BuildPlan {
               Platform.version,
             )) {
           buildLog.fullBuildBecause(FullBuildReason.incompatibleBuild);
-          // Delete old outputs.
+          // Mark old outputs for deletion.
           filesToDelete.addAll(
             previousAssetGraph.outputsToDelete(packageGraph),
           );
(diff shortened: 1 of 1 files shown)


build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -107,50 +128,165 @@ class BuildPlan {
...
         if (buildPhasesChanged ||
             pkgVersionsChanged ||
             enabledExperimentsChanged ||
-            !isSameSdkVersion(assetGraph.dartVersion, Platform.version)) {
+            !isSameSdkVersion(
+              previousAssetGraph.dartVersion,
+              Platform.version,
+            )) {
           buildLog.fullBuildBecause(FullBuildReason.incompatibleBuild);
+          // Delete old outputs.
```
jensjoha: Here and below --- wouldn't "mark for deleting" make more sense?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -107,50 +128,165 @@ class BuildPlan {
...
+        disabled: buildOptions.skipBuildScriptCheck,
+      );
+
+      final buildScriptUpdated =
+          !buildOptions.skipBuildScriptCheck &&
+          buildScriptUpdates.hasBeenUpdated(updates.keys.toSet());
+      if (buildScriptUpdated) {
+        buildLog.fullBuildBecause(FullBuildReason.incompatibleScript);
+        // Delete old outputs.
+        final deletedSourceOutputs = previousAssetGraph.outputsToDelete(
```
jensjoha: this variable name is also misleading.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Yes, for sure :) thanks, done.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Thanks--actually got rid of the variable since all of `filesToDelete` is removed from `inputSources` later.

PR #4211, opened 2025-09-15
Improve `package:build` API docs.

davidmorgan:
I probably want to deprecate `log` in favour of methods on `BuildStep`, but for now, document what it does :)

### DIFF
--- a/build/lib/src/logging.dart
+++ b/build/lib/src/logging.dart
@@ -10,7 +10,24 @@ const Symbol logKey = #buildLog;
 
 final _default = Logger('build.fallback');
 
-/// The log instance for the currently running BuildStep.
+/// `Logger` for use by a builder to log to the main `build_runner` output.
 ///
-/// Will be `null` when not running within a build.
+/// `build_runner` distinguishes three categories of log record:
+///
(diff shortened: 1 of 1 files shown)


build/lib/src/logging.dart
```diff
@@ -10,7 +10,24 @@ const Symbol logKey = #buildLog;
...
+/// `build_runner` distinguishes three categories of log record:
+///
+/// Below [Level.WARNING] is called "info".
+///
+/// Info is only shown if `--verbose` was explicitly requested on the command
+/// line.
+///
+/// At [Level.WARNING] but below [Level.SEVERE] is called a "warning".
+///
+/// Warnings are aways shown, and the final build status will indicate that
```
jensjoha, line 24: always?

jensjoha (approved):

build/lib/src/logging.dart
davidmorgan: Yes, you have a choice about showing infos but warnings are always shown.

jensjoha: That was not my point. My point was that it should say "always" not "aways".

PR #4217, opened 2025-09-17
Rewrite bootstrap not using `dart:mirrors`

davidmorgan:


davidmorgan:
Landed as https://github.com/dart-lang/build/pull/4217

PR #4226, opened 2025-09-23
New bootstrap code not using `dart:mirrors`

davidmorgan:
Also prepare for adding AOT compilation in parallel.

 - Stop using `frontend_server_client`, just run `dart compile`.
 - Stop running as an isolate, run as a process. Various changes to support running as a process:
   - Use stdin+stdout for messages between parent and child.
   - Tell the child about the parent console capabilities since they can't be queried directly.
   - Manage child process lifecycle: kill the child if the parent is killed.
 - Track compiled deps using the compiler-output depfile instead of mirrors; and, use the depfile instead of storing in the asset graph as "internal" nodes.
 - Move handling of changes from `watcher.dart` to `build_series.dart`.
 - Remove the `skipBuildScriptCheck` option and command line flag; the depfiles checks are always done from the child process, never done if not running as a child process.
  - Fix `PackageGraph.pathFor` returning paths with forward slashes on Windows, they should be backslashes so they match depfiles.


build_runner/lib/src/bootstrap/bootstrapper.dart
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /// Reads build configuration, writes the build script.
+  ///
+  /// Reads before write so the file is not written if there is no change.
+  Future<void> _writeBuildScript() async {
+    final buildScript = await generateBuildScript();
+    final path = '.dart_tool/build/entrypoint/build.dart';
+    final existingBuildScript =
+        File(path).existsSync() ? File(path).readAsStringSync() : '';
```
jensjoha: Why not just have it nullable?

build_runner/lib/src/bootstrap/bootstrapper.dart
```diff
@@ -0,0 +1,107 @@
...
+    if (buildScript != existingBuildScript) {
+      File(path)
+        ..createSync(recursive: true)
+        ..writeAsStringSync(buildScript);
+    }
+  }
+
+  /// Checks freshness of the entrypoint script compiled to kernel.
+  Future<FreshnessResult> checkKernelFreshness() async {
+    if (!ChildProcess.isRunning) return FreshnessResult(outputIsFresh: true);
```
jensjoha: Does this make sense? If nothing else a at least give a comment about why this is correct.

build_runner/lib/src/bootstrap/bootstrapper.dart
```diff
@@ -0,0 +1,107 @@
...
+    await _writeBuildScript();
+    return _compiler.checkFreshness();
+  }
+
+  /// Whether [path] is a dependency of the entrypoint script compiled to
+  /// kernel.
+  bool isKernelDependency(String path) {
+    if (!ChildProcess.isRunning) return false;
```
jensjoha: Similar here.

build_runner/lib/src/bootstrap/bootstrapper.dart
```diff
@@ -0,0 +1,107 @@
...
+          buildLog.error(
+            'Failed to compile build script. '
+            'Check builder definitions and generated script $scriptLocation.',
+          );
+          throw const CannotBuildException();
+        }
+      }
+
+      final result = await ParentProcess.runAndSend(
+        script: '.dart_tool/build/entrypoint/build.dart.dill',
```
jensjoha: This should probably be a constant somewhere.

build_runner/lib/src/bootstrap/kernel_compiler.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  /// Compiles the entrypoint script to kernel.
+  Future<CompileResult> compile({Iterable<String>? experiments}) async {
+    final dart = Platform.resolvedExecutable;
+    final result = await ParentProcess.run(dart, [
+      'compile',
+      'kernel',
+      '.dart_tool/build/entrypoint/build.dart',
+      '--output',
+      '.dart_tool/build/entrypoint/build.dart.dill',
```
jensjoha: (here the non-constant is too)

build_runner/lib/src/bootstrap/kernel_compiler.dart
```diff
@@ -0,0 +1,79 @@
...
+      if (experiments != null)
+        for (final experiment in experiments) '--enable-experiment=$experiment',
+    ]);
+
+    if (result.exitCode == 0) {
+      final stdout = result.stdout as String;
+
+      // Convert "unknown experiment" warnings to errors.
+      if (stdout.contains('Unknown experiment:')) {
+        if (File('.dart_tool/build/entrypoint/build.dart.dill').existsSync()) {
```
jensjoha: (and here)

build_runner/lib/src/bootstrap/kernel_compiler.dart
```diff
@@ -0,0 +1,79 @@
...
+          File('.dart_tool/build/entrypoint/build.dart.dill').deleteSync();
```
jensjoha: (and here)

build_runner/lib/src/bootstrap/kernel_compiler.dart
```diff
@@ -0,0 +1,79 @@
...
+        final messages = stdout
+            .split('\n')
+            .where((e) => e.startsWith('Unknown experiment'))
+            .map((l) => '$l\n')
+            .join('');
+        return CompileResult(messages: messages);
+      }
+    }
+
+    if (result.exitCode == 0) {
```
jensjoha: any reason this is not just inside the if above?

build_runner/lib/src/build/build_series.dart
```diff
@@ -243,5 +276,17 @@ class BuildSeries {
...
-  Future<void> beforeExit() => _resourceManager.beforeExit();
+  /// Ends the build series.
+  ///
+  /// First, any currently-running build is waited for.
+  ///
+  /// Then: resource cleanup is done, [buildResults] is closed and no further
+  /// builds are allowed.
+  Future<void> close() async {
+    if (_closedCompleter.isCompleted) return;
+    _closedCompleter.complete();
```
jensjoha: Can't we get into a weird situation here where it is currently closing down but other parts think it is already done?

build_runner/test/bootstrap/depfile_test.dart
```diff
@@ -0,0 +1,35 @@
...
+output: input1 input2
+'''),
+        ['output', 'input1', 'input2'],
+      );
+    });
+
+    test('parses when filenames have spaces and backslashes', () async {
+      expect(
+        Depfile.parse(r'''
+output: input1 input2 input\ with\ space input4 path\\input5
```
jensjoha: Will this not (potentially at least) fail on Windows? Wouldn't a line without a new line but with an explicit (written) `\n` be better?

build_runner/lib/src/bootstrap/bootstrapper.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/bootstrapper.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/kernel_compiler.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/kernel_compiler.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/kernel_compiler.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/kernel_compiler.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/bootstrapper.dart
davidmorgan: Done.

build_runner/lib/src/bootstrap/bootstrapper.dart
davidmorgan: Done.

build_runner/test/bootstrap/depfile_test.dart
davidmorgan: I'm not sure I understand what you're asking / asking for, could you clarify please?

The depfiles do differ between Linux and Windows: on Windows the path separator is `\` and it gets escaped to `\\`.

The parsing that I wrote handles `\ ` and `\\` specially, but everything else with `\` is untouched.

build_runner/lib/src/build/build_series.dart
davidmorgan: Good point, if something waited for `closed` then called `exit(0)` that would interrupt cleanup.

Currently nothing does that, the one place it's used is in `watcher.dart` to decide to stop scheduling new builds.

So, renamed `closed` -> `closing`.

`watcher.dart` uses `onDone` on the build results stream to decide when to exit, which seems good. Moved closing the stream last in this method and added comments.

davidmorgan (commented):
Thanks! One unresolved comment re: depfiles parsing, could you take a look please?

As discussed, checked re: bash on MacOS and added a comment.

build_runner/test/bootstrap/depfile_test.dart
jensjoha: Let me start by saying that - after looking in the spec - I'm wrong.

What I thought was that this would result in a string `output: input1 input2 input\ with\ space input4 path\\input5\n` when the line endings for the file is `\n` but in the string `output: input1 input2 input\ with\ space input4 path\\input5\r\n` when the line endings for the file is `\r\n`.

Being unsure what the default git settings are, how much is repository configured, how much is user setable, what wins if there are conflicting configs etc, what I initially suggested was that instead of

```
r'''
output: input1 input2 input\ with\ space input4 path\\input5
'''
```

which - I thought - had a difference at the end on Windows line endings in the file vs not, we do something like

```
r'output: input1 input2 input\ with\ space input4 path\\input5' '\n'
```

which does not.

The spec states though:

> Line breaks in a multiline string are represented by the ⟨LINE_BREAK⟩
> production. A line break introduces a single newline character (U+000A) into
> the string value.

Where

> ⟨LINE_BREAK⟩ ::= ‘\n’
> | ‘\r\n’
> | ‘\r’

So this already just produces the `\n`.

One could still do the more explicit thing to avoid (at least my) confusion --- but it's no longer super important :)

davidmorgan: Ah, that's what I was missing :) I had never thought about line endings and multiline strings.

I'm very happy to learn that they behave as I want them to.

Agreed re: making strings explicit in a parse test, done. Thanks.

PR #4240, opened 2025-10-03
Adding incremental Frontend Server-compatible DDC builders + supporting infrastructure

Markzipan:
Glues FES into build_runner, which is our first step towards hot reload + build_runner.
The high level workflow is:
1) A persistently running frontend server is initialized once when a build starts.
2) build_runner requests JS files based on locally modified/generated dart files (as usual). Builders that collect meta-information about DDC modules also - as a side effect - record the main app entrypoint and any locally modified files.
3) When a JS file is requested, the frontend server receives recompile requests via a proxy queue (to maintain communication order with the frontend server).
4) The frontend server processes compilation requests and serves compiled JS files back to `build_runner` (hot-reload ready).

Major changes:
* Adds a `DdcFrontendServerBuilder` to our set of DDC builders (enabled via the `web-hot-reload` config). This builder keeps a `PersistentFrontendServer` instance alive across rebuilds. Compile/recompile requests are queued via a `FrontendServerProxyDriver` resource.
* Uses `scratch_space` to record both 1) the main app entrypoint and 2) updated local files from the `entrypoint_marker` builder and the `module_builder` builder respectively. These are side effects that break certain stateful 'guarantees' of standard build_runner execution. The `entrypoint_marker` builder runs before any of the downstream DDC builders and finds the web entrypoint, as Frontend Server must receive the same entrypoint on every compilation request. 
* Requires that strongly connected components in both the frontend server and `build_runner` be disabled.

Test changes:
* Extends `build_test` to permit incremental builds. This involves passing the asset graph + asset reader/writer across build results and only performing cleanup operations after a series of rebuilds. 
* `build_test` doesn't support `runs_before` and other ordering rules in `build.yaml`, so the above changes allows a kind of imperative ordering, which is important for testing `entrypoint_marker`.

Minor changes:
* Added a flag to disable strongly connected components in build_web_compilers (implemented using raw ddc meta-modules over clean ddc meta-modules + enforcing fine module aggregation).
* Added disposal logic to `scratch_space` so that rebuilds only retain modified files.
* Updated `scratch_space` `package_config.json` specs (`packageUri` and `rootUri`). The previous values didn't seem to make sense to me, but I'm also not familiar with how that's standardized in `scratch_space`.
* Added `file` and `uuid` deps to `build_modules`.
* Moved around some helper functions.
* Ported some naming functions from the DDC runtime.

Currently doesn't support live-reloading (functionality appears to have been broken a while ago). This'll be added in an upcoming change and permit `webdev`-like auto-hot-reload on save (on top of manual). 

Enable this by adding the following to a project's `build.yaml`:
```
global_options:
  build_web_compilers|sdk_js:
    options:
      web-hot-reload: true
  build_web_compilers|entrypoint:
    options:
      web-hot-reload: true
  build_web_compilers|ddc:
    options:
      web-hot-reload: true
  build_web_compilers|ddc_modules:
    options:
      web-hot-reload: true
```


davidmorgan:
Thanks!

This will take me a little work to digest :) I guess I can get to it in 1-2 days.

It definitely makes sense to expand support for testing. The details are hard :) ... in particular, I would like to avoid exposing any `build_runner` internals, including `AssetGraph` or how it's detecting file changes for follow-on builds.

So I will see if I can come up with a slightly differently suggestion.

Currently `testBuilder` uses fake build configuration and that's quite awkward; it starts to be weird when you talk about follow-on builds because there is no clear line for what should stay the same in the fake configuration and what's allowed to change.

One possibility would be to go all the way to real builds, I recently added a new type of integration test that I'm pretty happy with

https://github.com/dart-lang/build/blob/17e38a62623a602c6babd59879fea557de5d60c2/build_runner/test/integration_tests/watch_command_test.dart#L1

I'll give it some thought :)

biggs0125:
> Requires that strongly connected components in both the frontend server and build_runner be disabled.

Can you say more about the affect this has on the structure of compilations? Are we doing more work by turning off SCCs? My assumption was that SCCs allowed us to limit the scope of an invalidation.

Markzipan:
@davidmorgan Thanks! It's a huge change, so please take your time. I wasn't able to configure it in a prettier way since I'm flying out for two weeks. 

I'm highly unopinionated wrt testing, but it was definitely a struggle to get `testBuilders` to cooperate with rebuilds. Please send any followup recs for better ways to represent this test. The important bits to me are: 1) respecting builder order (as per `build.yaml`) since that matters now, 2) being able to run rebuilds vs new builds (or determine build boundaries), 3) updating files passed to the builders.

@biggs0125 The SCCs here operate on a the import-graph level. DDC's old module system requires that cyclic dependencies be 'unified' into a single module. Both the Frontend Server and build_runner doing this non-deterministically might cause `a <--> b` to be merged into `a` in build_runner and `b` in the Frontend Server, which gives us "import not found" errors.

With SCCs enabled, `a.dart` would invoke a single builder on `module(a, b)`. Disabling them, two builders, `module(a)` and `module(b)`, are invoked. The net effect of this on some apps I've tested hasn't been large, as  most modules don't end up in an SCC anyway. The largest ramifications would be a hypothetical gigantic app that wants to bundle smaller JS files. But that would ideally be specified statically, so we'd be able to pass it consistently to both `build_runner` and the Frontend Server. The particularly annoying bit about SCCs is that both systems maintain their own independent algorithm without any configurability. I alternatively could've 1) poked a hole into the Frontend Server to allow per-compile module-to-lib maps and 2) had build_runner serialize its module-to-lib maps and send that per-compile to the Frontend Server. But I figured that was too much of a mess.

Re: failures. I think some tests are failing due to API changes without pinning new versions? I'll add those (maybe in a separate PR) if the current approach looks sound.

build_runner/lib/src/constants.dart
```diff
@@ -19,6 +19,9 @@ final String assetGraphPath = '$cacheDirectoryPath/asset_graph.json';
 /// The directory to which hidden assets will be written.
 String get generatedOutputDirectory => '$cacheDirectoryPath/generated';
 
+/// Relative path to the cache directory from the root package dir.
+const cacheDir = '.dart_tool/build';
```
davidmorgan: I don't see where this is used, remove?

scratch_space/lib/src/scratch_space.dart
```diff
@@ -21,12 +21,21 @@ class ScratchSpace {
   /// Whether or not this scratch space still exists.
   bool exists = true;
 
+  /// The built app's main entrypoint file.
+  ///
+  /// This must be set before any asset builders run when compiling with DDC and
+  /// hot reload.
+  late AssetId entrypointAssetId;
```
davidmorgan: This doesn't look particularly related to `ScratchSpace`, which I see is used by a few other builders.

I guess the problem here is of passing configuration between builders?

Maybe a new `Resource` class for that? It could also go in `scratch_space`.

scratch_space/lib/src/scratch_space.dart
```diff
@@ -21,12 +21,21 @@ class ScratchSpace {
...
+
   /// The `packages` directory under the temp directory.
   final Directory packagesDir;
 
   /// The temp directory at the root of this [ScratchSpace].
   final Directory tempDir;
 
+  /// Holds all files that have been locally modified in this build.
+  final changedFilesInBuild = <AssetId>{};
```
davidmorgan: Can this be private and expose an Iterable<AssetId>? Maybe a read-only view if `Set` is needed?

To avoid any possibility of corrupted shared state if a builder thinks it's fun to write to it :)

scratch_space/lib/src/scratch_space.dart
```diff
@@ -92,6 +101,8 @@ class ScratchSpace {
   /// Copies [assetIds] to [tempDir] if they don't exist, using [reader] to
   /// read assets and mark dependencies.
   ///
+  /// Locally updated assets will be recorded in [changedFilesInBuild].
```
davidmorgan: I suggest making this and the comment for `changedFilesInBuild` more precise, e.g. here

Assets that have changed since the last time they were seen by `ensureAssets` are added to `changedFilesInBuild`.

And for changedFilesInBuild, maybe: Assets that changed between calls to `ensureAssets`. Cleared at the end of every build.

scratch_space/lib/src/scratch_space.dart
davidmorgan, line 1: I think `scratch_space` needs a new version number and then anything that needs what you're adding here should get a min dep onto that version?

davidmorgan (commented):
This generally looks good, a few comments re: scratch_space.

Re: testing :)

You sort of don't need to pass `assetGraph`, it's serialized to the `readerWriter` and the next build will read it if it's there.

_But_, the next build will discard the asset graph if any builders changed, so you have to pass the exact same builders each time, which means you will get a full build the first time you call it anyway.

You commented with `build.yaml` ordering is not respected, and that's true. Adding workarounds for that is not too pretty, it's getting very close to a real build but now with quite a lot of configuration that doesn't exactly match a real build.

Would you be up for trying a different way of testing that _is_ a real build?

I added tests recently that make it easy to set up some packages, run `build_runner` for real and check the output. Since you want persistent processes, `watch` mode would make sense, so something like:

https://github.com/dart-lang/build/blob/9f07c7ea8425ac48420111afbc1abe84fe6388f8/build_runner/test/integration_tests/watch_command_test.dart#L1

There is an example web build using the same test infra here

https://github.com/dart-lang/build/blob/master/build_runner/test/integration_tests/web_compilers_test.dart

The test infra is currently internal to `build_runner`, mainly build_runner/test/common/build_runner_tester.dart, if it looks useful let's just hack around that for now and I'll work out how best to clean it up, I guess adding to build_test is one option.

build_runner/lib/src/constants.dart
Markzipan: Done

scratch_space/lib/src/scratch_space.dart
Markzipan: Done

scratch_space/lib/src/scratch_space.dart
Markzipan: Done!

scratch_space/lib/src/scratch_space.dart
Markzipan: Done!

scratch_space/lib/src/scratch_space.dart
Markzipan: Done!

Markzipan:
@davidmorgan Wow, the tests in `build/build_runner/test/integration_tests` are perfect for this. I've replaced the `build_web_compilers` FES driver test with additional `web_compilers_test.dart` tests, added a separate resource for FES-specific state, and bumped scratch_space's ver. 

Thanks for the review - back from vacation!

build_test/lib/src/internal_test_reader_writer.dart
```diff
@@ -274,4 +274,10 @@ class _ReaderWriterTestingImpl implements ReaderWriterTesting {
   void delete(AssetId id) => _readerWriter.filesystem.deleteSync(
     _readerWriter.assetPathProvider.pathFor(id),
   );
+
+  @override
+  void reset() {
```
davidmorgan: I think you don't need this any more?

build_test/lib/src/test_reader_writer.dart
```diff
@@ -92,4 +92,7 @@ abstract interface class ReaderWriterTesting {
 
   /// Deletes [id] from the [TestReaderWriter] in-memory filesystem.
   void delete(AssetId id);
+
+  /// Resets state in this [TestReaderWriter] between rebuilds.
```
davidmorgan: I think you don't need this any more?

build_runner/test/integration_tests/web_compilers_test.dart
```diff
@@ -148,5 +151,324 @@ void main() {
           '--define=build_web_compilers:dart_source_cleanup=enabled=true',
     );
     expect(tester.read('root_pkg/build/unused.dart'), null);
-  });
+  }, timeout: defaultTimeout);
+
+  test('DDC compiled with the Frontend Server', () async {
```
davidmorgan, line 159: Could you please add a TODO for me?

```
// TODO(davidmorgan): the remaining tests are integration tests of
// the web compilers themselves, support testing like this outside the
// `build_runner` package.
```

build_runner/test/common/build_runner_tester.dart
```diff
@@ -239,32 +239,63 @@ class BuildRunnerProcess {
   /// defaults to [BuildLog.failurePattern] so that `expect` will stop if the
   /// process reports a build failure.
   ///
+  /// if [expectFailure] is set, then both [pattern] and [failOn] must be
```
davidmorgan, line 242: Hmmmm multiple expectations is a bit awkward, I wonder if there's a nicer way.

Would it work to accumulate lines and return them, then the second expectation can be checked on the result?

Looking at how this method is used, I think maybe something like this:

 - most callers are not using the result, so how about the base method is `Future<void> expect`.
 - a few callers are using the single matching line, so how about `Future<String> expectAndGetLine` for that
 - and then I think you can do what you want with the full accumulated output until the match, how about `Future<List<String>> expectAndGetBlock`

?

build_modules/lib/src/frontend_server_driver.dart
```diff
@@ -0,0 +1,603 @@
...
+      );
+      metadata['$fileName.metadata'] = metadataView;
+    }
+    return updatedFiles;
+  }
+}
+
+/// Bundles information associated with a DDC library.
+class LibraryInfo {
+  final String moduleName;
```
nshahan: What is the `moduleName` in the new module system?

build_modules/lib/src/ddc_names.dart
```diff
@@ -1,14 +1,25 @@
...
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
 // for details. All rights reserved. Use of this source code is governed by a
 // BSD-style license that can be found in the LICENSE file.
 
 import 'package:path/path.dart' as p;
 
+/// Logic in this file must be synchronized with their namesakes in DDC at:
+/// pkg/dev_compiler/lib/src/compiler/js_names.dart
+
+bool isSdkInternalRuntimeUri(Uri importUri) {
```
nshahan: OK to make private?

build_modules/lib/src/ddc_names.dart
```diff
@@ -1,14 +1,25 @@
...
+  return importUri.isScheme('dart') && importUri.path == '_runtime';
+}
+
+String libraryUriToJsIdentifier(Uri importUri) {
```
nshahan: What is the `JsIdentifier` used for? I see it being used later in this change for `moduleName` but I'm getting confused from there.

build_modules/lib/src/frontend_server_driver.dart
```diff
@@ -0,0 +1,603 @@
...
+  Future<CompilerOutput> recompileAndRecord(
+    String entrypoint,
+    List<Uri> invalidatedFiles,
+    Iterable<String> filesToWrite,
+  ) async {
+    final compilerOutput = await recompile(entrypoint, invalidatedFiles);
+    if (compilerOutput == null) {
+      throw Exception('Frontend Server failed to recompile $entrypoint');
+    }
+    if (compilerOutput.errorCount != 0 || compilerOutput.errorMessage != null) {
```
nshahan: Do the reload rejection errors flow here and do these errors get handled somewhere else? We want to make sure the user can recover from them, make some edits and reload again right?

build_modules/lib/src/frontend_server_driver.dart
```diff
@@ -0,0 +1,603 @@
+// Copyright (c) 2025, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// ignore_for_file: constant_identifier_names
+
```
nshahan, line 6: At some point we need to unify all of the implementations of this type of library right? In creating this, have you seen any common parts we can factor out and reuse?

build_web_compilers/lib/src/common.dart
```diff
@@ -85,3 +102,67 @@ Future<void> fixAndCopySourceMap(
...
+        if (fileUri != null) {
+          libraryJson['fileUri'] = updatePath(fileUri);
+        }
+      }
+    }
+  }
+}
+
+/// The module name according to ddc for [jsId] which represents the real js
+/// module file.
```
nshahan: I can't quite make sense of what the module name according to DDC means. Is this the path to the Javascript file?

build_web_compilers/lib/src/common.dart
```diff
@@ -85,3 +102,67 @@ Future<void> fixAndCopySourceMap(
...
+String ddcModuleName(AssetId jsId) {
+  final jsPath =
+      jsId.path.startsWith('lib/')
+          ? jsId.path.replaceFirst('lib/', 'packages/${jsId.package}/')
+          : jsId.path;
+  return jsPath.substring(0, jsPath.length - jsModuleExtension.length);
+}
+
+String ddcLibraryId(AssetId jsId) {
```
nshahan, line 151: And this is the Dart library import uri? I'm nervous about recreating this from whatever the jsId is. Where does that come from?

davidmorgan:
It looks like you made some changes in response to my comments but I don't see any responses to the comments so I'm not sure if you're done--please let me know if/when I should take another look. Thanks!

davidmorgan (commented):
Welcome back!

Looking good so far.

There are some CI failures, could you take a look at those please?

build_test/lib/src/internal_test_reader_writer.dart
davidmorgan: (I think you already removed it, so I guess just close...)

davidmorgan:
> It looks like you made some changes in response to my comments but I don't see any responses to the comments so I'm not sure if you're done--please let me know if/when I should take another look. Thanks!

Whoops, looks like I had unpublished comments that you addressed by coincidence. Sorry about that! Now published.

build_modules/lib/src/ddc_names.dart
Markzipan: We definitely could (and probably should), but I copied this from the SDK here https://github.com/dart-lang/sdk/blob/de6b38c2fe07a8f4855951d8da641173da057286/pkg/dev_compiler/lib/src/compiler/module_builder.dart#L630

I'm slightly in favor of leaving it for consistency (since this file needs to be sync'd with its sdk namesake).

build_modules/lib/src/frontend_server_driver.dart
Markzipan: Reviewing this again, turns out I don't need these structures any more - deleted

build_modules/lib/src/ddc_names.dart
Markzipan: Ditto, this is unused now - removed

build_modules/lib/src/frontend_server_driver.dart
Markzipan: For sure - thankfully the ones in the sdk are relatively same-y (track modified libs/srcmaps/etc). Our hot reload tests need to accumulate/record input state (unlike this one). The flutter web implementation is a bit hairier though and has more custom logic.

build_modules/lib/src/frontend_server_driver.dart
Markzipan: They're handled in the wrapper that calls this function. This throw doesn't shut down the persistent FES process, and we automatically issue the `reject` signal. I've updated this pathway to be more clear.

build_web_compilers/lib/src/common.dart
Markzipan: This was copied from https://github.com/dart-lang/build/tree/f5ea13a17f91705294dd187e24ce223fe1de1709/build_web_compilers/lib/src
/dev_compiler_builder.dart to share logic, but this seems to be the key used in our DDC bootstrapper corresponding to that library. E.g., `/lib/fully/resolved/path/libraryNameNoExtension`.

build_web_compilers/lib/src/common.dart
Markzipan: Ditto - this is an existing bit of logic that I'm sharing. I think it's resolving an id to an asset name that can be looked up in the asset reader/resolver.

build_test/lib/src/internal_test_reader_writer.dart
Markzipan: Yep - done!

build_test/lib/src/test_reader_writer.dart
Markzipan: ditto

build_runner/test/integration_tests/web_compilers_test.dart
Markzipan: done!

build_runner/test/common/build_runner_tester.dart
Markzipan: I like that change, but I the newly added tests' reqs are:
1) we should be able to expect that a test throws 
2) we should be able to match on more complex patterns (more than one line of match)

The problem with returning a block is that `expect` waits on a stream until it times out for 30s or encounters `pattern`, so we don't have a good way to determine when a block is finished accumulating. Really it should just accept a `List<Matcher>` and return if all matchers are successful. Thoughts?

Markzipan:
> It looks like you made some changes in response to my comments but I don't see any responses to the comments so I'm not sure if you're done--please let me know if/when I should take another look. Thanks!

@davidmorgan Oh no - you didn't see my comments? I can ping you when things are fully ready, but I've been in and out since I'm sick with something mysterious lol

@nshahan Thanks for reviewing! I updated some of the formatting based on your feedback.

It looks like I've encountered a windows pathing issue in some SDK lib. I'll investigate tomorrow :(

build_runner/test/integration_tests/daemon_command_test.dart
```diff
@@ -143,6 +146,7 @@ void main() {
     results = StreamQueue(client.buildResults);
     expect((await results.next).results.single.status, BuildStatus.started);
     expect((await results.next).results.single.status, BuildStatus.succeeded);
+    print('bb');
```
davidmorgan: Looks like some debug prints got into the PR :)

davidmorgan:
> @davidmorgan Oh no - you didn't see my comments? I can ping you when things are fully ready, but I've been in and out since I'm sick with something mysterious lol

I think we're back in sync now, I'll keep commenting on changes/comments, and take a final look when we think it's done :) thanks.

build_runner/test/common/build_runner_tester.dart
davidmorgan: When you say "test throws" do you mean "logs an error"? It's supposed to be that no exceptions escape :)

build_runner should always run the build to completion then either print "build succeeded" or "build failed", so it should always work to use one of those or a RegExp that matches either to capture a block.

If this doesn't work could you please point to a specific test so I can understand? Thanks.

build_runner/test/integration_tests/daemon_command_test.dart
Markzipan: Woops - undone

build_runner/test/common/build_runner_tester.dart
Markzipan: Yep, logs an error (i.e., "build failed"). Hmm, watcher tests appear to stream the results of a build until some string or Regexp is matched - line by line. I'd like to both match specific text of a line and match the "build failed" text (which is output on a later line), which isn't supported by the current testing scheme (unless I've missed something). The `'can recompile incrementally after invalid edits'` test is an example of this. `await watch.expect(BuildLog.successPattern);`, which seems to be the canonical failure detecting check, doesn't let me check the error text.

davidmorgan: So currently you have this

```
    await watch.expect(
      'Hot reload rejected due to unsupported changes',
      expectFailure: true,
    );
```

I think you could add a new method `expectAndGetBlock`, that was the name given above, but maybe `linesUntil` would be better:

```
  final block = await watch.linesUntil(BuildLog.failurePattern);
  expect(block, contains(contains('Hot reload rejected due to unsupported changes')));
```

?

Per my original comment I think "expect" could be three versions: one that returns nothing, one that returns the matched line, and one that returns all the lines. Or maybe drop the single line version, you can always just do ".last" on the block?

davidmorgan (approved):
LGTM, feel free to address the e2e test suggestion before landing or not, it's easy enough to improve later if we want.

build_modules/lib/src/frontend_server_driver.dart
```diff
@@ -0,0 +1,551 @@
...
+      Directory.fromUri(outputDirectoryUri).existsSync(),
+      '$outputDirectoryUri does not exist.',
+    );
+    final filesToWrite = {...files, ...sourcemaps, ...metadata};
+    _writeToDisk(outputDirectoryUri, filesToWrite);
+  }
+
+  /// Writes [fileName] and its associated sourcemap and metadata files to
+  /// [outputDirectoryUri].
+  void writeFileToDisk(Uri outputDirectoryUri, String fileName) {
```
nshahan, line 343: What is `fileName` here? If there are associated sourcemaps it sounds like a JavaScript file, but it feels strange that the JavaScript file would have a `package:` scheme. 

Should `String fileName` be a `Uri` as well or at least documented as accepting some Uris? What do you expect besides a "package:" or "multiRootScheme:" Uri?

build_modules/lib/src/frontend_server_driver.dart
```diff
@@ -0,0 +1,551 @@
...
+      metadataFile: metadata[metadataFile]!,
+    };
+
+    _writeToDisk(outputDirectoryUri, filesToWrite);
+  }
+
+  /// Writes [rawFilesToWrite] to [outputDirectoryUri].
+  void _writeToDisk(
+    Uri outputDirectoryUri,
+    Map<String, Uint8List> rawFilesToWrite,
```
nshahan, line 373: How does the `String` key describe the file to be written? Should this be documented as a file path?

Markzipan: comment updated

build_modules/lib/src/frontend_server_driver.dart
Markzipan: Updated the comment. This is specifically a file generated by the frontend server (like `scheme://some/file/here.dart.lib.js`) that is explicitly being requested. I kept these as strings since we're directly comparing this name with string paths sent to us from the frontend server.

build_runner/test/common/build_runner_tester.dart
Markzipan: SG - lemme send out a followup change for tests that cleans these things up.

PR #4241, opened 2025-10-04
Add ability to force JS from query param instead of wasm.

biggs0125:
Adds a check for a `force_js` query param in the URL to switch between WASM and JS on-demand.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,7 +1,7 @@
-## 4.2.5
+## 4.3.0
 
-- Add support for `force_js=true` query parameter when both wasm and js builds
-  are enabled.
+- When both wasm and js builds are enabled you can now add force_js=true
+  to the URL in your browser to load the js build.
 
 ## 4.2.4
 
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 4.2.5
+
+- Add support for `force_js=true` query parameter when both wasm and js builds
+  are enabled.
+
 ## 4.2.4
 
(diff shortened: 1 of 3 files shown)


kevmoo:
Should, you know, document this and add a changelog entry. 😁

biggs0125:
@kevmoo Fineeee 😛

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.2.5
```
davidmorgan: It's a new feature, so it should be a new minor version --> `4.3.0` please.

You'll need to also update in `pubspec.yaml`, then once the PR is merged you can go ahead and release from the "publish" button that gets posted to the PR conversation. Or I can do it if you prefer.

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
...
+
+- Add support for `force_js=true` query parameter when both wasm and js builds
```
davidmorgan: I would write this the other way around, up to you

- When both wasm and js builds are enabled you can now add `force_js=true`
  to the URL in your browser to load the js build.

biggs0125: Updated wording!

build_web_compilers/CHANGELOG.md
biggs0125: Updated, thanks!

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 4.3.0
+
+- When both wasm and js builds are enabled you can now add force_js=true
```
kevmoo, line 3: Ha! we should probably have documented "set the query parameter" – since it's not just a random bit in the URL...but whatever 😄

PR #4243, opened 2025-10-04
Add `--debug-builders` flag

simolus3:
With recent changes to the bootstrapping logic, builders have become almost impossible to debug in real projects:

- Running `build_runner` itself with `--observe` doesn't do much because most of the build happens in a sub-process.
- The sub-process can't be started manually because it needs state sent to it from `stdin`, which is tricky to do when launching it from editors.

I've also started a discussion on this: https://github.com/dart-lang/build/discussions/4242. Originally I wanted to wait for feedback but this turned out to be fairly simple in the end, so here's a PR :)

This adds the `--debug-builders` flag. When set, the subprocess running the build will launch with `--observe` and `--pause-isolates-on-start`. This makes Dart print out the VM service and devtools URLs, so one can attach a debugger, set breakpoints and then resume the build.

I've tested these changes manually, I'm not sure how a good automated test for this could look like. Start a build, assert that a "VM service URL" line is printed on stdout and then kill it?

### DIFF
--- a/build_runner/test/integration_tests/processes_test.dart
+++ b/build_runner/test/integration_tests/processes_test.dart
@@ -34,6 +34,7 @@ Future<void> main() async {
   final processResult = await ParentProcess.runAndSend(
       script: 'bin/child.dart',
       arguments: [],
+      jitVmArgs: [],
       message: 'payload');
   stdout.write(
       'Parent received: '
@@ -78,6 +79,7 @@ Future<void> main() async {
   await ParentProcess.runAndSend(
       script: 'bin/child.dart',
       arguments: [],
+      jitVmArgs: [],
       message: 'payload');
 }
 ''');


davidmorgan:
Thanks!

I already felt the need for this myself :)

What do you think about instead of `--debug-builders`, having a multi-option `--dart-arg` that just passes through anything to `dart run` before the script? Then you would do `--dart-arg=--observe --dart-arg=--pause-isolates-on-start` or anything else you want. It can be wired up pretty much exactly as you already have it.

When AOT compile is added I guess the same flags won't apply to both, so maybe the name should be more explicit ... maybe `--dart-jit-vm-arg`?

For testing, it's easy to do a test that runs, checks output and kills using a recently-added test type, it would be like the "watch" test because that deals with a running process then killing it, too

https://github.com/dart-lang/build/blob/9f07c7ea8425ac48420111afbc1abe84fe6388f8/build_runner/test/integration_tests/watch_command_test.dart#L34

I suggest adding a new test named something like `build_command_vm_arg_test.dart`.

davidmorgan:
Thanks :) I guess I'll get to releasing a new version some time next week, feel free to nudge me if it starts to be too long :)

szotp:
Hi, I found a [more convenient way](https://github.com/szotp/injectable/tree/master/.vscode) of debugging (in vscode). You just press F5 and it launches with debugger attached. 

I had to write a script that maps the arguments from what DartCode provides to build_runner needs. I wonder what build_runner could do to make this simpler. Maybe a special command?

```
dart run build_runner debug $@
```

PR #4258, opened 2025-10-22
Speed up globbing by sorting paths and doing a prefix match.

davidmorgan:
For #3803

There is a scalability problem when a builder uses `findAssets(glob)`, because for N = the number of sources in the root package, the number of builder runs is O(N) and the cost of globbing is O(N), so the number of calls to `glob.matches` is O(N^2).

This shows clearly for built_value runs with N in the thousands, see benchmarks later.

Usually globs have a prefix, and the most important use of globs in the source_gen shared part builder actually has a prefix that pretty much just matches one file + parts that are output with the same base name plus different extensions.

So, optimize glob prefix matching. Keep a sorted list of paths in the package, extract the prefix from the glob, and do a binary search to find the range that matches the prefix. Then check with the full glob.

Introduce a new class `Nodes` to encapsulate lazily computed representations of the set of nodes in the graph.

Before this PR, if there is not much codegen going on then the shared part globbing dominates incremental runtime with 5000 files

```
> dart run _benchmark --generator=built_value --build-repo-path=$PWD/.. benchmark --size=5000 --mostly-no-codegen --shape=loop
clean/ms,no changes/ms,incremental/ms
41582,23393,27708
```

now with this PR

```
> dart run _benchmark --generator=built_value --build-repo-path=$PWD/.. benchmark --size=5000 --mostly-no-codegen --shape=loop
clean/ms,no changes/ms,incremental/ms
27783,7034,11892
```

### DIFF
--- a/build_runner/lib/src/build/asset_graph/nodes.dart
+++ b/build_runner/lib/src/build/asset_graph/nodes.dart
@@ -140,38 +140,13 @@ class Nodes {
     if (list == null) return const [];
     if (glob == null) return list;
 
-    final simplePrefix = _simpleGlobPrefix(glob);
+    final simplePrefix = simpleGlobPrefix(glob);
     Iterable<AssetId> result = list;
     if (simplePrefix.isNotEmpty) {
       result = list.filterToPrefix(simplePrefix);
     }
 
     return result.where((id) => glob.matches(id.path));
-  }
-
-  /// The prefix of the [glob] pattern up to but not including the first special
-  /// character.
-  String _simpleGlobPrefix(Glob glob) {
-    final pattern = glob.pattern;
-    for (var i = 0; i != pattern.length; ++i) {
-      final codeUnit = pattern.codeUnitAt(i);
-
-      /// Code units signalling the start of non-literal content in a glob
-      /// pattern, see: https://pub.dev/packages/glob
-      ///
-      /// Checked in nodes_test.dart:
-      ///
-      /// `expect(r'*?[\{'.codeUnits, [42, 63, 91, 92, 123])`
-      switch (codeUnit) {
-        case 42:
-        case 63:
-        case 91:
-        case 92:
-        case 123:
-          return pattern.substring(0, i);
-      }
-    }
-    return pattern;
   }
 
   /// Computes a `Map` from package names to sorted IDs of files in the package.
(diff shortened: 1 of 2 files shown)


build_runner/lib/src/build/asset_graph/nodes.dart
```diff
@@ -0,0 +1,239 @@
...
+    return result.where((id) => glob.matches(id.path));
+  }
+
+  /// The prefix of the [glob] pattern up to but not including the first special
+  /// character.
+  String _simpleGlobPrefix(Glob glob) {
+    final pattern = glob.pattern;
+    for (var i = 0; i != pattern.length; ++i) {
+      final char = pattern.codeUnitAt(i);
+      if (_specialGlobCharsCodeUnits.contains(char)) {
```
gmpassos: How about a `switch` to avoid `List.contains`?:

```dart
  switch (char) {
    case 42:  // *
    case 63:  // ?
    case 91:  // [
    case 123: // {
    case 92:  // \
      return pattern.substring(0, i);
  }
```

jensjoha: It searching a constant number of entries both will theoretically be O(1) --- that being said I agree: this field is inside a tight loop, but because it's a get of a final field it will add code that checks if the field has been initialized (again, inside the loop). Additionally it will have to load data to do the comparisons (vs just executing hard-coded comparisons).
Additionally, with the list only being used here it will be easier to see what actually happens.

build_runner/lib/src/build/asset_graph/nodes.dart
```diff
@@ -0,0 +1,239 @@
...
+        return pattern.substring(0, i);
+      }
+    }
+    return pattern;
+  }
+
+  /// Computes a `Map` from package names to sorted IDs of files in the package.
+  ///
+  /// An asset ID is a file if [AssetNode.isFile].
+  Map<String, List<AssetId>> _computeSortedFileIdsByPackage() {
```
jensjoha, line 155: How often is this done? It seems to me we invalidate much more than we have to, but if it's only done, say, once in practise anyway it obviously doesn't matter --- though maybe a comment somewhere would be good.

build_runner/lib/src/build/asset_graph/nodes.dart
davidmorgan: Thanks!

I like that code better, done.

The fact that there's no way to avoid "magic numbers" bugs me though, the best I could come up with was to add a test

```
expect(r'*?[\{'.codeUnits, [42, 63, 91, 92, 123]);
```

and refer to that in the comment.

gmpassos: \[EDITED] Added return line:
```dart
return pattern.substring(0, i);
```

build_runner/lib/src/build/asset_graph/nodes.dart
davidmorgan: It's at most once per build; and it's never used after it's invalidated.

So really it's three states: 1) not computed, 2) computed and still valid, 3) computed then discarded. If it's computed then discarded then when a new build starts it can be recomputed.

Added a bool flag to track the third state, and a big comment.

This is a bit complex, please let me know what you think :)

build_runner/lib/src/build/asset_graph/nodes.dart
```diff
@@ -0,0 +1,265 @@
...
+    for (var i = 0; i != pattern.length; ++i) {
+      final codeUnit = pattern.codeUnitAt(i);
+
+      /// Code units signalling the start of non-literal content in a glob
+      /// pattern, see: https://pub.dev/packages/glob
+      ///
+      /// Checked in nodes_test.dart:
+      ///
+      /// `expect(r'*?[\{'.codeUnits, [42, 63, 91, 92, 123])`
+      switch (codeUnit) {
```
jensjoha: Potentially you could add

```dart
const int $STAR = 42;
const int $QUESTION = 63;
const int $OPEN_SQUARE_BRACKET = 91;
const int $BACKSLASH = 92;
const int $OPEN_CURLY_BRACKET = 123;
```

(constants copied from the sdk @ pkg/_fe_analyzer_shared/lib/src/scanner/characters.dart) --- which could then make the switch more readable.

build_runner/lib/src/build/asset_graph/nodes.dart
jensjoha: 👍

build_runner/lib/src/build/asset_graph/nodes.dart
davidmorgan: Thanks! Done.

Issue #4264, opened 2025-10-28
build_runner is failing due to the missing files in another package


PiotrWpl:
Hi,

I have a monorepo project with multiple packages (27 to be more specific), and I'm using Melos to handle it. Unfortunately, after the recent project tech stack upgrade (Flutter from 3.29 to 3.35, Dart from 3.7.2 to 3.9.2, and different dependencies), I'm encountering a strange error while running the build_runner across the project with my melos steps.

```
buildrunner:clean:
    description: |
      Clean cache and run `build_runner` in all packages and apps
    steps:
      - for dir in apps packages; do find "$dir" -name ".dart_tool" -type d -exec rm -rf {} +; done
      - for dir in apps packages; do find "$dir" -name "build" -type d -exec rm -rf {} +; done
      - melos run pub:get
      - melos buildrunner:all

buildrunner:all:
    description: |
      Run `build_runner` to generate assets.
    run: |
      dart run build_runner build --delete-conflicting-outputs
    exec:
      failFast: true
      orderDependents: true
    packageFilters:
      dependsOn:
        - build_runner
```

Previously, with such code, we were able to successfully generate in parallel all the JSON serializations, api client, forms definition, local assets, and dependency injection files.
But now one builder is failing due to the non-existent files, which are part of another package (on which we have no dependency), and they are just not generated yet.

Example:
* We have a `domain_package`, `data_package` and `common_package`
* `domain` has no dependency on `data` and `common`
* `data` has dependency only on `domain`
* and `common` also has dependency on `domain`
* so there is no direct or transitive dependency between `common` and `data`

I run the build runner in a way as described above (a melos step), then:
* Previously generated files are being removed
* Next new files are successfully generated in the `domain` package
* And while generating dependency injection in `data` package and `common` package (they run in parallel, as there is no dependency between them), the build_runner in `common` fails due to the lack of .g.dart files in `data`

```
[common_package]: E injectable_generator:injectable_builder on data_package|lib/src/featureX/datasource/featureX_change_broadcaster_impl.dart:
[common_package]:   AssetNotFoundException: data_package|lib/src/featureX/dto/featureX_message_response_dto.g.dart 
```

My expectation would be not to fail the build runner due to a lack of files that are not only in the package on which we are not dependent, but also are not needed at all for the specific runner, and they will be generated later.

As a workaround, I had to redefine my melos steps to run the specific build runners in a sequence, but now the build time has increased a lot, so I would prefer to return to the solution where I can run them in parallel.

[...]

PR #4266, opened 2025-10-29
Prune package graph to actual deps.

davidmorgan:
WIP for #4264


davidmorgan:
Landing as #4278 instead

PR #4270, opened 2025-11-04
Splitting test expectations between expect, expectAndGetLine, and expectAndGetBlock.

Markzipan:
Following up on test changes mentioned earlier.

See https://github.com/dart-lang/build/pull/4240

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 2.10.2-wip
+## 2.10.3-wip
 
 - In "serve" and "watch" modes, retry failed compiles instead of exiting.
 


build_runner/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 2.10.3-wip
+
+- Testing: Split expects into three modes, which return nothing, a line, or several concatenated lines.
```
kevmoo: No need to document non-user facing changes in changelog!

build_runner/test/common/build_runner_tester.dart
```diff
@@ -239,62 +239,103 @@ class BuildRunnerProcess {
...
   ///
-  /// if [expectFailure] is set, then both [pattern] and [failOn] must be
-  /// encountered for the test to pass.
+  /// If the process exits instead, the test fails immediately.
+  ///
+  /// Otherwise, waits until [pattern] appears, then completes.
+  ///
+  /// Throws if the process appears to be stuck or done: if it outputs nothing
+  /// for 30s.
+  Future<void> expect(Pattern pattern, {Pattern? failOn}) async {
```
davidmorgan: Hmmmm I think this can just call `expectAndGetLine` and discard the result? :)

build_runner/CHANGELOG.md
davidmorgan: +1

build_runner/test/common/build_runner_tester.dart
```diff
@@ -239,62 +239,103 @@ class BuildRunnerProcess {
...
+  /// defaults to [BuildLog.failurePattern] so that `expect` will stop if the
+  /// process reports a build failure.
+  ///
+  /// If the process exits instead, the test fails immediately.
+  ///
+  /// Otherwise, waits until [pattern] appears, returns all text seen.
+  ///
+  /// Throws if the process appears to be stuck or done: if it outputs nothing
+  /// for 30s.
+  Future<String> expectAndGetBlock(Pattern pattern, {Pattern? failOn}) async {
```
davidmorgan, line 298: Hmmmm this could alternatively return `Future<List<String>>`, a list of lines.

Then the test would expect `contains(contains(...))` to match part of a line.

That's not really any better.

But it does mean `expectAndGetLine` can be just `(await expectAndGetBlock(...)).last`, which I kinda like.

Alternatively, it could stay as a String and `expectAndGetLine` could be `(await expectAndGetBlock(..)).split('\n).last`, which is a bit horrible but I think is okay for test code.

Up to you :) including leaving as-is.

Thanks!

build_runner/CHANGELOG.md
Markzipan: Done!

build_runner/test/common/build_runner_tester.dart
Markzipan: done!

build_runner/test/common/build_runner_tester.dart
Markzipan: I like that `.last` thing too, but I think it makes the actual test a bit too mysterious. The splits seem okay, but I'm getting windows-style line endings flashbacks. I'll leave it as-is until more tests using this pattern lead us down a more opinionated path.

davidmorgan:
LG, thanks!

PR #4273, opened 2025-11-07
Fixing DDC + FES not properly reading `entrypointAssetId` when re-serving an app.

Markzipan:
When this happens, we read from our `.web.entrypoint.json` file, which would have been generated in a previous compile.

### DIFF
--- a/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
+++ b/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
@@ -82,9 +82,10 @@ class DdcFrontendServerBuilder implements Builder {
           contents['entrypoint'] as String,
         );
       } else {
-        throw Exception(
+        log.severe(
           'Unable to read entrypoint when building ${buildStep.inputId}.',
         );
+        return;
       }
     }
 


build_web_compilers/lib/src/ddc_frontend_server_builder.dart
```diff
@@ -66,8 +66,30 @@ class DdcFrontendServerBuilder implements Builder {
...
+      );
+      if (await buildStep.canRead(webEntrypointAsset)) {
+        final contents =
+            json.decode(await buildStep.readAsString(webEntrypointAsset))
+                as Map<String, dynamic>;
+        frontendServerState.entrypointAssetId = AssetId.parse(
+          contents['entrypoint'] as String,
+        );
+      } else {
+        throw Exception(
```
davidmorgan: I guess you are throwing `Exception` so the stack trace doesn't get logged?

It would be maybe nicer to `log.severe` then return, which fails the build in the same way :) up to you.

Markzipan: ooh good point - updated!

PR #4279, opened 2025-11-12
Add logging of the compile step.

davidmorgan:
It seems like we should show what is going on with the compile, before adding any complexity about choosing to do an AOT compile vs a JIT compile :)

So, log it and how long it's taking.

Also improve logging around restarts, so they show with the same headers as a series of builds in watch mode.


davidmorgan:
Whoops, should have merged instead of rebased, sorry.

Force push was just updating version numbers in pubspecs+changelogs.

PR #4280, opened 2025-11-13
Add random benchmark shape.

davidmorgan:
Older build_runner versions stack overflow on very large max import "depth", so the other benchmark "shapes" don't work.

It would anyway be more realistic to have much more mixed imports.

So, add a "shape" with random imports. Use low-number-biased random numbers instead of flat random numbers as random flat imports are not realistic, either.

### DIFF
--- a/_benchmark/lib/shape.dart
+++ b/_benchmark/lib/shape.dart
@@ -73,7 +73,7 @@ enum Shape {
           if (_random.nextInt(libraryNumber + 1) == 0) 'app.dart',
           for (var i = 0; i != numberOfImports; ++i)
             Benchmarks.libraryName(
-              _random.nextInt(_random.nextInt(benchmarkSize + 1) + 1) + 1,
+              _random.nextInt(_random.nextInt(benchmarkSize) + 1),
               benchmarkSize: benchmarkSize,
             ),
         ];


_benchmark/lib/shape.dart
```diff
@@ -59,6 +64,21 @@ enum Shape {
...
+      case Shape.random:
+        // Use random.nextInt(random.nextInt(...)) to get a distribution of
+        // imports biased towards lower number libraries, which is more
+        // realistic than flat random imports.
+        final numberOfImports = _random.nextInt(_random.nextInt(10) + 1);
+        return [
+          if (_random.nextInt(libraryNumber + 1) == 0) 'app.dart',
+          for (var i = 0; i != numberOfImports; ++i)
+            Benchmarks.libraryName(
+              _random.nextInt(_random.nextInt(benchmarkSize + 1) + 1) + 1,
```
jensjoha: At least give a comment about all the +1s...
As I read it, if benchmarkSize is 9 this will give a number between 1 and 10 which `Benchmarks.libraryName` will then stringify to "between" `lib2.dart` and `lib11.dart` which might be right, but at least doesn't seem right (to me).

davidmorgan: Whoops, yes, it was too many +1s.

Only need one +1 to avoid passing 0 to nextInt; and libraryName does +1 as you pointed out.

Thanks.

PR #4281, opened 2025-11-13
Optimize library cycle loader.

davidmorgan:
The benchmark in https://github.com/dart-lang/build/pull/4280 shows another quadratic scaling problem: in the number of library cycles.

The code is currently doing work for every already-loaded library cycle for every library cycle.

A simple change only does the work for deps that are actually needed.

Edit+incremental build on "random" benchmark before this PR, aot
```
1000,2000,3000,4000,5000
5.17,24.8,83.22,217.17,424.04
```

Edit+incremental build on "random" benchmark after this PR, aot

```
1000,2000,3000,4000,5000
2.89,4.92,7.75,11.07,15.35
```

8j9fSNjWcErkMTU

### DIFF
--- a/build_runner/lib/src/build/library_cycle_graph/library_cycle_graph_loader.dart
+++ b/build_runner/lib/src/build/library_cycle_graph/library_cycle_graph_loader.dart
@@ -341,23 +341,20 @@ class LibraryCycleGraphLoader {
   ///
   /// A [_graphs] entry will be created for each ID in [newCycles].
   void _buildGraphs(int phase, {required List<LibraryCycle> newCycles}) {
-    // Build lookup from ID to [LibraryCycle] including new and existing cycles.
-    final existingCycles = <LibraryCycle>[];
-    for (final phasedCycle in _cycles.values) {
-      if (phasedCycle.isExpiredAt(phase: phase)) continue;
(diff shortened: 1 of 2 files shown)


gmpassos:
The yellow line in the chart was so low I only noticed it when I read the legend 😂

👏🏻👏🏻👏🏻

build_runner/lib/src/build/library_cycle_graph/library_cycle_graph_loader.dart
```diff
@@ -341,24 +341,21 @@ class LibraryCycleGraphLoader {
   ///
   /// A [_graphs] entry will be created for each ID in [newCycles].
   void _buildGraphs(int phase, {required List<LibraryCycle> newCycles}) {
-    // Build lookup from ID to [LibraryCycle] including new and existing cycles.
```
jensjoha, line 344: Am I reading this right, that this boils down to not pre-populating the map?

jensjoha (approved):

build_runner/lib/src/build/library_cycle_graph/library_cycle_graph_loader.dart
davidmorgan: Yes, exactly :)

Issue #4299, opened 2025-12-02
Expose build_web_compilers options to customize sdk and kernel paths


schultek:
TLDR;

Adding more customization options to build_web_compilers would allow Jaspr to support flutter embedding and flutter plugins without needing to maintain a custom fork of this package. These include changing the sdk-, kernel- and library paths of `ddc_builder`, adding to the list of platform libraries (to allow `dart:ui`) and a few more. No custom logic is needed beyond configuration options.

---

When I initially added flutter embedding support to Jaspr, I was forced to fork the `build_web_compilers` package because of the deep changes I needed to make to the compilation pipeline. The two main problems were:

1. Compiling a dart program that uses flutter (and thereby `dart:ui`).
2. Correctly initializing any depended-on flutter plugin.

The compilation is solved by:
- replacing a bunch of paths like the sdk path, sdk kernel path and `libraries.json` path with ones pointing to the respective assets from the installed flutter framework ([compare](https://github.com/schultek/build/pull/5/files#diff-3702e96e6a6ea2a0f1e3d4511a4deaa794e0ebb2dd34b036d1f8d4b801d41f16))
- adding `ui` and `ui_web` to the list of allowed platform libraries ([compare](https://github.com/schultek/build/pull/5/files#diff-6340b134ccb9374a6990f7e41e4a32fad96caafe15d933b645ca2958acf92ad0))
- ignoring checks on `dart:io` ([compare](https://github.com/schultek/build/pull/5/files#diff-507ff484ed0aabd52cde66af3db9cf701a824371780335498b08dc6afe0024c3L258))
- copying `dart_sdk.js` from flutter instead of compiling from source ([compare](https://github.com/schultek/build/pull/5/files#diff-9cd9d78c1562ccede76c2252b68efc5c90b16a238e5041b87e7827b59645db92))

The plugin initialization is solved by generating a custom entrypoint script that initializes all plugins, similar to what flutter is generating when running for web.

---

In the last release I made significant changes to the framework to reduce the custom logic (including moving all the plugin generation to `jaspr_builders`) and bring the fork much closer to the original, to a point where I think now there is a way to do everything Jaspr needs with a few additional build options on the original builders.

The following should be all options that would need to be added (preliminary names):

- `build_web_compilers:ddc`:
  - `platform_sdk_path` to be passed to `platformSdk` of `DevCompilerBuilder` and `KernelBuilder`
[...]

PR #4301, opened 2025-12-04
Add custom sdk build options

schultek:
For #4299

---

<details>

</details>

### DIFF
--- a/build_web_compilers/test/dart2js_bootstrap_test.dart
+++ b/build_web_compilers/test/dart2js_bootstrap_test.dart
@@ -223,7 +223,7 @@ void main() {
             const BuilderOptions({
               'compiler': 'dart2js',
               'native_null_assertions': false,
-              'unsafe_allow_unsupported_modules': true,
+              'unsafe-allow-unsupported-modules': true,
             }),
           ),
         ],


build_web_compilers/lib/src/platforms.dart
```diff
@@ -37,13 +36,29 @@ const _additionalWebLibraries = [
...
 
-final dart2wasmPlatform = DartPlatform.register('dart2wasm', [
-  ..._coreLibraries,
-  ..._additionalWasmLibraries,
-]);
+/// Registers the platforms with [DartPlatform].
+///
+/// Must be called before [ddcPlatform], [dart2jsPlatform], or
+/// [dart2wasmPlatform] is used.
+void initializePlatforms([List<String> additionalCoreLibraries = const []]) {
```
Markzipan: Is it important to pass in customizable allowed core libraries, or is a flag that enables a subset of them (like with `_additionalWebLibraries`) sufficient?

build_web_compilers/lib/builders.dart
```diff
@@ -194,20 +247,48 @@ bool _readWebHotReloadOption(BuilderOptions options) {
   return options.config[_webHotReloadOption] as bool? ?? false;
 }
 
+List<String> _readAdditionalCoreLibrariesOption(BuilderOptions options) {
+  return (options.config[_additionalCoreLibrariesOption] as List<Object?>?)
+          ?.cast<String>() ??
+      [];
+}
+
+String _readDdcKernelPathOption(BuilderOptions options) {
```
Markzipan, line 251: We likely want to extend checks in `_ensureSameDdcOptions` to these (if they're in just DDC + kernel builders).

I'm also not sure if we have a use case where we'd wanna pass in just a subset of these options, but it may be worth emitting a warning if one of them is missing.

build_web_compilers/lib/src/platforms.dart
schultek: I would only use it to set `ui` and `ui_web` and afaik there are no others, so yes we can also make this a bool flag.

build_web_compilers/lib/builders.dart
schultek: _ensureSameDdcOptions` already checks the whole options map so any option is always included in that check.

build_web_compilers/lib/src/platforms.dart
Markzipan: Sounds good! We can use something more general if the use case arises

build_web_compilers/lib/builders.dart
Markzipan: Oh true! And in hindsight, I don't think it makes sense to couple those options. So your current setup is fine.

schultek:
Thanks for the approval.

I did one renaming of the option related to allowing unsafe modules, because I had a read through the webdev issue and it seems like this is more related than I initially though. Since I don't want to conflict here with anything discussed there I opted for the renaming prefixed with "unsafe" and also not publicly documenting this flag.

Lmk if the changelog is ok like this, or if I should document the other added flags explicitly.

Markzipan:
Changelog looks fine! 'Non-user-visible' changes don't need to be included there, so your reasoning's sound. Pulling @bkonyi into this as an FYI; they've been following some efforts to allow these kinds of "flutter can use this lib but I can't" kind of situations.

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.4.6
+
+- Add build options to customize the SDK used for compiling to js and wasm.
```
kevmoo, line 3: Is this documented anywhere? How would one use it?

schultek: Currently not, but I'd like to. 
Where would be a good place to document this? I think this shouldn't be too visible (so not in Readme) since it's quite a special case.

kevmoo:
The other build options are in the readme, right?

It's fine to add an "advanced" section or similar. Just don't want this to get lost in the weeds!

build_web_compilers/CHANGELOG.md
schultek: @kevmoo I added this as a details section to the readme.

kevmoo:
Thank you, @schultek – at a minimum coding agents will appreciate having something to latch onto here!

build_web_compilers/CHANGELOG.md
kevmoo: Thanks!

schultek:
@kevmoo Great. Can you help me land this by approving one more time :)

schultek:
I would assume his approval is still valid, I didn't change any implementation details only docs.

Markzipan:
LGTM, let's land it!

davidmorgan:
I've been a bit snowed under so I didn't get a chance to look at this until now, it seems to have gone well without me, so, even better ;) thanks everyone.

bkonyi:
Thanks for looking into this! I think we'll actually be able to add this support generally to the CFE and DDC itself, so we might want to revisit this if https://dart-review.googlesource.com/c/sdk/+/465760 lands.

PR #4303, opened 2025-12-06
Adding support for DDC's Library Bundle module system.

Markzipan:
This module system - in conjunction with the Frontend Server - supports web hot reload. This is part of an effort to deprecate DDC's older module systems.

Notable updates:
* Plumbing a `ddc-library-bundle` flag through several builders.
* Copied/modified DDC tests
* Updated bootstrapper logic for the new module system.
* Updated some conflations between `webHotReload` and the new module system. These should be consistent now.

### DIFF
--- a/build_web_compilers/test/ddc_library_bundle_builder_test.dart
+++ b/build_web_compilers/test/ddc_library_bundle_builder_test.dart
@@ -16,6 +16,8 @@ final builderOptions = const BuilderOptions({
 });
 
 void main() {
+  initializePlatforms();
+
   group('DDC Library Bundle:', () {
     group('error free project', () {
       final startingAssets = {


build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -152,7 +152,7 @@ Future<void> _createDevCompilerModule(
   bool generateFullDill,
   bool emitDebugSymbols,
   bool canaryFeatures,
-  bool ddcModules,
+  bool ddcLibraryBundle,
```
davidmorgan: Optional, entirely up to you: this many adjacent positional bools makes me nervous, how about switching them to named?

build_web_compilers/test/ddc_library_bundle_bootstrap_test.dart
```diff
@@ -0,0 +1,275 @@
...
+          'a|web/index.digests': decodedMatches(contains('packages/')),
+        });
+        await testBuilders(
+          [...startingBuilders, builder],
+          startingAssets,
+          outputs: expectedOutputs,
+        );
+      });
+    });
+    group('regression tests', () {
```
davidmorgan, line 138: Super nitpicky optional nitpick: blank line here to give the group a bit of breathing space? :)

build_modules/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 5.1.5
+
+- Adding support for DDC's Library Bundle module system.
```
davidmorgan: Please use the direct form, "Add support for ...", or just "Support DDC's ..." would also work.

build_web_compilers/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-## 4.4.6
+## 4.4.7
+- Adding support for DDC's Library Bundle module system.
```
davidmorgan: Please use the direct form "Add support for DDC's ..." or just "Support DDC's ...".

Not sure if this is a complete change or part of the work? If it's complete then maybe it's worth saying a bit more about the feature from the user's point of view, if you don't plan to release then the version should be `4.4.7-wip` please :)

build_web_compilers/lib/src/dev_compiler_builder.dart
bkonyi: Yes please! I'd love to get to a place where we only use positional arguments for simple functions with one or two parameters and then default to using required named everywhere else, but obviously we don't have to do this all at once 😄

build_web_compilers/lib/src/web_entrypoint_builder.dart
```diff
@@ -126,9 +126,13 @@ final class EntrypointBuilderOptions {
 
   /// Whether or not to emit DDC entrypoints that support web hot reload.
   ///
-  /// Web hot reload is only supported for DDC's Library Bundle module system.
+  /// Only supported for DDC's Library Bundle module system.
```
bkonyi, line 129: Do we have any checks to enforce this? Or does `useWebHotReload == true` just do nothing if we're not using the new module system?

build_web_compilers/lib/src/dev_compiler_builder.dart
Markzipan: done!

build_modules/CHANGELOG.md
Markzipan: done!

build_web_compilers/test/ddc_library_bundle_bootstrap_test.dart
Markzipan: done!

build_web_compilers/CHANGELOG.md
Markzipan: This is 'formal' support, but the default hasn't been flipped. I'll add more details.

build_web_compilers/lib/src/web_entrypoint_builder.dart
Markzipan: Setting `useWebHotReload` automatically also enables the new module system. I manually added these checks a bit below during the flag resolution (~193).

PR #4306, opened 2025-12-10
Prepare to use DDC Library Bundle module system by default

Markzipan:
Plumbs some easily toggle-able flags through some tests + fixes some flags' logic. This isn't user visible.


gmpassos:
How to force it with the current published version? (So we can report any bug before it becomes the default.)

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.4.8-wip
+- Minor updates to flag resolution logic.
+
 ## 4.4.7
```
davidmorgan, line 1: Did you mean to release 4.4.7?

https://github.com/dart-lang/build/pull/4303/files#diff-6dc6d2a08a07bfe6a88bdaf089927dccbf638a138adddfea0664ca1ba2105851 put 4.4.7 in the CHANGELOG but did not update the pubspec, so there is no 4.4.7 released :) https://pub.dev/packages/build_web_compilers

To get back to a good state please do one of:

 - If you want a 4.4.7 release that does *not* include this PR, that will need a separate PR to update pubspec.yaml to 4.4.7, merge that, release, then this PR on top
 - If you want a 4.4.7 release that *does* include this PR, the new CHANGELOG entry here should be under 4.4.7, the pubspec needs updating to 4.4.7, then merge+release
 - If you don't want any release yet then the CHANGELOG should have 4.4.7-wip with two entries, and the pubspec should say 4.4.7-wip to match it

:)

davidmorgan (approved):

build_web_compilers/CHANGELOG.md
Markzipan: Woops - I do want this released in 4.4.7, so I'll update the changelog  + pubspec accordingly

Markzipan:
@gmpassos Disregard that commit message above - this doesn't flip the flag by default! It prepares the ecosystem for this, however, and we'll add instructions for reverting to a previous version (or other mitigations) prior to the swap.

gmpassos:
> @gmpassos Disregard that commit message above - this doesn't flip the flag by default! It prepares the ecosystem for this, however, and we'll add instructions for reverting to a previous version (or other mitigations) prior to the swap.

It’s very important to keep a flag for the new feature before it becomes the default, so we can test it properly and catch bugs or incompatibilities with other packages, especially for a package like this that is meant to build other packages.

Best regards.

PR #4318, opened 2026-01-07
Use builder names and factories instead of opaque functions.

davidmorgan:
For workspace support we need better management of build configuration; the `BuilderApplication` class gets in the way because it looks up the builder factory and discards information about the builder. Modify it to keep all the information, leaving the lookup for later.

This makes the `BuilderApplicaton` class more like a completed builder definition, so rename it to `BuilderDefinition`. It is a combination of the `BuilderDefinition` and `PostProcessBuilderDefinition` classes in `build_config`, which are direct from `build.yaml`.

`BuilderFactories` is now mentioned a lot in tests, make it take normal builders as a positional argument so the tests are more readable.

Move code for loading `BuilderDefinition`s out of `builder_factories.dart` into `builder_definition.dart` as it no longer has anything to do with factories.

Test coverage for "dev" and "release" defaults was missing, add it to e2e tests for normal and post builders.

The code in the new `BuildPhaseCreator` is largely identical to what was in `apply_builders.dart` and `build_phases.dart`, refactored to use methods instead of inline functions for clarity, and keeping the builder key instead of looking up factories.

Builder keys are now meaningful in `test_builder.dart` because they will be used to look up factories, where before the factories were passed directly. So, ensure that each name is unique and generate a unique name if not.

### DIFF
--- a/build_runner/lib/src/build_plan/builder_definition.dart
+++ b/build_runner/lib/src/build_plan/builder_definition.dart
@@ -18,92 +18,20 @@ import 'target_graph.dart';
 export 'package:build_config/build_config.dart'
     show AutoApply, TargetBuilderConfigDefaults;
 
-/// A builder definition read from `build.yaml` using
-/// [build_config.BuilderDefinition] or
-/// [build_config.PostProcessBuilderDefinition].
-class BuilderDefinition {
-  /// Whether this is a post process builder.
-  ///
-  /// If so, [hideOutput] is always `true`, [isOptional] is always false and
-  /// [autoApply] is always [AutoApply.none].
-  final bool isPostProcessBuilder;
+/// A [BuilderDefinition] or a [PostProcessBuilderDefinition].
+sealed class AbstractBuilderDefinition {
+  /// The builder key in the form `package:name`.
+  String get key;
 
   /// The package the builder is in.
-  final String package;
-
-  /// The builder key in the form `package:name`.
-  final String key;
-
-  /// Determines which packages a builder is automatically applied to.
-  final AutoApply autoApply;
-
-  /// Builder keys which, when applied to a target, will also apply this Builder
-  /// even if [autoApply] does not match.
-  final BuiltList<String> appliesBuilders;
-
-  /// Whether generated assets should be placed in the build cache.
-  final bool hideOutput;
-
-  /// Whether the builder is skipped if nothing uses its output.
-  final bool isOptional;
+  String get package;
 
   /// The defaults specified in `build.yaml` for this builder.
-  final TargetBuilderConfigDefaults targetBuilderConfigDefaults;
-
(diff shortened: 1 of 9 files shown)


build_runner/lib/src/build_plan/builder_definition.dart
```diff
@@ -0,0 +1,179 @@
...
+  final bool hideOutput;
+
+  /// Whether the builder is skipped if nothing uses its output.
+  final bool isOptional;
+
+  /// The defaults specified in `build.yaml` for this builder.
+  final TargetBuilderConfigDefaults targetBuilderConfigDefaults;
+
+  @visibleForTesting
+  BuilderDefinition(
```
jensjoha: Maybe the previous documentation

```
  /// Whether this is a post process builder.
  ///
  /// If so, [hideOutput] is always `true`, [isOptional] is always false and
  /// [autoApply] is always [AutoApply.none].
  final bool isPostProcessBuilder;
```

should be here (too or instead of, I'm not sure)?

build_runner/lib/src/build_plan/builder_definition.dart
```diff
@@ -0,0 +1,179 @@
...
+      hideOutput = builderDefinition.buildTo == build_config.BuildTo.cache,
+      isOptional = builderDefinition.isOptional,
+      targetBuilderConfigDefaults = builderDefinition.defaults;
+
+  BuilderDefinition.fromPostProcessConfig(
+    build_config.PostProcessBuilderDefinition builderDefinition,
+  ) : isPostProcessBuilder = true,
+      package = builderDefinition.package,
+      key = builderDefinition.key,
+      autoApply = build_config.AutoApply.allPackages,
```
jensjoha: This doesn't seem to fit with `If so [...] [autoApply] is always [AutoApply.none].`?

davidmorgan: Thanks!

These values shouldn't actually matter for post process builders, so it's unnecessarily confusing to have them.

I split the class into two so now these unused defaults are not needed, and the code gets more precise in a few places.

build_runner/lib/src/build_plan/builder_definition.dart
davidmorgan: Refactored so there's no list of hard coded fields for post process builders.

davidmorgan (commented):
Thanks!

Refactored to two separate classes, please see what you think.

Actually I think the main reason I didn't do this in the first place is we don't have a good name for a not-"post process"-builder, which makes it hard to pick a name for an "either-type-of"-builder. I went for `AbstractBuilderDefinition` + `BuilderDefinition` + `PostProcessBuilderDefinition`, if you have any better idea I'd be happy to hear it :)

PR #4325, opened 2026-01-16
Stop using "root package".

davidmorgan:
The "root package" is the package build_runner is run in. Currently, it's used for a mix of different things; split it up based on what it's doing, so that when workspace support is added the different uses are distinct.

 - The new `buildPackages.outputRoot` is
     - where `.dart_tool/build` will be written with the build script, compiled build script, generated output, serialized asset graph
     - if writing a performance log it's the root for the relative path
     - it's where `yaml` files are placed for global settings and global overrides
     - it's the root for the `serve` mode
 - The new `buildPackages.currentPackage` is the package `build_runner` was launched in, used to default to the current package when parsing/displaying; for a workspace it will be null
 - The new `buildPackages.packagesInBuild` is the packages for which output will be produced

Including specific changes

 - `BuildConfigs` holds all package configs by package name, not just the root package
 - `DuplicateAssetNodeException` no longer has special case `toString` for the current package, this saves passing the current package through a lot of `AssetGraph` methods
 - `AssetFinder`, `_findAssets` cleaned up to always require the package to search in, defaulting to "root package" is not needed
 - `AssetPathProvider`: do the mapping of paths to the build cache and the check for which packages can be written to here, because the main implementation is the `BuildPackages` class which knows how to do both
 - `GeneratedAssetHider`: return `isHidden` instead of the maybe-hidden ID; the main implementation is `AssetGraph` which knows whether to hide the asset but does not need to know where
 - Simplify `ReaderWriter` because `AssetPathProvider` now handles checking whether a delete is allowed

### DIFF
--- a/build_runner/lib/src/build_plan/build_packages.dart
+++ b/build_runner/lib/src/build_plan/build_packages.dart
@@ -55,7 +55,7 @@ class BuildPackages implements AssetPathProvider {
   final BuiltSet<String> packagesInBuild;
 
   BuildPackages._({
-    required this.current,
+    required BuildPackage this.current,
     required this.outputRoot,
     required Map<String, BuildPackage> allPackages,
   }) : asPackageConfig = _packagesToConfig(allPackages.values),
@@ -63,13 +63,11 @@ class BuildPackages implements AssetPathProvider {
          Map<String, BuildPackage>.from(allPackages)
            ..putIfAbsent(r'$sdk', () => _sdkPackage),
        ),
-       // TODO(davidmorgan): when workspace support is added there will be
-       // no current package, set the list of output packages.
-       packagesInBuild = {if (current != null) current.name}.build() {
+       packagesInBuild = {current.name}.build() {
     if (!current!.isInBuild) {
       throw ArgumentError('Current package must indicate `isInBuild`');
     }
-    if (allPackages.values.where((n) => n != current).any((n) => n.isInBuild)) {
+    if (allPackages.values.any((p) => p != current && p.isInBuild)) {
       throw ArgumentError(
         'No packages other than the current may indicate `isInBuild`',
       );
@@ -253,8 +251,11 @@ class BuildPackages implements AssetPathProvider {
     required bool hide,
     bool checkDeleteAllowed = false,
   }) {
-    if (checkDeleteAllowed && !hide) {
-      if (!packagesInBuild.contains(id.package)) {
+    if (checkDeleteAllowed) {
+      // Delete is allowed if it's a hidden output or if it's an output to a
+      // package in the build. This forbids deleting files that are part of the
+      // checked-in source tree of a dependency package.
+      if (!(hide || packagesInBuild.contains(id.package))) {
         throw InvalidOutputException(
           id,
           'Should not delete assets outside of package(s): '
--- a/build_runner/lib/src/build_plan/build_phases.dart
+++ b/build_runner/lib/src/build_plan/build_phases.dart
@@ -85,23 +85,23 @@ class BuildPhases {
 
   /// Checks that outputs are to allowed locations.
   ///
-  /// Valid outputs are hidden or to packages in [allowedOutputPackages].
+  /// Valid outputs are hidden or to packages in [packagesInBuild].
   ///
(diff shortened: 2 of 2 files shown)


build_runner/lib/src/build_plan/build_packages.dart
```diff
@@ -22,34 +22,63 @@ final _sdkPackage = BuildPackage(r'$sdk', sdkPath, null, isEditable: false);
...
+  }) : asPackageConfig = _packagesToConfig(allPackages.values),
+       allPackages = Map.unmodifiable(
+         Map<String, BuildPackage>.from(allPackages)
+           ..putIfAbsent(r'$sdk', () => _sdkPackage),
+       ),
+       // TODO(davidmorgan): when workspace support is added there will be
+       // no current package, set the list of output packages.
+       packagesInBuild = {if (current != null) current.name}.build() {
+    if (!current!.isInBuild) {
+      throw ArgumentError('Current package must indicate `isInBuild`');
```
jensjoha: This is very hard to parse (for me at least) - it took me a while to figure out that the first line is the last initializer and that the start bracket is the start of the body of the constructor :/

It's probably not made better by checking if `current` is non-null, then immediately forcing it not to be..? (as in you're preparing for it to be null with workspace, but also not really?)

build_runner/lib/src/build_plan/build_packages.dart
```diff
@@ -22,34 +22,63 @@ final _sdkPackage = BuildPackage(r'$sdk', sdkPath, null, isEditable: false);
...
     }
-    if (allPackages.values.where((n) => n != root).any((n) => n.isRoot)) {
+    if (allPackages.values.where((n) => n != current).any((n) => n.isInBuild)) {
```
jensjoha: while this isn't new it's just too convoluted. Why not just `if (allPackages.values.any((n) => n != current && n.isInBuild))`?

build_runner/lib/src/build_plan/build_packages.dart
```diff
@@ -211,8 +248,24 @@ class BuildPackages implements AssetPathProvider {
...
 
   @override
-  String pathFor(AssetId id) {
-    final package = this[id.package];
+  String pathFor(
+    AssetId id, {
+    required bool hide,
+    bool checkDeleteAllowed = false,
+  }) {
+    if (checkDeleteAllowed && !hide) {
```
jensjoha: A comment about why this logic is right would be good.

build_runner/lib/src/build_plan/build_phases.dart
```diff
@@ -85,25 +85,23 @@ class BuildPhases {
...
       // script generation filters these out.
       buildLog.error(
         'A build phase (${action.displayName}) is attempting '
-        'to operate on package "${action.package}", but the build script '
-        'is located in package "$root". It\'s not valid to attempt to '
-        'generate files for another package unless the BuilderDefinition '
-        'specified "hideOutput".',
+        'to operate on package "${action.package}" without "hideOutput". '
+        'Allowed non-hidden output packages are: '
+        '${allowedOutputPackages.join(', ')}',
```
jensjoha: Is outputting this list useful?

build_runner/lib/src/build_plan/build_packages.dart
davidmorgan: Oh, yes, that's a bit awkward. Made it just accept non-nullable `BuildPackage` for `current`. Forcing everyone else to deal with `BuildPackage?` is enough to be future-proof :)

Thanks.

build_runner/lib/src/build_plan/build_packages.dart
davidmorgan: Thanks, done.

build_runner/lib/src/build_plan/build_packages.dart
davidmorgan: Done.

build_runner/lib/src/build_plan/build_phases.dart
davidmorgan: Probably not, changed to "This is only allowed for packages in the build, no
for dependency packages." Thanks!

Issue #4312, opened 2025-12-24
`serve`: CMD console showed up after first change rebuild (dartvm.exe).


ykmnkmi:
`Dart SDK version: 3.10.6 (stable) (Tue Dec 16 06:05:04 2025 -0800) on "windows_x64"`

```
dev_dependencies:
  build_runner: ^2.10.4
  build_web_compilers: ^4.4.6
```

After I send `CTRL + C` into it to close it, next rebuild throws this (not always):

```
lib/blank.ddc.module build_web_compilers:ddc (lazy)
E SocketException: Write failed (OS Error: The channel is being closed, errno = 232), port = 0
```

But rebuilds after does not throw it. Local server is still working.

Before change:
<img width="1078" height="83" alt="Image" src="https://github.com/user-attachments/assets/d520ecc1-401b-4cc7-8a79-0fbff219ccd7" />
After change:
<img width="1097" height="135" alt="Image" src="https://github.com/user-attachments/assets/747b4c4d-552c-4749-8a35-e5b321faf002" />
After sending signal:
<img width="1349" height="100" alt="Image" src="https://github.com/user-attachments/assets/fa9188eb-2cca-4650-ac72-4563689873c2" />

davidmorgan:
@Markzipan could be to do with the recent build_web_compilers changes?

Markzipan:
Hmm, I don't suspect so at the moment, but I'll investigate. Do you have a sample app that's causing this issue @ykmnkmi ?

ykmnkmi:
@Markzipan fresh web application:
```
dart create -t web web_app
cd web_app
dart pub upgrade --major-versions --tighten
dart run build_runner serve
```

<img width="2091" height="1696" alt="Image" src="https://github.com/user-attachments/assets/5bfd6a8c-5ec8-46dc-9fb0-a85b6cef7f34" />

Markzipan:
Hmm, that's interesting. That additional process is just DDC spinning up to compile the implicated module. The command window opening and hanging implies there's some sort of comms problem causing the underlying process to hang, but I can't repro it on my local machine. Can you retry after restarting your machine? It's possible some other process is affecting DDC's ability to write.

ykmnkmi:
CMD windows didn't appear after a restart, but they did pop up after `dart run build_runner clean` again.

ykmnkmi:
I also upgraded my Flutter SDK, from Dart 3.10.4 to 3.10.7. Stable branch.

[...]

PR #4326, opened 2026-01-16
Use default process mode for workers to avoid spawning extraneous consoles in Windows. 

Markzipan:
Fixes https://github.com/dart-lang/build/issues/4312

This expression was added ~7 years ago to avoid spawning extraneous consoles, but now this behavior is handled deeper in the Dart stack. Avoiding spawning extraneous consoles in Win requires the process to be spawn non-detached. See https://github.com/dart-lang/sdk/blob/728f837e1872dfae0f1e200e59681db75702a5f4/runtime/bin/process_win.cc#L552

### DIFF
--- a/build_modules/CHANGELOG.md
+++ b/build_modules/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.1.7
+
+- Use default process modes for kernel workers to avoid spawning extraneous terminals. 
+
 ## 5.1.6
 
 - Allow `analyzer` 10.0.0.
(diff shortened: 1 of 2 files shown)


davidmorgan:
Edit: I did suggest releasing this, but please see the next comment :)

_Please go ahead with merge then publish if that's what you were planning :) thanks._

davidmorgan:
@Markzipan 

It looks like this also affects other places, see https://github.com/dart-lang/build/issues/4330

--we might want to hold off on merging+releasing this PR until we figure out what changed.

Markzipan:
Abandoning in favor of a deeper fix: https://dart-review.googlesource.com/c/sdk/+/475000

PR #4329, opened 2026-01-20
Improve BuildPackages tests, remove most checked in test fixtures.

davidmorgan:
The checked-in test fixtures to test reading packages/workspaces are awkward because they are hard to keep correct; use `BuildRunnerTester` package setup instead which uses yaml + the real `pub`.

Simplify the tests by using value comparison instead of matchers.

Delete test fixtures that are now unused, leaving test fixtures only for Flutter and for invalid file states.


davidmorgan:
@jensjoha rebased on top of the one you approved, PTAL :)

(Really I should have merged instead of rebasing, but, uh, I didn't figure out how to do that in jj yet).

PR #4331, opened 2026-01-21
Workspace support

davidmorgan:
Add `--workspace` flag that builds for all packages in the workspace instead of just the "current" package.

The recently-introduced concept "packages in the build" is renamed to "output packages". They are packages that generated files can be written to, in a workspace build they are exactly the packages in the workspace.

Places that do something different due to the `--workspace` flag:

- The entrypoint contains factories for builders needed to build any package, not just the "current" package
- The `.dart_tool/build` output and other outputs move to the workspace root
- Logging uses package names always instead of omitting the name if it matches the "current" package
- Auto-applying builders gets more subtle: a builder should only auto apply if it would be part of some single-package build; adding a package to the workspace that depends on a new builder should not cause that builder to apply to the rest of the workspace; packages in the "same single package build" are called "peers"

Other changes

 - Split `build_packages_loader.dart` out of `build_packages.dart` as the `BuildPackages` class was getting to be far too much about loading itself than doing something useful :)
 - Pubspec handling in build_runner/test/common/build_runner_tester.dart tweaked to use overrides less because packages in the same workspace aren't allowed to repeat overrides; update test setup in build_command_invalidation_test.dart so it works with the new approach.
 
 This PR leaves the feature substantially complete, I'm thinking it would be useful to do a dev release to get early testers before a stable release in case I missed something. Things that I know still need doing before a stable release:

 - Test coverage for watch, serve and daemon modes; test coverage for handling of yaml config in workspaces; and any needed fixes for issues found :)
 - Add to _benchmark, check performance is improved as expected

### DIFF
--- a/build_runner/test/integration_tests/build_command_workspace_test.dart
+++ b/build_runner/test/integration_tests/build_command_workspace_test.dart
@@ -106,12 +106,9 @@ workspace: [p1, p2, p3]
 ''');
 
     // Builders run.
-    await tester.run('p3', 'dart run build_runner build');
+    await tester.run('p3', 'dart run build_runner build --workspace');
     expect(tester.read('p3/lib/p3.txt.copy'), '1');
-    expect(
-      tester.read('p3/.dart_tool/build/generated/p3/lib/p3.txt.hidden'),
-      '1',
-    );
+    expect(tester.read('.dart_tool/build/generated/p3/lib/p3.txt.hidden'), '1');
 
     // Add a package to the workspace that are direct and indirect dependencies
     // of `p1`.


davidmorgan:
(Sorry, missed pushing some last cleanup--now done).

rrousselGit:
Thought: Should running `build_runner` at the workspace level bypass the need for `--workspace`?


As in, if you run `build_runner` next to a pubspec that contains `workspace: [...]`, I think it should be fair to assume that we want to run on the whole workspace

davidmorgan:
@rrousselGit probably, yes :) ... will wait for some real world feedback before making it default anywhere though.

Issue #4332, opened 2026-01-22
[build_test] TestReaderWriter writes to wrong AssetId


kerberjg:
When using testBuilder with TestReaderWriter I can't access the output assets via `readerWriter.testing.readString` or `readerWriter.readAsString`, even though the asset is visible in `readerWriter.testing.assetsWritten`.

The output is accessible with AssetId `a|.dart_tool/build/generated/a/lib/user_immutable.data.g.dart`, even though its intended AssetId is `a|lib/user_immutable.data.g.dart`, and it exists as such in `assetsWritten`

Example code:
```dart
/// Runs the crystallis builder in-memory and returns all output assets.
Future<Map<String, String>> _runBuilder({
  required String inputDartPath,
  required String inputDart,
}) async {
  final builder = crystallisBuilder(const BuilderOptions({}, isRoot: true));

  await preloadSources();

  final rw = TestReaderWriter(rootPackage: outputPackage);

  final result = await testBuilder(
    builder,
    {
      inputDartPath: inputDart,
      // Provide the package's own sources that the input imports.
      // build_test resolves package:crystallis/... against "crystallis|lib/..."
      ...sourceFiles,
    },
    outputs: null,
    readerWriter: rw,
  );

  // print all assets
  for (final asset in rw.testing.assetsWritten) {
    print('Generated asset: ${asset}');
    print('  path: ${asset.path}');
    print('  package: ${asset.package}');
    print('  uri: ${asset.uri}');
  }

  for (final asset in rw.testing.assets) {
    print('Asset in memory: ${asset}');
    print('  path: ${asset.path}');
    print('  package: ${asset.package}');
    print('  uri: ${asset.uri}');
  }

  // Collect all generated outputs.
  for (final asset in result.outputs) {
    print('Output asset id: ${asset}');
    print('  path: ${asset.path}');
    print('  package: ${asset.package}');
    print('  uri: ${asset.uri}');
  }

  /// Map of output asset paths to their contents.
  Map<String, String> outputs = {};

  for (final AssetId asset in result.outputs) {
    print("reading generated asset: ${asset}");
    print("  can read ${asset}: ${await rw.canRead(asset)}");
    print("  exists ${asset}: ${await rw.testing.exists(asset)}");

    final content = rw.testing.readString(asset);
    outputs['${asset.package}|${asset.path}'] = content;
  }

  return outputs;
}
```

...invoked like...

```dart
    test('immutable: set<T> returns new instance (signature check)', () async {
[...]

PR #4345, opened 2026-02-03
Added docs for accessing built `AssetId`s via `TestReaderWriter`

kerberjg:
Adds docs on how to read built files via `TestReaderWriter` with flattened `AssetId`
For details see https://github.com/dart-lang/build/issues/4332#issuecomment-3785487983

---

<details>

</details>

### DIFF
--- a/build_test/README.md
+++ b/build_test/README.md
@@ -55,43 +55,40 @@ test('should resolve a simple dart file', () async {
 ```
 
 ### Get generated source code as `String`
-You can use `TestReaderWriter` to read generated assets as `String`.
-Make sure to set `flattenOutput` to `true` in `testBuilder` to make sure the output assets are accessible:
+
+You can use `TestReaderWriter` to read generated assets as `String`. Read more about
+how this works in the [TestReaderWriter documentation][api:TestReaderWriter].
+
+Set `flattenOutput` to `true` in `testBuilder` to make the output assets accessible:
 
 ```dart
 final yourBuilder = ...;
 
-final rw = TestReaderWriter(rootPackage: 'test_package');
+final readerWriter = TestReaderWriter(rootPackage: 'test_package');
 
 final result = await testBuilder(
   yourBuilder,
   {...},
-  readerWriter: rw,
+  readerWriter: readerWriter,
   outputs: null,
   flattenOutput: true,
 );
 
 
-/// Map of output asset paths to their contents.
-Map<String, String> outputs = {};
-
+// You can read all of the generated assets like this...
 for (final AssetId asset in result.outputs) {
-  // print("reading generated asset: ${asset}");
-  // print("  can read ${asset}: ${await rw.canRead(asset)}");
-  // print("  exists ${asset}: ${await rw.testing.exists(asset)}");
-
-  final content = rw.testing.readString(asset);
-  outputs['${asset.package}|${asset.path}'] = content;
+  print('reading generated asset: ${asset}');
+  print('  can read: ${await readerWriter.canRead(asset)}');
+  print('  exists: ${await readerWriter.testing.exists(asset)}');
+  print('  content: \n${await readerWriter.testing.readString(asset)}');
 }
-```
 
-Now `outputs` contains all generated assets as strings. Access it like this:
-
-```dart
-final generatedCode = outputs['test_package|lib/source_file.g.dart'];
-
-// you can use it in tests like
-expect(generatedCode, contains('some expected content'));
+// ...or access a single generated asset directly, like this:
+final String generatedCode = await readerWriter.testing.readString(
+  AssetId('test_package', 'lib/source_file.g.dart')
+);
+// ...and, for example, use it in tests:
+expect(generatedCode, contains('callExpectedFunction();'));
 ```
 
 [api:PackageAssetReader]: https://pub.dev/documentation/build_test/latest/build_test/PackageAssetReader-class.html
--- a/build_test/lib/src/test_builder.dart
+++ b/build_test/lib/src/test_builder.dart
@@ -174,7 +174,7 @@ Future<TestBuilderResult> testBuilder(
 ///
 /// To access the outputs via `testing.readString` using the same
 /// `AssetId`s as returned by the outputs of this method,
-/// make sure to set [flattenOutput] to `true`. See [TestReaderWriter] for details.
+/// set [flattenOutput] to `true`. See [TestReaderWriter] for details.
 Future<TestBuilderResult> testBuilders(
   Iterable<Builder> builders,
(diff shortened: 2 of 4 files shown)


kerberjg:
cc @davidmorgan

build_test/lib/src/test_builder.dart
```diff
@@ -171,6 +171,10 @@ Future<TestBuilderResult> testBuilder(
 /// `build_runner` config pass [testingBuilderConfig] `false`. To read
 /// `build.yaml` files passed in `sourceAssets`, use [testBuilderFactories]
 /// directly.
+///
+/// To access the outputs via `testing.readString` using the same
+/// `AssetId`s as returned by the outputs of this method,
+/// make sure to set [flattenOutput] to `true`. See [TestReaderWriter] for details.
```
davidmorgan: Nitpick: please remove "make sure to" for conciseness, i.e. just "by the outputs of this method, set [flattenedOutput] to `true`".

build_test/lib/src/test_reader_writer.dart
```diff
@@ -18,6 +19,15 @@ import 'internal_test_reader_writer.dart';
...
 /// a build. This specifies which package `build_runner` is running in.
+///
+/// On disk "hidden" assets are placed under `.dart_tool/build/generated/<package>`
+/// while non-hidden assets are placed in the main source tree.
+/// Some parts of this API are from the point of view of a builder and use a
+/// unified namespace for [AssetId]s, while others see a split view
+/// of the hidden/non-hidden filesystem structure, and require an explicit
+/// mapping of [AssetId]s to the on-disk structure
+/// (eg. starting with `.dart_tool/build/generated/`) to access hidden files.
+/// See [testBuilders] `flattenOutput` parameter for more details.
```
davidmorgan, line 30: Please add a

///

line so "See [testBuilders]..." is a new paragraph.

build_test/README.md
```diff
@@ -54,6 +54,46 @@ test('should resolve a simple dart file', () async {
 });
 ```
 
+### Get generated source code as `String`
+You can use `TestReaderWriter` to read generated assets as `String`.
```
davidmorgan: Please add newlines so "You can use..." and "Make sure to..." are clearly separate paragraphs.

build_test/README.md
```diff
@@ -54,6 +54,46 @@ test('should resolve a simple dart file', () async {
...
+Make sure to set `flattenOutput` to `true` in `testBuilder` to make sure the output assets are accessible:
```
davidmorgan: Let's expand on this a bit, how about:

By default `TestReaderWriter` refers to generated assets as they would appear on disk, meaning "hidden" assets are under `.dart_tool`. Set `flattenOutput` to `true` to refer to them as they appear to builders, without `.dart_tool`:

build_test/README.md
```diff
@@ -54,6 +54,46 @@ test('should resolve a simple dart file', () async {
...
+
+```dart
+final yourBuilder = ...;
+
+final rw = TestReaderWriter(rootPackage: 'test_package');
```
davidmorgan: Nitpick: we usually encourage using full words in variable names, I suggest calling it just `readerWriter` please.

build_test/README.md
```diff
@@ -54,6 +54,46 @@ test('should resolve a simple dart file', () async {
...
+  outputs: null,
+  flattenOutput: true,
+);
+
+
+/// Map of output asset paths to their contents.
+Map<String, String> outputs = {};
+
+for (final AssetId asset in result.outputs) {
+  // print("reading generated asset: ${asset}");
```
davidmorgan: Nitpick: please use single quotes for strings throughout the example, i.e.

```
print('reading generated asset: ${asset}');
...
```

build_test/README.md
```diff
@@ -54,6 +54,46 @@ test('should resolve a simple dart file', () async {
...
+  yourBuilder,
+  {...},
+  readerWriter: rw,
+  outputs: null,
+  flattenOutput: true,
+);
+
+
+/// Map of output asset paths to their contents.
+Map<String, String> outputs = {};
```
davidmorgan: I'm not sure the Map adds much to the example, since you can go directly to `readerWriter.testing` instead of the map lookup ... how about the example being

1) the loop + prints that you have, maybe uncomment the prints and print the content length instead of adding to the map, this shows how to do something to all the output
2) show how to do some individual checks on files, I would just write it explicitly like

```
expect(
    readerWriter.testing.read(AssetId('test_package', 'lib/source_file.g.dart')),
    contains('some expected content'),
)
```

what do you think please?

build_test/lib/src/test_reader_writer.dart
```diff
@@ -74,6 +84,12 @@ abstract interface class ReaderWriterTesting {
   Iterable<AssetId> get assetsWritten;
 
   /// Whether [id] exists on the [TestReaderWriter] in-memory filesystem.
+  ///
+  /// [AssetId]s are from the point of view of the filesystem,
+  /// so "hidden" assets are under .dart_tool.
+  /// Also see:
```
davidmorgan: Please add a

///

to separate paragraphs; that makes it a bit tall, I think then packing into one paragraph is better:

```
  /// so "hidden" assets are under .dart_tool.
  ///
  /// See also [TestReaderWriter] class documentation and [testBuilders] `flattenOutput`.
```

and the same where this is repeated below.

build_test/lib/src/test_reader_writer.dart
kerberjg: I kept them in the same paragraph so it's unambiguous that the "See also" applies exclusively to this paragraph, and not any of the above ones.

May I suggest we keep it this way?

build_test/lib/src/test_reader_writer.dart
kerberjg: Same as above, the intention in keeping them close is preventing ambiguity.

build_test/README.md
kerberjg: That sounds better, yeah! I'd do both 1 and 2, adding a comment above each section explaining a the purpose

build_test/README.md
kerberjg: I'd like to suggest we keep the more in-depth explanation confined to the API docs (but perhaps link to them here?) and keep the README focused on the "how to achieve this use-case"

kerberjg:
@davidmorgan Pushed suggested changes

build_test/lib/src/test_reader_writer.dart
davidmorgan: Sure, that works. Thanks!

kerberjg:
@davidmorgan Thanks for reviewing! 🥳

PR #4352, opened 2026-02-05
fix dev bootstrap script on safari

schultek:
Fixes https://github.com/dart-lang/webdev/issues/1499 based on @eddyw 's solution.

I confirmed this works in Safari Version 26.2 (21623.1.14.11.9)

---

- [x] I’ve reviewed the contributor guide and applied the relevant portions to this PR.

<details>
  <summary>Contribution guidelines:</summary><br>

- See our [contributor guide](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md) for general expectations for PRs.
- Larger or significant changes should be discussed in an issue before creating a PR.
- Contributions to our repos should follow the [Dart style guide](https://dart.dev/guides/language/effective-dart) and use `dart format`.
- Most changes should add an entry to the changelog and may need to [rev the pubspec package version](https://github.com/dart-lang/sdk/blob/main/docs/External-Package-Maintenance.md#making-a-change).
- Changes to packages require [corresponding tests](https://github.com/dart-lang/.github/blob/main/CONTRIBUTING.md#Testing).

Many Dart repos have a weekly cadence for reviewing PRs - please allow for some latency before initial review feedback.

**Note**: The Dart team is trialing Gemini Code Assist. Don't take its comments as final Dart team feedback. Use the suggestions if they're helpful; otherwise, wait for a human reviewer.

</details>

PR #4356, opened 2026-02-06
Fixing issue with double loads in ddc library loader hot restarts

Markzipan:
Now that we emit the `onLoadEnd` bootstrap script, we need to remove its automatic invocation.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## 4.4.10
 
+- Fix stack trace file name extraction in Safari.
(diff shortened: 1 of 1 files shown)


build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.4.10
+
+- Remove extraneous reload logic in DDC's Library Loader bootstrapper.
```
davidmorgan, line 4: Could you please also describe the fix just merged in https://github.com/dart-lang/build/pull/4352 ? Then both can release together, yay.

Markzipan: Done!

PR #4374, opened 2026-02-25
Updating sourcemap logic in ddc library bundle sources

Markzipan:
Updates `.dart.lib.js` in inlined sourcemaps to `.ddc.js`.

Also updates the bootstrapper to invoke main after sources are loaded.

### DIFF
--- a/build_web_compilers/CHANGELOG.md
+++ b/build_web_compilers/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.4.13
+- Update sourcemap paths for DDC Library Bundle sources.
+- Update bootstrapper for DDC Library Bundle apps to auto-run main.
+
 ## 4.4.12
 - Remove `setStartAsyncSynchronously` in DDC's bootstrapper.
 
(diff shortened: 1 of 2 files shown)


build_web_compilers/lib/src/dev_compiler_builder.dart
```diff
@@ -277,7 +277,20 @@ Future<void> _createDevCompilerModule(
     }
 
     // Copy the output back using the buildStep.
-    await scratchSpace.copyOutput(jsId, buildStep);
+    if (ddcLibraryBundle) {
+      final jsFile = scratchSpace.fileFor(jsId);
+      final content = await jsFile.readAsString();
+      // This hack is required to make the source map's file name match that of
+      // the emitted JS file name (e.g., transform .lib.js into .ddc.js).
```
nshahan, line 284: IIUC correctly the embedded sourcemap string in the JavaScript contains the wrong name for the source file, which is the same file where it is located? How did this get to be wrong in the first place? I thought DDC would write whatever name was passed to it as the file name of the output file.

It makes me think there is code somewhere in here that is already rewriting the name and maybe we should correct it there?

Markzipan: That's the case for standard DDC compiles, but Frontend Server hardcodes the `.lib.js` ending. We can revisit this if we unify these names though

PR #4381, opened 2026-03-03
Fix for build_config 1.2.0.

### LINKED ISSUE #4380: [Bug] build_runner 2.12.0 fails: PostProcessBuilderDefinition has no buildTo getter
* The Dart SDK Version (`dart --version`)
  * From project constraints: `>=3.11.0 <4.0.0` (exact local output unavailable in this environment)
* What package(s) from this repo you are using, and the version
  * `build_runner: 2.12.0`
  * `build_config: 1.2.0`
  * `build: 4.0.4`
* What builder(s) you are using (or writing yourself)
  * `auto_route_generator: 10.5.0`
  * `freezed: 3.2.5`
  * `injectable_generator: 2.12.1`
  * `json_serializable: 6.13.0`
  * `retrofit_generator: 10.2.3`
  * `source_gen: 4.2.0`
* Whether you are using Windows, MacOSX, or Linux
  * macOS
* Whether you are using Chrome, Safari, Firefox, Edge
  * N/A (CLI failure)
* Any other packages or constraints we should know about
  * `sdk: '>=3.11.0 <4.0.0'`
  * Lockfile Flutter constraint: `>=3.41.3 <4.0.0`

## Actual behavior

`dart run build_runner build --delete-conflicting-outputs --force-aot` fails while compiling build_runner:

```text
Building package executable... (1.3s)
Failed to build build_runner:build_runner:
../../.pub-cache/hosted/pub.dev/build_runner-2.12.0/lib/src/build_plan/builder_definition.dart:218:38: Error: The getter 'buildTo' isn't defined for the type 'PostProcessBuilderDefinition'.
 - 'PostProcessBuilderDefinition' is from 'package:build_config/src/builder_definition.dart' ('../../.pub-cache/hosted/pub.dev/build_config-1.2.0/lib/src/builder_definition.dart').
Try correcting the name to the name of an existing getter, or defining a getter or field named 'buildTo'.
      hideOutput = builderDefinition.buildTo == build_config.BuildTo.cache,
```

## Expected behavior

`build_runner` should bootstrap and run builders successfully for this dependency set, or fail with a clear version-compatibility constraint message.

## Steps to reproduce

1. Use a project with `build_runner: 2.12.0` and the generators listed above.
2. Run: `dart run build_runner build --delete-conflicting-outputs --force-aot`
3. Observe the compile failure above.

davidmorgan:
Fix #4380 

`build_config 1.3.0` added a field, `build_runner` uses it so can no longer compile with `1.2.0`.

I retracted that version of the package as it will break if there is an upper bound forcing build_config 1.2.0 somewhere.

One option would be to add a min version to `build_runner` for `build_config 1.3.0`, but I prefer to use an extension method stub to keep compatibility with both. Tested by forcing build_config version to 1.2.0 and 1.3.0 and running tests.

Longer term ... in particular, if we make any big changes to config ... I'd like something less brittle.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 2.12.1
+
+- Bug fix: allow compilation with older `build_config 1.2.0`.
+
 ## 2.12.0
 
 - Add support for globs in the package list of workspace `pubspec.yaml`,
--- a/build_runner/lib/src/build_plan/builder_definition.dart
+++ b/build_runner/lib/src/build_plan/builder_definition.dart
@@ -218,3 +218,11 @@ class PostProcessBuilderDefinition implements AbstractBuilderDefinition {
       hideOutput = builderDefinition.buildTo == build_config.BuildTo.cache,
       targetBuilderConfigDefaults = builderDefinition.defaults;
 }
+
+/// Stubs for new `build_config` methods to allow compiling with 1.2.0.
+extension _PostProcessBuilderDefinitionStubs
+    on build_config.PostProcessBuilderDefinition {
+  // Added in `1.3.0`.
+  // ignore: unused_element
+  build_config.BuildTo? get buildTo => null;
+}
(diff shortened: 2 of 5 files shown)


davidmorgan:
Hmmm actually let me rethink this :)

davidmorgan:
Going to add a version constraint instead.

PR #4383, opened 2026-03-04
Set `isWindows` to false by default in DDC's bootstrapper for DDC-compiled apps

Markzipan:
This was causing library maps in DDC's library loader to have both `/` and `\`, invalidating lookups on Windows. We could multiplex between platforms, but sticking to `/` is easier.

Related: https://github.com/dart-lang/sdk/issues/62824

### DIFF
--- a/build_runner/test/integration_tests/watch_command_test.dart
+++ b/build_runner/test/integration_tests/watch_command_test.dart
@@ -11,7 +11,7 @@ import 'package:test/test.dart';
 import '../common/common.dart';
 
 void main() async {
-  test('watch command', () async {
+  test('watch command', timeout: const Timeout.factor(4), () async {
     final pubspecs = await Pubspecs.load();
     final tester = BuildRunnerTester(pubspecs);
 


build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.5.0
+- Setting 'isWindows' to false in DDC's bootstrapper to ensure that cross-platform path aren't interlaced.
```
davidmorgan: This looks like more of a bug fix than a user-visible feature? If so then `4.4.15`?

Or if it's user visible, the user-visible effect should be explained here :)

Markzipan: It shouldn't be overtly user-visible, but it's possible users of build_runner can see the side effects of this. Very unlikely though, so I can update the ver num

build_web_compilers/CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 4.4.15
+- Setting 'isWindows' to false in DDC's bootstrapper to ensure that cross-platform path aren't interlaced.
```
nshahan: I don't understand this entry. Would something like the following be accurate? It feels more focused on what the observable change the end use could expect.

`- Normalize URIs in DDC's bootstrapper on windows to always use forward slashes when running with the new library bundle module system.`

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
```diff
@@ -170,7 +169,9 @@ $librariesString
           'packages/build_web_compilers/src/dev_compiler/ddc_module_loader.js',
       mainBoostrapperUrl: bootstrapModuleName,
       mapperUrl: stackTraceMapperPath,
-      isWindows: Platform.isWindows,
+      // Ensure that `isWindows` is false to avoid DDC's bootstrapper appending
+      // backlashes to paths.
+      isWindows: false,
```
nshahan: Do you see this as being useful in the future? Maybe we should remove this required parameter and push this comment into the body where it would be used?

build_web_compilers/CHANGELOG.md
Markzipan: Oooh yes, I like that more - updated

build_web_compilers/lib/src/dev_compiler_bootstrap.dart
Markzipan: True - updated. We should keep this false going forwards - perhaps removing it from the module loader unless we have platform-specific hooks.

Issue #4302, opened 2025-12-04
build_web_compilers does not respect Dart language version from pubspec.yaml


greglittlefield-wf:
In packages that use a language version lower than the current SDK, via SDK constraints in pubspec.yaml, I've noticed compilation errors that seem to indicate that they're instead compiled with a newer language version.

- Dart versions I tried, which reproduced the issue: 3.10.3, 3.9.0
- build_web_compilers versions affected: >4.4.0 (more info at the bottom of the issue)

For example, given the following reduced test case that has a package language version of 2.19:
- pubspec.yaml
    ```yaml
    name: language_version_repro
    environment:
      # Lower bound sets a language version of 2.19
      sdk: '>=2.19.0 <4.0.0'
    dev_dependencies:
      build_web_compilers: ^4.4.5
    ```

- web/needs_less_than_3.0.dart
    ```dart
    class MixinClass  {
      foo() => 'foo';
    }
    
    // 3.0 no longer allows mixing in classes: https://dart.dev/resources/language/evolution#dart-3-0
    class B extends Object with MixinClass {}
    
    main() => print(B().foo());
    ```

- web/needs_less_than_3.7.dart
    ```dart
    // 3.7 no longer allows referencing `_`: https://dart.dev/resources/language/evolution#dart-3-7
    void function(dynamic _) => print(_);
    
    void main() => function('foo');
    ```

The build fails with the following errors:
```console
$ dart run build_runner build web
…
log output for build_web_compilers:ddc (lazy) on web/needs_less_than_3.0.ddc.module                                                              
E Error compiling dartdevc module:language_version_repro|web/needs_less_than_3.0.ddc.js                                                          
  web/needs_less_than_3.0.dart:5:29: Error: The class 'MixinClass' can't be used as a mixin because it isn't a mixin class nor a mixin.          
  class B extends Object with MixinClass {}                                                                                                      
                              ^               
…
log output for build_web_compilers:ddc (lazy) on web/needs_less_than_3.7.ddc.module                                                              
E Error compiling dartdevc module:language_version_repro|web/needs_less_than_3.7.ddc.js                                                          
  web/needs_less_than_3.7.dart:1:35: Error: Undefined name '_'.                                                                                  
  void function(dynamic _) => print(_);                                                                                                          
                                    ^     
…
```
and it also fails with similar errors when compiling in dart2js, via `--release`.

[...]

PR #4387, opened 2026-03-04
Updating synthetic package_config.json generation directory

Markzipan:
This was causing files in `web/` and other non-package roots from following versions in the generated `package_config.json`. 

Fixes https://github.com/dart-lang/build/issues/4302


davidmorgan (approved):
Thanks :)

PR #4392, opened 2026-03-09
Use asset graph in AnalysisDriverModel

davidmorgan:
It turns out there was still significant time spent walking deps graphs to find files to update in the analyzer's view of the filesystem. Simplify by sharing the full knowledge of the build, the asset graph, with `AnalysisDriverModel`.

On 5000 file "random" benchmark, AOT compiled, this reduces the overall time spent in `AnalysisDriverModel.updateDriver` from 30% to 0.37%, reducing initial build time from 36s to 22s.

Perf diff (internal only) go/perf-diff-4392

The difference is less for smaller projects: on the 1000 file benchmark it drops from 8% to 0.38%.

### DIFF
--- a/build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
+++ b/build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
@@ -48,11 +48,14 @@ class AnalysisDriverFilesystem implements UriResolver, ResourceProvider {
   /// Records changes due to the phase change in [changedPaths].
   set phase(int phase) {
     if (phase == _phase) return;
-    final oldPhase = _phase;
+    final previousPhase = _phase;
     _phase = phase;
 
     for (final entry in _pathByPhase.entries) {
-      if (oldPhase > entry.key != phase > entry.key) {
+      final previouslyWasVisible = previousPhase > entry.key;
+      final isVisible = phase > entry.key;
+
+      if (previouslyWasVisible != isVisible) {
         _changedPaths.addAll(entry.value);
       }
     }
@@ -85,7 +88,7 @@ class AnalysisDriverFilesystem implements UriResolver, ResourceProvider {
   ///
   /// Throws if ![exists].
   String read(String path) {
-    if (!exists(path)) throw StateError('!exists');
+    if (!exists(path)) throw StateError('Read of non-existent file.');
     return _data[path]!;
   }
 


build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -12,44 +12,105 @@ import 'package:analyzer/src/clients/build_resolvers/build_resolvers.dart';
...
+  /// A generated file is only visible if it was generated at an earlier phase.
+  ///
+  /// Records changes due to the phase change in [changedPaths].
+  set phase(int phase) {
+    if (phase == _phase) return;
+    final oldPhase = _phase;
+    _phase = phase;
+
+    for (final entry in _pathByPhase.entries) {
+      if (oldPhase > entry.key != phase > entry.key) {
```
jensjoha: Could you pull out each side as variables with a good name? This isn't exactly easy to read.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -12,44 +12,105 @@ import 'package:analyzer/src/clients/build_resolvers/build_resolvers.dart';
...
-
-  /// Deletes the data previously written to [path].
-  ///
-  /// Records the change in [changedPaths].
-  ///
-  /// Or, if it's missing, does nothing.
-  void deleteFile(String path) {
-    if (_data.remove(path) != null) _changedPaths.add(path);
+  String read(String path) {
+    if (!exists(path)) throw StateError('!exists');
```
jensjoha: Maybe a better error description?

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -12,44 +12,105 @@ import 'package:analyzer/src/clients/build_resolvers/build_resolvers.dart';
...
-    final oldContent = _data[path];
-    _data[path] = content;
-    if (content != oldContent) _changedPaths.add(path);
+    var wasAbsent = false;
+    final updatedContent = _data.putIfAbsent(path, () {
+      wasAbsent = true;
+      return content;
+    });
+    if (wasAbsent) {
+      if (_phase > _phaseOf(path)) {
```
jensjoha, line 106: Is it okay to write a file we don't know the phase of?

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
davidmorgan: Done.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
davidmorgan: Done.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
davidmorgan: Yes, if it doesn't have a phase then it's not a generated file, it's a source file.

davidmorgan:
Whoops, apparently I forgot the correct review changes workflow and force pushed. But my jj-foo is now sufficient to recover :) separated into two commits.

PR #4864, opened 2026-03-25
Add OSC 8 hyperlinks for logged input paths

### LINKED ISSUE #4295: emit OSC 8 terminal hyperlinks for file paths and error locations
### Feature Request: Add OSC 8 Hyperlink Support

`build_runner` currently prints file paths and error locations (both relative paths and `package:` URIs) as plain text.

[Modern terminal emulators and editors (including JetBrains + VS Code) support **OSC 8 hyperlinks**.](https://github.com/Alhadis/OSC8-Adoption/) These make file paths and error locations clickable, instantly opening the file (optionally, at an exact line and column.)

Adding this to `build_runner` would improve developer productivity.

### Official Specification & References

1. **Primary specification** – Xterm control sequences (updated June 2025)  
   https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Operating-System-Commands

2. **Detailed practical documentation** – mintty wiki  
   https://github.com/mintty/mintty/wiki/CtrlSeqs#hyperlinks

### Implementation Notes

- The Dart SDK already has reusable ANSI utilities (detect ANSI support via `stdout.supportsAnsiEscapes`). Optionally add a `--no-hyperlinks` flag for CI environments if necessary.
- Resolve the hyperlink target to an **absolute** `file://` URI (required for reliable opening in all terminals/IDEs):
  - **Relative paths** (e.g. `lib/src/utils.dart:45`) → resolve against the project root using the `path` package or `Directory.current`.
  - **package: URIs** (e.g. `package:my_app/utils.dart:120`) → resolve via `package_config` or `Uri.base.resolve()` to the real on-disk location (local package or `.pub-cache`).
- Keep the **display text exactly as currently printed** (relative or package: form) — only the invisible hyperlink target changes.

Example:
```
lib/src/utils.dart:45:8      → clickable → file:///Users/me/project/lib/src/utils.dart:45:8
package:my_app/foo.dart:23   → clickable → file:///Users/me/.pub-cache/hosted/pub.dev/my_app-1.2.3/lib/foo.dart:23
```

Thank you for considering this enhancement!

kyungilcho:
## Summary
- add OSC 8 hyperlinks for build phase progress paths
- add OSC 8 hyperlinks for builder log context paths
- keep the scope to `file://` links for logged file paths only

## Testing
- /Users/jogyeong-il/Desktop/dev/flutter/bin/cache/dart-sdk/bin/dart test build_runner/test/logging/ansi_buffer_test.dart build_runner/test/logging/build_log_console_mode_test.dart build_runner/test/logging/build_log_line_mode_test.dart build_runner/test/logging/build_log_messages_test.dart

Fixes #4295


build_runner/lib/src/logging/build_log.dart
```diff
@@ -659,6 +684,78 @@ class BuildLog {
...
+    }
+
+    try {
+      return AssetId(singleOutputPackage, context);
+    } on ArgumentError {
+      return null;
+    }
+  }
+
+  bool _looksLikeAssetPath(String value) =>
```
davidmorgan: The `context` is passed in via `loggerFor` or `loggerForOther`, so actually we _know_ whether it's an `AssetId` or not but throw away that information.

So: I suggest making `BuildLogLogger` explicitly take an `AssetId` for context, with a backup `String` context for cases where there is no `AssetId` to pass.

build_runner/lib/src/logging/build_log.dart
```diff
@@ -659,6 +684,78 @@ class BuildLog {
...
+      !value.contains(' ') &&
+      (value.contains('/') || p.extension(value).isNotEmpty);
+
+  Uri? _targetUriForAssetId(AssetId id) {
+    final packageRoot =
+        (_packageRootsByName ??= _loadPackageRoots())[id.package];
+    return packageRoot?.resolve(id.path);
+  }
+
+  Map<String, Uri> _loadPackageRoots() {
```
davidmorgan: Instead of loading the package roots here, they can be passed in.

The class `BuildPackages` is how they're usually handled, it has a method `pathFor` that converts an `AssetId` to a path.

So I think `BuildLog` should have a `BuildPackages? buildPackages` field. It can be set in the `startPhases` method, which is only called in one place which already has a `buildPackages`, and then cleared in `finishBuild`. The hyperlink rendering should only happen if there is non-null `buildPackages`.

How does that sound?

davidmorgan (commented):
Looking good so far, thanks! A couple of suggestions, please see what you think.

build_runner/lib/src/logging/build_log.dart
kyungilcho: Yep, this is cleaner. I removed the package-config lookup and now pass `BuildPackages` into `BuildLog` from `startPhases`, then use `pathFor` when building the `file://` link. Hyperlink rendering now only happens when `buildPackages` is available.

build_runner/lib/src/logging/build_log.dart
kyungilcho: That makes sense. I changed `BuildLogLogger` to carry an `AssetId? contextId`, and `BuildLogMessages` now preserves that instead of trying to recover it from the rendered string. Plain string context is still used in the cases where there isn't an `AssetId`.

build_runner/lib/src/logging/ansi_buffer.dart
```diff
@@ -15,9 +15,20 @@ class AnsiBuffer {
   static const reset = '\x1B[0m';
   static const bold = '\x1B[1m';
   static const boldRed = '\x1B[1;31m';
+  static const _hyperlinkPrefix = '\x1B]8;;';
+  static const _hyperlinkTerminator = '\x1B\\';
+  static final RegExp _hyperlinkPattern = RegExp(
+    '${RegExp.escape(_hyperlinkPrefix)}[^\x1B]*'
```
davidmorgan: I'm not sure this regexp is correct, doesn't [^\x1B]* mean "characters other than \x, 1, B"? In which case it won't match if your hyperlink has a 1 in it :)

How about: `${Regexp.escape(_hyperlinkPrefix)}.*?${Regexp.escape(_hyperlinkTerminator)}`?

build_runner/lib/src/logging/ansi_buffer.dart
```diff
@@ -15,9 +15,20 @@ class AnsiBuffer {
...
+    '${RegExp.escape(_hyperlinkTerminator)}',
+  );
   static const _nbspCodeUnit = 0xA0;
   static const _spaceCodeUnit = 32;
 
+  static String openHyperlink(Uri uri) =>
+      '$_hyperlinkPrefix${uri.toString()}$_hyperlinkTerminator';
```
davidmorgan: No need to "toString" in a string interpolation, can be just `$uri`.

build_runner/lib/src/logging/ansi_buffer.dart
```diff
@@ -173,7 +187,8 @@ class AnsiBufferLine {
 bool _isAnsi(String item) =>
     item == AnsiBuffer.reset ||
     item == AnsiBuffer.bold ||
-    item == AnsiBuffer.boldRed;
+    item == AnsiBuffer.boldRed ||
+    AnsiBuffer._hyperlinkPattern.hasMatch(item);
```
davidmorgan: This is an easier match than the one in `removeAnsi` since it must match the whole of `item`. It doesn't need to search. You could use `matchAsPrefix`.

But how about instead

```
||
(item.startsWith(_hyperlinkPrefix) && item.endsWith(_hyperlinkTerminator))
```

?

build_runner/lib/src/logging/build_log.dart
```diff
@@ -104,6 +105,8 @@ class BuildLog {
   /// For `watch` and `serve` modes, the build number.
   int _buildNumber = 1;
 
+  BuildPackages? buildPackages;
```
davidmorgan, line 109: Please add comment

/// For looking up asset paths when rendering asset IDs.

build_runner/lib/src/logging/build_log.dart
```diff
@@ -659,6 +691,15 @@ class BuildLog {
       'https://github.com/dart-lang/build/discussions/4349',
     );
   }
+
+  Uri? _targetUriForAssetId(AssetId id) {
```
davidmorgan: This is only used by `renderLinkedId`, I suggest moving it next to that method.

build_runner/lib/src/logging/build_log.dart
```diff
@@ -659,6 +691,15 @@ class BuildLog {
...
+    try {
+      final path = buildPackages?.pathFor(id, hide: false);
+      return path == null ? null : Uri.file(path);
+    } on Exception {
```
davidmorgan: I think the only exception that needs catching here is PackageNotFoundException, which could happen if `buildPackages` gets set multiple times due to running multiple builds concurrently ... only possible in tests or if using `build_test` from a builder.

So I suggest `on PackageNotFoundException`.

davidmorgan (commented):
Looking good, thanks! Some minor suggestions.

Could you please also add an entry to build_runner/CHANGELOG.md

```
## 2.14.0-wip

- Add OSC 8 hyperlinks for logged input paths.
```

and change the version in build_runner/pubspec.yaml to `2.14.0-wip` to match.

And, the presubmit shows an analyzer error about sorting directives, please sort+format as needed :)

Thanks!

build_runner/lib/src/logging/ansi_buffer.dart
lrhn: It's correct (in that regard, haven't checked in detail), RegExp character classes can contain character escapes.

This one actually doesn't, because the string is not raw, so the RegExp Source contains the literal 0x1B byte. But that's OK, it still means `any character not == U+001B`.

Using `.*?` seems a little unsafe. If the `\x1b]8;;` occurs in the text without a matching
end, then `.*?` may match until the end of a later wrapped hyperlink.

Using `.` doesn't match line terminators. 
Do we _want_ to allow line terminators in hyperlinks?
If not, maybe use `r'[^\x1b\P{Zl}]*'` instead (and a unicode regexp).

lrhn: Looking more: The `RegExp.escape` should make the pre-post patterns safe.
I'd personally split the string into raw and non-raw parts, so everything that can be raw, is.
```dart
  ... = RegExp( 
    '${RegExp.escape(_hyperlinkPrefix)}' r'[^\x1b]*' '${RegExp.escape(_hyperlinkSuffix)}',
  );
```
It's sooo easy to forget that a string isn't raw when editing a RegExp, even if you knew it when you wrote it, so I prefer to make everything that isn't an interpolation be a raw string.
(Can get messy, though.)

davidmorgan: Thanks! That makes sense, I had incorrectly read the `\x1b` as three characters, of course it's a single U+001B as you say.

Then: the original was correct. I like Lasse's suggestion to split into three strings for clarity, please follow that unless you would prefer to keep the current form :) thanks!

build_runner/lib/src/logging/ansi_buffer.dart
lrhn: That's different from matching the regexp.

If you have two hyperlinks concatenated, then that will match the start of one and the end of the oher.

davidmorgan: They can't be concatenated; we take care to keep "rendered items" separated, that's why `item == AnsiBuffer.bold` and so on are correct.

The codepath that removes ansi codes from arbitrary strings is only used in testing, so the tests can render with ansi codes turned on then make assertions about the output ignoring ansi codes.

kyungilcho:
Thanks, I pushed an update addressing these suggestions, including the changelog/version updates, the analyzer cleanup, and the follow-up changes in `ansi_buffer.dart` and `build_log.dart`.

I also updated the `build_test` workspace dependency so the workspace resolves cleanly.

kevmoo:
I asked Gemini about whe/how to enable this...

1.  **Check `FORCE_HYPERLINK`:** If it's `1`, enable them.
2.  **Check `isatty`:** If `STDOUT` is not a terminal, disable them.
3.  **Check `TERM`:** If it’s `dumb`, disable them.
4.  **Check `DOMINANT_TERM_VARS`:** If `VTE_VERSION` is present or `TERM_PROGRAM` matches a known compatible terminal, enable them.
5.  **Default:** Usually, it's safest to default to **OFF** unless you are sure, or provide a `--hyperlink=auto/always/never` flag for the user.

davidmorgan:
Looks good, thanks!

Per Kevin's comment: we already disable ansi codes altogether if there isn't a terminal.

It seems support for OSC links is pretty widespread https://github.com/Alhadis/OSC8-Adoption/ my preference is to just turn it on with the other ansi codes, we can add support for configuring later if there is feedback that it's needed.

So I'll go ahead and merge it.

Not sure when the next release will be, but I guess within a month :)

davidmorgan (approved):

davidmorgan:
@kyungilcho released in 2.14 :)

PR #4874, opened 2026-04-03
build_runner: add locking to prevent multiple build_runners

kevmoo:
Avoids a lot of ugly broken things

### DIFF
--- a/build_runner/test/integration_tests/watch_command_test.dart
+++ b/build_runner/test/integration_tests/watch_command_test.dart
@@ -43,18 +43,14 @@ void main() async {
     tester.write('root_pkg/web/a.txt', 'updated');
     await watch.expect('wrote 0 outputs');
 
-    // State on disk is updated so `build` knows to do nothing.
-    var output = await tester.run('root_pkg', 'dart run build_runner build');
-    expect(output, contains('wrote 0 outputs'));
+    // State on disk is updated by watch.
 
     // New file.
     tester.write('root_pkg/web/b.txt', 'b');
     await watch.expect(BuildLog.successPattern);
     expect(tester.read('root_pkg/web/b.txt.copy'), 'b');
 
-    // State on disk is updated so `build` knows to do nothing.
-    output = await tester.run('root_pkg', 'dart run build_runner build');
-    expect(output, contains('wrote 0 outputs'));
+    // State on disk is updated by watch.
 
     // Deleted file.
     tester.delete('root_pkg/web/b.txt');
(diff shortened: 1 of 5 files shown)


build_runner/test/integration_tests/watch_command_test.dart
```diff
@@ -43,18 +43,14 @@ void main() async {
     tester.write('root_pkg/web/a.txt', 'updated');
     await watch.expect('wrote 0 outputs');
 
-    // State on disk is updated so `build` knows to do nothing.
```
kevmoo, line 46: interleaving is now forbidden...which is good...right?

davidmorgan:
Thanks! It's kinda surprising build_runner has been around for so long without this, it definitely seems worth having.

Some thoughts :)

I suggest moving the implementation out of `build_runner.dart` and into ... how about `bootstrap/lock.dart`?

For correctness, it will need to take into account workspaces. Within a workspace, build_runner can either run in workspace mode, in which case it wants to lock the whole workspace, or in package mode, in which case it only needs to lock a single package. When running in a workspace only the workspace `.dart_tool` is used, when running in an individual package it's that package's `dart_tool`.

Regarding usability, I think how I would want it to work is

 - if there is a one-shot command already running like `build`, it waits for the other command to complete then runs
 - if there is a continuous command running like `watch` or `serve` it tells that command to exit, immediately if there is no build running, or after the next build; then runs
 
so it will always be able to do something. I guess this might be not too hard to do with a "request exit" file next to the lock file. `watch` and `serve` are already looking for file changes so they can notice it and exit. The file can be deleted if it exists after the lock is acquired, and recreated if it doesn't exist while waiting for the lock.

For extra fun we could have an `--interrupt` flag that tries to terminate an already-running build_runner process instead of waiting for it. We can also have `build run build_runner stop` and `build run build_runner stop --interrupt` for completeness :) ... these could wait for later though.

The "watch" tests that were checking state by running a build alongside a watch are a bit awkward :) I guess they can do the same check by copying the entire folder and running the build inside the copy?

Re: releasing the lock before deleting the folder, that sounds like it might lead to a race--I guess it should also work to delete everything except the lock, then the lock will be released on exit?

rrousselGit:
We'll still be able to run multiple build_runner if they aren't in the same dir, right?

davidmorgan:
> We'll still be able to run multiple build_runner if they aren't in the same dir, right?

Yes, the lock should be by package, or by workspace if `--workspace` is used.

kevmoo:
@davidmorgan – game to make it more fancy in the future. Now I have a `build` crashing through my `serve` which crashes my next request.

Do you like this as-is for a starting spot?

davidmorgan:
> @davidmorgan – game to make it more fancy in the future. Now I have a `build` crashing through my `serve` which crashes my next request.
> 
> Do you like this as-is for a starting spot?



> @davidmorgan – game to make it more fancy in the future. Now I have a `build` crashing through my `serve` which crashes my next request.
> 
> Do you like this as-is for a starting spot?

Since it can block people and affects basic interaction with the tool I think it's worth aiming quite high to release: workspace support and the "waits until exit, causes existing process to exit". I don't want to have the possibility of showing a "something else has the lock" failure, we should figure out how to always make progress.

You have my interest now :) it would make the whole build much more well defined to not step on other `build_runner`s, it seems like a basic usability feature. So I could take it from here if you like, I am maybe optimistically thinking it's probably possible to wrap it up in a day :) which seems well worth it.

davidmorgan:
https://github.com/dart-lang/build/pull/4879 is a work in progress PR adding the fancy version :)

"Work in progress" in the sense that it basically works but needs cleanup and some checking.

PR #4876, opened 2026-04-06
Retain generated analyzer filesystem contents across incremental builds

777genius:
This keeps generated file contents in `AnalysisDriverFilesystem` across incremental builds instead of clearing them unconditionally on every `startBuild()`.

Refs #4875.

Today `startBuild()` clears all cached file contents. In long-lived watch sessions that means generated files that still exist in the next asset graph are removed and then re-added to the analyzer-facing filesystem on every incremental build.

This patch narrows that behavior:

- generated files that are still present in the next asset graph are retained,
- generated files are removed only if they actually disappear,
- source file handling stays unchanged,
- generated paths are still reported as changed when:
  - file content changes,
  - visibility changes because of phase transitions,
  - a generated path disappears.

This is intended as a watch / incremental performance optimization only. It does not change the broader build planning model or generated output semantics.

Tests added:

- non-generated files are still cleared at `startBuild()`,
- disappeared generated files are removed and reported as changed,
- unchanged generated files survive `startBuild()`,
- rewriting a generated file with different content across builds reports the change,
- visibility changes for retained generated files are still reported correctly.

### DIFF
--- a/build_runner/lib/src/build/build.dart
+++ b/build_runner/lib/src/build/build.dart
@@ -49,6 +49,23 @@ final ResolversImpl _defaultResolvers = ResolversImpl(
   analysisDriverModel: AnalysisDriverModel(),
 );
 
+class _AssetGraphSourceUpdates {
+  final bool isInitialBuild;
+  final Set<AssetId> changedSources;
+  final Set<AssetId> deletedSources;
+
+  const _AssetGraphSourceUpdates({
+    required this.isInitialBuild,
+    required this.changedSources,
+    required this.deletedSources,
+  });
+
+  const _AssetGraphSourceUpdates.initial()
+    : isInitialBuild = true,
+      changedSources = const {},
+      deletedSources = const {};
+}
+
 /// A single build.
 class Build {
   final BuildPlan buildPlan;
@@ -220,16 +237,23 @@ class Build {
     return result;
   }
 
-  Future<void> _updateAssetGraph(Set<AssetId> updates) async {
+  _AssetGraphSourceUpdates _initialSourceUpdates() =>
+      const _AssetGraphSourceUpdates.initial();
+
+  Future<_AssetGraphSourceUpdates> _updateAssetGraph(
+    Set<AssetId> updates,
+  ) async {
     changedInputs.clear();
     deletedAssets.clear();
 
     // Check what actually changed for each asset in `updates`.
     readerWriter.cache.invalidate(updates);
     final resolvedUpdates = <AssetId, ChangeType>{};
+    final previousSourceNodes = <AssetId, bool>{};
     final newDigests = <AssetId, Digest>{};
     for (final id in updates) {
       final oldNode = assetGraph.get(id);
+      previousSourceNodes[id] = oldNode?.type == NodeType.source;
       final oldExisted =
           oldNode != null && oldNode.type != NodeType.missingSource;
       final oldDigest = oldNode?.digest;
@@ -280,6 +304,29 @@ class Build {
         nodeBuilder.digest = entry.value;
       });
     }
+
+    final changedSources = <AssetId>{};
+    final deletedSources = <AssetId>{};
+    for (final entry in resolvedUpdates.entries) {
+      final id = entry.key;
+      final changeType = entry.value;
+      if (changeType == ChangeType.REMOVE) {
+        if (previousSourceNodes[id] == true) {
+          deletedSources.add(id);
+        }
+        continue;
+      }
+      final updatedNode = assetGraph.get(id);
+      if (updatedNode?.type == NodeType.source) {
+        changedSources.add(id);
+      }
+    }
+
+    return _AssetGraphSourceUpdates(
+      isInitialBuild: false,
+      changedSources: changedSources,
+      deletedSources: deletedSources,
+    );
   }
 
   /// Runs a build inside a zone with an error handler and stack chain
@@ -288,9 +335,10 @@ class Build {
     final done = Completer<BuildResult>();
     runZonedGuarded(
       () async {
-        if (!assetGraph.cleanBuild) {
-          await _updateAssetGraph(idsToCheck);
-        }
+        final sourceUpdates =
+            assetGraph.cleanBuild
+                ? _initialSourceUpdates()
+                : await _updateAssetGraph(idsToCheck);
         for (final id in assetGraph.sources) {
           final node = assetGraph.get(id)!;
           if (node.digest == null && node.primaryOutputs.isNotEmpty) {
@@ -300,7 +348,12 @@ class Build {
             });
           }
         }
-        await resolversImpl?.takeLockAndStartBuild(assetGraph);
+        await resolversImpl?.takeLockAndStartBuild(
+          assetGraph,
+          isInitialBuild: sourceUpdates.isInitialBuild,
+          changedSources: sourceUpdates.changedSources,
+          deletedSources: sourceUpdates.deletedSources,
+        );
         final result = await _runPhases();
 
         assetGraph.previousBuildTriggersDigest =
--- a/build_runner/lib/src/build/build.dart
+++ b/build_runner/lib/src/build/build.dart
@@ -49,23 +49,6 @@ final ResolversImpl _defaultResolvers = ResolversImpl(
   analysisDriverModel: AnalysisDriverModel(),
 );
 
-class _AssetGraphSourceUpdates {
-  final bool isInitialBuild;
-  final Set<AssetId> changedSources;
-  final Set<AssetId> deletedSources;
-
-  const _AssetGraphSourceUpdates({
-    required this.isInitialBuild,
-    required this.changedSources,
-    required this.deletedSources,
-  });
-
-  const _AssetGraphSourceUpdates.initial()
-    : isInitialBuild = true,
-      changedSources = const {},
-      deletedSources = const {};
-}
-
 /// A single build.
 class Build {
   final BuildPlan buildPlan;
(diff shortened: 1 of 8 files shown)


build_runner/lib/src/build/build.dart
```diff
@@ -49,6 +49,23 @@ final ResolversImpl _defaultResolvers = ResolversImpl(
   analysisDriverModel: AnalysisDriverModel(),
 );
 
+class _AssetGraphSourceUpdates {
+  final bool isInitialBuild;
+  final Set<AssetId> changedSources;
```
davidmorgan: I think `changedSources` and `deletedSources` end up being handled the same? In both cases the filesystem just removes the source.

So they could be combined into one `invalidatedSources`.

If you like, you could use `Set<AssetId>? invalidatedSources` and have `null` mean "initial build". Or, keep a separate bool for that if you prefer.

build_runner/lib/src/build/build.dart
```diff
@@ -220,16 +237,23 @@ class Build {
...
+  Future<_AssetGraphSourceUpdates> _updateAssetGraph(
+    Set<AssetId> updates,
+  ) async {
     changedInputs.clear();
     deletedAssets.clear();
 
     // Check what actually changed for each asset in `updates`.
     readerWriter.cache.invalidate(updates);
     final resolvedUpdates = <AssetId, ChangeType>{};
+    final previousSourceNodes = <AssetId, bool>{};
```
davidmorgan: I think this can be a set <AssetId>{}, only add to the set if it's a source node?

build_runner/lib/src/build/build.dart
```diff
@@ -280,6 +304,29 @@ class Build {
...
+    final deletedSources = <AssetId>{};
+    for (final entry in resolvedUpdates.entries) {
+      final id = entry.key;
+      final changeType = entry.value;
+      if (changeType == ChangeType.REMOVE) {
+        if (previousSourceNodes[id] == true) {
+          deletedSources.add(id);
+        }
+        continue;
+      }
```
davidmorgan: I don't think we need to do anything for ChangeType.ADD, so instead of the `continue` this can be

```
else if (changeType == ChangeType.MODIFY) {
...
}
```

build_runner/lib/src/build/build.dart
```diff
@@ -280,6 +304,29 @@ class Build {
...
+      final updatedNode = assetGraph.get(id);
+      if (updatedNode?.type == NodeType.source) {
```
davidmorgan: I don't think it's necessary to check the new node type, whatever the new type we anyway want to clear out the old one.

build_runner/lib/src/build/resolver/analysis_driver_model.dart
```diff
@@ -46,8 +46,23 @@ class AnalysisDriverModel {
...
   /// If another build has the lock, waits for it to finish.
-  Future<void> takeLockAndStartBuild(AssetGraph assetGraph) async {
+  Future<void> takeLockAndStartBuild(
+    AssetGraph assetGraph, {
+    required bool isInitialBuild,
+    required Set<AssetId> changedSources,
+    required Set<AssetId> deletedSources,
+  }) async {
     _lock = await _pool.request();
+    if (isInitialBuild) {
```
davidmorgan: I suggest moving this logic into `filesystem.startBuild`, i.e. just pass everything through to it.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -59,23 +60,79 @@ class AnalysisDriverFilesystem
       final isVisible = phase > entry.key;
 
       if (previouslyWasVisible != isVisible) {
-        _changedPaths.addAll(entry.value);
+        for (final path in entry.value) {
+          if (_data.containsKey(path)) {
```
davidmorgan: This check is needed because _pathByPhase might contain sources that were removed?

I think it would be better to remove from _pathByPhase at the same time, then, so this check is not needed :)

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -59,23 +60,79 @@ class AnalysisDriverFilesystem
...
   }
 
   /// Initializes a new filesystem that will have files added due to the
   /// build described by [generatedNodes].
   void startBuild(Iterable<AssetNode> generatedNodes) {
-    _changedPaths.addAll(_data.keys);
-    _data.clear();
-    _phaseByPath.clear();
-    _pathByPhase.clear();
+    final previousPhaseByPath = Map<String, int>.from(_phaseByPath);
```
davidmorgan: Better than copying these would be to reassign, by making _phaseByPath non-final. Then you can do

final previousPhaseByPath = _phaseByPath;
_phaseByPath = <String, int>{};

and use `_phaseByPath` instead of a new variable `nextPhaseByPath`.

Similarly for _pathByPhase.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -59,23 +60,79 @@ class AnalysisDriverFilesystem
...
       final phase = node.generatedNodeConfiguration!.phaseNumber;
       final idAsPath = node.id.asPath;
-      _phaseByPath[idAsPath] = phase;
-      _pathByPhase.putIfAbsent(phase, () => []).add(idAsPath);
+      nextPhaseByPath[idAsPath] = phase;
+      nextPathByPhase.putIfAbsent(phase, () => []).add(idAsPath);
+    }
+
+    final removedGeneratedPaths =
+        previousPhaseByPath.keys.toSet()..removeAll(nextPhaseByPath.keys);
```
davidmorgan: Instead of doing the set operations how about

```
// Handle generated files that were removed.
for (final path in previousPhaseByPath.keys) {
  if (_phaseByPath.containsKey(path)) continue;

   ... handle remove ...
}
```

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -59,23 +60,79 @@ class AnalysisDriverFilesystem
...
+    for (final path in removedGeneratedPaths) {
+      final previousPhase = previousPhaseByPath[path]!;
+      final wasVisible = _phase > previousPhase;
+      if (_data.remove(path) != null && wasVisible) {
+        _changedPaths.add(path);
+      }
+    }
+
+    for (final entry in previousPhaseByPath.entries) {
```
davidmorgan, line 103: This could actually be combined with the previous for loop, i.e. we loop over all previous generated files and either remove or update.

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -59,23 +60,79 @@ class AnalysisDriverFilesystem
...
+    if (_phaseByPath.containsKey(path)) {
+      throw ArgumentError.value(path, 'path', 'Must not be a generated path.');
+    }
+    if (_data.remove(path) != null) {
+      _changedPaths.add(path);
+    }
+  }
+
+  /// Clears all source file contents from the in-memory filesystem.
+  void clearSourcePaths() {
```
davidmorgan: For the "initial build" case I suggest to just clear everything, as it was before: otherwise we could end up with generated data stored that has no connection at all to the current build.

davidmorgan (commented):
Thanks! I checked and I agree, this saves a lot of time on the 5000 file "watch" rebuild :)

I made a pile of minor suggestions to minimize/simplify, please take a look.

Could you please add a CHANGELOG entry, something like

```
 - Performance: further improvements to management of files for analysis
   for 2x faster incremental builds.
```

Thanks :)

777genius:
I really appreciate the detailed review. I went through the simplification suggestions and updated the PR accordingly.

The current version now:
- uses a single `invalidatedSources` set
- moves source invalidation handling into `AnalysisDriverFilesystem.startBuild`
- restores full-clear semantics for the initial build
- keeps the incremental invalidation path focused on the analyzer filesystem layer
- adds the requested CHANGELOG entry

After the cleanup, I reran the synthetic benchmark on `_benchmark` with `built_value`, `random`, `size=5000`.

For clean `build`, the branch is now basically at parity with `master` across 3 cold runs per branch:
- logged build time median: `25.13s` vs `25.81s`

For `watch`, the main incremental signal is still clearly positive across 2 cold runs per branch with 3 rebuilds per run:
- rebuild duration from the verbose build log: `9.11s` on `master` vs `4.46s` here on average, about `2.04x` faster
- edit-to-rebuild wall time: `9.66s` vs `5.17s` on average, about `1.87x` faster

Initial `watch` startup was a bit slower in this sample (`30.62s` vs `32.84s` on average), while the main incremental metrics remained clearly improved.

I really appreciate the guidance here :)

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -102,19 +146,35 @@ class AnalysisDriverFilesystem
...
+    if (content.contentHash == previousContent.contentHash) {
+      if (!_allCachedContentsAreFromCurrentBuild) {
+        _writtenPathsThisBuild.add(path);
       }
+      return;
+    }
+
+    if (_allCachedContentsAreFromCurrentBuild ||
+        _writtenPathsThisBuild.contains(path)) {
+      throw StateError('Different write to $path.');
```
davidmorgan: Thanks for the updates! Looking great. Just one last point for discussion.

I'm considering whether it would make sense to remove this check altogether.

Originally I added a check for conflicting writes because AnalysisDriverModel was shared. In normal builds it would only be used by one build at a time, but in tests via `package:build_test` it could be used by multiple builds. And so tests would occasionally hit the check and throw.

But in a later PR I added locking to ensure AnalysisDriverModel is only used by one build at a time. In other places we check and ensure that an output is only written once--it would already be a bug if there are multiple writes. So there should be no way to hit this check.

It would simplify the code here quite a bit to remove it.

There is also a middle ground, we could have a check but only when asserts are enabled. A slightly simpler check would be to only check changes, ignoring duplicate writes with the same content. Then it would be just

```
final _changedPathsThisBuild = <String>{};

... clear it on start of build ...

void writeContent(...) {
    if (!content.exists) throw ArgumentError('content must exist');
    final path = content.path;
    final previousContent = _data[path];
    final isVisible = _phase > _phaseOf(path);    
    if (previousContent == null || content.contentHash != previousContent.contentHash) {
      _data[path] = content;
      if (isVisible) {
        _changedPaths.add(path);
      }
      assert(_changedPathsThisBuild.add(path), path);
    }
}
```

What do you think?

Thanks :)

davidmorgan:
@777genius it's been a few days, just wanted to check you saw my comment.

No rush from my side, but looking forward to landing this :)

777genius:
Sorry for the slow follow-up here.

I think the middle ground is the right choice here too, and I just updated the PR in that direction. It now keeps only an `assert` for repeated changed writes and removes the runtime bookkeeping around this check.

That feels like the best tradeoff here now that `AnalysisDriverModel` is locked per build - we still keep the invariant visible in debug/test mode, but the hot path stays simpler and cheaper.

I also added a small test for the duplicate changed-write case.

And thanks for the nudge :)

davidmorgan (approved):
Looks great, thanks!

There are quite a few improvements ready to release, I guess this will go out in the next week or so.

777genius:
It feels great to contribute to such an important tool! Thanks for the direction. Wow, I didn’t expect it to make it into the release that quickly - that’s awesome! 🚀 🚀 🚀

davidmorgan:
@777genius released in 2.14 :)

777genius:
Very good! 🤩 🤩 🤩 🚀🚀🚀

PR #4877, opened 2026-04-08
Remove use of Expression.expression getter

natebosch:
See https://github.com/dart-lang/tools/issues/2371

This `.expression` call was likely always unnecessary and unnoticed. The
`Expression.expression` API is marked with `@visibleForOverriding` and
is not intended for use from other packages, but the diagnostics are
suppressed because of some overrides that are not annotated.

Remove the call to `.expression` in preparation for it to be deprecated
and eventually removed.

### DIFF
--- a/builder_pkgs/mockito/lib/src/builder.dart
+++ b/builder_pkgs/mockito/lib/src/builder.dart
@@ -1453,7 +1453,7 @@ class _MockClassInfo {
             referImported(
               'throwOnMissingStub',
               'package:mockito/mockito.dart',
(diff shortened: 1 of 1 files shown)


davidmorgan (approved):
Thanks :)

Any urgency on publishing / syncing to google3?

natebosch:
> Any urgency on publishing / syncing to google3?

No urgency on publishing.
No significant urgency on sync to google3, I have a few other internal cases that are will need to be fixed too. If it can land in the next week or 2 that works for me.

PR #4879, opened 2026-04-10
Add process lock.

davidmorgan:
Inspired by https://github.com/dart-lang/build/pull/4874.

Introduce `BuildPaths` which is the current package path, the workspace path (if there is one) and whether to build the single package or the workspace.

Add locking to the `build_runner` process so only one command can run at once.

If a `watch` or `serve` has the lock, notify it so it will exit: immediately if no build is running, or after the build completes.

In a workspace, lock the whole workspace for a workspace build or just the package for a single package build.

The `daemon` command does not take or check the lock.

Add `stop` command that just requests the lock then exits.

Add `--workspace` flag to `clean` command.

Update an end to end test that was using concurrent build during `watch` for checking state: copy the workspace to do the concurrent build.

### DIFF
--- a/build_runner/lib/src/bootstrap/build_process_lock.dart
+++ b/build_runner/lib/src/bootstrap/build_process_lock.dart
@@ -1,0 +1,225 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:path/path.dart' as p;
+import 'package:watcher/watcher.dart';
+
+import '../build_plan/build_paths.dart';
+import '../logging/build_log.dart';
+
+const _packageLockName = 'build_runner.lock';
+const _workspaceLockName = 'build_runner.workspace.lock';
+
+/// File lock to ensure only one `build_runner` command runs at a time on the
+/// same files.
+///
+/// Needs to know whether the build is a whole workspace or package build to
+/// use the correct lock granularity.
+///
+/// If a lock can't be taken, requests the lock. Notifies of requests for the
+/// lock from another process via [setLockRequestCallback].
+///
+/// Use via `BuildProcessState` which handles passing the configuration to the
+/// child process.
+class BuildProcessLock {
+  final BuildPaths paths;
+
+  BuildProcessLock(this.paths);
+
+  void Function()? _onLockRequested;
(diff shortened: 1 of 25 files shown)


build_runner/lib/src/bootstrap/build_process_lock.dart
```diff
@@ -0,0 +1,225 @@
...
+            break;
+          }
+
+          // Release the workspace lock before retrying the package lock.
+          workspaceLock?.unlock();
+        }
+      }
+
+      if (!logged) {
+        buildLog.flushAndPrint('Waiting for already-running build_runner.');
```
jensjoha, line 76: Am I reading this right that if there is another running instance this will print 10 times per second?

build_runner/lib/src/bootstrap/build_process_lock.dart
```diff
@@ -0,0 +1,225 @@
...
+      _requestLock(file.path);
+      _randomAccessFile.closeSync();
+      rethrow;
+    }
+  }
+
+  /// Releases the lock and closes the file.
+  void unlock() {
+    try {
+      _randomAccessFile.unlockSync();
```
jensjoha: Should this be split into two try's?

build_runner/lib/src/bootstrap/build_process_lock.dart
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  /// Creates `$path.requested` file.
+  static void _requestLock(String path) {
+    final requestedFile = File('$path.requested');
+    try {
+      requestedFile.parent.createSync(recursive: true);
+      // Write even if the file already exists to trigger a file watch event in
+      // the process that has the lock.
+      requestedFile.writeAsStringSync('');
```
jensjoha: Is writing enough or does it also have to flush?

jensjoha (approved):

build_runner/lib/src/bootstrap/build_process_lock.dart
davidmorgan: We set `logged = true` so it's only printed once :)

build_runner/lib/src/bootstrap/build_process_lock.dart
davidmorgan: Sounds good, thanks. Done.

build_runner/lib/src/bootstrap/build_process_lock.dart
davidmorgan: Can't hurt to flush; done. Thanks.

PR #4881, opened 2026-04-15
Compile builders with AOT by default.

### LINKED ISSUE #4386: Use AOT builds by default
Currently they are enabled by `--force-aot`.

Reasons _not_ to use AOT:

 - builders that use `dart:mirrors` will not compile
 - builders that depend on a detail of the JIT runtime environment such as `Isolate.packageConfig` will fail at runtime
 - 20s or so of additional startup time on first build

But these should not be a problem in _most_ situations.

davidmorgan:
Fix #4386.

Compile with AOT first, fall back to JIT if there is a failure due to mirrors.

The `build_runner` child process sometimes needs `Platform.resolvedExecutable`, pass it in `build_process_state`.

The `run` command doesn't work with AOT mode, use JIT always and remove the --force-jit and --force-aot flag for that command.

`testBuilders` now explicitly turns off builder freshness checking, the builders are passed rather than compiled so it anyway didn't make sense; it was only passing by coincidence, now it can fail if the test code tries to check the wrong compile type.

Add `--force-jit` in most e2e tests as we want the faster compile time.

Add e2e test for the fallback to JIT.

### DIFF
--- a/build_runner/lib/src/bootstrap/compile_type.dart
+++ b/build_runner/lib/src/bootstrap/compile_type.dart
@@ -31,7 +31,10 @@ enum CompileStrategy {
   /// Try AOT compilation, fall back to JIT if it fails.
   tryAot,
 
-  alwaysJit;
+  /// For the `run` command: JIT is required, so there are no flags and
+  /// `build_runner` knows to always use JIT and to check the JIT compile for
+  /// freshness.
+  commandForcesJit;
 
   CompileType get initialCompileType {
     switch (this) {
@@ -39,7 +42,7 @@ enum CompileStrategy {
       case CompileStrategy.tryAot:
         return CompileType.aot;
       case CompileStrategy.forceJit:
-      case CompileStrategy.alwaysJit:
+      case CompileStrategy.commandForcesJit:
         return CompileType.jit;
     }
   }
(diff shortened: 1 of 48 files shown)


build_runner/lib/src/bootstrap/compile_type.dart
```diff
@@ -0,0 +1,46 @@
...
+
+enum CompileStrategy {
+  forceJit,
+
+  forceAot,
+
+  /// Try AOT compilation, fall back to JIT if it fails.
+  tryAot,
+
+  alwaysJit;
```
jensjoha: The `alwaysJit` confused me.
If we can't come up with a better name could we at least add a comment that this is (assuming I finally understood correctly) used by commands (currently only `run`) that can't be forced to run in a specific way by the user?

jensjoha (approved):

build_runner/lib/src/bootstrap/compile_type.dart
davidmorgan: Thanks! Renamed to `commandForcesJit` with a comment.

/// For the `run` command: JIT is required, so there are no flags and
/// `build_runner` knows to always use JIT and to check the JIT compile for
/// freshness.
commandForcesJit;

davidmorgan:
CI failure is unrelated, looks like the latest SDK dev release has the incremental compiler issue that we saw in google3. I'll merge anyway.

PR #4892, opened 2026-04-24
Remove old performance-related options.

davidmorgan:
I've long suspected that `--low-resources-mode` is not useful: it turns off caching at the filesystem level but we anyway cache files that get handed to the analyzer. I compared with/without the flag on watch mode with 10 builds of the 5000 file benchmark and didn't see any difference.

So, remove it. This is simpler: the presence/absence of filesystem cache changed when files are read and written, which could change behavior.

I haven't heard of anyone using the old performance logging, and I don't find references to it on Google search either, so I think it can be removed too.

The large-looking changes to `build.dart` are just removing the log hook, no actual changes.

Add warning if a removed option is used.


build_runner/lib/src/build_runner.dart
```diff
@@ -112,6 +112,14 @@ class BuildRunner {
       );
     }
 
+    final removedOptionsUsed = commandLine.removedOptionsUsed;
+    if (removedOptionsUsed.isNotEmpty) {
+      buildLog.warning(
+        'These options have been removed and were ignored: '
+        '${removedOptionsUsed.map((o) => '--$o').join(', ')}',
```
jensjoha, line 119: Should there for good measure be a test for this?

jensjoha (approved):

build_runner/lib/src/build_runner.dart
davidmorgan: Sounds good, done. Thanks.

PR #4910, opened 2026-05-05
Fix post process builder error handling.

davidmorgan:
Both normal builders and post process builders store their build step results in a weird way right now: they are copied over all output nodes. I was in the middle of refactoring to store the build results in a better way when I discovered that post process errors aren't handled properly, there is a crash when deserializing them.

Add test coverage, fix it.

Prepare for the upcoming refactoring about build results by moving post process build results to be stored once per execution instead of on all the outputs. Introduce a new node type, postGenerated, for these generated nodes where the execution result is stored elsewhere. The ~next PR will do the same for normal builders.

### DIFF
--- a/build_runner/lib/src/build/asset_graph/graph.dart
+++ b/build_runner/lib/src/build/asset_graph/graph.dart
@@ -606,10 +606,14 @@ class AssetGraph implements GeneratedAssetHider {
 
   /// Returns outputs that were written to the source tree.
   Iterable<AssetId> outputsToDelete(BuildPackages buildPackages) {
-    // Checks if `id` is in a known package. If not, maybe the package was
-    // renamed: move it to the single package being built, if there is one,
-    // to delete for that case. If the ID is in an unknown package and can't
-    // be fixed, returns `null`.
+    // Checks if `id` is in a known package. If so, returns it.
+    //
+    // If not, and a single package is being built, returns `id` moved to that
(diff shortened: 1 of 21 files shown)


build_runner/lib/src/build/asset_graph/graph.dart
```diff
@@ -599,22 +606,43 @@ class AssetGraph implements GeneratedAssetHider {
 
   /// Returns outputs that were written to the source tree.
   Iterable<AssetId> outputsToDelete(BuildPackages buildPackages) {
+    // Checks if `id` is in a known package. If not, maybe the package was
+    // renamed: move it to the single package being built, if there is one,
+    // to delete for that case. If the ID is in an unknown package and can't
```
jensjoha: "to delete for that case"?

jensjoha (approved):

build_runner/lib/src/build/asset_graph/graph.dart
davidmorgan: Expanded the comment :) hopefully it's clearer now

```
    // Checks if `id` is in a known package. If so, returns it.
    //
    // If not, and a single package is being built, returns `id` moved to that
    // package. This allows old generated output to be deleted if the package
    // was renamed since the last build.
    //
    // If `id` is not in a known package and a single package is not being
    // built, returns `null`.
```

PR #4918, opened 2026-05-11
Move build_modules into build_web_compilers

Markzipan:
`build_modules` is only used by `build_web_compilers` to handle web-specific builds. Merging these simplifies releases and will help with our incoming web builder changes to support hot reload.


davidmorgan (approved):
Looks good, thanks for the cleanup :)

PR #4921, opened 2026-05-12
Reduce watches for nested packages

### LINKED ISSUE #4346: File watching with --workspace could be a single watch
Currently the workspace + packages are watched individually, the watch on the workspace should be enough.

It'd be interesting to check if there is any noticeable performance gain from fewer watches.)

davidmorgan:
Fix #4346.

When packages are nested, watch only the outer level package and resolve events to the correct package from that.

### DIFF
--- a/build_runner/lib/src/commands/watch/build_packages_watcher.dart
+++ b/build_runner/lib/src/commands/watch/build_packages_watcher.dart
@@ -6,6 +6,8 @@ import 'dart:async';
 import 'dart:io';
 
 import 'package:async/async.dart';
+import 'package:build/build.dart';
+import 'package:path/path.dart' as p;
 
 import '../../build_plan/build_package.dart';
 import '../../build_plan/build_packages.dart';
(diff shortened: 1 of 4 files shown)


build_runner/lib/src/commands/watch/build_packages_watcher.dart
```diff
@@ -43,58 +44,128 @@ class BuildPackagesWatcher {
...
-    return (change) => !ignorePaths.any(change.id.path.startsWith);
+  /// Builds a forest of [WatchablePackageTree] with all watchable packages.
+  List<WatchablePackageTree> _buildPackageForest() {
+    final watchablePackages = <BuildPackage>[];
+    for (final p in _buildPackages.packages.values) {
+      if (p.watch) watchablePackages.add(p);
+    }
+    watchablePackages.sort((a, b) => a.path.length.compareTo(b.path.length));
+    final result = <WatchablePackageTree>[];
+    for (final package in watchablePackages) {
```
jensjoha, line 90: How many packages do we think we have? This is an O(n^2) algorithm in the number of (watched) packages... (I might add that package_config used to have an O(n^2) algorithm in the number of packages too --- it became noticeable with the number of packages in the package config for internal packages at least, see https://github.com/dart-archive/package_config/pull/117).

jensjoha (approved):

build_runner/lib/src/commands/watch/build_packages_watcher.dart
davidmorgan: That should be okay, we don't watch deps packages, only packages that are in the current workspace. Thanks.

PR #4922, opened 2026-05-12
Extend build_web_compilers to support expression evaluation in Frontend Server + Build Daemon environment

Markzipan:
Part 2 of https://github.com/dart-lang/build/pull/4918

For context, see: https://github.com/dart-lang/build/pull/4898

Moves `build_frontend_server` into build_web_compilers. I'll have some followup fixes to resolve some of the wonkiness with naming.


builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
nshahan, line 47: Leaving a note here as a reminder of our conversation. We will need to have separate "Recompile" and "Recompile Restart" requests.

This is to ensure we can support recompiles for hot reload and recompiles for hot restart. 

If it is the first request, both should coerce to a what we think of as the regular initial compile. 

"Recompile" will recompile a module for a hot reload which includes different output and performs additional checks with potential errors that recommend a hot restart as a recovery. 

"Recompile Restart" should recompile the module as if it were the first compile. This should clear the slate for the next request.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:build_web_compilers/src/build_frontend_server/build_frontend_server.dart';
+import 'package:build_web_compilers/src/build_frontend_server/common.dart';
+import 'package:path/path.dart' as p;
+
+void main(List<String> args) async {
```
davidmorgan, line 19: Worth a comment?

I'd at least say that it's only for internal use by the builder.

```
/// Starts or connects to a frontend server instance.
///
/// Not for direct use, this is an implementation detail of the web builders.
```

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+  print('Frontend Server Manager listening on port ${server.port}');
+
+  server.listen((socket) {
+    socket
+        .cast<List<int>>()
+        .transform(utf8.decoder)
+        .transform(const LineSplitter())
+        .listen((line) async {
+          try {
+            final request = jsonDecode(line) as Map<String, dynamic>;
```
davidmorgan: Please move this lambda body to a top level private method (I guess, whatever works) so this doesn't have to be so very indented.

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/build_frontend_server.dart
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
```
davidmorgan: Optional: you could just directly import these as needed, now? And remove this file.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+                );
+            }
+          } catch (e) {
+            socket.writeln(jsonEncode({'error': e.toString()}));
+          }
+        });
+  });
+
+  // Handle shutdown
+  ProcessSignal.sigint.watch().listen((_) async {
```
davidmorgan, line 67: Is the idea to ensure the frontend server is killed if this process is killed?

FWIW, I don't think this approach works across all ways a process can be killed and all platforms, I don't quite remember the details. Here's what I ended up doing for build_runner, feel free to copy it :)

https://github.com/dart-lang/build/blob/932dfa2f66d27535ed431887e8a7a806795a713e/build_runner/lib/src/bootstrap/processes.dart#L156

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
```diff
@@ -159,28 +228,91 @@ class _RecompileRequest extends _CompilationRequest {
...
 
   static Future<PersistentFrontendServer> start({
     required String sdkRoot,
     required Uri fileSystemRoot,
     required Uri packagesFile,
   }) async {
+    final file = File(p.join(Directory.current.path, fesWorkerPortPath));
+    if (await file.exists()) {
+      try {
+        final content = await file.readAsString();
```
davidmorgan: I would do

```
String? content;
try {
if (file.existsSync()) content = file.readAsStringSync();
} catch (_) {}
if (content == null) {
} else {
}
```

to reduce indentation and the size of try/catch, and because sync I/O is usually simpler+faster.

Feel free to use either / both / none :)

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
```diff
@@ -235,29 +385,103 @@ class PersistentFrontendServer {
...
+          'scriptUri': scriptUri,
+          'line': line,
+          'column': column,
+          'jsModules': jsModules,
+          'jsFrameValues': jsFrameValues,
+          'moduleName': moduleName,
+        },
+      }),
+    );
+    try {
```
davidmorgan, line 495: I would reduce the scope of the try to the assignment to `result`, you already have `result != null` next so that just works.

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
```diff
@@ -235,29 +385,103 @@ class PersistentFrontendServer {
...
+  /// Reads the content of a file generated by the Frontend Server.
+  ///
+  /// This is called by the worker in `workers.dart` in response to a
+  /// `READ_OUTPUT_FILE` request from the daemon in `daemon_builder.dart`.
+  Future<String> readOutputFile(String path) async {
+    final file = File(path);
+    if (await file.exists()) {
+      return await file.readAsString();
+    }
+    throw Exception('File not found: $path');
```
davidmorgan: You could just do file.readAsString(), or file.readAsStringSync(), and let the exception that throws escape if it's not found?

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
```diff
@@ -91,7 +96,10 @@ class DdcFrontendServerBuilder implements Builder {
       }
     }
 
-    final scratchSpace = await buildStep.fetchResource(scratchSpaceResource);
+    final scratchSpace =
+        scratchSpaceDir != null
+            ? ScratchSpace.existing(Directory(scratchSpaceDir!))
```
davidmorgan, line 101: I think you need a new min version for `scratch_space` in `build_web_compilers/pubspec.yaml` for this?

davidmorgan (approved):
Looks good, mostly optional/style comments.

Except the min version of `scratch_space`, I think you need that :)

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:build_web_compilers/src/build_frontend_server/build_frontend_server.dart';
+import 'package:build_web_compilers/src/build_frontend_server/common.dart';
+import 'package:path/path.dart' as p;
+
+void main(List<String> args) async {
+  if (args.length < 3) {
+    print('Usage: fes_manager <sdkRoot> <fileSystemRoot> <packagesFile>');
```
nshahan, line 21: Why not use the `package:args`? Seeing an invocation in our tooling with three file/directory paths without any option names will be somewhat harder to decipher what each of them are and since this is intended for tooling the terseness doesn't give any benefits.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+  final sdkRoot = args[0];
+  var fileSystemRoot = Uri.parse(args[1]);
+  if (!fileSystemRoot.path.endsWith('/')) {
+    fileSystemRoot = fileSystemRoot.replace(path: '${fileSystemRoot.path}/');
+  }
+
+  final packagesFile = Uri.parse(args[2]);
+
+  // Delete the port file if it exists. Frontend Server finds a port and
+  // writes it.
```
nshahan: This file is used by other parts of the tooling to know what port to communicate with the frontend server? Are they intended to watch the file to see if it changes or are we assuming no service has read it yet?

Also, the comment is confusing me because it looks like this program gets a port and writes the file a few lines below.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -0,0 +1,151 @@
...
+  if (portFile.existsSync()) portFile.deleteSync();
+
+  final fes = await PersistentFrontendServer.start(
+    sdkRoot: sdkRoot,
+    fileSystemRoot: fileSystemRoot,
+    packagesFile: packagesFile,
+  );
+
+  final driver = FrontendServerProxyDriver()..init(fes);
+  final server = await ServerSocket.bind(InternetAddress.loopbackIPv4, 0);
```
nshahan, line 44: Do we ever close this socket?

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_resources.dart
```diff
@@ -18,6 +19,9 @@ class FrontendServerState {
   /// hot reload.
   AssetId? entrypointAssetId;
 
+  /// The scratch space used by the Frontend Server if owned by webdev.
```
nshahan: Can you elaborate the comment a little more to describe what is written by fes and what reads from the scratch space?

builder_pkgs/build_web_compilers/lib/src/build_modules/common.dart
```diff
@@ -13,6 +13,11 @@ const webHotReloadOption = 'web-hot-reload';
 final sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
 final packagesFilePath = p.join('.dart_tool', 'package_config.json');
 
+/// The file path containing the Frontend Server worker's port.
+///
+/// Used for expression evaluation.
+final fesWorkerPortPath = p.join('.dart_tool', 'build', 'fes_worker_port');
```
nshahan: Should we create this with a different name? It seems like it got expanded already to contain more than just a port.

Maybe `fes_manager_config` or something like that?

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/common.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2026, the Dart project authors.  Please see the AUTHORS file
+// for details. All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+import 'dart:io';
+import 'package:path/path.dart' as p;
+
+const multiRootScheme = 'org-dartlang-app';
+final sdkDir = p.dirname(p.dirname(Platform.resolvedExecutable));
+final fesWorkerPortPath = p.join('.dart_tool', 'build', 'fes_worker_port');
```
nshahan: Is there another service somewhere we call the "frontend server worker" or does the "worker" in this name refer to the "frontend server manager" that writes the file?

builder_pkgs/build_web_compilers/lib/src/build_modules/workers.dart
```diff
@@ -135,14 +135,26 @@ final frontendServerProxyDriverResource = Resource<FrontendServerProxyDriver>(
 
 PersistentFrontendServer? __persistentFrontendServer;
```
nshahan, line 136: Are the two leading underscores in this name intentional? Is there some convention in this file already?

builder_pkgs/build_web_compilers/lib/src/web_entrypoint_marker_builder.dart
```diff
@@ -32,25 +40,36 @@ class WebEntrypointMarkerBuilder implements Builder {
     final frontendServerState = await buildStep.fetchResource(
       frontendServerStateResource,
     );
-    final frontendServerStateWasLoaded = await frontendServerState
-        .checkAndDeserializeState(buildStep);
-    if (frontendServerStateWasLoaded) return;
 
-    final webAssets = await buildStep.findAssets(Glob('web/**')).toList();
+    // Start the Frontend Server early to record its port.
```
nshahan: IIUC this triggers fes_manager to be spun up which writes the file with the port in it but why is that important? What reads it?

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
```diff
@@ -0,0 +1,151 @@
...
+      tempDir = await Directory.systemTemp.createTemp('fes-manager-test');
+      packageConfig = File(
+        p.join(tempDir.path, '.dart_tool', 'package_config.json'),
+      );
+      packageConfig.createSync(recursive: true);
+      packageConfig.writeAsStringSync(
+        jsonEncode({'configVersion': 2, 'packages': <String>[]}),
+      );
+
+      final portFile = File(p.join(Directory.current.path, fesWorkerPortPath));
```
nshahan: What is `Directory.current` when you run this test, does the test package control it? Do we need to change the working directory to be under the tempDir?

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
```diff
@@ -0,0 +1,151 @@
...
+              .transform(const LineSplitter())
+              .first;
+      final response = jsonDecode(responseLine) as Map<String, dynamic>;
+      expect(response.containsKey('error'), isTrue);
+
+      await socket.close();
+      process.kill();
+      await process.exitCode;
+
+      if (portFile.existsSync()) portFile.deleteSync();
```
nshahan: Should we move this to `tearDown()`?

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
```diff
@@ -0,0 +1,151 @@
...
+              .first;
+
+      final response = jsonDecode(responseLine) as Map<String, dynamic>;
+      expect(response.containsKey('error'), isTrue);
+
+      await socket.close();
+      process2.kill();
+      await process2.exitCode;
+
+      if (portFile.existsSync()) portFile.deleteSync();
```
nshahan: ditto

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: For sure - updated!

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/build_frontend_server.dart
Markzipan: Good catch - updated

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
Markzipan: I broke this out into a helper method + used the sync version. I kinda like the try catches as they are since we can add more helpful error messages.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: done

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: Yep - exactly. Thanks for the pointer. That function's private, so I just copied the logic. We can share the two in a future update if needed

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
Markzipan: true - updated!

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
Markzipan: Good catch - updated

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
Markzipan: done

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: Since this is purely used for internal tooling (ideally no users except us and other maintainers), adding package:args as an extra dep doesn't seem too worthwhile. Lmk if you have strong opinions here - I'm mostly ambivalent

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: Woops, that's a badly written/incomplete comment. It should've been "this frontend server manager" or something. Basically the 'first' invocation of this process should 1) clean up stale port files, 2) spin up FES, and 3) write its port to a port file. Comment updated

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: Nope - I'll add some logic for explicitly closing these

builder_pkgs/build_web_compilers/lib/src/build_modules/common.dart
Markzipan: Oof, those field were added from a previous impl, and I forgot to clean them up. I like the idea of calling it a config though (for future proofing). Name updated

builder_pkgs/build_web_compilers/lib/src/build_modules/workers.dart
Markzipan: Yep, I borrowed that convention for its ilk like `__frontendDriver`. No idea why though

builder_pkgs/build_web_compilers/lib/src/web_entrypoint_marker_builder.dart
Markzipan: Good catch! This was from an old iteration that wrote the port builder-first. It's stale now, so deleted

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
Markzipan: Ooof good catch - updated

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
Markzipan: Moving this to the temp dir deletes these files too - so done

builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
Markzipan: Moving this to the temp dir deletes these files too - so done

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_resources.dart
Markzipan: done

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/frontend_server_driver.dart
Markzipan: Thanks for the review! I'll add these functions in an incoming change + more tests

nshahan (approved):

PR #4926, opened 2026-05-14
Cache wasm module loads

biggs0125:
Different loadLibrary calls can result in the same wasm module getting requested multiple times. The wasm runtime leaves it up to the embedder to deduplicate these requests.

This updates the build_web_compilers generated entrypoint to handle this module loading deduplication. We now cache the Promise that resolves after the module is initialized. Subsequent requests can then block on this same Promise.


Markzipan (approved):
LGTM!

PR #4935, opened 2026-05-19
Move generated output information from AssetNode to AssetGraph

davidmorgan:
- Remove `AssetNode` type `NodeType.generated`, all the information is now in `AssetGraph`.
- Digests of outputs, and which files were output, go in `BuildStepResult`.
- Declared outputs go in a new map in `AssetGraph`.
- Add getters on `AssetGraph` in preference to checking things with `AssetNode`


build_runner/lib/src/io/asset_tracker.dart
```diff
@@ -70,16 +69,12 @@ class AssetTracker {
...
-          return true;
-        })
-        .map((n) => n.id)
-        .where((id) => !allSources.contains(id));
+    final removedAssets = [
+      for (final node in assetGraph.allNodes)
+        if (node.isFile) node.id,
+      for (final id in assetGraph.buildStepsByDeclaredOutput.keys)
+        if (assetGraph.isActualOutput(id)) id,
+    ].where((id) => !allSources.contains(id));
```
jensjoha: Why not just include this check in the ifs above?

build_runner/lib/src/io/build_output_reader.dart
```diff
@@ -90,32 +89,31 @@ class BuildOutputReader {
...
-    if (!node.isFile) return UnreadableReason.assetType;
+    if (!_assetGraph.isFile(id)) return UnreadableReason.assetType;
 
-    if (node.type == NodeType.postGenerated) {
+    if (_assetGraph.isActualPostOutput(id)) {
       return null;
     }
-    if (node.type == NodeType.generated) {
-      if (_processedOutputs?.contains(node.id) == false) {
+    if (_assetGraph.buildStepsByDeclaredOutput.containsKey(id)) {
```
jensjoha: Whether it matters or not I don't know, but there are a few entries of this pattern of
```dart
if (foo.containsKey(bar)) {
  var value = foo[bar];
  // ...
}
```

where unless the value is nullable (which I haven't checked) could instead be
```dart
var value = foo[bar];
if (value != null) {
  // ...
}
```
which would save a lookup.

build_runner/lib/src/io/build_output_reader.dart
```diff
@@ -153,59 +151,53 @@ class BuildOutputReader {
...
-          return node.id;
-        })
-        .whereType<AssetId>()
-        .toList();
+    final assetGraph = _assetGraph;
+    if (assetGraph == null) return [];
+    return [
+      for (final node in assetGraph.allNodes) node.id,
+      ...assetGraph.buildStepsByDeclaredOutput.keys,
+    ].where((id) => !_shouldSkipId(assetGraph, id, rootDir)).toList();
```
jensjoha: (I'm not a fan of these `.where` etc and this _could_ be included in the generation as well... don't feel obligated to change stuff you like just because I don't like it though)

jensjoha (approved):

build_runner/lib/src/io/asset_tracker.dart
davidmorgan: Done.

build_runner/lib/src/io/build_output_reader.dart
davidmorgan: Done, and in various other places. Thanks!

build_runner/lib/src/io/build_output_reader.dart
davidmorgan: Agreed, did it with normal for loops instead. Thanks.

davidmorgan (commented):
Thanks!

PR #4937, opened 2026-05-20
Remove NodeType.placeholder, just match paths.

davidmorgan:
We need to populate the graph with outputs generated for placeholders, but don't need the actual placeholders.

### DIFF
--- a/build_runner/lib/src/build_plan/build_packages.dart
+++ b/build_runner/lib/src/build_plan/build_packages.dart
@@ -16,6 +16,7 @@ import '../io/asset_path_provider.dart';
 import 'build_package.dart';
 import 'build_packages_loader.dart';
 import 'build_paths.dart';
+import 'placeholders.dart';
 
 /// The SDK package, we filter this to the core libs and dev compiler
 /// resources.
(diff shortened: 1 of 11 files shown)


build_runner/lib/src/build_plan/build_packages.dart
```diff
@@ -263,6 +263,15 @@ class BuildPackages implements AssetPathProvider {
   /// Builders can only auto apply from a package in the same build.
   BuiltSet<String> peersOf(String packageName) => _peerPackages[packageName]!;
 
+  Iterable<AssetId> get placeholderIds => packages.keys.expand(
+    (package) => [
+      AssetId(package, r'lib/$lib$'),
```
jensjoha: We now have these 4 strings in (at least) two places - would it make sense to have them as constants or something somewhere?

jensjoha (approved):

build_runner/lib/src/build_plan/build_packages.dart
davidmorgan: Thanks, added `placeholders.dart`. This forced me to actually figure out how they work and write a comment explaining :)

PR #4938, opened 2026-05-21
Remove AssetNode

davidmorgan:
Remove `AssetNode`, rename `Nodes` to `Sources` and handle all sources and their Digests there.

Remove AssetGraph serialization test, it's better covered by functional tests.

Remove AssetGraph "tracks dependencyGraph" test, it's better covered by functional tests.

I'll send a follow-up PR cleaning up comments that still mention "node" and tidying up the APIs and comments generally.


build_runner/lib/src/build/asset_graph/graph.dart
```diff
@@ -135,11 +130,16 @@ class AssetGraph implements GeneratedAssetHider {
       isSource(id) || isDeclaredOutput(id) || isActualPostOutput(id);
 
   /// Whether [id) is a source file.
```
jensjoha: Probably `[id]` not `[id)`? (also below and above and whatnot).

jensjoha (approved):

build_runner/lib/src/build/asset_graph/graph.dart
davidmorgan: Done. Thanks!

PR #4947, opened 2026-05-27
Document JSON parsing of --define values.

### LINKED ISSUE #4310: --define includes literal quotes in value on Windows
I ran into a weird issue where it seems that when providing build options like this:

```
dart run build_runner build --define=some_package:some_builder=some_option="Priority 6"
```

on Windows the value of that option in the builder includes the `"` quotes as part of the String.

---

- Dart SDK version: 3.10.4 (stable) (Tue Dec 9 00:01:55 2025 -0800) on "windows_x64"
- build_runner 2.10.4

davidmorgan:
Fix #4310, add test coverage.

The "parse as JSON, fall back to string" is quite surprising, I would prefer without the fallback. But at least now it's documented :)

### DIFF
--- a/build_config/README.md
+++ b/build_config/README.md
@@ -413,9 +413,9 @@ And when running the build:
 dart run build_runner build --define=some_package:some_builder=some_option="Priority 6"
 ```
 
-If values are valid JSON they are parsed to objects, making them as expressive
-as YAML files. For example, it's possible to pass lists and maps. If values are
-not valid JSON they are used as exact strings.
+If values are valid JSON they are parsed to objects. For example, it's possible
+to pass lists and maps. If values are not valid JSON they are used as exact
+strings.
 
 ### How can I include additional sources in my build?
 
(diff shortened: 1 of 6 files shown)


build_config/README.md
```diff
@@ -413,6 +413,10 @@ And when running the build:
 dart run build_runner build --define=some_package:some_builder=some_option="Priority 6"
 ```
 
+If values are valid JSON they are parsed to objects, making them as expressive
+as YAML files. For example, it's possible to pass lists and maps. If values are
```
jensjoha: "making them as expressive as YAML files." --- I don't think that's technically true.

jensjoha (approved):

build_config/README.md
davidmorgan: Thanks! On checking it looks like this is subtle for different reasons: for example yaml allows non-String map keys, but build_runner rejects them. I guess it probably enforces yaml files that are equivalent to JSON.

But I don't see any benefit to trying to explain that here :) so I just removed the sentence. Thanks.

PR #4953, opened 2026-05-28
Simplify build_runner instructions in README

jonasfj:
Removed redundant command from README for clarity.

Credits @sigurdm in https://github.com/dart-lang/build/commit/056424f57ceade852839cc9338fb5a8c0efb3b34 :rofl:


davidmorgan (approved):
Ah yes I think that's an improvement. Thanks!

PR #4955, opened 2026-05-29
Move declared outputs from BuildState to BuildStepPlan.

davidmorgan:
While working on the benchmarks I noticed that the web build asset_graph.json is huge, this turns out to be because it has optional outputs for each compiler and right now optional-but-not-written outputs do get mentioned in the serialized BuildState.

Fortunately all the recent refactorings should make it easy to separate declared outputs from actual outputs: move them from `BuildState` to new `BuildStepPlan` in `BuildPlan` and stop serializing them, they can always be computed.

Note that a change to `BuildPhases` always causes the old build to be discarded, we can rely on `BuildPhases` being the same for an incremental build, allowing the declared outputs to be computed correctly.

 - `BuildStepPlan` takes over from `BuildState` as the `GeneratedAssetHider` that knows whether an asset is generated or not
 - A few different places in `BuildState` were computing transitive declared outputs, pull out one method and simplify

Cuts JSON size for a small web build from ~10MB to ~1MB.

### DIFF
--- a/build_runner/lib/src/build/build.dart
+++ b/build_runner/lib/src/build/build.dart
@@ -18,11 +18,13 @@ import '../build_plan/build_options.dart';
 import '../build_plan/build_packages.dart';
 import '../build_plan/build_phases.dart';
 import '../build_plan/build_plan.dart';
+import '../build_plan/build_step_plan.dart';
 import '../build_plan/phase.dart';
 import '../build_plan/testing_overrides.dart';
 import '../constants.dart';
 import '../io/build_output_reader.dart';
 import '../io/create_merged_dir.dart';
+import '../io/generated_asset_hider.dart';
 import '../io/reader_writer.dart';
 import '../logging/build_log.dart';
 import '../logging/timed_activities.dart';
@@ -53,11 +55,11 @@ final ResolversImpl _defaultResolvers = ResolversImpl(
 
 /// A single build.
 class Build {
-  final BuildPlan buildPlan;
+  BuildPlan _buildPlan;
 
   // Collaborators.
   final ResourceManager resourceManager;
-  final ReaderWriter readerWriter;
+  ReaderWriter _readerWriter;
   final LibraryCycleGraphLoader previousLibraryCycleGraphLoader =
       LibraryCycleGraphLoader();
   final AssetDepsLoader? previousDepsLoader;
@@ -125,11 +127,13 @@ class Build {
   BuildOutputReader? _buildOutputReader;
 
   Build({
-    required this.buildPlan,
-    required this.readerWriter,
+    required BuildPlan buildPlan,
+    required ReaderWriter readerWriter,
     required this.resourceManager,
     required this.buildState,
-  }) : previousDepsLoader =
+  }) : _buildPlan = buildPlan,
+       _readerWriter = readerWriter,
+       previousDepsLoader =
            buildPlan.previousPhasedAssetDeps == null
                ? null
                : AssetDepsLoader.fromDeps(buildPlan.previousPhasedAssetDeps!),
@@ -140,11 +144,14 @@ class Build {
          _ => null,
        };
 
+  BuildPlan get buildPlan => _buildPlan;
+  ReaderWriter get readerWriter => _readerWriter;
   BuildOptions get buildOptions => buildPlan.buildOptions;
   TestingOverrides get testingOverrides => buildPlan.testingOverrides;
   BuildPackages get buildPackages => buildPlan.buildPackages;
   BuildConfigs get buildConfigs => buildPlan.buildConfigs;
-  BuildPhases get buildPhases => buildPlan.buildPhases;
+  BuildPhases get buildPhases => buildPlan.buildStepPlan.buildPhases;
+  BuildStepPlan get buildStepPlan => buildPlan.buildStepPlan;
 
   BuildOutputReader get buildOutputReader =>
       _buildOutputReader ??= BuildOutputReader(
@@ -162,8 +169,8 @@ class Build {
     if (result.status == BuildStatus.success) {
       final failedSteps = <BuildStepId>{};
       for (final output in processedOutputs) {
-        if (!buildState.isDeclaredOutput(output)) continue;
-        final step = buildState.stepForDeclaredOutput(output);
+        final step = buildStepPlan.stepForDeclaredOutputOrNull(output);
+        if (step == null) continue;
         final stepResult = buildState.stepResult(step);
         if (stepResult.failed) {
           failedSteps.add(step);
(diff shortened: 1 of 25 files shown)


build_runner/lib/src/build/build.dart
```diff
@@ -904,7 +926,11 @@ class Build {
 
         // If the primary input succeeded but was not output, this build is
         // skipped.
-        if (!buildState.isActualOutput(step.primaryInput)) {
+
```
jensjoha, line 929: The empty line seems weird.

build_runner/lib/src/build/build_series.dart
```diff
@@ -269,9 +280,10 @@ class BuildSeries {
 
     _currentBuildResult = build.run(updates);
     final result = await _currentBuildResult!;
-    _buildPlan = _buildPlan.forNextBuild(
+    _buildPlan = build.buildPlan.updateForResult(
       previousPhasedAssetDeps: result.phasedAssetDeps,
     );
+    _readerWriter = build.readerWriter;
```
jensjoha, line 286: Maybe add a comment why this is needed. Possibly move it up a few lines to the call where it changes (and is this a bug-fix instead of really related to this refactor?).

build_runner/lib/src/build/build.dart
```diff
@@ -1178,10 +1205,12 @@ class Build {
...
       for (final id in generatedFileInputs) {
-        final stepResult = buildState.stepResult(
-          buildState.stepForDeclaredOutput(id),
+        final stepResult = buildState.stepResultOrNull(
+          buildStepPlan.stepForDeclaredOutput(id),
         );
-        if (buildState.isActualOutput(id) && stepResult.succeeded) {
+        if (stepResult != null &&
+            stepResult.outputs.containsKey(id) &&
+            stepResult.succeeded) {
```
jensjoha, line 1213: Possibly move the `stepResult.succeeded` before `stepResult.outputs.containsKey(id)`.

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -284,18 +285,40 @@ class BuildPlan {
       // Files marked for deletion are not inputs.
       inputSources.removeAll(filesToDelete);
 
+      // Build steps are computed twice: first to identify inputs that are
+      // actually outputs and remove them from the input set.
```
jensjoha, line 289: This comment reads oddly to me --- it's like I'm about to be told two things, but then I'm only told one.

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -284,18 +285,40 @@ class BuildPlan {
...
+      final declaredOutputs =
+          BuildStepPlan.compute(
+            buildPhases: buildPhases,
+            placeholderIds: buildPackages.placeholderIds,
+            sources: inputSources,
+          ).declaredOutputs;
+      final inputSourcesWithoutOutputs = Set<AssetId>.from(inputSources);
+      for (final output in declaredOutputs) {
+        inputSourcesWithoutOutputs.remove(output);
```
jensjoha, line 298: I suppose if all of these `remove` calls return `false` we could reuse the `BuildStepPlan.compute` call if we wanted. I don't know if it's worth it though.

build_runner/lib/src/io/asset_tracker.dart
```diff
@@ -70,9 +84,7 @@ class AssetTracker {
     final newSources = inputSources.difference(buildState.sources.toSet());
     addUpdates(newSources, ChangeType.ADD);
     final removedAssets = [
-      for (final id in buildState.sources.followedBy(
-        buildState.declaredAndActualOutputs,
-      ))
+      for (final id in buildState.sources.followedBy(declaredAndActualOutputs))
```
jensjoha, line 87: Wouldn't

```
      for (final id in buildState.sources)
        if (!allSources.contains(id)) id,
      for (final id in declaredAndActualOutputs)
        if (!allSources.contains(id)) id,
```

be the same, easier to read and better performing?

build_runner/lib/src/io/build_output_reader.dart
```diff
@@ -182,10 +197,12 @@ class BuildOutputReader {
...
-    if (buildState.isDeclaredOutput(id)) {
-      final step = buildState.stepForDeclaredOutput(id);
-      final stepResult = buildState.stepResult(step);
-      if (!buildState.isActualOutput(id) || stepResult.failed) {
+    final step = _buildStepPlan?.stepForDeclaredOutputOrNull(id);
+    if (step != null) {
+      final stepResult = buildState.stepResultOrNull(step);
+      if (stepResult == null ||
+          !stepResult.outputs.containsKey(id) ||
+          stepResult.failed) {
```
jensjoha, line 205: (if moving the `stepResult.succeeded` previously maybe also move `stepResult.failed`)

build_runner/test/build/build_state/build_state_test.dart
```diff
@@ -190,8 +200,9 @@ void main() {
       });
 
       test('overlapping build phases cause an error', () async {
+        // Statically compute overlapping outputs to verify it throws
```
jensjoha, line 203: What does "Statically" mean here?

jensjoha (approved):

build_runner/lib/src/build/build.dart
davidmorgan: Done.

build_runner/lib/src/build/build.dart
davidmorgan: Done.

build_runner/lib/src/build/build_series.dart
davidmorgan: _readerWriter used to use the BuildState, now it uses the BuildStepPlan which is updated in slightly different places.

So `Build` was updating it and `BuildSeries` was grabbing the updated version.

But actually that can be simplified: we can just pass the `BuildStepPlan` before every build, in `Build`, and remove everywhere else. Done.

Thanks :)

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Done.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: I think it's not worth it: I've never noticed the build step computation in profiling.

build_runner/lib/src/io/asset_tracker.dart
davidmorgan: Done.

build_runner/lib/src/io/build_output_reader.dart
davidmorgan: Thanks, done :)

build_runner/test/build/build_state/build_state_test.dart
davidmorgan: Whoops, sorry. It's a word that Gemini likes to use that I didn't catch. Thanks!

PR #4960, opened 2026-06-03
Show import path to libraries with disallowed deps.

### LINKED ISSUE #4959: Make web build "unsupported platform" error actionable
It should give a way to find why there is a dependency on the problem libraries, or print it itself.

`import_path` can apparently do it https://github.com/dart-lang/build/issues/2719#issuecomment-1703117602

davidmorgan:
Fix #4959.

This is AI generated. I'm not super familiar with the code, as far as I can tell the change makes sense.

Initial prompt: "Please fix https://github.com/dart-lang/build/issues/4959"

Then I asked "What does the output look like if there are multiple files with disallowed imports?" and it added a second test with multiple disallowed imports.

--> first commit.

Then I said "Hmm I think you could clean up that rendering code in errors.dart with a few extension methods on AssetId?" and it cleaned up; I manually removed the one-disallowed-import test case as I don't think it adds much. --> second commit.

### DIFF
--- a/builder_pkgs/build_web_compilers/lib/src/build_modules/errors.dart
+++ b/builder_pkgs/build_web_compilers/lib/src/build_modules/errors.dart
@@ -192,10 +192,7 @@ class UnsupportedModules implements Exception {
                 .toList();
         if (unsupportedSdkDeps.isEmpty) continue;
 
-        final targetAssetId = AssetId(
-          library.id.package,
-          library.id.path.replaceFirst(moduleLibraryExtension, '.dart'),
-        );
+        final targetAssetId = library.id.dartAssetId;
         final path = await _findImportPath(
           entrypoint,
           targetAssetId,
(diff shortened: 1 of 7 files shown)


builder_pkgs/build_web_compilers/lib/src/build_modules/errors.dart
```diff
@@ -163,8 +165,111 @@ class UnsupportedModules implements Exception {
...
+    return buffer.toString();
+  }
+
   @override
   String toString() =>
       'Some modules contained libraries that were incompatible '
       'with the current platform.';
 }
+
+Future<List<AssetId>?> _findImportPath(
```
Markzipan, line 222: I'm fine with crawling the dep graph 'late', but we could save some work if we attached backpointers during the original crawl in https://github.com/dart-lang/build/blob/master/builder_pkgs/build_web_compilers/lib/src/build_modules/modules.dart in `computeTransitiveDependencies` and `computeTransitiveAssets`

davidmorgan: I think I prefer the current version--precomputing something that will only be used if there is a serious error seems like a net loss. Up to you, what would you like me to do? :) thanks.

Markzipan: I'm not opinionated so lgtm

Markzipan (approved):

PR #4961, opened 2026-06-03
Provide dummy values for Int32 and Int64

natebosch:
These are value types that should not be faked. An upcoming change will
mark these classes as final. Since these are types from a core package
owned by the Dart team, centralize support for using the static `ZERO`
fields as the dummy values.

The implementation and test intentionally avoid a dependency on
`package:fixnum`.

### DIFF
--- a/builder_pkgs/mockito/CHANGELOG.md
+++ b/builder_pkgs/mockito/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 5.8.0-wip
+
+* Provide dummy values for `Int32` and `Int64` from `package:fixnum`.
+
 ## 5.7.0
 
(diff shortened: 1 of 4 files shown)


davidmorgan:
Thanks! Change makes sense but PTAL at google3 presubmit failures

cl/926515114

--in particular whether it's landable with reasonable test fixes or whether there are changes in the PR that could help.

natebosch:
> PTAL at google3 presubmit failures

I'm working through the failures surfaced by a TAP global run. cl/925685112

davidmorgan (approved):
Approved, please be sure it's landable in google3 before merging :)

Issue #4963, opened 2026-06-04
`build_web_compilers` >= 4.5.0 is resulting in a build error when compiling Jaspr projects


Archais:
Depending on `build_web_compilers` version 4.5.0 (or later) results in the error below:

```console
andrew@mbp jaspr_site % dart run build_runner build -v
Building package executable...
Built build_runner:build_runner.
11s compiling builders/aot

build_runner
E Builders build_web_compilers:module_library and build_modules:module_library outputs collide: package:collection/collection.module.library
```

This error still occurs with the latest `build_web_compilers` version I tested, 4.7.0. Reverting `build_web_compilers` to 4.4.19 resolves the error and allows the project to build successfully.

The issue appears to be related to the change introduced in `build_web_compilers` 4.5.0:

> Move `module_library` and `module_cleanup` builders from `build_modules` to `build_web_compilers`.

## Environment

* Dart: 3.12.1
* Build packages:
    * `build_runner`: 2.15.0
    * `build_web_compilers`: >=4.5.0
    * `build` (transitively): 4.0.6
    * `build_modules` (transitively): 5.1.12
* Builders: I'm using `jaspr_builder` and `jaspr_tailwind`, these are used to compile Jaspr projects and provide support for TailwindCSS respectively.
* OS: macOS

Below is a sample of my pubspec for reference:

<details>
<summary>pubspec.yaml</summary>

```yaml
environment:
  sdk: ^3.12.0

dependencies:
  dotenv: ^4.2.0
  http: ^1.6.0
  jaspr: ^0.23.1
  jaspr_content: ^0.5.2
  jaspr_riverpod: ^0.4.5
  jaspr_router: ^0.8.2
  shelf: ^1.4.2
  shelf_router: ^1.1.4
  universal_web: ^1.1.1+1

dev_dependencies:
  build_runner: ^2.15.0
  build_web_compilers: ^4.7.0 # Results in error; pinning to version 4.4.19 works
  jaspr_builder: ^0.23.1
  jaspr_tailwind: ^0.3.6
  lints: ^6.1.0

jaspr:
  mode: server
```

</details>

davidmorgan:
I believe the problem is that you still have a transitive dependency onto `build_modules`, but these outputs are now generated by `build_web_compilers`. 

[...]

PR #4964, opened 2026-06-04
Add build_modules exports to build_web_compilers

Markzipan:
`build_modules` was moved back into `build_web_compilers`. It didn't have many imports (based on pub.dev), but this broke projects using [jaspr_tailwind](https://pub.dev/packages/jaspr_tailwind) since it depends on `scratchSpace` and `scratchSpaceResource`.

I've added the following exports back to `build_web_compilers`:
```
        DartPlatform
        MissingModulesException
        Module
        ModuleBuilder
        UnsupportedModules
        moduleExtension
        scratchSpace
        scratchSpaceResource
```

Only the last two are strictly necessary, but the others seemed reasonable for someone interacting with `build_modules` artifacts.

See: https://github.com/dart-lang/build/issues/4963

### DIFF
--- a/builder_pkgs/build_web_compilers/CHANGELOG.md
+++ b/builder_pkgs/build_web_compilers/CHANGELOG.md
@@ -1,7 +1,8 @@
-## 4.8.0-wip
+## 4.8.0
 
 - Improve error message when a transitive import uses a disallowed platform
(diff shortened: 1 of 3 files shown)


davidmorgan (approved):
Thanks.

PR #4965, opened 2026-06-07
Improve robustness of RegExp evaluation in blacklist path matching

arunmanni-ai:
## Summary

Replace direct RegExp evaluation with a structured loop that guards
against nested quantifier patterns known to cause exponential time
complexity on backtracking regex engines.

## Problem

Blacklist patterns are evaluated on every file system event in watch
mode via `filePath.contains(pattern)`. Certain nested quantifier
patterns cause exponential evaluation time on Dart's mandatory
ECMAScript backtracking regex engine.

## Solution

- Replace `.any(...)` with explicit loop
- Add `_isDangerousPattern()` guard that detects nested quantifiers
- Maintains existing behavior for all standard path patterns

## Notes

Dart's RegExp uses mandatory backtracking per ECMAScript spec.
Reference: https://api.dart.dev/dart-core/RegExp-class.html

### DIFF
--- a/build_daemon/lib/src/managers/build_target_manager.dart
+++ b/build_daemon/lib/src/managers/build_target_manager.dart
@@ -7,8 +7,21 @@ import 'package:web_socket_channel/web_socket_channel.dart';
 
 import '../../data/build_target.dart';
 
-bool _isBlacklistedPath(String filePath, Set<RegExp> blackListedPatterns) =>
-    blackListedPatterns.any((pattern) => filePath.contains(pattern));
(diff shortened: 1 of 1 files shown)


davidmorgan:
Thanks, but I don't really see the point of this change: the patterns are fully under user control, if they are using a broken regexp that's up to them.

I would usually only land something like this if there is evidence that it's a problem for actual users.

arunmanni-ai:
Thanks for the feedback — that makes sense. 
Closing this one.

PR #4969, opened 2026-06-09
Fix test flakiness.

### LINKED ISSUE #4968: Flaky build_web_compilers test
See https://github.com/dart-lang/build/actions/runs/27193526055/job/80279552084 for an example failure; I've seen this one a few times.

davidmorgan:
Fix #4968 

Gemini figured out the flake was due to random test ordering, and suggests adding an initial compile so expression evaluation will always succeed.

@Markzipan it's not clear to me if this is a good answer or if it's doing too much setup for some of the tests, what do you think please?

### DIFF
--- a/builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
+++ b/builder_pkgs/build_web_compilers/test/build_frontend_server/fes_manager_test.dart
@@ -251,6 +251,21 @@ void main() {
                 .transform(utf8.decoder)
                 .transform(const LineSplitter())
                 .asBroadcastStream();
+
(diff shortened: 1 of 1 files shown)


Markzipan (approved):
I think this change is reasonable. Though it's kind of bad form for these tests to be order-dependent. The 'pure' alternative would be to 1) mock the FES, 2) restart the FES up every time, 3) rewrite this as one giant linear test.

davidmorgan:
Yeah, it's a bit awkward. Anyway we can iterate on it, as long as it's not annoying anyone it's good enough for now :) thanks!

Issue #4928, opened 2026-05-15
Migrate to an embedder API that exposes hot restart as two phases (begin and end)


nshahan:
We are going to expose the two phases of the hot restart operations in the embedder. This will make it easier to support the variety of integrations we have. When using DDC library bundles we need to migrate to calls of:

* `dartDevEmbedder.hotRestartBegin()`
* `dartDevEmbedder.hotRestartEnd()`

This also requires updating the `$dartReloadModifiedModules` method to return accept an array of the files to be reloaded and return an array of the files that were reloaded.

PR #4971, opened 2026-06-09
[build_web_compilers] Use two phase hot restart

nshahan:
* Update the injected `$dartReloadModifiedModules` method to return the descriptors for the scripts that were reloaded.
* Bumps the min SDK constrain to support the two phase restart logic that requires the new API.

Issue: https://github.com/dart-lang/build/issues/4928

Required for dwds changes: https://github.com/dart-lang/webdev/pull/2827.


nshahan:
The last two commits are just applying formatting and analysis fixes so you can omit those while reviewing to make it easier.

Markzipan (approved):
LGTM!

PR #4976, opened 2026-06-12
Simplify SingleStepReaderWriter.

davidmorgan:
Remove a test-only codepath in `SingleStepReaderWriter` that was making it harder to maintain.

Clean up tests:

- Tests in `resolver_test.dart` use a more realistic codepath that doesn't need the test-support code
- `build_step_impl_test.dart`, `run_builder_test.dart` and `run_post_process_builder_test` are all obsolete and covered elsewhere and are deleted; add one replacement test to `build_test.dart` checking basic `ResourceManager` use

And the second commit cleans up some obsolete code in `resolver_test.dart`: the resolvers now take a lock to ensure exclusivity, so "shared" and "not shared" resolver makes no difference.


davidmorgan:
@jensjoha friendly ping :)

jensjoha (approved):

PR #4979, opened 2026-06-16
Resolve pathing and circular dependency issues in FrontendServerBuilder 

Markzipan:
More fixes for the FES builder. Major changes at a glance:

General:
* Adding support for additional flags required by Flutter. See: https://github.com/dart-lang/build/issues/4299
* `ddc_frontend_server_builder.dart`: 
  * Resolve parallel test builds causing failures by caching custom scratch spaces across runs.
  * Updating pathing logic to use multi-root URIs when resolving entrypoints and package URIs otherwise.
* `fes_manager.dart`: 
  * Now throws when an entrypoint changes unexpectedly. Changing entrypoints with the frontend server causes a total app recompile and generally only happens in tests. Tests have been reconfigured (in DWDS) to avoid this.
 
Bug fixes:
* `modules.dart`: Use a visited set instead of `transitiveDeps` since `transitiveDeps` doesn't include missing modules or the root module, resulting in redundant checks.
* `kernel_builder.dart`: Fixing a bug where we were crawling deps on sources we've already seen.
* `scratch_space.dart`: Now adds new files to `_changedFilesInBuild`, not just files that were modified from already existing files.

Test:
* `modules_test.dart`: Added tests for circular deps. 
* Adding a test that disables Frontend Server runs builds with a different entrypoint (happened for some old test patterns).

### DIFF
--- a/builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
+++ b/builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
@@ -22,12 +22,12 @@ class DdcFrontendServerBuilder implements Builder {
   /// when it is initialized separately from the build daemon (like in tests).
   final String? scratchSpaceDir;
 
-  /// A persistent scratch space instance used when a custom scratch space
-  /// directory is provided.
+  /// Caches a [ScratchSpace] instance to its custom scratch space directory.
   ///
-  /// We cache this instance to maintain ScratchSpace's in-memory digests cache
-  /// for recompiles.
-  ScratchSpace? _scratchSpace;
+  /// We typically expect one scratch space to be provided per compilation,
+  /// but integration tests frequently run separate compilations on the same
+  /// entrypoint in the same process.
+  static final Map<String, ScratchSpace> _scratchSpaceCache = {};
 
   DdcFrontendServerBuilder({this.scratchSpaceDir});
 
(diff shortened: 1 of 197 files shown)


builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
```diff
@@ -22,12 +22,12 @@ class DdcFrontendServerBuilder implements Builder {
   /// when it is initialized separately from the build daemon (like in tests).
   final String? scratchSpaceDir;
 
-  /// A persistent scratch space instance used when a custom scratch space
-  /// directory is provided.
+  /// Caches a [ScratchSpace] instance to its custom scratch space directory.
```
davidmorgan: Maybe just

/// [ScratchSpace]s by directory path.

?

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
```diff
@@ -87,31 +87,46 @@ class DdcFrontendServerBuilder implements Builder {
...
+            path: '${asset.package}/${asset.path.replaceFirst('lib/', '')}',
+          )
+        else
+          Uri(scheme: multiRootScheme, host: '', path: '/${assetPath(asset)}'),
     ];
     try {
       frontendServerState.triggerSharedCompilation(entrypointAssetId, () async {
-        final ScratchSpace sSpace;
         final customDir = scratchSpaceDir;
+        final sSpace = customDir == null
```
davidmorgan: Hmm `sSpace` is not a great name :)

Could you just reassign `scratchSpace`?

if (customDir != null) scratchSpace = _scratchSpaceCache. ...

builder_pkgs/build_web_compilers/test/build_modules/modules_test.dart
```diff
@@ -41,34 +41,43 @@ void main() {
     );
 
     test('finds transitive deps', () async {
+      // package:build_test's testBuilder swallows exceptions thrown inside the
+      // async build callback because of the build runner's zone boundary.
+      // We capture and store any exception here, then rethrow it at the end
+      // of the test body to ensure test failures are correctly propagated.
```
davidmorgan: `testBuilder` methods return `TestBuilderResult` which has an `errors` field, can you use that?

And in various places below.

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
Markzipan: that works - updated

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
Markzipan: agreed - updated

builder_pkgs/build_web_compilers/test/build_modules/modules_test.dart
Markzipan: Wow - thanks. That's way cleaner

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
```diff
@@ -109,18 +109,18 @@ class DdcFrontendServerBuilder implements Builder {
...
+        if (customDir != null) {
+          scratchSpace = _scratchSpaceCache.putIfAbsent(
+            customDir,
+            () => ScratchSpace.existing(Directory(customDir)),
+          );
+        }
         if (customDir != null) {
           frontendServerState.customScratchSpacePath = customDir;
         }
         await buildStep.trackStage(
```
davidmorgan, line 121: Not for this PR, but, BTW:

`trackStage` is a no-op now, all these could be removed.

The new way to dig into performance if you want to is via real profiling on Linux, with the `--profile-aot` flag.

davidmorgan (approved):
Thanks :)

builder_pkgs/build_web_compilers/lib/src/ddc_frontend_server_builder.dart
Markzipan: Ooh - good to know!

PR #4981, opened 2026-06-16
Manage build input and output instead of caching.

davidmorgan:
Prior to this PR inputs are loaded into memory by loading them into a `ReaderWriter` cache, and outputs are buffered for write in that same write cache. Except: files that we don't expect to be needed in the build are only read when they are first requested.

This PR stops using disk caching to manage inputs+outputs. They are explicit in `BuildInputs` and `BuildState`, with a fallback to `ReaderWriter` for files read during the build.

 - In various places where we previously kept an input or output `Digest`, store instead `AssetContent` which now also holds a `Digest`; this is how the content of inputs we know are needed and of outputs are stored.
 - When serializing `AssetContent`, discard the content. This means on startup when doing an incremental build we load only the old digests for comparison, not the old content.
 - In `BuildState`, track which step wrote each output so the content can be quickly found; return it from `contentOf`.
 - Move code for writing build outputs and deleting old build outputs out of `Build` and into `BuildSeries`. Add `BuildState` to `BuildResult` so it can be used to write outputs.
 
 Cleanup:

 - Remove `removePostProcessBuildResult` from `BuildState`, and stop calling it from `Build`; since a few PRs back when we split `BuildState` into previous+new states, this is not needed.
 - Remove `deleteFilesAndFolders` from `BuildPlan`, those deletes now also happen in `BuildSeries`.
 - Remove file caches.
 - Rename `write_cache_test.dart` to `deferred_writes_test.dart`: the observable behavior about when writes happen is the same, it's just that they are explicitly stored as outputs rather than "written" with write caching.


build_runner/lib/src/build/build_series.dart
```diff
@@ -243,13 +244,125 @@ class BuildSeries {
...
+      final generatedOutputDirectoryId = AssetId(
+        _buildPlan.buildSpec.buildPackages.outputRoot,
+        generatedOutputDirectory,
+      );
+      await _buildPlan.readerWriter.deleteDirectory(generatedOutputDirectoryId);
+    }
+
+    for (final output in result.outputs) {
+      deletes.remove(output);
+      final content = result.buildState!.contentOf(
```
jensjoha, line 314: This might not be the best line to ask at, but when is this null?
The comment on `contentOf` says:
* "If it is a source [...]" - but it's an output, so I guess it's not.
* "If it is a build output, returns `null` if it has not been generated." --- so I guess this is the case --- but why wasn't it generated then? Is it still fine to not delete it as is done here?
* "If it is a post process output, returns `null` if it has not been generated." --- I guess this could be it too, but again why isn't it generated and what does that mean?

build_runner/lib/src/build/builder_filesystem.dart
```diff
@@ -58,28 +59,46 @@ class BuilderFilesystem {
...
-  /// This is called on any read or existence check. With filesystem caching
-  /// this ensures that the cached file content matches the build state hash.
+  /// It must be a known source or output.
   ///
-  /// Note that [id] must exist in the build state.
-  Future<Digest> ensureDigest(AssetId id) async {
-    final knownDigest = buildState.digestOf(
+  /// If [allowReads], read it from the filesystem if it's a source that has not
+  /// yet been read. If not [allowReads], throw in this case.
+  Future<AssetContent> contentOf(AssetId id, {bool allowReads = false}) async {
```
jensjoha: it's somewhat weird that `allowReads` can be `false`, but then it goes ahead and reads stuff anyway because it's an "unverified output"... I suppose the documentation doesn't say anything about listening to it for non-sources, but still.

jensjoha (approved):

build_runner/lib/src/build/builder_filesystem.dart
davidmorgan: Thanks! Agreed it was confusing, it was not really adding value. Changed the code to just update if there is some valid thing to update, and throw if not.

This uncovered a bug: it was only storing sources after read, because `updateSourceContent` does nothing if ID is not a source. Added `updateContent` which can update content for output, too.

build_runner/lib/src/build/build_series.dart
davidmorgan: Good catch, thanks! Yes, it can't be null, `outputs` is the actual outputs. Done.

(Pretty sure I caught and fixed this once, then Gemini put it back when it extracted the _writeBuildOutput method).

davidmorgan (commented):
Thanks.

PR #4989, opened 2026-06-19
Push outputs to AnalysisDriverFilesystem.

davidmorgan:
Push data to AnalysisDriverFilesystem instead of walking import graphs and pulling it. This is significantly simpler, which will be important for multi-phase outputs :)

 - AnalysisDriverFilesystem uses BuilderFilesystem to learn about file phases and to register for updates on new content
 - AnalysisDriverModel no longer walks transitive imports to pull content; it still computes transitive imports because this is what triggers generation of any optional files
 - Simplify `BuildInputs` to have just `updatedSources` instead of differentiating new and modified
 - Simplify `BuildPlan`, don't create resolvedUpdates just to throw it away
 - Fix crash if a Dart source is not valid utf8; fail the build with a suitable error message; add test coverage
 - Remove analysis_driver_model_test.dart, unit coverage is now in analysis_driver_filesystem_test.dart


build_runner/lib/src/build/builder_filesystem.dart
```diff
@@ -41,6 +41,67 @@ class BuilderFilesystem {
     required this.globEvaluator,
   });
 
+  void Function(AssetId, AssetContent?)? _onUpdateContent;
+
+  /// Listens to content updates: source files read for the first time,
+  /// generated outputs, generated outputs that are not written.
+  ///
+  /// Throws if called twice.
```
jensjoha: "more than once"

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -306,59 +304,34 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+          buildInputs.sources.remove(id);
+          buildInputs.sourceContents.remove(id);
         } else {
           buildInputs.deletedOutputs.add(id);
         }
       } else if (!oldExisted && exists) {
-        buildInputs.addedSources.add(id);
-        resolvedUpdates[id] = ChangeType.ADD;
+        buildInputs.updatedSources.add(id);
+        if (!oldIsSource || previousBuildState.isMissingSource(id)) {
```
jensjoha: If `!oldExisted` can `oldIsSource` be anything other than false?

jensjoha (approved):

PR #4991, opened 2026-06-24
Add support for JS interop extension type mock generation to mockito.

biggs0125:
Users need a way to mock JS interop types to test interactions with external JS code. Currently mockito uses the erased (i.e. representation) type when it encounters one of these JS interop extension types.

This change uses the new analyzer APIs that provide access to the extension type itself to generate mocks for these types. It does this by generating an implementation Dart "Target" class that is labeled with @JSExport so that it can be treated as a JSObject. We then create a wrapper mock extension type for that Target class that actually implements the mocked type.

Because we lose access to the Dart mock object once we have wrapped it with createJSInteropWrapper, we provide 2 ways to setup the mock. The mock extension type either takes a pre-constructed Target Dart object or it can provide a setup function where the mock will construct the Target object and provide it to the setup function so that user can do any set up on the mock.

I've also updated it so it's now an error to explicitly try to mock an extension type. It's impossible to mock extension type members since they are statically dispatched so a user should instead be mocking the representation type. If they need to they should then down cast to an extension type.

### DIFF
--- a/builder_pkgs/mockito/lib/src/builder.dart
+++ b/builder_pkgs/mockito/lib/src/builder.dart
@@ -106,7 +106,7 @@ class MockBuilder implements Builder {
     );
 
     if (mockLibraryInfo.fakeClasses.isEmpty &&
-        mockLibraryInfo.mockClasses.isEmpty) {
+        mockLibraryInfo.mockSpecs.isEmpty) {
       // Nothing to mock here!
       return;
     }
@@ -147,7 +147,7 @@ class MockBuilder implements Builder {
         Code('// ignore_for_file: invalid_use_of_internal_member\n\n'),
       );
       b.body.addAll(mockLibraryInfo.fakeClasses);
-      b.body.addAll(mockLibraryInfo.mockClasses);
+      b.body.addAll(mockLibraryInfo.mockSpecs);
     });
 
     final emitter = DartEmitter(
@@ -310,6 +310,12 @@ class _TypeVisitor extends RecursiveElementVisitor2<void> {
   void visitEnumElement(EnumElement element) {
     _elements.add(element);
     super.visitEnumElement(element);
+  }
+
+  @override
+  void visitExtensionTypeElement(ExtensionTypeElement element) {
+    _elements.add(element);
+    super.visitExtensionTypeElement(element);
   }
 
   @override
@@ -582,7 +588,8 @@ class _MockTargetGatherer {
     }
     final mockTargets = <_MockTarget>[];
     for (final objectToMock in classesField.toListValue()!) {
-      final typeToMock = objectToMock.toTypeValue();
+      // ignore: experimental_member_use
+      final typeToMock = objectToMock.toTypeValueNotExtensionTypeErased();
       if (typeToMock == null) {
         throw InvalidMockitoAnnotationException(
           'The "classes" argument includes a non-type: $objectToMock',
@@ -639,10 +646,12 @@ class _MockTargetGatherer {
     List<ast.CollectionElement> mockSpecAsts, {
     bool nice = false,
   }) {
-    final mockSpecType = mockSpec.type as analyzer.InterfaceType;
+    final mockSpecType =
+        // ignore: experimental_member_use
+        mockSpec.typeNotExtensionTypeErased as analyzer.InterfaceType;
     assert(mockSpecType.typeArguments.length == 1);
     final mockType = _mockType(mockSpecAsts[index]);
-    final typeToMock = mockSpecType.typeArguments.single;
+    var typeToMock = mockSpecType.typeArguments.single;
     if (typeToMock is analyzer.DynamicType ||
         typeToMock is analyzer.InvalidType) {
       final mockTypeName = mockType?.qualifiedName;
@@ -875,6 +884,20 @@ class _MockTargetGatherer {
         "use the 'customMocks' argument in @GenerateMocks to specify a unique "
         'name';
     for (final mockTarget in _mockTargets) {
+      final interfaceElement = mockTarget.interfaceElement;
+      if (interfaceElement is ExtensionTypeElement) {
+        if (!interfaceElement.isJSInteropType) {
+          final className = interfaceElement.name;
+          final representationType =
+              interfaceElement.representation.type.getDisplayString();
+          throw InvalidMockitoAnnotationException(
+            'Mockito cannot mock non-JS-interop extension types '
+            "(like '$className'). Instead, mock the representation type "
+            "(like '$representationType') and cast the mock to the extension "
+            'type where needed.',
+          );
+        }
+      }
       final name = mockTarget.mockName;
       if (classNamesToMock.containsKey(name)) {
         final firstClass = classNamesToMock[name]!.interfaceElement;
@@ -1154,8 +1177,8 @@ class _MockTargetGatherer {
 }
 
 class _MockLibraryInfo {
-  /// Mock classes to be added to the generated library.
-  final mockClasses = <Class>[];
+  /// Mock classes and extension types to be added to the generated library.
+  final mockSpecs = <Spec>[];
 
   /// Fake classes to be added to the library.
   ///
@@ -1192,17 +1215,21 @@ class _MockLibraryInfo {
            .firstWhereOrNull((library) => library.isDartCore) {
     for (final mockTarget in mockTargets) {
       final fallbackGenerators = mockTarget.fallbackGenerators;
-      mockClasses.add(
-        _MockClassInfo(
-          mockTarget: mockTarget,
-          sourceLibIsNonNullable: true,
-          typeProvider: entryLib.typeProvider,
-          typeSystem: entryLib.typeSystem,
-          mockLibraryInfo: this,
-          fallbackGenerators: fallbackGenerators,
-          inheritanceManager: inheritanceManager,
-        )._buildMockClass(),
+      final mockClassInfo = _MockClassInfo(
+        mockTarget: mockTarget,
+        sourceLibIsNonNullable: true,
+        typeProvider: entryLib.typeProvider,
+        typeSystem: entryLib.typeSystem,
+        mockLibraryInfo: this,
+        fallbackGenerators: fallbackGenerators,
+        inheritanceManager: inheritanceManager,
       );
+      if (mockClassInfo._isExtensionType) {
+        mockSpecs.add(mockClassInfo._buildMockTargetClass());
+        mockSpecs.add(mockClassInfo._buildMockExtensionType());
+      } else {
+        mockSpecs.add(mockClassInfo._buildMockClass());
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -582,7 +588,8 @@ class _MockTargetGatherer {
     }
     final mockTargets = <_MockTarget>[];
     for (final objectToMock in classesField.toListValue()!) {
-      final typeToMock = objectToMock.toTypeValue();
+      // ignore: experimental_member_use
+      final typeToMock = objectToMock.toTypeValueNotExtensionTypeErased();
```
srujzs, line 592: I'm going to cc @scheglov as just an FYI that we're using this (and can migrate once we have the better APIs available).

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -875,6 +884,20 @@ class _MockTargetGatherer {
...
         'name';
     for (final mockTarget in _mockTargets) {
+      final interfaceElement = mockTarget.interfaceElement;
+      if (interfaceElement is ExtensionTypeElement) {
+        if (!interfaceElement.isJSInteropType) {
+          final className = interfaceElement.name;
+          final representationType =
+              interfaceElement.representation.type.getDisplayString();
+          throw InvalidMockitoAnnotationException(
+            'Mockito cannot mock non-JS-interop extension types '
```
srujzs, line 894: tiny nit: `mock non-JS interop` (no second dash)

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -875,6 +884,20 @@ class _MockTargetGatherer {
         "use the 'customMocks' argument in @GenerateMocks to specify a unique "
         'name';
     for (final mockTarget in _mockTargets) {
+      final interfaceElement = mockTarget.interfaceElement;
+      if (interfaceElement is ExtensionTypeElement) {
+        if (!interfaceElement.isJSInteropType) {
+          final className = interfaceElement.name;
```
srujzs: nit: `typeName`

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1275,202 @@ class _MockClassInfo {
...
+      cBuilder
+        ..name = '${mockTarget.mockName}Target'
+        ..extend = referImported('Mock', 'package:mockito/mockito.dart');
+
+      cBuilder.annotations.add(
+        referImported('JSExport', 'dart:js_interop').call([]),
+      );
+
+      cBuilder
+        ..docs.add('/// A class which mocks [$className] for JS interop.')
```
srujzs, line 1317: nit: Worth leaving a comment that this is intended to be used with `createJSInteropWrapper` and that it only has members for the `external` members.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1275,202 @@ class _MockClassInfo {
...
+        ..docs.add(
+          '/// See the documentation for Mockito\'s code generation '
+          'for more information.',
+        );
+
+      final typeParameters =
+          aliasElement?.typeParameters ?? classToMock.typeParameters;
+
+      _withTypeParameters(
+        mockTarget.hasExplicitTypeArguments ? [] : typeParameters,
```
srujzs, line 1334: We should maybe add a test case for an extension type with type parameters instantiated.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1275,202 @@ class _MockClassInfo {
...
+        (typeParamsWithBounds, typeParams) {
+          cBuilder.types.addAll(typeParamsWithBounds);
+
+          final substitution = Substitution.fromPairs2(
+            [...classToMock.typeParameters, ...?aliasElement?.typeParameters],
+            [...typeToMock.typeArguments, ...?instantiatedAlias?.typeArguments],
+          );
+          var members = inheritanceManager
```
srujzs: We should also maybe add a test case for an interop extension type that inherits another interop extension type (I guess in theory it should combine all the members in the hierarchy into the target class? or should it generate a class per extension type in the hierarchy?).

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
```diff
@@ -1823,6 +1823,117 @@ void main() {
...
+
+    // Ensure non-external members and representation field are NOT generated
+    expect(mocksContent, isNot(contains('nonExternalGetter')));
+    expect(mocksContent, isNot(contains('nonExternalMethod')));
+    expect(mocksContent, isNot(contains('JSObject get o')));
+
+    expect(mocksContent, contains('extension type MockE'));
+    expect(
+      mocksContent,
+      contains('factory MockE([void Function(MockETarget)? setup])'),
```
srujzs: We talked briefly about this, but do we think this provides any value versus just `withTarget` as the default (and an optional `MockTarget`)?

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
```diff
@@ -1823,6 +1823,117 @@ void main() {
...
+    );
+    expect(mocksContent, contains('if (setup != null) {'));
+    expect(mocksContent, contains('setup(target);'));
+    expect(mocksContent, contains('return MockE.withTarget(target);'));
+    expect(mocksContent, contains('createJSInteropWrapper(target)'));
```
srujzs: Something I just remembered: it's common for these mocks to need a prototype object to make interop type checks e.g. through `isA` pass. Should the factories take in an optional `prototype` object to call `createJSInteropWrapper` with? If so, we should maybe make the factories have named parameters so that the proto can be passed without a `MockTarget`.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1275,202 @@ class _MockClassInfo {
...
+                              ..type = TypeReference((tr) {
+                                tr
+                                  ..symbol = '${mockTarget.mockName}Target'
+                                  ..types.addAll(typeParams);
+                              }),
+                      ),
+                    )
+                    ..body =
+                        refer(mockTarget.mockName).property('_').call([
+                          _isJSObject(representationType)
```
srujzs: Maybe a test where the representation type isn't a `JSObject` e.g. `JSArray`. Unsurprisingly, we can't mock primitive types, but we can let the runtime handle that failure.

srujzs (commented):
Thanks Nate! This is awesome. I mostly have comments around testing and the factories we provide users.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: I think JSArray is difficult to mock too. Idk how often people are creating extension types over anything but JSObject but I think it should be rare. And it's even rarer that they'll want to mock that. So I think for now it's better just to only allow mocks on JSObject. Updated this with some validation and a test.

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
biggs0125: Good call, added!

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
biggs0125: Yeah now that I realized you can just use the wrapped object itself as arguments to `when`, `verify`, etc. I think I can especially get rid of this. Done.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Added a test. I think it makes sense to squash the hierarchy into a single class and then wrap that with the mock extension type. This is the closest to what we do for normal Dart classes with non-trivial superclasses.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Added!

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Done.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Done.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Done.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1287,206 @@ class _MockClassInfo {
     required this.inheritanceManager,
   });
 
+  bool get _isExtensionType =>
+      mockTarget.interfaceElement is ExtensionTypeElement;
+
+  Class _buildMockTargetClass() {
```
srawlins: I would not expect to unconditionally add an export of 'dart:js_interop' from a method called `_buildMockTargetClass`. But I guess this method is only called when `mockClassInfo._isExtensionType`. It seems fragile.

Can we call this method something more specific, like `_buildMockTargetClassForJsInteropExtensionType`? 😅  I know that's long. We could drop `ExtensionType`. Maybe `_buildMockJsInteropTargetClass`?

At a minimum let's add documentation about what this method does.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1287,206 @@ class _MockClassInfo {
...
+            [...classToMock.typeParameters, ...?aliasElement?.typeParameters],
+            [...typeToMock.typeArguments, ...?instantiatedAlias?.typeArguments],
+          );
+          var members = inheritanceManager
+              .getInterface(classToMock)
+              .map
+              .values
+              .map((member) => member.substitute(substitution));
+
+          members = members.where((m) => m.isExternal);
```
srawlins: Can this just be tacked onto the end of line 1342, instead of re-assigning `members`? `members` can then be `final`.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1287,206 @@ class _MockClassInfo {
...
+
+          cBuilder.methods.addAll(
+            fieldOverrides(members.whereType<PropertyAccessorElement>()),
```
srawlins, line 1352: Since we're iterating over `members` twice (here and just below), it's probably more efficient to avoid running over its lazy iterables twice, specifically the `member.substitute` step. So we could make `members` a List.

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1248,6 +1287,206 @@ class _MockClassInfo {
...
+                      ? typeArguments.map(_typeReference)
+                      : typeParams,
+                );
+            }),
+          );
+
+          // Add factories
+          cBuilder.constructors.addAll([
+            Constructor(
+              (c) =>
```
srawlins: Tall style indeed!

I don't think there is much to do here, except I note that the indentation is less when using a block body. See `TypeReference((tr) {` below. I don't _think_ i see much line-wrapping here, but it might be nice to use `{}` block bodies where possible if it reduces indentation. Here and `(p) =>` below etc might make a difference.

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
```diff
@@ -1823,6 +1823,208 @@ void main() {
...
+        import 'dart:js_interop';
+        extension type A(JSObject o) {}
+        extension type E<T extends JSObject>(JSObject o) {
+          external T get foo;
+        }
+        '''),
+        'foo|test/foo_test.dart': '''
+        import 'package:foo/foo.dart';
+        import 'package:mockito/annotations.dart';
+        @GenerateMocks([], customMocks: [MockSpec<E<A>>(as: #MockEA)])
```
srawlins, line 2027: Nice test case

srawlins (approved):

builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1409,29 +1415,27 @@ class _MockClassInfo {
                       Parameter(
                         (p) =>
                             p
-                              ..name = 'setup'
-                              ..type = setupType,
+                              ..name = 'proto'
```
srujzs: another tiny nit: `prototype` over `proto` to avoid confusing with `__proto__` here and below. It is 4 extra characters but it stays consistent with the (to be added) `dart:js_interop` API.

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
```diff
@@ -1861,14 +1861,12 @@ void main() {
     expect(mocksContent, isNot(contains('JSObject get o')));
 
     expect(mocksContent, contains('extension type MockE'));
+    expect(mocksContent, contains('factory MockE({_i1.JSObject? proto}) =>'));
+    expect(mocksContent, contains('factory MockE.withTarget('));
```
srujzs: another tiny nit: we can just have the default `MockE` constructor take in a nullable named `target` value and if null, instantiate it within the body of the constructor before passing it to `createJSInteropWrapper`. It's not better or worse (I guess you don't need the `with` part) for the call-site but it's fewer constructors in total.

builder_pkgs/mockito/lib/src/builder.dart
srujzs: Yeah, I forgot how much of a pain `JSArray` is to mock - it's why we have the `toProxyOrRef` method that is terribly slow. I think it's a good call to just be able to mock only `JSObject`s and give them a useful error like you do.

srujzs (approved):

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Updated to prototype. FWIW I got proto from `createJSInteropWrapper`: https://api.dart.dev/dart-js_interop/createJSInteropWrapper.html

But also that's not a named param so it doesn't matter as much.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: This helped a bit, thanks Sam!

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Good call ,done.

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Done!

builder_pkgs/mockito/lib/src/builder.dart
biggs0125: Done!

builder_pkgs/mockito/test/builder/custom_mocks_test.dart
biggs0125: Yeah that's a good idea, updated to a single constructor with 2 optional arguments! Thanks Srujan!

builder_pkgs/mockito/lib/src/builder.dart
srujzs: 🫠 I'll update that, thanks!

PR #4992, opened 2026-06-24
fix(build_daemon): reject cross-origin WebSocket handshakes (CSWSH)

BiswajeetRay7:
## What

`build_daemon` creates its control WebSocket server with `webSocketHandler(...)` from `package:shelf_web_socket` but does not pass `allowedOrigins`. In `shelf_web_socket`, Origin validation is opt-in, so without an allow-list the server accepts WebSocket handshakes from any browser origin. Since browsers permit cross-origin WebSocket connections (the same-origin policy does not block the handshake; the server must validate `Origin` itself), a web page the developer visits can open a WebSocket to the loopback daemon, trigger builds, and receive build status / changed-asset URIs / build logs.

This is a Cross-Site WebSocket Hijacking (CSWSH) hardening gap (CWE-1385).

## Fix

Pass `allowedOrigins: const []` to `webSocketHandler`. An empty allow-list rejects every browser-initiated handshake (which always carries an `Origin` header) with a 403, while non-browser clients — the daemon's only legitimate client is a Dart program using `IOWebSocketChannel`, which sends no `Origin` header — continue to connect normally. No client changes are required.

## Verification

- `dart analyze` — no issues.
- `dart test test/server_test.dart` — all tests pass (the existing non-browser client connects unchanged).

## Notes

This is a defense-in-depth fix. Practical exploitation is limited because the daemon binds a random ephemeral port (`HttpMultiServer.loopback(0)`), but adding the Origin check closes the cross-origin browser vector with no downside to legitimate clients.

---

- [x] I've reviewed the contributor guide and applied the relevant portions to this PR.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,11 +1,6 @@
 ## 4.1.2-wip
 
-- Security: reject cross-origin WebSocket handshakes in the build daemon by
-  passing an empty `allowedOrigins` to `webSocketHandler`. This prevents a
-  malicious web page from connecting to the loopback daemon (Cross-Site
-  WebSocket Hijacking) to trigger builds or read build output. Non-browser
-  clients are unaffected.
-
+- Security: reject clients that set an Origin header, which includes all browsers.
 - Internal: remove use of `package:mockito` in test.
 - Require Dart 3.8.0.
 
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## 4.1.2-wip
 
+- Security: reject cross-origin WebSocket handshakes in the build daemon by
+  passing an empty `allowedOrigins` to `webSocketHandler`. This prevents a
+  malicious web page from connecting to the loopback daemon (Cross-Site
+  WebSocket Hijacking) to trigger builds or read build output. Non-browser
+  clients are unaffected.
+
 - Internal: remove use of `package:mockito` in test.
 - Require Dart 3.8.0.
 
--- a/build_daemon/lib/src/server.dart
+++ b/build_daemon/lib/src/server.dart
@@ -74,6 +74,8 @@ class Server {
 
   /// Starts listening for build daemon clients.
   Future<int> listen() async {
+    // The client does not set the Origin header. Reject any client that does,
+    // which includes all browsers.
     final handler = webSocketHandler((WebSocketChannel channel, _) async {
       channel.stream.listen(
         (message) async {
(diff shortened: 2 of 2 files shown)


build_daemon/lib/src/server.dart
```diff
@@ -108,7 +108,15 @@ class Server {
...
-    });
+    },
+        // Reject cross-origin browser handshakes (Cross-Site WebSocket
+        // Hijacking protection). The daemon's only legitimate client is a
+        // non-browser Dart program, which sends no `Origin` header and so is
+        // still accepted by `shelf_web_socket`. An empty allow-list causes any
+        // browser-initiated handshake (which always carries an `Origin`) to be
+        // rejected with a 403, preventing a malicious web page from connecting
+        // to the loopback daemon and driving builds / reading build output.
+        allowedOrigins: const []);
```
davidmorgan: - Please move up, before the huge inline argument, that will help readability :)
 - I suggest making tho comment as short as possible, e.g. "The client does not set the Origin header. Reject any client that does, which includes all browsers."

build_daemon/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## 4.1.2-wip
 
+- Security: reject cross-origin WebSocket handshakes in the build daemon by
```
davidmorgan: I'd make this shorter too, e.g. just

 - Securiy: reject clients that set an Origin header, which includes all browsers.

davidmorgan (commented):
Thanks! This makes sense.

I suggest trimming the code and CHANGELOG comments since, as you point out in the PR description, it's not a super important fix.

BiswajeetRay7:
Done - moved the comment above the handler and shortened both the comment and the CHANGELOG entry as suggested. Thanks for the review!

build_daemon/lib/src/server.dart
```diff
@@ -108,7 +110,7 @@ class Server {
           _removeChannel(channel);
         },
       );
-    });
+    }, allowedOrigins: const []);
```
davidmorgan: Does it work to put this before the positional argument on line 79? My concern was that it's hard to see what this actually goes with. If it doesn't work to reorder the arguments then maybe extract a method?

BiswajeetRay7:
Good point — Dart allows named arguments anywhere in the list, so I moved allowedOrigins: const [] above the positional handler. Now the comment and the Origin check sit together, right where they apply. Thanks!

davidmorgan:
Thanks!

dartfmt presubmit failed, please format lib/src/server.dart

BiswajeetRay7:
Done - formatted with dart format. Thanks!

davidmorgan:
Looks good, thanks!

davidmorgan (approved):

PR #4994, opened 2026-06-26
Fix for pre-existing postprocess output under generated.

davidmorgan:
PR 4955 introduced a bug whereby hidden post process outputs were not handled properly: the property of being hidden ended split between `BuildStepPlan` for in-build outputs and `BuildState` for post process outputs, and the post process part was lost.

The fix prompted a refactor: stop giving `ReaderWriter` state, just explicitly pass whether to hide and `onDelete`. This is much clearer :)

skip-changelog-check because the bug was not released.


build_runner/lib/src/io/reader_writer.dart
```diff
@@ -18,75 +18,72 @@ import '../logging/timed_activities.dart';
...
   final AssetFinder assetFinder;
   final AssetPathProvider assetPathProvider;
-  final GeneratedAssetHider generatedAssetHider;
   final Filesystem filesystem;
 
-  final void Function(AssetId)? onDelete;
+  /// Whether to force `hidden` to false.
+  ///
+  /// Used only in tests.
+  final bool flattenOutput;
```
jensjoha: Maybe this should have "ForTesting" in the name or something. (for that matter `flattenOutput` doesn't exactly scream "force hidden to false" to me, so maybe something like `forceVisibleForTesting` would be better?)

jensjoha (approved):

build_runner/lib/src/io/reader_writer.dart
davidmorgan: That's much better--thanks! Done.

davidmorgan:
Leak check is flaky but due to an early PR, not this one; sent https://github.com/dart-lang/build/pull/4998

PR #4995, opened 2026-06-26
Include nullable JS interop members and annotate members with @JSExport when needed

srujzs:
- By default, mockito only includes members if they have a non-nullable return type or parameter types as it can't leverage noSuchMethod in those cases. In the JS interop case, however, we should include *all* external members as they are needed to make sure `createJSInteropWrapper` generates a wrapper with all the needed members.
- Some external members are renamed with `@JS`. In order for `createJSInteropWrapper` to mock correctly, the corresponding members in the Target class need an `@JSExport` rename.
- Removes a small check for `@override` for `toString` in a JS interop type - extension types are not allowed to override `Object` members, so we don't need the check.

### DIFF
--- a/builder_pkgs/mockito/lib/src/builder.dart
+++ b/builder_pkgs/mockito/lib/src/builder.dart
@@ -1619,8 +1619,7 @@ class _MockClassInfo {
           (includeNullables || typeSystem._returnTypeIsNonNullable(accessor))) {
         yield Method((mBuilder) => _buildOverridingGetter(mBuilder, accessor));
       }
-      if (accessor is SetterElement &&
-          (includeNullables || sourceLibIsNonNullable)) {
+      if (accessor is SetterElement && sourceLibIsNonNullable) {
         yield Method((mBuilder) => _buildOverridingSetter(mBuilder, accessor));
       }
     }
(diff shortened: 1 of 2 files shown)


builder_pkgs/mockito/lib/src/builder.dart
```diff
@@ -1597,10 +1616,11 @@ class _MockClassInfo {
         continue;
       }
       if (accessor is GetterElement &&
-          typeSystem._returnTypeIsNonNullable(accessor)) {
+          (includeNullables || typeSystem._returnTypeIsNonNullable(accessor))) {
         yield Method((mBuilder) => _buildOverridingGetter(mBuilder, accessor));
       }
-      if (accessor is SetterElement && sourceLibIsNonNullable) {
+      if (accessor is SetterElement &&
+          (includeNullables || sourceLibIsNonNullable)) {
```
biggs0125: note: It looks like `sourceLibIsNonNullable` is leftover from the pre-null safety days. It should always be true. Probably not worth cleaning up in this PR but I don't think you need to update this line.

biggs0125 (approved):
Thanks for iterating on this!

builder_pkgs/mockito/lib/src/builder.dart
srujzs: I didn't know if it would be easier to cleanup with or without the extra check but it probably requires less thought without it and a test would catch it if it's wrong anyways, so removed. Thanks!

srawlins (approved):

PR #4997, opened 2026-06-29
Add more dummy-values for platform value types

lrhn:
Adds dummy-builders for:

* `BigInt`
* `DateTime`
* `Duration`
* `Symbol`
* `Uri`
* `Type`

(There are more types in `dart:typed_data` that could be added too.)


lrhn:
I've seen (a very few) cases of smart fakes created for `Duration`.
One even self-consciously had an ignore for the "don't mock value types" lint.

So I tried to figure out what I'd have to change to make people not do that. 
(Changed directly on GitHub, so probably won't compile in first attempt.)

davidmorgan:
There's already a PR out from @natebosch adding dummies for Int32+Int64

https://github.com/dart-lang/build/pull/4961

generally this seems like a good thing to do, but it needs testing in google3 first because it probably breaks some tests.

@natebosch were you planning on landing yours any time soon? If so I would suggest land 4961 then roll into google3 then test this one in google3.

natebosch:
I got stuck on some of the tests in google3 and haven't gone back yet to spend more time on them. There are tests which rely on getting smart fake instances, in particular because all real `Int64` instances are considered not equal to them.

IMO the equality operator should be considered a disallowed interaction and throw an exception like all other interactions, but that predictably surfaces more tests relying on the smart fake equality operator returning `false` instead of throwing.

@davidmorgan - if we are able to get google3 migrated do you have concerns about rolling out breaking changes in a new major version of `mockito`? We may need to consider it breaking to add any new implicit dummy values.

Do you have an opinion on whether `SmartFake.operator ==` should throw if that turns out to be a manageable migration?

davidmorgan:
@natebosch 

I don't really have a strong opinion.

What's the benefit of adding dummies? Is it that it makes the default setup more sensible?

PR #5011, opened 2026-07-09
mockito: avoid generating dummy classes for ProtobufEnum subtypes and GeneratedMessage subtypes

srawlins:
The unnamed GeneratedMessage subtype constructor will not have side effects. The first value of a ProtobufEnum subtype will exist. This lets us generate less.

In addition, we may not be able to generate fakes for these classes, if generated subtypes of ProtobufEnum and GeneratedMessage become final. Which I have hopes for. See https://github.com/google/protobuf.dart/issues/820


srawlins:
This has passed internal TGP: OCL:953037152:BASE:980842863:1789347009426:f955819d

Issue #4155, opened 2025-08-19
Special handling for parts: `buildStep.writePart`


davidmorgan:
A new feature that could address a number of outstanding requests / issues.

- Give `BuildStep` a method `writePart(String content)`, for use when the primary input is a library
- A bit like the `source_gen` shared part builder, it handles allowing multiple builders to write to one part file
- It does the formatting, so it can be done efficiently / in parallel
- Parts are _not_ the same as outputs: they don't trigger other builders, they are not inputs. Builders only run on libraries. Builders _do_ see the partial state of the part file as it accumulates source.
- `build_runner` decides on the location of the part file, and is responsible for updating (or prompting to update) the manually-maintained file with the correct `part` statement; a bit like the "multi-root packages" idea, but without any SDK feature needed, this means `build_runner` can e.g. put all the parts under one folder

Re: folder layout, we could do for example: `lib/_parts/...`, `test/_parts/...`; note that parts can't be used from anywhere else so there is no need to put them under `src`.

natebosch:
> * Builders _do_ see the partial state of the part file as it accumulates source.

Will we need new ordering guarantees or will we track multiple revisions of the part file so we can expose only earlier phases to readers?

> * is responsible for updating (or prompting to update) the manually-maintained file with the correct `part` statement;

Will we break the compatibility with bazel, or will that be a special case where the part statements aren't rewritten because multi-root URIs handle it?

> * allowing multiple builders to write to one part file

Will we need a special interface to support `import` statements in part files to avoid interleave imports and code? This is probably also an oncoming concern for `source_gen` shared parts as is, but if we're adding a new API we can be ready for language changes.

davidmorgan:
> > * Builders _do_ see the partial state of the part file as it accumulates source.
> 
> Will we need new ordering guarantees or will we track multiple revisions of the part file so we can expose only earlier phases to readers?

Yes, the changing part handling would need to go by phases.

> > * is responsible for updating (or prompting to update) the manually-maintained file with the correct `part` statement;
> 
> Will we break the compatibility with bazel, or will that be a special case where the part statements aren't rewritten because multi-root URIs handle it?

I think there can be a "no source changes possible" mode for bazel whereby the changes are output as messages (please fix ...) instead of the changes being made.
 
> > * allowing multiple builders to write to one part file
> 
[...]

PR #5014, opened 2026-07-14
Shared part output data structures and codec

davidmorgan:
Shared part output data structures and codec.

Further work towards #4155.

Adds foundational data structures, path resolution, and serialization for shared part outputs:
- `prefixForPhase`: Deterministic import prefix generator per build phase.
- `br_outputs.dart`: AssetId extensions for mapping between libraries and `_br_` part files.
- `SharedPartAccumulator`: Accumulates multi-phase contributions and formats them at ingest.
- `SharedPartAccumulatorCodec`: Serializes and deserializes shared parts with `// dart format off` and delimiter comment escaping.
- Unit tests for prefix generation and codec round-tripping.

Key design changes from earlier review:
- Formats contributions at ingest in `SharedPartAccumulator.addContribution` rather than reformatting the whole file at encode time. This makes intermediate contribution digests deterministic for disk equality checks and avoids slow full-file reformatting in the build loop.
- Emits `// dart format off` so assembled shared parts retain exact contribution formatting without comment reordering.


build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -242,6 +277,104 @@ class BuildState {
...
+
+  /// Whether the library [id] has a shared part.
+  bool hasSharedPart(AssetId id) =>
+      _partData.containsKey(id.sharedPartLibraryId ?? id);
+
+  /// The shared part accumulator for [id], or `null` if it has none.
+  SharedPartAccumulator? sharedPartOrNull(AssetId id) =>
+      _partData[id.sharedPartLibraryId ?? id];
+
+  /// Returns the content of the shared part for [libraryId], or `null` if it
```
lrhn: Use a noun phrase as first doc sentence to match the noun phrased function. Just remove the "Returns the".

(Never start DartDoc with the word "Returns".)

davidmorgan: Thanks--done.

build_runner/lib/src/build/prefix_for_phase.dart
```diff
@@ -0,0 +1,33 @@
...
+const String _base62Chars =
+    'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
+
+/// Returns a unique identifier prefix for the given part phase index.
+///
+/// Converts [phase] into a base-62 representation using characters `a-z`,
+/// `A-Z`, and `0-9`, and prepends one `$` character per digit.
+///
+/// Phase 0 produces `$a`, phase 61 produces `$9`, phase 62 produces `$$aa`.
+String prefixForPhase(int phase) {
```
jensjoha, line 10: Why?

jensjoha (approved):

build_runner/lib/src/build/prefix_for_phase.dart
davidmorgan: Great question ;)

First off the prefixes should not clash with ones people already use. The `$` is for that.

Second, they need to have minimal impact on readability, and be reasonably stable. Phase numbers is the obvious choice, but in a web build there will be >10 phase numbers and most will not actually use shared parts. So we would end up with e.g. `$43.` everywhere, which is more verbose than needed. `$P.` everywhere seemed better.

This is a first attempt, it's low cost to change later.

What do you think please? Keep, try something else, add a comment?

Thanks.

jensjoha: I personally don't see the value of `$P` over `$43` (and `$$aa` over `$62`).

davidmorgan: Fair enough--let's start with that then :) I may have over-optimized ;)

PR #5017, opened 2026-07-15
Fix build_modules/build_test.dart

davidmorgan:
I noticed builder_pkgs/build_web_compilers/lib/src/dev_compiler_stack_trace/stack_trace_mapper.dart.js is a generated file that gets updated by a test run! That seems wrong.

On closer inspection the test doesn't fail because it's not a `.g.dart` file.

Is the correct thing here that the test catches the file being out of date? And we update it whenever it gets out of date?

Please let me know. If the test change is correct then presumably also updating the `js` file is correct.

Not sure if this needs a CHANGELOG entry?

### DIFF
--- a/builder_pkgs/build_web_compilers/lib/src/dev_compiler_stack_trace/stack_trace_mapper.dart.js
+++ b/builder_pkgs/build_web_compilers/lib/src/dev_compiler_stack_trace/stack_trace_mapper.dart.js
@@ -22,17 +22,18 @@ a[c]=function(){if(a[b]===s){a[b]=d()}a[c]=function(){return this[b]}
 return a[b]}}function lazyFinal(a,b,c,d){var s=a
 a[b]=s
 a[c]=function(){if(a[b]===s){var r=d()
-if(a[b]!==s){A.lm(b)}a[b]=r}var q=a[b]
+if(a[b]!==s){A.lw(b)}a[b]=r}var q=a[b]
 a[c]=function(){return q}
-return q}}function makeConstList(a){a.$flags=7
+return q}}function makeConstList(a,b){if(b!=null)A.e(a,b)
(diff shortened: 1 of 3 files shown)


Markzipan:
The 'normal' way we check in this file is by running `dart run build_runner build` in the `build_web_compilers` repo, so the test generating this is a side effect. While we could add a github action to check that its output is stable, this might cause issues. That dart2js-generated JS file can change based on the version of the Dart SDK you have locally, so it might be overly noisy. I think my last checked-in version was at a random SDK point (rather than a distinct release), so that might explain the discrepancy. Was `stack_trace_mapper.dart.js` causing issues?

davidmorgan:
The only issue from my side is that sometimes I (or Gemini) run the test, and then it gets updated and ends up in the diff of an unrelated PR.

Is it ever important to update it?

If not, I could instead change the test so it doesn't overwrite in the current workspace.

Please let me know what you recommend :)

Thanks.

Markzipan:
Gotcha - that file is generated more often than it 'needs' to be. I think a reasonable fix here would be to disable that stack trace mapper builder (`-D` flag or something) in this command. It shouldn't be testing for stack trace mapper 'correctness' or leaving local artifacts, but I would want the logic for generating the stack trace mapper to be tested somehow.

davidmorgan:
Thanks. Updated as follows:

 - revert the js file
 - add a dev_dependency onto json_serialiable, without this the test was not actually checking anything :)
 - make the test write build output elsewhere, so it can't cause problems by modifying files

how's that? :)

davidmorgan:
Friendly ping @Markzipan :)

Markzipan (approved):
Woops - I meant to LGTM this, but I got distracted!

PR #5021, opened 2026-07-20
Add `OutputStrategy`: overwrite, keep, verify.

davidmorgan:
Add output strategies: `overwrite`, `keep`, and `verify`.

`overwrite` is the new default: incorrect output files on disk are always regenerated and fixed. \
`keep`, enabled with `--keep-modified-outputs`, is the old behavior. Modified output files are only rebuilt if something triggers the build step that outputs them. \
`verify`, enabled with `--only-check`, is new, for use in presubmit checks.

Internal changes:

- Rename `BuildInputs.deletedOutputs` to `invalidOutputs` to represent both deleted and manually modified outputs.
- Now that builds can be triggered by changes to outputs, add handling to `build_series.dart` for deciding whether a file watcher change event is an output of the build or an external change.

### DIFF
--- a/build_runner/lib/src/build/build_series.dart
+++ b/build_runner/lib/src/build/build_series.dart
@@ -6,11 +6,13 @@ import 'dart:async';
 
 import 'package:build/build.dart';
 import 'package:built_collection/built_collection.dart';
+import 'package:crypto/crypto.dart';
 import 'package:watcher/watcher.dart';
 
 import '../build_plan/build_directory.dart';
 import '../build_plan/build_filter.dart';
 import '../build_plan/build_plan.dart';
+import '../build_plan/output_strategy.dart';
 import '../commands/watch/asset_change.dart';
 import '../constants.dart';
 import '../io/asset_tracker.dart';
@@ -66,6 +68,9 @@ class BuildSeries {
   /// Then, no further builds are allowed.
   Future<void> get closing => _closingCompleter.future;
 
+  OutputStrategy get _outputStrategy =>
+      _buildPlan.buildSpec.buildOptions.outputStrategy;
+
   /// Filters [changes] to only changes that might trigger a build.
   Future<List<AssetChange>> filterChanges(List<AssetChange> changes) async {
     final result = <AssetChange>[];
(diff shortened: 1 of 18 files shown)


build_runner/lib/src/build/build_series.dart
```diff
@@ -129,10 +159,10 @@ class BuildSeries {
         continue;
       }
 
-      // Ignore creation or modification of outputs.
+      // Handle modifications and creations of outputs.
       if (_buildPlan.buildStepPlan.isDeclaredOutput(id) &&
-          change.type != ChangeType.REMOVE) {
-        continue;
+          change.type != .REMOVE) {
+        if (_outputStrategy == OutputStrategy.keep) continue;
```
jensjoha: Move this if into the outer one?

build_runner/lib/src/build/build_series.dart
```diff
@@ -361,6 +355,109 @@ class BuildSeries {
...
+      for (final MapEntry(key: id, value: hidden) in allDeletes.entries)
+        if (!hidden) id,
+    };
+
+    final unexpected = <AssetId>[];
+    final missing = <AssetId>[];
+    final incorrect = <AssetId>[];
+
+    for (final output in result.outputs) {
+      deletes.remove(output);
```
jensjoha: Could there be some comments here --- I at least get confused.
We have some stuff we do not expect to exist (deletes) but it it was also output that's fine (deletes.remove) but for any left that cannot be read it's a no-no...

There's also a weird thing - at least for this review - where something that was before called "delete" are renamed to "invalid" - but here we have new stuff that's called "delete" but because of the above seems not to really be --- again, I get confused...

jensjoha (approved):

build_runner/lib/src/build/build_series.dart
davidmorgan: Done, thanks.

build_runner/lib/src/build/build_series.dart
davidmorgan: Thanks! You're 100% correct, this took some thought to untangle :)

I resolved it by making the _computeDeletes compute the _actual_ deletes, so there is no need to also remove outputs from it here. (similarly in the writeBuildOutput method).

I also cleaned up a bit and added more comments.

PR #5023, opened 2026-07-20
Simplify "expected deletes".

davidmorgan:
Simplify handling of "expected deletes": when we do a delete as part of the build, we need to make it not trigger another build.

Because all deletes are now done in `build_series.dart` which also decides what file changes can trigger builds, we can move the logic entirely to `build_series.dart` instead of having it be distributed.

Remove a test that was of dubious value anyway :) it was using the `onDelete` hook to assert that nothing happened, which doesn't guarantee anything much.

Helps with the next PR which needs to track modifications in a similar way to differentiate user-made edits to outputs from build-output writes to outputs.

### DIFF
--- a/build_runner/lib/src/build/build_series.dart
+++ b/build_runner/lib/src/build/build_series.dart
@@ -44,6 +44,10 @@ class BuildSeries {
 
   final Completer<void> _closingCompleter = Completer();
 
+  /// Deletes that are part of build output, so the resulting file watch events
+  /// can be ignored.
+  final Set<AssetId> _expectedDeletes = {};
+
   /// Whether the next build is the first build.
   bool firstBuild = true;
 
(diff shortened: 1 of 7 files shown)


build_runner/lib/src/build/build_series.dart
```diff
@@ -81,12 +79,13 @@ class BuildSeries {
         result.add(change);
         continue;
       }
-      if (id.path.startsWith(entrypointDirectoryPath)) {
+
+      // Ignore any expected delete once.
+      if (change.type == ChangeType.REMOVE && _expectedDeletes.remove(id)) {
```
johnniwinther, line 84: Not new for this CL, but it seems dangerous to have side effects for calling `filterChanges`. I would expect it to be idempotent.

johnniwinther (approved):

build_runner/lib/src/build/build_series.dart
davidmorgan: Thanks! I agree.

The file watching API currently is a stream of events, which means we end up with event handling code like this. We do one delete, so we expect one delete event.

I would prefer that the file watching API also talks about the current state. We would get a notification that the state changed then compare the new state against what we expect. That's pretty much the `package:watcher` rewrite I was talking about last year :)

PR #5026, opened 2026-07-22
Check builder package in build.yaml parsing.

davidmorgan:
Various parts of `build_runner` expect that a package can only define builders in its own package; enforce it in `build_config`, and reject duplicate config in `build_runner` that could be caused by two packages defining the same builder.

### DIFF
--- a/build_runner/test/bootstrap/builder_ordering_test.dart
+++ b/build_runner/test/bootstrap/builder_ordering_test.dart
@@ -219,6 +219,32 @@ void main() {
       final orderedKeys = orderedBuilders.map((b) => b.key);
       expect(orderedKeys, ['a:self_cycle']);
     });
+
+    test('throws on duplicate builder keys', () {
+      final buildConfigs = parseBuildConfigs({
+        'a': {
+          'builders': {
+            'b': {
+              'builder_factories': ['createBuilder'],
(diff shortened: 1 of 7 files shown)


build_runner/test/bootstrap/builder_ordering_test.dart
```diff
@@ -219,6 +219,32 @@ void main() {
       final orderedKeys = orderedBuilders.map((b) => b.key);
       expect(orderedKeys, ['a:self_cycle']);
     });
+
+    test('throws on duplicate builder keys', () {
+      final buildConfigs = parseBuildConfigs({
+        'a': {
+          'builders': {
+            'b': {
```
johnniwinther, line 228: I clearly don't understand build configs well enough. How is the a duplicate builder key?

johnniwinther (approved):

build_runner/test/bootstrap/builder_ordering_test.dart
davidmorgan: Very good question :) actually there can't be a duplicate here, it wouldn't parse; we introduce one below.

Added a comment and an intermediate variable so it's clear.

Thanks.

johnniwinther (approved):

PR #5027, opened 2026-07-22
Fix flaky test.

davidmorgan:
Fix a flake caused by `0s compiling builders` not always being logged due to timing.

Sometimes `1s compiling builders` is the first line logged.

Just check for `compiling builders`.

### DIFF
--- a/build_runner/test/integration_tests/watch_command_test.dart
+++ b/build_runner/test/integration_tests/watch_command_test.dart
@@ -204,7 +204,7 @@ $yaml
     // The first build after retry is marked as a new build.
     await watch.expect('Starting build #5');
     // But the next identical build failure is not logged as a new build.
-    final block = await watch.expectAndGetBlock('0s compiling builders');
+    final block = await watch.expectAndGetBlock('compiling builders');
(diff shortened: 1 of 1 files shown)


build_runner/test/integration_tests/watch_command_test.dart
```diff
@@ -204,7 +204,7 @@ $yaml
     // The first build after retry is marked as a new build.
     await watch.expect('Starting build #5');
     // But the next identical build failure is not logged as a new build.
-    final block = await watch.expectAndGetBlock('0s compiling builders');
+    final block = await watch.expectAndGetBlock('1s compiling builders');
```
johnniwinther: From the description shouldn't `'1s compiling builders'` be `'compiling builders'` instead?

johnniwinther (approved):

build_runner/test/integration_tests/watch_command_test.dart
davidmorgan: Yes, thanks! I forgot to push the latest version. Done.

johnniwinther (approved):

PR #5030, opened 2026-07-23
Put lock file under `.dart_tool` by default.

davidmorgan:
We put other lock files and tracking under `.dart_tool` in the workspace; make `build_daemon` consistent.

Add a way to get the old behavior, but it's not exposed in any CLI, it's for users of `build_daemon` as a library.

Update tests to use real temp directories for workspaces, because now they might be used as real directories. Update tests to check both cases, lock files under workspace and lock files under shared temp.

Tested webdev with this change here: https://github.com/dart-lang/webdev/pull/2851, all tests pass.


build_daemon/lib/daemon.dart
```diff
@@ -126,21 +139,32 @@ class Daemon {
...
     return lock;
   } on FileSystemException {
     return null;
   }
 }
 
-void _createDaemonWorkspace(String workingDirectory) {
+void _createDaemonWorkspace(
+  String workingDirectory, {
+  String? daemonSharedPath,
```
johnniwinther: Given that this is internal to the library, maybe make this required to avoid accidentally missing this argument on call sites.

build_daemon/lib/client.dart
```diff
@@ -19,8 +19,13 @@ import 'data/server_log.dart';
 import 'data/shutdown_notification.dart';
 import 'src/file_wait.dart';
 
-Future<int> _existingPort(String workingDirectory) async {
-  final portFile = File(portFilePath(workingDirectory));
+Future<int> _existingPort(
+  String workingDirectory, {
+  String? daemonSharedPath,
```
johnniwinther: Given that this is internal to the library, maybe make this required to avoid accidentally missing this argument on call sites.

build_daemon/lib/daemon.dart
```diff
@@ -126,21 +139,32 @@ class Daemon {
   }
 }
 
-RandomAccessFile? _tryGetLock(String workingDirectory) {
+RandomAccessFile? _tryGetLock(
+  String workingDirectory, {
+  String? daemonSharedPath,
```
johnniwinther: Given that this is internal to the library, maybe make this required to avoid accidentally missing this argument on call sites.

build_daemon/test/daemon_test.dart
```diff
@@ -187,11 +276,15 @@ Future<String> _statusOf(Process daemon, {String logPrefix = ''}) async {
 
 Future<Process> _runDaemon(
   String workspace, {
+  String? daemonSharedPath,
```
johnniwinther: Given that this is internal to the library, maybe make this required to avoid accidentally missing this argument on call sites.

johnniwinther (approved):

build_daemon/lib/client.dart
davidmorgan: Ah yes that's much better, thanks.

Done.

build_daemon/lib/daemon.dart
davidmorgan: Done.

build_daemon/lib/daemon.dart
davidmorgan: Done.

davidmorgan: Done.

build_daemon/test/daemon_test.dart
davidmorgan: Done.

johnniwinther (approved):

PR #5031, opened 2026-07-23
Check packages and package names.

davidmorgan:
In `package:build`, check package names in `AssetId` are valid.

In `package:build_runner`, check that `AssetId` passed from builder code is actually `AssetId`, if not, copy it. This means the normalization is actually guaranteed to run :)

In `package:build_runner`, check that outputs are to known packages. This catches malformed package names, like the AssetId changes, but additionally catches attempts to "write" to packages not known to the build, which don't make sense.

### DIFF
--- a/build/test/asset_id_test.dart
+++ b/build/test/asset_id_test.dart
@@ -9,6 +9,19 @@ import 'package:test/test.dart';
 
 void main() {
   group('constructor', () {
+    test('validates the package parameter', () {
+      expect(() => AssetId('..', 'a'), throwsArgumentError);
+      expect(() => AssetId('a..b', 'a'), throwsArgumentError);
+      expect(() => AssetId('.a', 'a'), throwsArgumentError);
+      expect(() => AssetId('a.', 'a'), throwsArgumentError);
+      expect(() => AssetId('a/b', 'a'), throwsArgumentError);
+      expect(() => AssetId('a\\b', 'a'), throwsArgumentError);
+      expect(AssetId('', 'a').package, '');
+      expect(AssetId('a.b', 'a').package, 'a.b');
+      expect(AssetId('a\$b', 'a').package, 'a\$b');
+      expect(AssetId('a-b', 'a').package, 'a-b');
+    });
+
     test('normalizes the path', () {
       final id = AssetId('app', r'path/././/to/drop/..//asset.txt');
       expect(id.path, equals('path/to/asset.txt'));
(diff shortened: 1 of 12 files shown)


davidmorgan:
There are some further desirable improvements from this point: the path validation could use some tweaks, and `AssetId` has constructors that bypass the path normalization. I'll send those as a follow-up :)

build/test/asset_id_test.dart
```diff
@@ -9,6 +9,19 @@ import 'package:test/test.dart';
...
+      expect(() => AssetId('..', 'a'), throwsArgumentError);
+      expect(() => AssetId('a..b', 'a'), throwsArgumentError);
+      expect(() => AssetId('.a', 'a'), throwsArgumentError);
+      expect(() => AssetId('a.', 'a'), throwsArgumentError);
+      expect(() => AssetId('a/b', 'a'), throwsArgumentError);
+      expect(() => AssetId('a\\b', 'a'), throwsArgumentError);
+      expect(AssetId('', 'a').package, '');
+      expect(AssetId('a.b', 'a').package, 'a.b');
+      expect(AssetId('a\$b', 'a').package, 'a\$b');
+      expect(AssetId('a-b', 'a').package, 'a-b');
```
johnniwinther: I don't think package names can contain `-`.

build_runner/lib/src/build/resolver/asset_ids.dart
```diff
@@ -8,6 +8,14 @@ import '../../build_plan/build_step_plan.dart';
 import '../build_state/build_state.dart';
 
 extension AssetIdExtension on AssetId {
+  /// Normalizes this [AssetId].
+  ///
+  /// If the type of `this` is exactly `AssetId` then normalization of package
+  /// and path was already done by the constructor and `this` is returned.
+  ///
+  /// Otherwise returns a new `AssetId` constructed from [package] and [path].
+  AssetId normalize() => runtimeType == AssetId ? this : AssetId(package, path);
```
johnniwinther: You can actually override `runtimeType` so this is not a safe test. Maybe just always create a new `AssetId`  or maybe mark `AssetId` as `final`?

johnniwinther (approved):

build/test/asset_id_test.dart
davidmorgan: Thanks! Checked carefully: `dart pub get` allows `.`, rejects `-` or `$`. But `build_runner` uses `$` internally.

So disallowing `-` looks correct: done.

build_runner/lib/src/build/resolver/asset_ids.dart
davidmorgan: Thanks, yeah, I forgot about that one.

I don't want to make it `final` as that can break things.

So, "always copy" it is :)

johnniwinther (approved):

PR #5057, opened 2026-08-03
Remove redundant per-class // ignore: must_be_immutable comment.

### LINKED ISSUE #4588: Duplicate ignore `must_be_immutable`
At the start of the generated mock files, Mockito puts a `// ignore_for_file: must_be_immutable` directive.
For immutable classes Mockito additionally puts another `// ignore: must_be_immutable`.

This generates the following analyzer warning:

The diagnostic 'must_be_immutable' doesn't need to be ignored here because it's already being ignored.
Try removing the name from the list, or removing the whole comment if this is the only name in the list.
dart([duplicate_ignore](https://dart.dev/diagnostics/duplicate_ignore ))

Either the local ignore (for specific classes) or the global ignore (for the entire class) needs to be removed. It does not seem to matter which one, though keeping the global ignore for the entire file and removing the one specific to immutable classes seems to be the easier way that reduces complexity.

davidmorgan:
Fix #4588.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.1.4
+
+- In the client, check the workspace build daemon version and throw
+  `VersionSkew` if there is a mismatch.
+- Add descriptive messages and custom `toString` overrides to
+  `MissingPortFile`, `OptionsSkew`, and `VersionSkew`.
+
 ## 4.1.3
 
 - Bug fix: fix race condition when a new daemon immediately starts after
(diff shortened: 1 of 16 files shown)


jensjoha (approved):
I assume we have tests somewhere that verify that `// ignore_for_file: must_be_immutable` is added?

davidmorgan:
> I assume we have tests somewhere that verify that // ignore_for_file: must_be_immutable is added?

Yes, that's covered by golden tests since it's always added. Thanks.

Issue #5052, opened 2026-08-01
[build_daemon] Client 4.1.3 against daemon 4.1.2 fails with MissingPortFile — daemon exits silently with code 0


gmpassos:
## Summary

A `build_daemon` **client** at 4.1.3 talking to a **daemon** at 4.1.2 fails: the daemon process starts, exits immediately with code 0, prints nothing to stdout or stderr, and never writes its port file. The client then throws `MissingPortFile`.

Because the two sides are resolved independently — the client from `dart pub global activate webdev`, the daemon from the project's own `pubspec.lock` — this broke CI for projects whose lockfile pinned `build_daemon` below 4.1.3, with no change on their side. A workflow that was green on 2026‑07‑27 fails on 2026‑08‑01 with the same Dart SDK (3.12.2) and same webdev (3.8.1); re-running the identical commit reproduces it.

## Reproduction

Any web project built with `webdev build` whose `pubspec.lock` pins `build_daemon: 4.1.2`:

```
$ webdev build
[INFO] Connecting to the build daemon...
[INFO] Starting daemon...
Unhandled exception:
Instance of 'MissingPortFile'
#0      _existingServerInfo (package:build_daemon/client.dart:30:37)
<asynchronous suspension>
#1      BuildDaemonClient.connectUnchecked (package:build_daemon/client.dart:234:24)
<asynchronous suspension>
#2      BuildCommand.run (package:webdev/src/command/build_command.dart:71:22)
<asynchronous suspension>
```

Aligning the project's `build_daemon` to match the client fixes it outright:

```
$ dart pub upgrade build_daemon     # 4.1.2 -> 4.1.3
$ webdev build
[INFO] Starting daemon...
[INFO] Built with build_runner/aot in 13s; wrote 1173 outputs.
```

Reverting the lock back to 4.1.2 reproduces the failure again, so it is that version pair specifically, not stale state — the failure also occurs on a completely fresh CI checkout.

| | version |
|---|---|
| client (`dart pub global activate webdev`) | build_daemon **4.1.3** |
| daemon (project `pubspec.lock`) | build_daemon **4.1.2** |
| webdev | 3.8.1 |
| build_runner | 2.15.2 |
| build_web_compilers | 4.8.0 |
| Dart SDK | 3.12.2 (stable) |

Reproduced on **macOS** (`darwin arm64`, local) and **Linux** (`ubuntu-latest` on GitHub Actions). Both use the AOT path (`compiling builders/aot`).

[...]

PR #5061, opened 2026-08-03
Check project build_daemon version in connect and release 4.1.4.

davidmorgan:
For https://github.com/dart-lang/build/issues/5052; one possible improvement we can make+release.

This catches the case where the client is ahead of the current workspace, and fails with a version error.

Add detail to the various errors.

### DIFF
--- a/build_daemon/CHANGELOG.md
+++ b/build_daemon/CHANGELOG.md
@@ -1,3 +1,10 @@
+## 4.1.4
+
+- In the client, check the workspace build daemon version and throw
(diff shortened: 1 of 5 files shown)


Markzipan (approved):
Thanks for the fix here! LGTM. I wonder if we could add a test that checks compatibility between the current and last major release. Though in effect this is kinda rare and could have used a non-minor version bump. 

I'll merge and release 4.1.4, then add some webdev fixes.

PR #5063, opened 2026-08-03
Reduce "compiling" output for non-interactive builds

### LINKED ISSUE #5029: build_runner: progress line repeated once per second when stdout is not a terminal (floods CI logs)
### Summary

When stdout is not a terminal (CI logs, pipes), `build_runner build` emits one progress line **per second** for the whole builder-compile phase. On a cold cache that is dozens of consecutive lines that differ only by the elapsed-second counter, which dominates the CI log.

Real example from a Jenkins job (builder compile takes ~73s there):

```
  0s compiling builders/aot
  1s compiling builders/aot
  2s compiling builders/aot
  ... (~70 more)
```

Locally, in a real terminal, the same build is fine — the line updates in place. The problem is specific to non-interactive stdout.

### Reproduction

`build_runner` 2.15.1, Dart 3.12.2, Linux.

```sh
rm -f .dart_tool/build/entrypoint/build.dart.aot .dart_tool/build/entrypoint/build.dart.dill
dart run build_runner build > out.log 2>&1
wc -l < out.log                                     # 40
grep -c 'compiling builders/aot' out.log            # 29
sed -E 's/^ *[0-9]+s //' out.log | grep 'compiling builders/aot' | sort -u | wc -l   # 1
```

29 of 40 lines (72%) are the same message repeated; stripping the counter leaves exactly **one** distinct string.

### Why it happens

`displayingBlocks` requires a terminal (`lib/src/logging/log_display.dart:44-49`):

```dart
bool get displayingBlocks =>
    buildLog.configuration.forceConsoleWidthForTesting != null ||
    (buildLog.configuration.onLog == null &&
        buildLog.configuration.mode == BuildLogMode.build &&
        buildProcessState.stdio.hasTerminal &&
        buildProcessState.stdio.supportsAnsiEscapes);
```

With no terminal it takes the `else` branch and prints unconditionally, with no comparison against what was printed last (`lib/src/logging/build_log.dart:309-317`):

```dart
if (_shouldShowProgressNow || lastTick) {
  if (_display.displayingBlocks) {
    _display.block(render());
  } else {
    _display.message(
      Severity.info,
      _renderCompiling(compileType: compileType).toString(),
    );
  }
}
```

…once per second (`lib/src/logging/build_log.dart:707-718`):

```dart
final interval =
    configuration.mode == BuildLogMode.build && _display.displayingBlocks
    ? const Duration(milliseconds: 100)
    : const Duration(seconds: 1);
```

Since `_renderCompiling` renders the same text apart from the duration, every tick is a near-duplicate line.

### Why the obvious workarounds don't apply

- **No flag or env var.** Nothing in `build_runner build --help` or the global options disables progress. `BuildLogConfiguration.throttleProgressUpdates` is internal-only and set to `true`; setting it `false` would emit *more*.
- **Giving CI a pty makes it worse, not better.** Satisfying `hasTerminal` switches to block mode, which ticks at **100ms instead of 1s** and rewrites the entire block each time prefixed with a cursor-move (`lib/src/logging/log_display.dart:64-65`). CI consoles (e.g. the Jenkins AnsiColor plugin) are stream filters that handle SGR colour codes but do not implement cursor repositioning, so those re-renders are not collapsed — roughly 10x more output plus stray escape sequences.
- **Filtering with `grep -v`** works but breaks the exit code in a shell pipeline unless `pipefail` is set, so it is easy to get wrong in CI.

### Possible fixes

Either would solve it; happy to send a PR for whichever you prefer (or both):

1. **Don't repeat unchanged progress lines when not displaying blocks.** Keep the last emitted string and skip the write if the new one is identical apart from the duration. Needs no new API, no opt-in, and helps every CI user by default. Would collapse the example above from 29 lines to 1.
2. **Add a `--no-progress` flag** (or honour an env var) that suppresses intermediate progress while keeping per-phase results and the final summary. Explicit and discoverable, but opt-in, so existing CI setups stay noisy until they adopt it.

My preference is 1 as the default behaviour with 2 as an explicit escape hatch, but I'd rather follow whatever you consider the right shape before writing the patch.

davidmorgan:
Fix #5029.

### DIFF
--- a/build_runner/CHANGELOG.md
+++ b/build_runner/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 2.16.1-wip
 
+- Reduce logged compile progress output for non-interactive builds, for example
+  presubmits.
+
 ## 2.16.0
 
 - New default output behavior: always fix incorrect generated files. For
--- a/build_runner/lib/src/logging/build_log.dart
+++ b/build_runner/lib/src/logging/build_log.dart
@@ -306,10 +306,11 @@ class BuildLog {
       }
     }
 
-    if (_shouldShowProgressNow || lastTick) {
+    if (_shouldShowProgressNow || firstTick || lastTick) {
       if (_display.displayingBlocks) {
         _display.block(render());
       } else {
+        if (!firstTick && !lastTick) return;
         _display.message(
           Severity.info,
           _renderCompiling(compileType: compileType).toString(),
@@ -649,8 +650,8 @@ class BuildLog {
         ? _aotCompileProgress!
         : _jitCompileProgress!;
     return AnsiBufferLine([
-      renderDuration(progress.duration),
-      ' ',
+      if (progress.duration != Duration.zero)
+        '${renderDuration(progress.duration)} ',
       AnsiBuffer.bold,
       'compiling builders',
       AnsiBuffer.reset,
--- a/build_runner/test/integration_tests/builder_changes_test.dart
+++ b/build_runner/test/integration_tests/builder_changes_test.dart
@@ -37,7 +37,10 @@ void main() async {
       (script) => script.replaceAll(r"'$extraContent'", "'(v1)'"),
     );
     await watch.expect('Starting build #2 with updated builders.');
-    await watch.expect(RegExp(r'1s compiling builders/jit'));
+    await watch.expect('compiling builders/jit');
+    // Wait a second so the build has definitely started.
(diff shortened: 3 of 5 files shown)


jensjoha (approved):

qeepcologne:
code looks valid, cant test in my real projects at the moment, this version works only on flutter 3.47 and other dependency updates.

PR #5078, opened 2026-08-13
Deduplicate compiler process execution and logging helpers

kevmoo:
- Introduce BaseCompiler in build_runner to share process execution, depfile management, and experiment handling between AOT and Kernel compilers.
- Extract _runBuild in _benchmark to consolidate repeated build execution and stopwatch measurement.
- Deduplicate input tracking filtering in build_test via _trackedFor.
- Extract common progress display rendering in build_runner BuildLog.

### DIFF
--- a/_benchmark/lib/workspace.dart
+++ b/_benchmark/lib/workspace.dart
@@ -80,63 +80,39 @@ void main() {
     return result;
   }
 
-  Future<void> _measure(PendingResult result) async {
-    // Clean build.
+  Future<Duration?> _runBuild(String description, PendingResult result) async {
     final stopwatch = Stopwatch()..start();
-    var process = await Process.start('dart', [
+    final process = await Process.start('dart', [
       'run',
       'build_runner',
       'build',
       '-d',
     ], workingDirectory: directory.path);
-    var exitCode = await process.exitCode;
+    final exitCode = await process.exitCode;
     if (exitCode == 0) {
-      result.cleanBuildTime = stopwatch.elapsed;
+      return stopwatch.elapsed;
     } else {
       final stdout = await process.stdout.transform(utf8.decoder).join();
       final stderr = await process.stderr.transform(utf8.decoder).join();
-      result.failure = 'Initial build failed:\n$stdout\n$stderr';
-      return;
+      result.failure = '$description build failed:\n$stdout\n$stderr';
+      return null;
     }
+  }
+
+  Future<void> _measure(PendingResult result) async {
+    // Clean build.
+    result.cleanBuildTime = await _runBuild('Initial', result);
+    if (result.isFailure) return;
 
     result.graphSize = _readGraphSize();
 
     // Build with no changes.
-    stopwatch.reset();
-    process = await Process.start('dart', [
-      'run',
-      'build_runner',
-      'build',
-      '-d',
-    ], workingDirectory: directory.path);
-    exitCode = await process.exitCode;
-    if (exitCode == 0) {
-      result.noChangesBuildTime = stopwatch.elapsed;
-    } else {
-      final stdout = await process.stdout.transform(utf8.decoder).join();
-      final stderr = await process.stderr.transform(utf8.decoder).join();
-      result.failure = 'No changes build failed:\n$stdout\n$stderr';
-      return;
-    }
+    result.noChangesBuildTime = await _runBuild('No changes', result);
+    if (result.isFailure) return;
 
     // Incremental build after a change.
     edit('lib/app.dart');
-    stopwatch.reset();
-    process = await Process.start('dart', [
-      'run',
-      'build_runner',
-      'build',
-      '-d',
-    ], workingDirectory: directory.path);
-    exitCode = await process.exitCode;
-    if (exitCode == 0) {
-      result.incrementalBuildTime = stopwatch.elapsed;
-    } else {
-      final stdout = await process.stdout.transform(utf8.decoder).join();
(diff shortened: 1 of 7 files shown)


kevmoo:
https://github.com/kevmoo/kevmoo_skills/blob/main/skills/deslop-duplication-audit/SKILL.md - FYI

build_test/lib/src/internal_test_reader_writer.dart
```diff
@@ -209,14 +209,9 @@ class _ReaderWriterTestingImpl implements ReaderWriterTesting {
...
 
-  @override
-  Iterable<AssetId> get inputsTracked => InputTracker
-      .inputTrackersForTesting[_readerWriter.filesystem]!
-      .expand((tracker) => tracker.inputs)
-      .toSet();
-
-  @override
-  Iterable<AssetId> inputsTrackedFor({
+  /// Returns tracked inputs or resolver entrypoints matching the filter.
```
davidmorgan: I don't think this refactoring is an improvement--the extracted function is just a bit too general and at the same time specific to extract for just two uses.

build_runner/test/common/build_runner_tester.dart
```diff
@@ -156,6 +157,19 @@ class BuildRunnerTester {
     final file = File(p.join(tempDirectory.path, path));
     final data = file.readAsStringSync();
     file.writeAsStringSync(update(data));
+    _deletePubspecLocksFor(path);
+  }
+
+  void _deletePubspecLocksFor(String path) {
```
davidmorgan: Thanks for commenting on the issue :) ... do you need this workaround for something? I don't think it relates to the refactoring?

build_runner/lib/src/logging/build_log.dart
```diff
@@ -420,17 +398,26 @@ class BuildLog {
...
+    if (_display.displayingBlocks) {
+      _display.block(render());
+    } else {
+      _display.message(Severity.info, _renderPhase(phaseName).toString());
+    }
+  }
+
+  /// Displays progress if [progress] for [phaseName] has finished.
+  void _displayProgressIfFinished(_PhaseProgress progress, String phaseName) {
     // Usually the next step will immediately run and update with more useful
```
davidmorgan: This one's an improvement, thanks.

This comment no longer makes sense here, it depends on the caller. Please just remove it.

build_runner/lib/src/bootstrap/compiler.dart
```diff
@@ -24,6 +32,87 @@ abstract class Compiler {
...
+      _outputDepfile.checkFreshness(digestsAreFresh: digestsAreFresh);
+
+  @override
+  bool isDependency(String path) => _outputDepfile.isDependency(path);
+
+  /// Cleans up compiler-specific stderr messages.
+  String cleanStderr(String stderr);
+
+  @override
+  Future<CompileResult> compile({Iterable<String>? experiments}) async {
```
davidmorgan: I don't think this is a good refactor.

The AOT compiler and JIT compiler are different tools, they are not for example required to output the same "Unknown experiment" messages formatted in the same way. It's a coincidence that they do.

So pulling that into a base class makes it more strongly linked in the code than it actually is.

I would instead consider moving shared code down instead of up: to "helper" methods.

But on balance I think it's better to leave it separate, a list of actions that we take to run and process the output of each tool is simple and easy.

_benchmark/lib/workspace.dart
```diff
@@ -80,63 +80,40 @@ void main() {
     return result;
   }
 
-  Future<void> _measure(PendingResult result) async {
-    // Clean build.
+  /// Runs `build_runner` in [directory] and records elapsed time or failure.
```
davidmorgan, line 83: This is an improvement--thanks.

davidmorgan (commented):
Thanks! I like 2/4 of the refactors, and the other I'm not sure if you need it / if it belongs in this PR.

build_runner/test/common/build_runner_tester.dart
davidmorgan: Oh, I see--it's an SDK regression that broke the tests.

I'll land this separately to get the CI green again.

davidmorgan: Actually let's pin the SDK version instead https://github.com/dart-lang/build/pull/5082

build_test/lib/src/internal_test_reader_writer.dart
kevmoo: Reverted the _trackedFor helper extraction.

build_runner/test/common/build_runner_tester.dart
kevmoo: Removed the workaround from this PR so #5082 can address it.

build_runner/lib/src/logging/build_log.dart
kevmoo: Removed the stale comment.

build_runner/lib/src/bootstrap/compiler.dart
kevmoo: Reverted the BaseCompiler extraction and kept the compiler implementations separate.

_benchmark/lib/workspace.dart
kevmoo: Thanks!

kevmoo:
@davidmorgan PTAL

davidmorgan (approved):
Thanks :)

PR #5080, opened 2026-08-17
build_web_compilers: require Dart 3.13.0.

### LINKED ISSUE #5077: build_web_compilers ships a dev-build SDK floor that leaks into downstream lockfiles
`build_web_compilers` declares its SDK constraint with dev-build prerelease bounds, apparently carried over from whatever dev SDK it was built against, rather than the plain `^3.x.0` that every other package in this repo uses:

| package | latest | `environment.sdk` |
| --- | --- | --- |
| build | 4.0.10 | `^3.11.0` |
| build_runner | 2.16.0 | `^3.11.0` |
| build_config | 1.3.2 | `^3.8.0` |
| build_daemon | 4.1.5 | `^3.11.0` |
| build_resolvers | 3.0.4 | `^3.7.0` |
| build_test | 3.5.19 | `^3.11.0` |
| **build_web_compilers** | 4.8.9 | `>=3.13.0-107.0.dev <3.14.0-z` |

To be clear about scope: this is cosmetic. Resolution works — `build_web_compilers: any` resolves fine on stable Dart 3.13.0, and stable `3.13.0` satisfies the `3.13.0-107.0.dev` floor since a release outranks its own prereleases. I'm not reporting a broken build.

The annoyance is that `pubspec.lock`'s `sdks:` block records the tightest constraint in the graph, so the dev-build floor surfaces verbatim in projects that only pull this in transitively. Upgrading Flutter in such a repo produces:

```diff
 sdks:
-  dart: ">=3.12.0 <3.13.0-z"
+  dart: ">=3.13.0-107.0.dev <3.14.0-z"
```

That reads as though something in the project targets a dev SDK, which invites a "why is this here?" round trip in review. It came up for us in supabase/supabase-flutter#1705.

Normalizing the published constraint to `^3.13.0`, matching the rest of the repo, would avoid it. More generally, rounding prerelease bounds away at publish time would stop the dev SDK used during development from leaking into the released pubspec.

### Aside on the upper bound

Not asking for anything here, just noting it since it's the same mechanism. The `<3.13.0-z` form that earlier versions used does exclude stable `3.13.0`, so pinning an exact version in that range fails:

```console
$ dart pub add build_web_compilers:4.8.4
Because unpinned depends on build_web_compilers >=4.1.0-beta.3 <4.8.5 which
requires SDK version >=3.6.0-165.0.dev <3.13.0-z, version solving failed.
```

In practice this is invisible to almost everyone, since unpinned resolution just selects 4.8.5+, where the bound is already `<3.14.0-z`. Mentioning it only because it suggests these bounds are generated mechanically rather than chosen.

Checked on Dart 3.13.0 (stable) / Flutter 3.47.0.

davidmorgan:
Fix #5077.

There was a min bound on a dev SDK, put it on the released SDK.

### DIFF
--- a/builder_pkgs/build_web_compilers/CHANGELOG.md
+++ b/builder_pkgs/build_web_compilers/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.8.10
+
+- Require Dart `3.13.0`.
+
 ## 4.8.9
 - Fix an issue where `DdcFrontendServerBuilder` accumulates changed files across builds.
 
--- a/builder_pkgs/build_web_compilers/pubspec.yaml
+++ b/builder_pkgs/build_web_compilers/pubspec.yaml
@@ -1,11 +1,11 @@
 name: build_web_compilers
-version: 4.8.9
+version: 4.8.10
 description: Builder implementations wrapping the dart2js and DDC compilers.
 repository: https://github.com/dart-lang/build/tree/master/builder_pkgs/build_web_compilers
 resolution: workspace
 
 environment:
-  sdk: '>=3.13.0-107.0.dev <3.14.0-z'
+  sdk: '>=3.13.0 <3.14.0-z'
 
 dependencies:
   analyzer: '>=13.3.0 <15.0.0'


jensjoha (approved):

davidmorgan:
CI is broken due to SDK issue https://github.com/dart-lang/sdk/issues/61950, I will send a separate workaround

PR #5081, opened 2026-08-17
build_runner: remove stale pubspec.lock on pubspec.yaml writes.

davidmorgan:
Workaround for https://github.com/dart-lang/sdk/issues/61950.

Gets the CI green; must be removed when the SDK is fixed.


davidmorgan:
I think I prefer to pin the SDK version used on CI, it's clearer that it's not intended to work like this https://github.com/dart-lang/build/pull/5082

PR #5087, opened 2026-08-18
Add an authentication token to fes_manager

Markzipan:
For more context, see a related change for `build_daemon`: https://github.com/dart-lang/build/pull/5041

### DIFF
--- a/builder_pkgs/build_web_compilers/bin/fes_manager.dart
+++ b/builder_pkgs/build_web_compilers/bin/fes_manager.dart
@@ -73,16 +73,12 @@ void main(List<String> args) async {
   // Write the Frontend Server's config to file so we can create connections to
   // it later in the build.
   final configDir = configFile.parent;
-  final isNew = !configDir.existsSync();
-  if (isNew) {
-    configDir.createSync(recursive: true);
-    FilePermissions.makeUserPrivate(configDir.path);
-  }
+  configDir.createSync(recursive: true);
+  FilePermissions.makeUserPrivate(configDir.path);
 
   final serverInfo = FesServerInfo(server.port, token);
   final tempConfigFile = File('${configFile.path}.tmp');
   serverInfo.writeToFile(tempConfigFile);
-  FilePermissions.makeUserPrivate(tempConfigFile.path);
   tempConfigFile.renameSync(configFile.path);
 
   print('Frontend Server Manager listening on port ${server.port}');
--- a/builder_pkgs/build_web_compilers/bin/fes_manager.dart
+++ b/builder_pkgs/build_web_compilers/bin/fes_manager.dart
@@ -5,6 +5,8 @@
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:build_web_compilers/src/build_frontend_server/fes_server_info.dart';
+import 'package:build_web_compilers/src/build_frontend_server/file_permissions.dart';
 import 'package:build_web_compilers/src/build_frontend_server/frontend_server_driver.dart';
 import 'package:build_web_compilers/src/build_frontend_server/package_config.dart';
 import 'package:build_web_compilers/src/common.dart';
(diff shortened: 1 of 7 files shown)


builder_pkgs/build_web_compilers/bin/fes_manager.dart
```diff
@@ -65,13 +67,22 @@ void main(List<String> args) async {
...
+    configDir.createSync(recursive: true);
+    FilePermissions.makeUserPrivate(configDir.path);
+  }
+
+  final serverInfo = FesServerInfo(server.port, token);
   final tempConfigFile = File('${configFile.path}.tmp');
-  tempConfigFile.createSync(recursive: true);
-  tempConfigFile.writeAsStringSync(jsonEncode({'port': server.port}));
+  serverInfo.writeToFile(tempConfigFile);
+  FilePermissions.makeUserPrivate(tempConfigFile.path);
```
davidmorgan: Making the file private is not useful because it can't be created private--there is a race. That's why the directory is made private: you can't read a file in a private directory, so it's already private when it's created.

In the build_daemon change we also changed the location of the directory from the previously released version, so we could rely on it being created private. In this case it might have already been created. So I think the simple thing is to always make the directory private.

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/fes_server_info.dart
```diff
@@ -0,0 +1,42 @@
...
+
+  /// Reads and parses [FesServerInfo] from [file].
+  ///
+  /// Returns `null` on parse failure.
+  static FesServerInfo? fromFile(File file) {
+    try {
+      final content = file.readAsStringSync();
+      final json = jsonDecode(content) as Map<String, dynamic>;
+      final port = json['port'] as int?;
+      final token = json['token'] as String?;
```
davidmorgan: What happens if there is version skew? We want that there is some reasonable prompt to the user rather than e.g. a retry loop.

davidmorgan (approved):
Looks good, thanks! A few questions to resolve before it goes in though.

builder_pkgs/build_web_compilers/bin/fes_manager.dart
Markzipan: Oooh, good catch - updated!

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/fes_server_info.dart
Markzipan: I added some logic to detect port files without a token + an error message

Markzipan:
Thanks! One last check @davidmorgan

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/fes_server_info.dart
```diff
@@ -26,17 +26,30 @@ class FesServerInfo {
...
+          '(missing auth token). Please upgrade to build_web_compilers '
+          '>=4.8.11 and restart.',
+        );
+      }
+
       if (port != null && token != null) {
         return FesServerInfo(port, token);
       }
-    } catch (_) {}
+    } on FormatException {
```
davidmorgan: It's better to keep the "catch (_) { return null; }", that way you don't have to worry about tedious stuff like valid JSON with the wrong primitive types.

You can do that by

```
int? port;
String? token;
try {
...
} catch (_) { return null; }
```
then choose whether to throw, return FesServerInfo, or return null.

davidmorgan (approved):
Thanks! One last improvement suggested.

builder_pkgs/build_web_compilers/lib/src/build_frontend_server/fes_server_info.dart
Markzipan: Good idea - I unrolled the logic + added the checks below

PR #5089, opened 2026-08-20
Add immutable FinishedBuildState.

### LINKED ISSUE #5079: Incremental build deadlocks when a generated file is on disk but not in the asset graph
A `.g.dart` that exists on disk but isn't recorded in `asset_graph.json` gets classified as a build **source**. `freezed` (phase 0) then picks it up as a primary input while `source_gen:combining_builder` (phase 2) declares it as an output. Phase 0 can't read it until phase 2 writes it, so the build parks forever — 0% CPU, empty Dart stack, no error, no timeout.

Clean builds handle this fine. Only incremental is affected.

## Repro

`pubspec.yaml`:

```yaml
name: repro
publish_to: none
environment:
  sdk: ">=3.8.0 <4.0.0"
dependencies:
  freezed_annotation: any
  json_annotation: any
dev_dependencies:
  build_runner: 2.15.1
  freezed: 3.2.5
  json_serializable: any
```

`lib/alpha.dart`:

```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'alpha.freezed.dart';
part 'alpha.g.dart';

@freezed
abstract class Alpha with _$Alpha {
  const factory Alpha({required String id}) = _Alpha;
  factory Alpha.fromJson(Map<String, dynamic> json) => _$AlphaFromJson(json);
}
```

```sh
dart pub get
dart run build_runner build                 # ok

# add a second model plus a stray output file
sed 's/alpha/beta/g; s/Alpha/Beta/g' lib/alpha.dart > lib/beta.dart
touch lib/beta.g.dart

dart run build_runner build                 # hangs forever
```

Last line printed is `freezed on 3 inputs; lib/alpha.dart` — 3, because `beta.g.dart` became a source.

`dart run build_runner clean && dart run build_runner build` succeeds in the same state.

The empty `beta.g.dart` stands in for what an interrupted build leaves behind: outputs written, graph not yet updated. That's how this shows up in practice — and because the deadlock forces a Ctrl-C, every retry re-arms it. Ours sat broken for a week before we worked out what it was.

## Expected

Either strip/delete the conflicting output the way a clean build does, or fail with the old conflicting-outputs error. Not a silent hang.

## Cause

`lib/src/build_plan/build_plan.dart`. The clean path removes declared outputs from the source set:

```dart
final declaredOutputs = BuildStepPlan.compute(...).declaredOutputs;
inputSources.removeAll(declaredOutputs);
```

`_createIncremental` has no equivalent. `filesToCheck` is just the union of `assetTrackerInputSources`, `cacheDirSources`, `previousBuildState.sources` and `previousBuildState.actualOutputs`, and `withCompatiblePreviousBuild` does `..conflictingOutputs.clear()`.

Dumping `BuildStepImpl` instances over the VM service during a hang shows the cycle (captured on the real project where we hit this, hence the different names):

| phase | builder | primary input |
|---|---|---|
| 2 | combining_builder | `leaderboard.dart` |
| 0 | freezed | `leaderboard.g.dart` ×2 |
| 1 | json_serializable | `leaderboard.g.freezed.dart` |

`leaderboard.g.freezed.dart` doesn't exist on disk — it only exists as freezed's declared output *from a generated file*.

## Versions

Reproduced on:

- build_runner 2.15.1, freezed 3.2.5, analyzer 10.2.0
- build_runner 2.16.0, freezed 4.0.0-dev.3, analyzer 13.3.0

Dart 3.12.2, macOS 26.5.1. `--delete-conflicting-outputs` doesn't help — it's accepted and ignored since 2.15.

davidmorgan:
Add SerializedBuildState and FinishedBuildState.

After a build, BuildState turns into FinishedBuildState which has a BuildStepPlan and a SerializedBuildState.

When starting from deserialization, SerializedBuildState is only turned into FinishedBuildState if there is a compatible build spec. Otherwise it's used to identify stale outputs to delete then discarded.

Build planning uses the previous FinishedBuildState.

BuildState can now have a BuildStepPlan, simplifying the API and allowing it to fully validate outputs.

Helps with but does not fully fix #5079.

Publishing the build result as immutable FinishedBuildState highlighted a surprising behavior. When a file not read in the build was read by BuildOutputReader, it would push a digest back into BuildState. This would cause the file to be considered "interesting", and any change to the file to cause a no-op rebuild.

("Interesting" was the old terminology: after this PR we say "consumed outside the build").

This is useful: users of `serve` and `daemon` commands get "new build happened" notifications when something that they read changes, so they will read it again.

But, it's confusing. Instead, 1) track files that are consumed outside the build in `BuildOutputReader`, 2) tell `watch` and `daemon` modes about files that did not trigger a new build; they can intersect with the set of files in `BuildOutputReader` to know whether to notify about the change; they notify by publishing as if there was a no-op build.

This preserves the notification behavior without running a build or mutating build state after the build.


build_runner/lib/src/build/build_state/asset_graph_json.dart
```diff
@@ -16,24 +16,27 @@ import 'serializers.dart';
...
-    required this.buildState,
+    required this.serializedBuildState,
     required this.phasedAssetDeps,
   });
 
   /// Serializes for `asset_graph.json`.
   static Uint8List serialize({
     required BuildSpecDigest buildPlanDigest,
-    required BuildState buildState,
+    required SerializedBuildState buildState,
```
jensjoha: on the face of it at least it reads weirdly that we serialize a "serialized build state".

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -226,6 +218,50 @@ class BuildState {
...
+      }
+    }
+
+    for (final outer in _buildStepResultsByPrimaryInput.entries) {
+      final input = outer.key;
+      for (final inner in outer.value.entries) {
+        builder.buildStepResults[BuildStepId(
+              primaryInput: input,
+              phaseNumber: inner.key,
+            )] =
```
jensjoha: is this really the formatting output of dart format?

build_runner/lib/src/commands/daemon/daemon_builder.dart
```diff
@@ -215,6 +229,85 @@ class BuildRunnerDaemonBuilder implements DaemonBuilder {
...
+                reader?.wasSourceConsumedOutsideBuild(change.id) ?? false,
+          )
+          .map((change) => change.id)
+          .toSet();
+
+      // If neither accepted changes nor consumed rejected changes exist, do
+      // nothing.
+      if (filtered.accepted.isEmpty &&
+          consumedRejected.isEmpty &&
+          fileChanges.isNotEmpty) {
```
jensjoha: Why do we need to continue if the file changes are empty?

build_runner/lib/src/io/build_output_reader.dart
```diff
@@ -108,35 +116,53 @@ class BuildOutputReader {
...
+    if (!_isFile(id)) {
       throw AssetNotFoundException(id);
     }
+
+    final bytes = await readerWriter.readAsBytes(
+      id,
+      hidden: buildState.isHidden(id),
+    );
+    _recordSourceConsumedOutsideBuild(id);
+    return cached != null ? cached.withBytes(bytes).bytes : bytes;
```
jensjoha: So I was about to suggest

```
cached?.withBytes(bytes);
return bytes;
```

but then looked up what `withBytes` actually do...
Unless I'm reading something completely wrong this is just a weird way to make it slower to return `bytes`..?

(similar below)

jensjoha (approved):

build_runner/lib/src/build/build_state/asset_graph_json.dart
davidmorgan: That's fair: renamed to IncrementalBuildState

davidmorgan: Good point, thanks. Renamed -> IncrementalBuildState

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: Yes. Pulled out local ID variables to make it less bad.

build_runner/lib/src/commands/daemon/daemon_builder.dart
davidmorgan: This comes from a weird API point of the build_daemon API: in auto build mode an empty `fileChanges` means that it's _not_ a file-watch-based build, it's an explicit build request. So we must return a response.

Split out to a boolean `isExplicitBuild` in the caller; it would be nice to split out earlier but that would be a public API change. Documented in `package:build_daemon` instead.

And: tweaked the case below. If it's an explicit build request but nothing actually changed, and there is a cached build result, we emit a "synthetic result", better than doing a no-op build.

build_runner/lib/src/io/build_output_reader.dart
davidmorgan: Good catch, thanks--this is incorrectly kept from the mutable version which can update with the new data.

But actually digging into this shows a bunch of subtle issues: by this point in the build all the data should already be loaded, but there are some cases where that is not happening correctly, unnoticed because of a fallback to just reading the files. I'll send a follow-up PR rather than try to tack on a fix here.

For this PR just returned the values directly.

PR #5090, opened 2026-08-24
Improve build_runner_build.dart.

davidmorgan:
Improvements to the "run build_runner in build_runner" script.

Remove capability to override to a local build_runner instance: that was only needed because there was no suitable published build_runner.

Make it add the generator deps, otherwise I often forget to add them. (They're not checked in because occasional circular dependency problems are annoying).

Make it work when run anywhere in the depot.

This change is AI generated and only lightly reviewed; I tried it and it worked.

### DIFF
--- a/tool/build_runner_build.dart
+++ b/tool/build_runner_build.dart
@@ -40,7 +40,7 @@ void main(List<String> arguments) {
   );
   tempDirectory.createSync(recursive: true);
   File.fromUri(tempDirectory.uri.resolve('pubspec.yaml')).writeAsStringSync('''
-name: none
+name: fake_package
 environment:
   sdk: ^3.7.0
 dependencies:
@@ -59,6 +59,11 @@ dependencies:
         as Map<String, Object?>,
   );
 
+  final packageConfigFile = File.fromUri(
+    repoRoot.uri.resolve('.dart_tool/package_config.json'),
+  );
+  final packageConfig = packageConfigFile.readAsStringSync();
+
   // Merge `pubConfig` into the local package config.
   //
   // `build_runner` expects to run with the local package config because it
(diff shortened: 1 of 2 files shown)


tool/build_runner_build.dart
```diff
@@ -70,54 +70,135 @@ $maybeOverride
...
+
+  final workspaceDotDartToolUri = repoRoot.uri.resolve('.dart_tool/');
+  for (final package in mergedConfig.packages) {
+    package.rootUri = workspaceDotDartToolUri
+        .resolve(package.rootUri)
+        .toString();
+  }
+
+  // Override build packages with published versions from temp pubConfig.
+  final currentPackageName = RegExp(
```
jensjoha: It feels weird to parse the yaml with regex when doing all sorts of yaml stuff below with packages...

tool/build_runner_build.dart
```diff
@@ -70,54 +70,135 @@ $maybeOverride
...
+      }
     }
   }
-  final mergedConfigFile = File.fromUri(
-    tempDirectory.uri.resolve('package_config.json'),
+
+  // Add all other packages from pubConfig not present in mergedConfig, such as
+  // built_value_generator, json_serializable, and their dependencies.
+  for (final package in pubConfig.packages) {
+    if (package.name != 'none' &&
```
jensjoha: what is this `none`?

tool/build_runner_build.dart
```diff
@@ -70,54 +70,135 @@ $maybeOverride
...
+        mergedConfig.packageNamedOrNull(package.name) == null) {
+      mergedConfig.addPackage(package);
+    }
+  }
+
+  final repoRootPackageConfigFile = File.fromUri(
+    repoRoot.uri.resolve('.dart_tool/package_config.json'),
   );
-  mergedConfigFile.writeAsStringSync(json.encode(mergedConfig));
+  final originalPackageConfig = repoRootPackageConfigFile.readAsStringSync();
```
jensjoha: these files are read twice. It probably doesn't matter too much in this tool, but it is sort of confusing.

jensjoha (approved):

tool/build_runner_build.dart
davidmorgan: It's the fake package created to do a "pub get". Renamed "fake_package".

tool/build_runner_build.dart
davidmorgan: Done, and improved variable names a bit.

tool/build_runner_build.dart
davidmorgan: Done.

davidmorgan (commented):
Thanks!

PR #5092, opened 2026-08-24
Reject non-loopback Host in build_runner serve

herdiyana256:
`build_runner serve` binds an HTTP server that streams build outputs, and by default binds to localhost so it is only reachable from the local machine. The request pipeline in `ServeHandler.handlerFor` did not validate the Host or Origin header on the asset serving path; only the live reload WebSocket had an origin check. A loopback HTTP server with no Host validation is reachable through DNS rebinding: a page the developer visits can rebind its own hostname to a loopback address and then read the served application and its source maps cross origin.

Concretely, with the server on `localhost:8080`, a page on `attacker.example.com` that rebinds its DNS to `127.0.0.1` can issue:

```
GET /main.dart.js HTTP/1.1
Host: attacker.example.com:8080

before: 200 OK          (serves the compiled app and its .js.map source maps)
after:  403 Forbidden
```

This validates the Host and Origin headers when the server is bound to a loopback interface, rejecting requests whose Host is absent or non loopback and requests carrying a non loopback Origin. It mirrors the daemon asset server, which already does this:

```dart
if (restrictToLoopback) {
  pipeline = pipeline.addMiddleware(_loopbackOnly);
}
```

`serve_command.dart` enables it whenever the bind hostname is loopback (the `localhost` default). When the user has explicitly bound a routable address such as `--hostname 0.0.0.0` the restriction is skipped, so intentional network serving is unaffected.

Added tests in `serve_handler_test.dart` covering missing Host, non loopback Host, non loopback Origin, the allowed loopback case, and that the default unrestricted behavior is unchanged.


build_runner/lib/src/commands/serve/server.dart
```diff
@@ -71,6 +72,16 @@ class ServeHandler {
...
     });
     var pipeline = const shelf.Pipeline();
+    // When serving on a loopback interface reject requests whose `Host` header
+    // is not itself loopback (or that carry a non-loopback `Origin`). This
+    // blocks DNS-rebinding attacks, where a page the developer visits rebinds
+    // its hostname to a loopback address to read build outputs cross-origin.
+    // Mirrors the daemon asset server. Skipped when the user has bound the
+    // server to a routable address, which is an explicit opt-in to network
+    // serving.
+    if (restrictToLoopback) {
```
davidmorgan: Please put the restriction after the logging middleware.

build_runner/lib/src/commands/serve/server.dart
```diff
@@ -71,6 +72,16 @@ class ServeHandler {
       return assetHandler.handle(request, rootDir: rootDir);
     });
     var pipeline = const shelf.Pipeline();
+    // When serving on a loopback interface reject requests whose `Host` header
+    // is not itself loopback (or that carry a non-loopback `Origin`). This
+    // blocks DNS-rebinding attacks, where a page the developer visits rebinds
+    // its hostname to a loopback address to read build outputs cross-origin.
+    // Mirrors the daemon asset server. Skipped when the user has bound the
+    // server to a routable address, which is an explicit opt-in to network
+    // serving.
```
davidmorgan: I don't think the detailed comment is needed--it's clear from the code what it's doing.

build_runner/lib/src/commands/serve_command.dart
```diff
@@ -79,13 +79,20 @@ class ServeCommand implements BuildRunnerCommand {
       }
       final handler = ServeHandler(watcher);
 
+      // Bound to a loopback interface the server is reachable only from the
+      // local machine, so reject requests carrying a non-loopback `Host` header
+      // to block DNS-rebinding attacks. When the user has bound a routable
+      // address they have opted into network serving, so leave it unrestricted.
```
davidmorgan: I think this is likewise already understandable without a comment.

build_runner/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - Reduce logged compile progress output for non-interactive builds, for example
   presubmits.
+- The `serve` command now rejects non-loopback `Host` and `Origin` headers when
+  bound to a loopback interface, blocking DNS-rebinding access to build outputs.
```
davidmorgan: I would cut this off after the comma--no need to mentions the specific type of attack this is for.

davidmorgan (commented):
In https://github.com/dart-lang/build/issues/5097 the `_rejectCrossOrigin` check addresses this problem but just for websockets; the version in this PR is better, as it only rejects if serving on loopback.

Please remove `_rejectCrossOrigin` from `BuildUpdatesWebSocketHandler`.

davidmorgan:
Failure on CI is because of stale generated files unrelated to this PR--please merge from master and uploaded to get the needed updates so CI can run. Or if you prefer to rebase onto master that's okay, it's a small PR.

Thanks!

build_runner/lib/src/commands/serve/server.dart
```diff
@@ -252,6 +263,36 @@ String _buildUpdatesInjectedJS(String scriptName) =>
 window.\$dartLoader.forceLoadModule('packages/build_runner/src/commands/serve/$scriptName');
 ''';
 
+/// Whether [host] refers to a loopback interface.
+bool isLoopbackHost(String host) =>
+    host == 'localhost' ||
+    (InternetAddress.tryParse(host)?.isLoopback ?? false);
```
davidmorgan, line 247: Rather than checking the host string, check the server's actual resolved serving address: `server.address.isLoopback`.

herdiyana256:
Thanks for the review. Addressed all points and rebased onto master so CI has the up to date generated files:

- Removed `_rejectCrossOrigin` from `BuildUpdatesWebSocketHandler` (and its test group); the loopback middleware now covers that path.
- Moved the restriction after the logging middleware, so rejected requests are still logged.
- Dropped the extra explanatory comments in `server.dart` and `serve_command.dart`.
- Trimmed the CHANGELOG line after the comma.
- `restrictToLoopback` is now decided from the actual bound address via `server.address.isLoopback` instead of parsing the hostname string. The middleware still validates the incoming `Host`/`Origin` header values, since those are the request-controlled strings being checked.

build_runner/test/commands/serve/serve_handler_test.dart
```diff
@@ -275,6 +275,73 @@ void main() {
...
+    });
+
+    test('serves requests with a loopback host', () async {
+      addSource('a|web/index.html', 'content');
+      final response =
+          await serveHandler.handlerFor('web', restrictToLoopback: true)(
+            Request(
+              'GET',
+              Uri.parse('http://localhost/index.html'),
+              headers: {'host': 'localhost'},
```
davidmorgan, line 324: Please also add test case for host: localhost, origin localhost:8080, serving is allowed.

davidmorgan (commented):
Thanks!

build_runner/test/commands/serve/serve_handler_test.dart
herdiyana256: Added, as `serves requests with a loopback host and origin`: `host: localhost` plus `origin: http://localhost:8080`, asserting 200 and the served body.

Also rebased onto master. Since 2.16.1 shipped in #5111 the CHANGELOG entry had to move out of that released section, so this opens `2.16.2-wip` and bumps `build_test` to match, following the pattern in #5049. Happy to drop those version bumps if you would rather handle them separately.

Local check: `dart analyze` clean, `dart format` clean, `serve_handler_test.dart` 33/33 passing.

herdiyana256:
Rebased onto master to resolve the conflicts.

Master already bumped to `build_runner` 2.16.2-wip and `build_test` 3.5.21-wip, so the version bumps from this branch are gone and the pubspec changes are dropped. The changelog entry is now just a line in the existing 2.16.2-wip section of build_runner/CHANGELOG.md, and build_test/CHANGELOG.md is back to master.

The PR is down to 4 files, with the code and test diff unchanged.

davidmorgan (approved):
Looks good, thanks :)

davidmorgan:
I managed to get ahead of the flow of PRs and merge it, hurrah :)

Thanks again for this, I expect I can release it in the next week or so.

herdiyana256:
Thanks David! great working with you on this. 🎉

PR #5093, opened 2026-08-26
Separate asset content from incremental build digests.

davidmorgan:
Due to the previous change I discovered a few issues around `AssetContent`s dual behavior: it carries content but loses it on deserialization. On startup, there were cases where the content was not correctly restored, causing it to be loaded lazily later.

So, remove the dual-use AssetContent.

Make AssetContent always contain bytes or a string. Store digests separately in IncrementalBuildState, retain content in FinishedBuildState, and use PreviousBuild to hold partially loaded content during planning.

Fix build planning so it retains new content instead of discarding it after checking it.

Fix build planning so that externally modified post-process outputs are rebuilt, respecting --keep-modified-outputs and --only-check.

Replace `Sources`: move indexing code  `BuildFileIndex`, do source and missing source tracking in `BuildState` directly.

As a consequence

 - fix cases in build planning where content was read, checked but discarded: after build planning all files we have digests for now also have content
 - after the build, all files that were used in the build now have content, so there is no case where BuildOutputReader might want to cache a newly-read file with digest but no content
 - fix handling of changes to post process output outside the build: like normal outputs, rebuild to correct outputs unless --keep-modified-outputs or --only-check

Other cleanup

 - remove redundant BuilderFilesystem.buildStepPlan, delegate to buildState.buildStepPlan


davidmorgan:
(markdown presubmit fail is unrelated)

build_runner/lib/src/build/build_file_index.dart
```diff
@@ -111,20 +37,9 @@ class Sources {
   }
 
   /// Computes a `Map` from package names to sorted IDs of files in the package.
-  Map<String, List<AssetId>> _computeSortedFileIdsByPackage(
-    Iterable<AssetId> generatedIds,
-  ) {
-    if (_sortedFileIdsByPackageWasComputed) {
```
jensjoha, line 117: is it no longer the case it's only calculated once, or do you just not feel the check makes sense anymore?

build_runner/lib/src/build/builder_filesystem.dart
```diff
@@ -56,60 +58,43 @@ class BuilderFilesystem {
...
+      result: result,
+      contents: contents,
+    );
 
-    final declaredOutputsForStep =
-        buildStepPlan.declaredOutputsByStep[buildStepId];
+    final declaredOutputsForStep = buildStepPlan.declaredOutputsByStep[step];
     for (final declaredOutput in declaredOutputsForStep) {
-      if (!result.outputs.containsKey(declaredOutput)) {
+      if (!result.outputs.contains(declaredOutput)) {
```
jensjoha: I think the `buildState.addBuildStepResult` call above will throw if content and results don't agree. Couldn't this just be `_onUpdateContent?.call(declaredOutput, contents[declaredOutput]);`?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -264,34 +264,32 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
     final readerWriter = buildSpec.readerWriter;
     final buildInputs = BuildInputsBuilder()..cleanBuild = false;
 
-    final previousBuildState = previousBuild.state!;
+    final incrementalBuilder = previousBuild.incrementalState!.toBuilder();
+    final previousContents = previousBuild.contents.toBuilder();
```
jensjoha: Isn't this updated below to become current? (I find the name confusing).

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -300,28 +298,57 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+      if (oldIsSource) {
+        if (changed) buildInputs.updatedSources.add(id);
+        buildInputs.sourceContents[id] = newContent;
+        previousContents[id] = newContent;
+        incrementalBuilder.digests[id] = newContent.digest;
+        continue;
+      }
+
+      if (!changed) {
+        previousContents[id] = newContent;
```
jensjoha: Maybe a comment about why this is needed?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -300,28 +298,57 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+        continue;
+      }
+
+      if (buildSpec.buildOptions.outputStrategy == OutputStrategy.keep) {
+        final existingContent = previousBuild.contentOf(id);
```
jensjoha: Also here... The output has changed so we make sure to update the bytes but we use the old digest (although it might be possible for existingContent to now have the digest?)

build_runner/test/commands/serve/asset_handler_test.dart
```diff
@@ -54,13 +54,19 @@ void main() {
...
+          deletedPrimaryInput: true,
+        ),
       );
     }
     buildState.addSourceForTest(
       parsedId,
-      digest: AssetContent.digest(computeDigest(parsedId, 'a')),
+      content: AssetContent.string(
+        content,
+        digest: computeDigest(parsedId, 'a'),
```
jensjoha: I realize that this hasn't changed, but is there a reason why the content is not passed along but `a` is given instead?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -351,7 +381,13 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
 
     return BuildPlan((b) {
       b.buildSpec.replace(buildSpec);
-      b.previousBuild.replace(previousBuild);
+      b.previousBuild.replace(
```
jensjoha: This is not new, but I get confused on the naming... BuildPlan now has a  `previousBuild` that was updated to fit with the current world?

jensjoha (approved):

build_runner/lib/src/build/build_file_index.dart
davidmorgan: Because this class is no longer mutable, `_sortedFileIdsByPackage` is never reset to null and this can only be called once.

build_runner/lib/src/build/builder_filesystem.dart
davidmorgan: Thanks! Done.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Thanks.

// It's an output of the previous build that has remained unchanged on
// disk. Store it as `previousContents` so the next build can reuse it
// if it decides it does not need to be rebuilt.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Good point, thanks.

Restructured:

 - no mutation of PreviousBuild
 - new field `retainedOutputContents` in BuildInputs that does the work PreviousBuild was doing: carry outputs from the previous build that might be reused because they have not changed or because `--keep`.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: There was an "if (!changed) continue" above, so it's guaranteed that the digest doesn't match.

Added a comment

        // `--keep` mode retains external modifications to outputs as if the
        // content matches: copy the old digest.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Now addressed--previousBuild is no longer updated :) thanks!

build_runner/test/commands/serve/asset_handler_test.dart
davidmorgan: Stale test setup :) fixed, thanks.

PR #5094, opened 2026-08-27
Introduce AssetFile to differentiate source and cache locations

### LINKED ISSUE #5079: Incremental build deadlocks when a generated file is on disk but not in the asset graph

davidmorgan:
Introduce AssetFile to differentiate source and cache locations.

Fix #5079. Track asset locations in AssetTracker and BuildPlan so that source+cache conflicts are correctly cleaned up.

Add unit tests for AssetFile and incremental planning.

Add tests verifying that BuildState content resolution respects physical location and that incremental planning detects displaced outputs and changes.

### DIFF
--- a/build_runner/lib/src/build/build_series.dart
+++ b/build_runner/lib/src/build/build_series.dart
@@ -383,10 +383,12 @@ class BuildSeries {
       }
     }
 
-    final forceVisible =
+    // With `forceVisibleForTesting` all hidden outputs are forced to non-hidden
+    // outputs for simpler test assertions.
+    final forceVisibleForTesting =
         _buildPlan.buildSpec.testingOverrides.forceVisibleForTesting;
     for (final file in _buildPlan.conflictingOutputs) {
-      final outputIsHidden = forceVisible
+      final outputIsHidden = forceVisibleForTesting
           ? false
           : currentState.isHidden(file.id);
       final isMatchingActualOutput =
--- a/build_runner/lib/src/build/build_series.dart
+++ b/build_runner/lib/src/build/build_series.dart
@@ -9,6 +9,7 @@ import 'package:built_collection/built_collection.dart';
 import 'package:crypto/crypto.dart';
 import 'package:watcher/watcher.dart';
 
+import '../build_plan/asset_file.dart';
 import '../build_plan/build_directory.dart';
 import '../build_plan/build_filter.dart';
 import '../build_plan/build_plan.dart';
@@ -106,13 +107,19 @@ class BuildSeries {
         if ((change.type == .ADD || change.type == .MODIFY) &&
             _buildPlan.buildStepPlan.isDeclaredOutput(id)) {
           final expectedDigest = previousBuild.digestOf(id);
+          final isHidden = _buildPlan.buildStepPlan.isHidden(id);
           if (expectedDigest != null) {
             try {
               final bytes = await _buildPlan.readerWriter.readAsBytes(
                 id,
-                hidden: _buildPlan.buildStepPlan.isHidden(id),
+                hidden: isHidden,
               );
-              if (md5.convert(bytes) == expectedDigest) {
+              // TODO(davidmorgan): alternative is to track location in events.
+              final conflictingExists = await _buildPlan.readerWriter.canRead(
+                id,
+                hidden: !isHidden,
+              );
+              if (!conflictingExists && md5.convert(bytes) == expectedDigest) {
                 continue;
               }
             } catch (_) {}
@@ -146,10 +153,10 @@ class BuildSeries {
 
       // Changes to files that are part of the build.
 
-      // If not copying to a merged output directory, ignore changes to files
+      // If not copying to a merged output directory, ignore changes to sources
       // with no outputs.
       if (!_buildPlan.buildSpec.buildOptions.anyMergedOutputDirectory &&
-          !previousBuild.isMissingSource(id) &&
+          previousBuild.isSource(id) &&
           previousBuild.digestOf(id) == null) {
         rejected.add(change);
         continue;
@@ -177,18 +184,16 @@ class BuildSeries {
               'build.${_buildPlan.buildSpec.buildOptions.configKey}.yaml');
 
   Future<List<WatchEvent>> checkForChanges() async {
-    final updates =
-        await AssetTracker(
-          _buildPlan.readerWriter,
-          _buildPlan.buildSpec.buildPackages,
-          _buildPlan.buildSpec.buildConfigs,
-        ).collectChanges(
-          buildStepPlan: _buildPlan.buildStepPlan,
-          previousBuild: _buildPlan.previousBuild,
-        );
+    final updates = await AssetTracker(
+      _buildPlan.readerWriter,
+      _buildPlan.buildSpec.buildPackages,
+      _buildPlan.buildSpec.buildConfigs,
+    ).collectChanges(previousBuild: _buildPlan.previousBuild);
 
     return List.of(
-      updates.entries.map((entry) => WatchEvent(entry.value, '${entry.key}')),
+      updates.entries.map(
+        (entry) => WatchEvent(entry.value, '${entry.key.id}'),
+      ),
     );
   }
 
@@ -259,7 +264,11 @@ class BuildSeries {
     );
 
     if (!firstBuild || updates.isNotEmpty) {
-      _buildPlan = await _buildPlan.updateForFileChanges(updates);
+      final filesToCheck = <AssetFile>{
+        for (final id in updates) AssetFile.source(id),
+        for (final id in updates) AssetFile.cache(id),
+      };
+      _buildPlan = await _buildPlan.updateForFileChanges(filesToCheck);
     }
 
     final build = Build(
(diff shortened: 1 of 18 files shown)


build_runner/lib/src/build/build_series.dart
```diff
@@ -388,6 +383,28 @@ class BuildSeries {
       }
     }
 
+    final forceVisible =
```
jensjoha: I suppose this is more like `forceVisibleForTesting`? Also there should probably be a comment explaining what's going on.

build_runner/lib/src/build_plan/asset_file.dart
```diff
@@ -0,0 +1,29 @@
...
+  bool get hidden;
+
+  factory AssetFile(AssetId id, {required bool hidden}) =>
+      _$AssetFile._(id: id, hidden: hidden);
+
+  /// An asset in the source directory.
+  factory AssetFile.source(AssetId id) => _$AssetFile._(id: id, hidden: false);
+
+  /// An asset in the hidden cache directory.
+  factory AssetFile.cache(AssetId id) => _$AssetFile._(id: id, hidden: true);
```
jensjoha, line 26: I find it somewhat confusing that we use multiple terms for the same thing. source == hidden and cache == !hidden - couldn't we stick to one name?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -185,24 +190,26 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
   }) async {
     final readerWriter = buildSpec.readerWriter;
     final result = BuildInputsBuilder()..cleanBuild = true;
 
-    for (final id in filesToCheck) {
-      if (await readerWriter.canRead(id)) {
-        result.sources.add(id);
+    for (final file in filesToCheck) {
+      if (await readerWriter.canRead(file.id, hidden: file.hidden)) {
+        if (!file.hidden) {
```
jensjoha: Is there a reason for doing the `canRead` call if `file.hidden`?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -282,20 +293,33 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+    for (final file in filesToCheck) {
+      final id = file.id;
       final oldIsSource = previousBuild.isSource(id);
-      final oldExisted = previousBuild.isFile(id);
-      final oldDigest = previousBuild.digestOf(id);
+      final oldExisted =
+          oldIsSource ||
+          previousBuild.isActualOutput(id) ||
+          previousBuild.isActualPostOutput(id);
+      final oldFile = !oldExisted
```
jensjoha: This isn't easy to read.

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -282,20 +293,33 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+          ? null
+          : oldIsSource
+          ? AssetFile.source(id)
+          : AssetFile(id, hidden: previousBuild.isHidden(id));
```
jensjoha: Previously in this PR, at least, it seemed like something not being a source meant it was hidden, here it's not a source - when will it not be hidden?

build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -282,20 +293,33 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
...
+      final oldExistedHere = oldExisted && oldFile == file;
```
jensjoha: Maybe `oldExistedSameLocation`?

build_runner/lib/src/io/asset_tracker.dart
```diff
@@ -65,61 +67,41 @@ class AssetTracker {
...
     }
 
-    final preExistingOutputs = declaredAndActualOutputs.toSet().intersection(
-      allSources,
-    );
-    for (final id in preExistingOutputs) {
-      final originalDigest = previousBuild.digestOf(id);
+    for (final file in previousFiles.intersection(diskFiles)) {
+      final originalDigest = previousBuild.digestOf(file.id);
       if (originalDigest == null) continue;
```
jensjoha, line 94: I realize this isn't new, but when is this the case?

build_runner/test/integration_tests/build_command_invalidation_test.dart
```diff
@@ -118,5 +118,74 @@ void main() async {
...
+    await tester.run('root_pkg', 'dart run build_runner build --force-jit');
+    final cacheFile =
+        'root_pkg/.dart_tool/build/generated/root_pkg/web/a.txt.hidden';
+    expect(tester.read(cacheFile), 'a');
+
+    // Create a source file with the exact same logical ID.
+    tester.write('root_pkg/web/a.txt.hidden', 'source content');
+
+    // Next build recognizes the new source file and handles the conflict:
+    // the conflicting source file is deleted and hidden output remains in
```
jensjoha, line 146: Didn't notice that in the code - but doesn't that seem quite dangerous?

jensjoha (approved):

build_runner/lib/src/build/build_series.dart
davidmorgan: Renamed, added comment. Thanks.

build_runner/lib/src/build_plan/asset_file.dart
davidmorgan: That's very fair. This is worth a check+cleanup codebase wide--I'll send as a follow-up PR.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Thanks--moved the check before the read.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Cleaned up; we can use null `oldFile` to signal missing instead of needing `oldExisted`.

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: No change in behavior; the confusion is because we overload the word "source": in some places it means "source tree" and it others it means "build input".

In this case "not build input" -> "output" -> hidden or not depending on the build configuration.

I will include this terminology cleanup in the promised next PR :)

build_runner/lib/src/build_plan/build_plan.dart
davidmorgan: Done, thanks.

build_runner/lib/src/io/asset_tracker.dart
davidmorgan: Builders can read arbitrary files from their own package and transitive deps packages, but usually don't read everything; so they're ready lazily.

`orginalDigest == null` means the file was unused in the previous builds.

build_runner/test/integration_tests/build_command_invalidation_test.dart
davidmorgan: We only delete in packages that are an active part of the build; in deps packages a conflict is an error.

It used to be that you had to pass `--delete-conflicting-outputs` to do the delete, but every single build_runner guide recommended using the flag, so I made it the default.

This PR makes incremental build deletes consistent with clean build deletes.

davidmorgan (commented):
Thanks :)

davidmorgan:
Mockito failure was due to some release causing different whitespace in generated code; updated.

PR #5095, opened 2026-08-28
[build_web_compilers] Standardize web app entrypoint selection

parlough:
Provides a stable order for finding web entrypoints when hot reload is enabled. This fixes an issue we were hitting with Jaspr serving where a test file with `main` was being preferred over an entrypoint in the `web` directory. Now the following priority is followed, while still ensuring they are entrypoints:

1. The specified order of directories. So by default: `web` -> `test` -> `example` -> `benchmark`.
2. Files closer to the root. So considers `web/main.dart` before `web/a/main.dart`.
3. Files named `main.dart`.
4. Files with an earlier alphanumeric value.

This also updates the `web-assets-path` option to support a YAML list of values rather than just a comma separated list and adds additional validation to the option.


davidmorgan:
@Markzipan

builder_pkgs/build_web_compilers/lib/src/web_entrypoint_marker_builder.dart
```diff
@@ -90,4 +81,52 @@ class WebEntrypointMarkerBuilder implements Builder {
...
+/// Compares [a] and [b] by how likely each is to be a web app's entrypoint.
+///
+/// Assets closer to the root of the searched directory sort first,
+/// then those named `main.dart`, then those that sort first alphabetically.
+int _compareEntrypointPriority(AssetId a, AssetId b) {
+  // A top-level entrypoint is usually the app itself,
+  // while a nested one is more often a secondary target,
+  // such as a debug or example page.
+  final depthComparison = p
+      .split(a.path)
```
Markzipan: IIRC asset ids always use a URL paths (no Windows specific conventions), so we should be using p.url.split to avoid introducing accidental Windows-specific splits.

Markzipan (approved):
LGTM - with a small change.

Note that the final changes to hot reload webdev-side are being added in this change: https://github.com/dart-lang/webdev/pull/2867

after some DWDS-specific glue is landed in our SDK. This may delay formal hot reload support unless it's been handled separately in Jaspr.

builder_pkgs/build_web_compilers/lib/src/web_entrypoint_marker_builder.dart
davidmorgan: Correct, AssetId is normalized to be a relative path with forward slashes. Those guarantees were beefed up recently, so they now also apply on deserialization, and build_runner fixes any weird values passed in from builders.

https://github.com/dart-lang/build/blob/master/build/lib/src/asset_id.dart

parlough: Thanks for calling this out. Updated!

parlough:
Thanks for taking a look! Updated.

Note the failing test seems to be an unrelated flake that's also timing out on `main`.

Markzipan (approved):
LGTM!

PR #5096, opened 2026-08-31
Clarify source/cache/visible/hidden.

davidmorgan:
Clarify source/cache/visible/hidden.

`source` was used to mean two things: "at package path" and "build input". Use "at package path" for files at their package paths, keep "source" to mean "build input".

`cache` was used to mean many things. Stop using "build cache" to mean "in .dart_tool/build/generated", except in public APIs. Use "artifact tree" instead.

`visible` was used to mean "at package path". Use "at package path" instead.

`hidden` was used to mean "in .dart_tool/build/generated". Use "in artifact tree" instead.

Minor code changes:
 - cleaned up `BuildStepPlan.isHidden`: remove obsolete prefix check, rename to indicate that it's not a general predicate but can only return true for declared outputs --> `isDeclaredOutputInArtifactTree`


build_runner/lib/src/build/build_series.dart
```diff
@@ -366,37 +368,40 @@ class BuildSeries {
   /// longer declared outputs, plus files on disk that match declared outputs
   /// that were not actually output.
   ///
-  /// Set [onlyVisible] to skip hidden files.
-  Set<AssetId> _computeDeletes(BuildResult result, {bool onlyVisible = false}) {
+  /// Set [onlyAtPackagePath] to skip files in the artifact tree.
+  Set<AssetId> _computeDeletes(
+    BuildResult result, {
+    bool onlyAtPackagePath = false,
```
jensjoha: maybe `includeInArtifactTree = true` would be more clear?

build_runner/lib/src/build_plan/build_phases.dart
```diff
@@ -82,24 +82,25 @@ class BuildPhases {
...
   void checkOutputLocations(BuiltSet<String> packagesInBuild) {
     for (final action in inBuildPhases.cast<BuildAction>().followedBy(
       postBuildPhase.builderActions,
     )) {
-      if (action.hideOutput) continue;
+      if (action.outputsToArtifactTree) continue;
       if (packagesInBuild.contains(action.package)) continue;
-      // This should happen only with a manual build script since the build
-      // phases generation filters these out.
+      // Phase generation should have filtered these out; only possible if
```
jensjoha: Phase generation?

build_runner/lib/src/build_plan/previous_build.dart
```diff
@@ -51,7 +51,7 @@ abstract class PreviousBuild
   BuiltList<bool> get phaseOptionsChangedList;
   BuiltList<bool> get postBuildOptionsChangedList;
 
-  /// If the previous build is not compatible, source tree outputs from it to
+  /// If the previous build is not compatible, package path outputs from it to
```
jensjoha: there's something weird about this sentence (there was before too).

build_runner/lib/src/build_plan/previous_build.dart
```diff
@@ -149,7 +151,7 @@ abstract class PreviousBuild
   /// detail on compatibility.
   ///
   /// If the previous build cannot be used for an incremental build then
-  /// [incompatibleBuildOutputsToDelete] is filled with its source tree outputs
+  /// [incompatibleBuildOutputsToDelete] is filled with its package path outputs
```
jensjoha, line 154: this sentence too.

jensjoha (approved):

build_runner/lib/src/build/build_series.dart
davidmorgan: Yes, that's better; or rather, used `includeArtifactTree`, and make it `required` rather than having a default so it's clear at call sites. Thanks!

build_runner/lib/src/build_plan/build_phases.dart
davidmorgan: // BuildPhaseCreator filters these out; this should only be reachable
    // if there is a bug.

build_runner/lib/src/build_plan/previous_build.dart
davidmorgan: Thanks, improved to

/// If the previous build is not compatible, these are the package path
/// outputs from it that should be deleted.

build_runner/lib/src/build_plan/previous_build.dart
davidmorgan: Thanks, improved to

```
- /// to delete.
+ /// that should be deleted.
```

davidmorgan (commented):
Thanks.

PR #5098, opened 2026-08-31
Build runner with contracts

davidmorgan:
build_runner experiment with contracts.

### DIFF
--- a/build_runner/lib/src/build_plan/build_plan.dart
+++ b/build_runner/lib/src/build_plan/build_plan.dart
@@ -10,6 +10,7 @@ import '../build/asset_content.dart';
 import '../build/build_state/finished_build_state.dart';
 import '../build/library_cycle_graph/phased_asset_deps.dart';
 import '../constants.dart';
+import '../contracts.dart';
 import '../exceptions.dart';
 import '../io/asset_tracker.dart';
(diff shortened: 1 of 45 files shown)


build_runner/lib/src/build_plan/build_plan.dart
```diff
@@ -461,7 +469,8 @@ abstract class BuildPlan implements Built<BuildPlan, BuildPlanBuilder> {
       buildInputs.retainedOutputContents.remove(id);
     }
     for (final id in newArtifactTreeFiles) {
-      if (!finalSources.contains(id)) {
+      if (!finalSources.contains(id) &&
+          buildInputs.retainedOutputContents[id] == null) {
```
jensjoha: this seems more like a fix?

build_runner/lib/src/build_plan/previous_build.dart
```diff
@@ -66,6 +79,23 @@ abstract class PreviousBuild
     b.triggersChanged = false;
     b.incrementalState.replace(finishedBuildState.incremental);
     b.buildStepPlan.replace(finishedBuildState.buildStepPlan);
+    b.phaseOptionsChangedList.replace(
```
jensjoha: here too

PR #5101, opened 2026-08-31
Improve `BuildOutputReader`; add `BuildOutputReadResult`.

davidmorgan:
Improve `BuildOutputReader`; add `BuildOutputReadResult`.

Previously, `BuildOutputReader` could return stale data, and all callers did checks to avoid reading it.

Make `BuildOutputReader` throw on stale data, and tidy up the checking work by encapsulating in `BuildOutputReadResult`.

This was found via the contract programming experiment, which identified that the responsibility of avoiding publishing stale data was on the callers of `BuildOutputReader`.

### DIFF
--- a/build_runner/lib/src/commands/serve/server.dart
+++ b/build_runner/lib/src/commands/serve/server.dart
@@ -288,14 +288,14 @@ class AssetHandler {
         result = candidate;
         break;
       }
+      // Ensure some result, if it's not readable then an error will be reported
+      // against it.
       result ??= candidate;
     }
-    // Or if none exists, report an error about the first one.
-    result ??= await reader.read(assetIds.first);
 
     try {
       try {
-        if (!result.canRead) {
+        if (!result!.canRead) {
           switch (result.unreadableReason!) {
             case UnreadableReason.failed:
               return shelf.Response.internalServerError(
--- a/build_runner/lib/src/commands/serve/server.dart
+++ b/build_runner/lib/src/commands/serve/server.dart
@@ -15,6 +15,7 @@ import 'package:shelf_web_socket/shelf_web_socket.dart';
 import 'package:web_socket_channel/web_socket_channel.dart';
 
 import '../../build/build_result.dart';
+import '../../io/build_output_read_result.dart';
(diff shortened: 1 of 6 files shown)


build_runner/lib/src/commands/serve/server.dart
```diff
@@ -284,31 +281,32 @@ class AssetHandler {
...
+    BuildOutputReadResult? result;
     for (final id in assetIds) {
-      if (await reader.canRead(id)) {
-        assetId = id;
+      final candidate = await reader.read(id);
+      if (candidate.canRead) {
+        result = candidate;
         break;
       }
+      result ??= candidate;
```
jensjoha, line 293: why? Shouldn't it at least make line 294 irrelevant? (modulo possibly having to use a bang somewhere)

build_runner/test/io/build_output_reader_test.dart
```diff
@@ -382,18 +387,151 @@ void main() {
...
+        readerWriter: buildPlan.readerWriter,
+        buildState: buildState.toFinishedBuildState(),
+      );
+
+      final readableResult = await reader.read(readableId);
+      expect(readableResult, isA<BuildOutputReadResult>());
+      expect(readableResult.canRead, isTrue);
+      expect(readableResult.unreadableReason, isNull);
+      expect(readableResult.readAsString(), 'content');
+      expect(
```
jensjoha, line 522: I suppose this is a test of it being cached? Maybe write that in a comment because it reads a little weird...

jensjoha (approved):

build_runner/test/io/build_output_reader_test.dart
davidmorgan: Done.

build_runner/lib/src/commands/serve/server.dart
davidmorgan: Yes, it was an attempt to do line 294 without a double read, but line 294 was left in error.

Removed and added a comment.

And a bang :)

davidmorgan (commented):
Thanks.

PR #5110, opened 2026-09-02
Import built_value into built_types/.

davidmorgan:
Import exact files from google/built_value.dart at release-8-13-0, commit 54b4ec6e4f17. Small commit on top to make CI continue to pass.

I propose to import without full git history and without moving all the existing built_value issues in.

It's been in maintenance mode for a long time: we can add new issues here based on new work we want to do--primarily, keeping it up to date with new build_runner and language features.

Also re: git history, Gemini is really good at tracing git history--I don't think it will be a problem for it to hop over to the archive repo.

But, happy to go for the approach we used for mockito (copy history, move issues) if you prefer; or some hybrid.

Cleanup to follow directly: refactor to workspace, add to CI.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -18,6 +18,7 @@ jobs:
     with:
       write-comments: false
       sdk: dev
+      ignore-packages: 'built_types/**'
     permissions:
       id-token: write
       pull-requests: write
--- a/built_types/.github/dependabot.yaml
+++ b/built_types/.github/dependabot.yaml
@@ -1,0 +1,8 @@
+# Dependabot configuration file.
+version: 2
+
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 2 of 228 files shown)


jensjoha:
https://github.com/google/built_value.dart/commit/5582a42fffe6abfbc21ae8998c47350b1ce2c1b3 I suppose... There are a few diffs in CHANGELOG.md and chat_example/README.md when comparing to that...

I don't think I have any opinion on whether to copy or move...

jensjoha (approved):

davidmorgan:
Thanks!

> https://github.com/google/built_value.dart/commit/5582a42fffe6abfbc21ae8998c47350b1ce2c1b3 I suppose... There are a few diffs in CHANGELOG.md and chat_example/README.md when comparing to that...

The changes are in the second commit: minimal changes to get CI to stay green.

PR #5123, opened 2026-09-04
Add testing/bin/build_runner entrypoint for fast integration tests.

davidmorgan:
This doesn't yet speed up the full test suite because there are a small pile of tests doing compiles that are now the slowest thing running.

Those can be refactored next :)

But, this already significantly reduces load running the suite.


davidmorgan:
Closing for now as this is tougher than it looks :)

PR #5129, opened 2026-09-04
Require analyzer >=14.3.0, use AnalysisOptionsBuilder.

scheglov:


build_runner/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ platforms:
   macos:
 
 dependencies:
-  analyzer: '>=13.3.0 <15.0.0'
+  analyzer: '>=14.3.0 <15.0.0'
```
davidmorgan, line 16: I think here is the only place the new requirement is needed.

Other packages can continue to work with earlier `analyzer` versions and earlier `build_runner` releases, their min analyzer version will be updated when there is something they need.

davidmorgan (commented):
Thanks. I'm guessing you want this rolled into google3?

Happy to do so--how urgent is it?

scheglov:
Yes, I need it rolled into google3 so that I can land https://dart-review.googlesource.com/c/sdk/+/529541
Not urgent.

davidmorgan (approved):
Thanks :) I'll look at rolling into google3 this week.

davidmorgan:
@scheglov google3 roll is done--please let me know if you're still blocked.

PR #5135, opened 2026-09-07
Copy built collection

davidmorgan:
Import exact files from google/built_collection.dart at origin/master (4e9db27a94df).

Then small fixes on top to get CI to pass.


davidmorgan:
I should have said--this package sees almost no changes, and should be almost no work; it's convenient to have it here as it reduces the Morgan-owned-side-depot-deps to zero.

jensjoha (approved):

PR #5140, opened 2026-09-09
Allow hyphens in package names.

davidmorgan:
Unblocks rolling to google3.


jensjoha (approved):
Should this be a google3-specific thing somehow instead with it not actually being a valid package name?

davidmorgan:
> Should this be a google3-specific thing somehow instead with it not actually being a valid package name?

Since we added this restriction without a major version bump, I think it's kinder to relax it again :) as we would if anyone else pinged us saying "hey we're using that".

PR #5144, opened 2026-09-11
Copy CHANGELOG.md to built_value_generator and built_value_test

davidmorgan:
Replace symlinks to built_value/CHANGELOG.md with standalone copies in built_value_generator and built_value_test so package changelogs can be tracked and updated independently.

We'll want to separate the version numbers when we update built_value_generator to use shared parts: it will be a breaking change to the generator but not breaking for built_value.


jensjoha (approved):
rubberstamp

PR #5146, opened 2026-09-14
Add CODEOWNERS to route reviews automatically.

davidmorgan:
Ownership is a suggestion here, as it is in dart-lang/core, not a merge requirement. No branch protection rule refers to it.

@Markzipan what do you think please? I'll still look at web compilers PRs, but I route them to you first either way :) so I figured we might as well automate it.


Markzipan (approved):
Makes sense - ty for routing!

PR #5147, opened 2026-09-14
Add a single CI job that summarises the whole run.

davidmorgan:
Adds one job that runs after all the others and fails if any of them failed or was cancelled. Its name does not change when the jobs beneath it change, so it can be used for repo automation like autosubmit.


mono_repo.yaml
```diff
@@ -7,3 +7,14 @@ merge_stages:
 github:
   permissions:
     contents: read
+  on_completion:
+    - name: ci
+      runs-on: ubuntu-latest
+      # Runs whatever happened above, because a job that simply depends on a
```
jensjoha: I don't understand the comment.

jensjoha (approved):

mono_repo.yaml
davidmorgan: Thanks--improved :)

PR #5148, opened 2026-09-14
Shared part output engine integration and phase execution

davidmorgan:
Shared part output engine integration.

Further work towards #4155.

Wires the shared part data structures landed in #5014 into the build engine, so
builders that opt in with `adds_to_library: true` can write to a shared part via
`BuildStep.librarySourceSink`.

- `LibrarySourceSinkImpl` and `BuildStepImpl.librarySourceSink`: collect a
  contribution and its imports for one build step, with a per-phase import
  prefix.
- `BuildState`: accumulates contributions per library across phases and serves
  shared part content, including content truncated to a phase.
- `BuilderFilesystem` and `AnalysisDriverFilesystem`: expose the shared part to
  the analyzer so a builder resolving a library sees the contributions from
  earlier phases only, and hide `_br_` assets from `AssetReader` calls.
- `BuildPlan` and `PreviousBuild`: track shared parts across incremental builds,
  reusing unchanged contributions, invalidating changed ones, and deleting parts
  whose contributors are gone.
- `BuildOutputReader`: reads generated shared parts.
- `build_test`: `addsToLibraryBuilders` and `addsToLibraryBuilderFactories` so
  builders using `librarySourceSink` can be tested.

Tests in `part_build_test.dart` cover multi-phase composition, incremental reuse
and removal of contributions, language version comments, phase visibility during
resolution, recovery after the asset graph is discarded, and deletion when
contributors are removed by a configuration change.

More extensive integration tests coming in the next PR.


build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
```diff
@@ -137,6 +155,30 @@ class AnalysisDriverFilesystem
...
+      if (_data.remove(path) != null) {
+        _changedPaths.add(path);
+      }
+      return;
+    }
+    _writeContent(
+      BuildRunnerFileContent(
+        path: id.asPath,
+        exists: true,
+        content: partContent.dartStringValueOrEmptyFail(id: id),
```
jensjoha: why? I mean most places it justs calls `stringValue()` directly; below it assumes to get the digest without throwing and if it were to throw, would using the empty string make sense?

build_runner/lib/src/build_plan/previous_build.dart
```diff
@@ -136,10 +137,24 @@ abstract class PreviousBuild
...
 
   Digest? digestOf(AssetId id) => digests[id];
 
+  // -- Shared parts.
+
+  Iterable<AssetId> get sharedPartIds =>
+      digests.keys.where((id) => id.isBrSharedPart);
+
+  Iterable<AssetId> get sharedPartLibraryIds =>
+      sharedPartIds.map((id) => id.sharedPartLibraryId!).whereType<AssetId>();
```
jensjoha: does the whereType add anything?

build_runner/test/build/part_build_test.dart
```diff
@@ -0,0 +1,1025 @@
...
+          'a|lib/a.dart': '',
+          'a|lib/b.txt': 'initial_b',
+          'a|lib/c.txt': 'initial_c',
+        },
+        outputs: {'a|lib/_br_/a.part.dart': expectedGeneratedPart},
+      );
+
+      // Now we do an incremental build where we ONLY change b.txt.
+      // builder1 will run again, builder2 will be cached.
+      final expectedGeneratedPart2 = '''
```
jensjoha: Why not use `r'''`? (several places)

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -192,14 +215,32 @@ class BuildState {
 
     builder.globResults.addAll(_globResults);
 
+    for (final entry in _partData.entries) {
+      final partId = entry.key.sharedPartId!;
+      final digest =
```
jensjoha: This dance with _retainedPartContent ?? finalContent() is done in several places. Maybe a helper would make things more readable?

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -242,6 +283,115 @@ class BuildState {
     _contents.addAll(contents);
   }
 
+  // -- Shared parts.
+
+  /// Adds [sharedPart].
+  ///
+  /// Throws if a shared part for the same library already exists.
+  void addSharedPart(SharedPartAccumulator sharedPart) {
```
jensjoha: This is called twice from non-tests, in both cases it's guarded by `hasSharedPart` before it (logically) calls `addPartContribution`. Would it make sense to just combine them?

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -242,6 +283,115 @@ class BuildState {
...
+    if (imports == null && contribution == null) return;
+    if (!hasSharedPart(libraryId)) {
+      addSharedPart(
+        SharedPartAccumulator(
+          libraryId,
+          languageVersion ?? fromPart.languageVersion,
+        ),
+      );
+    }
+    _partData[libraryId]!.addContribution(
```
jensjoha: This isn't calling `addPartContribution` but adds directly. It doesn't call `markPartRebuilt` is that on purpose on a mistake?

build_runner/lib/src/build/build.dart
```diff
@@ -803,6 +875,14 @@ class Build {
         }
       }
 
+      if (_partPhaseIndices[step.phaseNumber] != null) {
```
jensjoha: Here (and above I guess) vs line 985 seems to disagree on how to check what at least feels like the same thing.
For that matter it feels like a misuse of the map to use it like it's used here.

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -242,6 +283,115 @@ class BuildState {
...
+    }
+    return partData.contentAt(upToPhase);
+  }
+
+  /// The previous build's on-disk content of the shared part for [libraryId].
+  ///
+  /// Returns `null` if there is none, or if the part is rebuilt in this build
+  /// and so the previous content no longer applies.
+  AssetContent? _retainedPartContent(AssetId libraryId) {
+    if (_librariesWithRebuiltPart.contains(libraryId)) return null;
```
jensjoha: Would it make sense to remove things from `_retainedOutputContents` instead of this? For  that matter it's only ever called with ?? and a SharedPartAccumulator --- so maybe that is a hint of doing something else too. (the last part is basically reiterating a comment from above, but I'll leave it anyway)

build_runner/lib/src/build/build_state/build_state.dart
```diff
@@ -242,6 +283,115 @@ class BuildState {
...
+      addSharedPart(
+        SharedPartAccumulator(
+          libraryId,
+          languageVersion ?? fromPart.languageVersion,
+        ),
+      );
+    }
+    _partData[libraryId]!.addContribution(
+      phase,
+      builderKey ?? '',
```
jensjoha, line 374: What's the story with the builder key? Are they optional or is a defensive style or..?

jensjoha (approved):
Something about this feels overly complex, but I can't quite put my finger on it (except for the comments I've made - maybe fixing them will help, I don't know)...

build_runner/lib/src/build/resolver/analysis_driver_filesystem.dart
davidmorgan: Good catch, thanks--it's not needed here as there can't be invalid UTF8 here.

build_runner/lib/src/build_plan/previous_build.dart
davidmorgan: Thanks--removed.

build_runner/test/build/part_build_test.dart
davidmorgan: That's better, thanks--done.

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: Yes, thanks: done.

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: Thanks--done.

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: Yes: if copied, it's not rebuilt, and doesn't need writing to disk later.

This distinction is important but was not covered by any test--test coverage added.

build_runner/lib/src/build/build.dart
davidmorgan: Re: misuse of a map; yes; add _isPartPhase helper so it's clearer.

The checks are different because here we assume the primary input is a .dart file, while below we're iterating nextGraph.root.ids which don't have to be .dart files, someone might import 'foo.txt'. This difference is not currently enforced: will add a constraint that shared part builders must run on .dart primary inputs.

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: Partly addressed for other comment--factor out `_partContent`.

`BuildState` is append-only by design. It starts empty and fills up with correct content: it never holds values that someone has to invalidate or remove later. That makes correctness much easier to reason about :) ... anecdota: the previous version of this PR did have removal, and switching to append-only caught a bug.

build_runner/lib/src/build/build_state/build_state.dart
davidmorgan: They're required but it's not clear with the code as it is; will send follow-up.

davidmorgan (commented):
Agreed re: seems complex; this is the most complex step in a highly complex new feature. It's taken quite some effort to get it down to just this :)

I'll be on the lookout for ways to simplify once I understand better how it works in practice.

Thank you for the excellent review, as always :)

build_runner/lib/src/build/build.dart
davidmorgan: ("will add constraint"-->second follow-up PR).

PR #5149, opened 2026-09-14
Fix the remaining small lint groups in built_types

davidmorgan:
Removes thirteen repo-wide lint suppressions from
built_types/analysis_options.yaml and fixes the violations they hid.

Most are real fixes. A handful are deliberate test fixtures or intentional imports of another package's internals, and get a targeted ignore with an explanation.

TAG=agy
CONV=9eabb90a-3e69-4742-a07f-41be531b3bf3


built_types/built_collection/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Require Dart SDK `^3.11.0`.
 - Fix unnecessary non-null assertions in internal copy-on-write collections.
 - Migrate to `package:dart_flutter_team_lints`.
+- Fix remaining analysis issues and remove the lint suppressions that hid them.
```
jensjoha: I don't think "remaining" adds anything. (also several places below)

built_types/built_value_chat_example/pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   shelf: ^1.0.0
   shelf_proxy: ^1.0.0
   shelf_web_socket: ^1.0.0
+  test: any
```
jensjoha: From what I can tell this is done because "lib/testing" uses "package:test" --- but wouldn't it make more sense to have that in "test" instead? I realize this is an example thing, but still.

jensjoha (approved):

built_types/built_collection/CHANGELOG.md
davidmorgan: Done, thanks.

built_types/built_value_chat_example/pubspec.yaml
davidmorgan: Done, thanks.

PR #5151, opened 2026-09-16
Write the SDK summary deps file atomically.

davidmorgan:
The SDK summary cache under `.dart_tool/build_resolvers` is two files, `sdk.sum` and `sdk.sum.deps`. The summary was written to a temporary file and then renamed into place, but the deps file was created empty and then written to.

So, a second build process running concurrently in the same package can see the deps file in the window between `create` and `writeAsString`, pass the `exists` check, and then fail decoding an empty file:

FormatException: Unexpected end of input (at character 1)

Move the temporary file and rename into `writeAtomically` and use it for both files. The deps file is written last, so it stays the marker that the summary next to it is complete and current.

A deps file can still be left corrupt on disk by a process killed partway through the old write, so extract the comparison as `depsMatch` and have it treat content that is not a JSON map as not matching. A corrupt deps file now costs a rebuild rather than failing the build.


build_runner/test/io/atomic_write_test.dart
```diff
@@ -0,0 +1,66 @@
...
+      expect(file.readAsStringSync(), 'hello');
+    });
+
+    test('leaves no temporary files behind', () async {
+      final file = File(p.join(tempDir.path, 'file.txt'));
+      await writeAtomically(file, utf8.encode('hello'));
+      expect(tempDir.listSync().map((entity) => entity.path), [file.path]);
+    });
+
+    test('does not create the file until it has the full content', () async {
```
jensjoha, line 49: I'm not sure this test actually does anything?

jensjoha (approved):

build_runner/test/io/atomic_write_test.dart
davidmorgan: Agent says it's reliably red before this PR and green after

  15 runs red before the fix, 15 runs green after. Not a stress test: the                                                                                                                                         
  create/write gap spans event loop turns, so the poller always lands                                                                                                                                             
  in it.

PR #5152, opened 2026-09-17
Shared part output invalidation and CLI integration tests

davidmorgan:
Shared part output invalidation and CLI integration tests

Further work towards #4155.

The shared part engine landed in #5148 with unit test coverage. Add end to
end coverage: the real `build_runner` CLI, and the invalidation tester,
running builders that write shared parts.

`build_command_write_part_test.dart` drives the CLI: one builder writing a
part, two builders concatenating into one part, a later phase resolving a
library that has a generated part, deleting a source so its generated part
goes away, and a contribution that adds imports which a later phase then
resolves under `--enable-experiment=enhanced-parts`. All the builders live
in packages `root_pkg` depends on, and each section of the test enables the
ones it needs, so the whole file shares one workspace and one build script.

`part_writer_invalidation_test.dart` pins down what an incremental build
rewrites and rebuilds. A no-op build rewrites nothing; a builder that
resolves the shared part rebuilds when the part content changes but not
when an unrelated input changes; deleting the source deletes the part;
optional builders can write shared parts.

`invalidation_tester.dart` gains `addsToLibraryBuilders`, `writesPart`,
`partWrite` and `partGraph` to describe those cases.

`output_strategy_test.dart` covers the output strategies for a generated
part: overwrite, `--keep-modified-outputs`, `--only-check`, and restoring a
part deleted from disk.

`build_command_config_validation_test.dart` checks that
`adds_to_library: true` with `is_optional: true` is rejected by the CLI.


build_runner/test/invalidation/part_writer_invalidation_test.dart
```diff
@@ -0,0 +1,166 @@
...
+            '_br_/c.part',
+            'a.2',
+            'b.2',
+            'c.2',
+          ],
+        ),
+      );
+    });
+
+    test('no-op build does not rebuild c.2', () async {
```
jensjoha: `c.2` is more like anything or..? (also below)

jensjoha (approved):
Consider if it makes sense to move all the dart-code-as-strings into files.

If not, maybe at least using `r'''` could avoid having to do `\$`.

Another possibility (related to the first) would also be - and that probably wouldn't be for this cl - to have an expectation test framework like the cfes strong test for instance. That is essentially what this is, only here it's strings inside a dart file which doesn't make things easier to read.

build_runner/test/invalidation/part_writer_invalidation_test.dart
davidmorgan: Thanks--done.

davidmorgan (commented):
Thanks--`r'''` is strictly better, sorry I missed it again.

Re: code as strings; for some tests it works well like this, where it lets you see code + yaml + structure in a short space; as they grow, it gets less convincing. I'll give it some thought--thanks for highlighting.

PR #5154, opened 2026-09-17
Fix BuiltMap equality for different keys with null values

hoc081098:
### 1. Fixes `BuiltMap` equality when two maps have:

- the same length;
- the same hash code;
- different keys; and
- `null` values.

Previously, equality only compared values using `operator []`. Since a missing
key and a key mapped to `null` both return `null`, maps with different keys
could incorrectly compare equal after passing the length and hash-code checks.

The regression test compares `{1: null}` and `{2: null}` using overridden,
matching hash codes. This ensures equality reaches the entry comparison and
verifies that the maps are not equal.

### 2. Validation:

- `dart format --output=none --set-exit-if-changed .`
- `dart analyze --fatal-infos`
- `dart test`

---

<details>

</details>


built_types/built_collection/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix unnecessary non-null assertions in internal copy-on-write collections.
 - Migrate to `package:dart_flutter_team_lints`.
 - Fix analysis issues and remove the lint suppressions that hid them.
+- Fix `BuiltMap` equality when maps with nullable values have different keys.
```
davidmorgan: It's narrower than that, let's be specific, e.g.:

    - Fix `BuiltMap` equality incorrectly returning true for maps with null values,                                                                                                                                
      different keys and colliding hash codes.

built_types/built_collection/lib/src/map/built_map.dart
```diff
@@ -104,6 +104,8 @@ abstract class BuiltMap<K, V> {
     if (other.hashCode != hashCode) return false;
     for (final key in keys) {
       if (other[key] != this[key]) return false;
+      // A lookup can't distinguish a missing key from one mapped to null.
+      if (!other._map.containsKey(key)) return false;
```
davidmorgan: We can save a lookup for all cases but the null one:

```
    for (final key in keys) {                                                                                                                                                                                      
      final value = this[key];                                                                                                                                                                                     
      if (other[key] != value) return false;                                                                                                                                                                       
      // A lookup cannot distinguish a missing key from one mapped to null.                                                                                                                                        
      if (value == null && !other._map.containsKey(key)) return false;                                                                                                                                             
    }  
```

built_types/built_collection/test/map/built_map_test.dart
```diff
@@ -267,10 +266,26 @@ void main() {
       );
     });
 
+    test('compares not equal to different keys with null values', () {
```
davidmorgan, line 269: Please also test two equal content maps with null values compare as equal.

davidmorgan (commented):
This is a great find, thank you--just a few suggestions.

hoc081098:
> This is a great find, thank you--just a few suggestions.

Thanks for the review!
I've addressed your suggestions in https://github.com/dart-lang/build/pull/5154/changes/440fa42738d8f8320ad1c8d8b72c537b3cfc4c08

